PR #18, opened 2026-05-17
fix(RemoteLlmEngine): pin Java HttpClient to HTTP/1.1 (hangs with non-Apache-class HTTP/1.1 LLM servers)

pmanko:
`RemoteLlmEngine.getHttpClient()` builds the Java HttpClient with no `.version()` call, so it defaults to `HTTP_2` and attempts an `h2c` upgrade on the first plaintext-`http://` request. Against several lightweight HTTP/1.1 LLM servers commonly used with the `remote` engine — notably LM Studio's llama.cpp-derived server, but the same shape applies to `llama-cpp-server`, some `vLLM` configurations, and `text-generation-inference` — the upgrade handshake wedges the client. `HttpClient.send()` parks on its internal CompletableFuture and never completes; `chartsearchai.llm.timeoutSeconds` does not recover because the wedge is before the request actually leaves the client (per-request `.timeout()` only covers in-flight sockets).

Confirmed with a thread dump: `http-nio-*-exec-*` parked at `RemoteLlmEngine.infer:67 → HttpClient.send → CompletableFuture.get`, zero TCP connections in the container's `/proc/net/tcp` to the endpoint, and zero `chat/completions` entries in the server's request log. With this one-line change pinning `HTTP_1_1`, the same setup (LM Studio + gemma-4-e2b-it, a 25k-token chart prompt) streams an answer with citations in ~12s.

`https://` endpoints (OpenAI, Anthropic, etc.) are unaffected — Java's `client.version()` preference is overridden by ALPN, so TLS connections keep negotiating HTTP/2 as before. The change only affects plaintext `http://` endpoints, which is exactly the case where the wedge happens.

CI failures on this PR are the pre-existing `querystore-api:1.0.0-SNAPSHOT` resolution error introduced in `0215d81` — independent of this change.

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/api/impl/RemoteLlmEngine.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/api/impl/RemoteLlmEngine.java
@@ -206,6 +206,7 @@ public class RemoteLlmEngine implements LlmEngine {
 	private synchronized HttpClient getHttpClient() {
 		if (httpClient == null) {
 			httpClient = HttpClient.newBuilder()
+					.version(HttpClient.Version.HTTP_1_1)
 					.connectTimeout(Duration.ofSeconds(30))
 					.build();
 		}


dkayiwa:
I verified this against current main (5223f92) and want to take it.

The gap is still there: RemoteLlmEngine.getHttpClient() builds its client with only connectTimeout (lines 231-238), no version pin, so the default HTTP_2-with-h2c-upgrade behavior you diagnosed is still in play for plaintext endpoints. Your one-line diff applies cleanly to today's main (checked with `git apply --check`). The CI failures on this branch were the querystore-api SNAPSHOT resolution problem of that era, which is gone: the artifact is published on mavenrepo.openmrs.org (snapshot build 74, last updated 2026-07-14) and main's build matrix is green.

Two accuracy notes from testing around this, neither of which blocks:

1. The PR text says https endpoints are unaffected because ALPN overrides the client preference. It is actually the reverse: with version(HTTP_1_1) the client stops offering h2 in ALPN, so TLS endpoints also drop to HTTP/1.1. That is harmless for OpenAI-style APIs (long-lived single streaming requests), just worth recording accurately here for the history.

2. The wedge looks specific to certain server implementations rather than to HTTP/1.1 LLM servers generally: LocalLlmEngine drives the bundled llama.cpp server over plaintext http with the same unpinned java.net.http.HttpClient and has never hung, so cpp-httplib evidently tolerates the h2c upgrade header while LM Studio's server does not. Your thread-dump evidence stands, and the pin is the right defensive default for the remote engine either way.

Could you rebase so CI runs green here? If you would rather not, say so and I will cherry-pick the commit with attribution. I will smoke-test the public demo (which runs the remote engine) right after this lands.

PR #19, opened 2026-05-17
fix(QueryStoreChartBuilder): access querystore-api reflectively, drop Maven dep

pmanko:
`QueryStoreChartBuilder` imports `org.openmrs.module.querystore.api.QueryStoreService` and `org.openmrs.module.querystore.model.QueryDocument` directly, with a corresponding `querystore-api:1.0.0-SNAPSHOT` compile-time dependency in `api/pom.xml`. querystore is a SNAPSHOT-only sibling module not published to `mavenrepo.openmrs.org`, so this dependency makes chartsearchai unbuildable for anyone who doesn't have querystore-api in their local Maven repo. The repo's own GitHub Actions CI has been failing on every main-branch run since 0215d81 introduced the dep on 2026-05-15.

Refactor resolves both classes via `Class.forName` + reflective `Method.invoke`, and removes the querystore-api dependency from `api/pom.xml`. The existing graceful-degradation path (return empty chart) already handled `APIException` and `LinkageError` when querystore wasn't installed at runtime; this just adds `ClassNotFoundException` to the same handler. When querystore IS installed, reflection finds the service and the search path runs identically to the direct-typed code.

The reflective accessor lookup is grouped in a small `QueryDocumentAccess` holder for readability; the public surface of `QueryStoreChartBuilder.build(Patient, String)` is unchanged.

Verified: `api` + `omod` build standalone with no `querystore-api` in the local Maven repo; chartsearchai module starts and serves `/search/stream` against a real patient with querystore-api intentionally absent from the runtime classpath.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -52,12 +52,11 @@
 			<version>8.11.2</version>
 			<scope>provided</scope>
 		</dependency>
-		<dependency>
-			<groupId>org.openmrs.module</groupId>
-			<artifactId>querystore-api</artifactId>
-			<version>1.0.0-SNAPSHOT</version>
-			<scope>provided</scope>
-		</dependency>
+		<!-- querystore-api is accessed reflectively in QueryStoreChartBuilder so
+		     chartsearchai compiles and starts without it. When the querystore
+		     omod is installed at runtime, reflection finds the service. When
+		     it's not, the chart-build falls back to an empty result, same as
+		     a search returning zero hits. -->
 		<dependency>
 			<groupId>org.openmrs.api</groupId>
 			<artifactId>openmrs-api</artifactId>
--- a/api/src/main/java/org/openmrs/module/chartsearchai/api/impl/QueryStoreChartBuilder.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/api/impl/QueryStoreChartBuilder.java
@@ -9,6 +9,8 @@
  */
 package org.openmrs.module.chartsearchai.api.impl;
 
+import java.lang.reflect.Method;
+import java.time.LocalDate;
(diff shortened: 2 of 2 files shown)


dkayiwa:
Both problems this solves got solved the opposite way since May, so I am closing it.

The build problem is gone. querystore-api 1.0.0-SNAPSHOT is published on mavenrepo.openmrs.org (the snapshot metadata shows build 74, lastUpdated 2026-07-14), and main's CI is green with the typed provided-scope dependency still in api/pom.xml (lines 26-31 at 5223f92).

Runtime optionality was explicitly rejected rather than implemented. Since 41efea3 (#56, implementing #51), config.xml declares querystore as a required module, so the module manager guarantees it is present before chartsearchai starts; there is no supported querystore-absent runtime state to degrade into.

With that premise gone, porting the reflection forward would be a net loss on what is now our only retrieval path. QueryStoreChartBuilder today imports three querystore types (QueryStoreService, QueryDocument, and QueryStoreConstants for the obs-group metadata fields used at lines 274-275) and calls getPatientChart plus metadata accessors, a wider surface than the four getters wrapped here. Reflection would convert any API drift between the two modules from a compile error into runtime empty-chart degradation, which is silent, and exactly the kind of failure nobody notices until a clinician gets an empty answer.

The branch also conflicts with main; the file was rewritten around getPatientChart and buildFocused since this was cut. Thanks for the clean, small diff. This one was simply overtaken by the direction change.

pmanko:
Closing. #157 keeps the typed querystore-api dependency and consumes it directly; the reflective access was deliberately reverted in favour of the typed bridge, and the paired build in #157's CI exists to prove that pinned source combination. Not carried, by decision.

PR #20, opened 2026-05-18
feat(chat): multi-turn chat with frozen-session chart in stable prefix

pmanko:
## Summary

Adds a server-side multi-turn chat path to chartsearchai so the AI panel can carry conversation context across turns. Existing `/search` endpoints stay single-shot and unchanged; `/chat` is new.

The design follows the prompt-cache rule that Anthropic, OpenAI, and llama.cpp all converge on in their caching docs: **static content at the top of the prompt, variable content at the end.** The patient chart is the large per-session static block, so it lives in a stable prefix (byte-identical every turn of a session) rather than the variable tail:

```
[ system,
  user(chart envelope — frozen on session create),
  ...prior user/assistant pairs...,
  user(current question) ]
```

A naive design that puts the chart in the current-user message would both miss the cache on every turn AND would force priors to be stripped of the chart to avoid duplication, which breaks the LLM's view of the conversation. This shape avoids both failure modes — verified against LM Studio (`cache_prompt=true`): turn 3's "how many medications did you list?" answered "11" only by reading turn 1's assistant message, with backend logs showing priors flowing and `chart_chars` byte-stable across turns.

## What's new

**Schema** (two changesets — kept separate for already-deployed downstream instances):
- `chartsearchai-007`: `chartsearchai_chat_session` + `chartsearchai_chat_message` tables. Per-`(patient, user)` thread header; ordered turns with `parent_message_id` + `is_summary` reserved for future branching / summarization; assistant rows link back to the existing `chartsearchai_audit_log` via `audit_log_id`.
- `chartsearchai-008`: `chart_snapshot` + `chart_mappings_json` + `chart_built_at` columns on `chat_session`. Chart is built once on session create (or first `chat()` call on legacy sessions) and replayed verbatim every subsequent turn.

**REST**:
- `POST /ws/rest/v1/chartsearchai/chat` — sync, body `{patient, session?, question}`; returns `{answer, references, session, messageId}`.
- `POST /chat/stream` — SSE variant; emits `X-ChartSearchAi-Session` response header before the stream opens so SPA clients can pin subsequent posts to the same conversation.
- `POST /chat/new?patient=...` — closes current active session, opens fresh one (clinician's "New chat" affordance).
- `GET /chat?patient=...` — hydrate the SPA on mount; returns the active session uuid + ordered message list.

**Code**:
- `ChartBuildingStrategy.buildChartUnfiltered(Patient)` — full chart, bypasses `chartsearchai.embedding.preFilter`. Pre-filter remains the right behavior for `/search` single-shot, but in chat the per-question chart variation would break the prefix-cache contract.
- `LlmInferenceService.buildSessionChart(Patient)` produces the LLM-ready envelope ChatService stores on the session row.
- `LlmInferenceService.chat / chatStreaming` take `chartEnvelope` + `mappings` parameters (sourced from `chat_session.chart_snapshot` / `chart_mappings_json`). No per-turn build.
- `ChatMessages.assembleChat` builds the four-section message array with drop-oldest pair trim on the conversation tail only (chart + system reserved up front along with a response budget).
- `LlmEngine` gains `infer(ArrayNode, int)` / `inferStreaming` overloads for prebuilt messages; existing `(systemPrompt, userMessage)` forms delegate so the search path is unchanged.
- `AuditLogPurgeTask` extended to purge `chartsearchai_chat_message` rows older than the new `chartsearchai.chat.retentionDays` GP (default 90).

**Config (new global properties)**:
- `chartsearchai.chat.maxContextTokens` (default 30000). Caps the assembled prompt. The default suits typical chartsearchai deployments using LARGE-context models (default `chartsearchai.llm.contextSize` is already 32768). A too-small value here is a silent footgun — javadoc on the constant documents the failure mode.
- `chartsearchai.chat.retentionDays` (default 90). Clinical-utility retention for chat content; separate from the audit-of-record (`chartsearchai_audit_log`).

## Trade-off

Chat sessions don't pick up new lab results mid-session — the chart snapshot is frozen until the clinician clicks "New chat". This is the deliberate cost of cache-eligibility, and matches the user-facing model (opening a new chat = a fresh context view). Documented in the `ChatService.openOrLoadActiveSession` / `populateChartSnapshot` Javadocs.

## Tests

- `MessagesArrayShapeTest` (7 cases) covers:
  - empty priors → 3-message shape
  - priors sit between chart and current question
  - **first-two-messages byte-identical across turns** (the load-bearing cache-hit invariant)
  - chart never leaks into prior content
  - system always at index 0 even under tight budget
  - drop-oldest pair trim respects budget + response reserve
- RED verified by stubbing `assembleChat` to single-shot — 7/7 fail including the byte-identity test. GREEN restored.
- Full api suite stays at 463 pass.

## Test plan

- [ ] Liquibase changesets 007 + 008 run cleanly against a fresh DB
- [ ] `POST /chat` with no session uuid → creates session, persists user + assistant rows, returns session uuid
- [ ] `POST /chat` with existing session uuid → reuses session, replays priors
- [ ] `POST /chat/stream` emits `X-ChartSearchAi-Session` header before SSE body
- [ ] `GET /chat?patient=` hydrates an existing session
- [ ] `POST /chat/new` closes current session (`status='closed'`, `ended_at` set), opens fresh
- [ ] `AuditLogPurgeTask` purges chat-history rows past retention without affecting audit log rows
- [ ] Multi-turn referential question (e.g. "how many medications did you list?") returns a number derived from the prior assistant turn — proves priors reach the LLM
- [ ] `chart_chars` log line shows byte-stable count across turns of one session — proves prefix-cache eligibility

## Added — refresh chart snapshot (a89b33d)

`POST /chat/refresh-chart` rebuilds the frozen chart snapshot for the active session — picking up clinical data entered since it opened — **without** clearing the transcript (the deliberate counterpart to `/chat/new`). `GET /chat` now also returns `chartBuiltAt` so the SPA can offer "refresh" vs "new chat". Rebuilding changes the chart bytes, invalidating the session prefix cache — the explicit price of freshness. Covered by `ChatServiceRefreshTest` (real ChatService + ChatDAO + a real patient, no mocks).

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/ChartSearchAiConstants.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/ChartSearchAiConstants.java
@@ -239,6 +239,28 @@ public class ChartSearchAiConstants {
 
 	public static final int DEFAULT_AUDIT_LOG_RETENTION_DAYS = 90;
 
+	public static final String GP_CHAT_MAX_CONTEXT_TOKENS = "chartsearchai.chat.maxContextTokens";
+
+	/**
+	 * Default budget for the multi-turn prompt assembly ({@code system} + prior
+	 * turns + current user message with the patient chart). The chart alone is
+	 * routinely ~10K tokens for a real patient, so this default has to be sized
+	 * to the operator's configured {@link #GP_LLM_CONTEXT_SIZE} (default 32768)
+	 * minus the LLM's expected response width
+	 * ({@link #DEFAULT_LLM_MAX_OUTPUT_TOKENS}, default 1024) and some headroom.
+	 *
+	 * <p>A too-small value here is a silent footgun: {@code ChatMessages.fromTurns}
+	 * drops every prior turn when the chart already exceeds the budget, so every
+	 * "chat" turn becomes effectively single-shot and the assistant ignores the
+	 * conversation. Set this to a small value only when running against a
+	 * deliberately tiny model context.
+	 */
+	public static final int DEFAULT_CHAT_MAX_CONTEXT_TOKENS = 30000;
+
+	public static final String GP_CHAT_RETENTION_DAYS = "chartsearchai.chat.retentionDays";
+
+	public static final int DEFAULT_CHAT_RETENTION_DAYS = 90;
+
 	public static final String GP_LLM_ENGINE = "chartsearchai.llm.engine";
 
 	public static final String LLM_ENGINE_LOCAL = "local";
--- a/api/src/main/java/org/openmrs/module/chartsearchai/api/AuditLogPurgeTask.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/api/AuditLogPurgeTask.java
@@ -14,6 +14,7 @@ import java.util.Date;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.chartsearchai.ChartSearchAiConstants;
 import org.openmrs.module.chartsearchai.api.AuditLogService;
+import org.openmrs.module.chartsearchai.api.db.ChatDAO;
 import org.openmrs.scheduler.tasks.AbstractTask;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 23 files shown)


dkayiwa:
There is real product thinking here, and multi-turn chat is a genuine gap (nothing on main handles follow-up questions). I am closing it anyway, for reasons that are about the design's central premise and about what main now requires of every answer path.

The frozen-chart premise is superseded. Main gets the same KV-cache reuse without freezing anything: LlmProvider.buildUserMessage keeps the records section byte-identical across queries for the same patient (the byte-prefix contract is documented at LlmProvider.java lines 638-654; a warmup with an empty question produces a byte-prefix of every real query), and freshness stays event-driven, because ChartSearchEventListener routes chart writes through IndexingHelper.onChartWrite, which invalidates cached answers and re-pins the prewarmed KV (#70, #71). The frozen chart_snapshot gives that up. I searched this PR's full diff and there is no write-driven invalidation for sessions, only the manual /chat/refresh-chart and /chat/new endpoints. Merged as-is, a lab result written mid-session is silently invisible to every later turn of that conversation. For a clinical tool that is the worst kind of wrong: no error, a plausible answer, stale data.

Safety parity would also break the moment this lands. The full diff contains zero references to CitationGroundingVerifier, DrugSafetyValidator or DrugReferenceInjector (all landed after this was cut, and all now run on the /search path). So /chat answers would be ungrounded and drug-safety-unchecked while /search answers are checked, an inconsistency clinicians cannot see.

Mechanically it no longer fits either: buildChartUnfiltered calls chartSerializer.serialize(patient) against ChartBuildingStrategy fields that were removed in the querystore migration, and GitHub reports the branch conflicting.

Recording what I want to keep from this design for when chat is prioritized: the session/message schema (ordinal-ordered messages, the audit_log_id bridge, finish_reason for aborted streams), the drop-oldest-pair trimming with a response-token reserve in assembleChat, the LlmEngine messages-array overloads, the X-ChartSearchAi-Session SSE header, and a chat retention purge separate from the audit log. A fresh design should sit on per-question retrieval plus the existing prefix-cache machinery, with grounding and drug safety on the chat path from day one.

pmanko:
Closing as superseded by #157, which rebuilds multi-turn chat as ConversationService / ClinicalConversation with /chat/new, GET /chat and /chat/stream, persisted per provider. The frozen-session chart in a stable prompt prefix is not carried: prompt-prefix reuse is deferred in the dual-provider roadmap and would return as its own change.

PR #21, opened 2026-05-18
fix(activator): provision privileges + role bindings on every startup

pmanko:
## Summary

- Activator's `started()` now idempotently ensures the two chartsearchai privileges exist AND binds them to standard admin roles (`System Developer`, `Privilege Level: Full`, `Organizational: System Administrator`).
- Adds `ChartSearchAiModuleActivatorTest` covering: priv-wipe-then-restart, role-binding, and double-call idempotency.

## Why

OpenMRS auto-creates module-declared privileges from `config.xml` **only on initial install**. If the `privilege` table is later wiped — by a demo-data SQL dump predating the module, a backup restore, a fresh clone+seed, or any other DB reset — subsequent module restarts will NOT re-create them. The SPA's privilege-gated extensions then silently disappear with no error path.

We saw this on both the harness and a cloud deploy: `data/large-demo-data-2-7-0.sql` truncates and repopulates `privilege` + `role_privilege`, wiping the chartsearchai entries. The AI chat button vanished. Manual SQL inserts worked around it but don't survive future reseeds.

Without the role-binding piece, `System Developer` admins also can't see chartsearchai extensions in the SPA: SD is a backend super-role that bypasses `Context.requirePrivilege()`, but the REST `/session` response only enumerates **explicitly bound** privileges, so the SPA's `userHasAccess()` returns false.

The activator is now the source of truth for chartsearchai's privilege set. Demo-data dumps and seed scripts stay vanilla; this module owns its own onboarding.

## Test plan

- [x] `mvn -pl api -am test` — 471 tests pass (3 new, 468 prior)
- [x] Wipe `privilege` table → restart module → privilege re-appears
- [x] Run `provisionPrivilegesAndRoles()` twice → no duplicate-key error
- [x] `System Developer.hasPrivilege("AI Query Patient Data")` is true after one call
- [ ] Deploy `.omod` to a running OpenMRS, restart module, confirm SPA shows AI button for admin/Admin123 without manual privilege setup


dkayiwa:
I traced both premises of this fix through the actual platform and SPA code, and neither holds anymore, so I am closing it.

Privilege re-creation already happens on every startup, not only at initial install. In openmrs-core, ModuleFactory.startModuleInternal calls Context.checkCoreDataset() whenever the starting module declares privileges (ModuleFactory.java, around line 713 on current master; the comment there even notes that it duplicates work done at initial app startup). checkCoreDataset re-creates any missing privilege via UserService.savePrivilege using OpenmrsUtil.getCorePrivileges(), which merges module privileges back in every time ("always add the module core privileges back on", iterating ModuleFactory.getPrivileges()). So a demo reseed that truncates the privilege table self-heals at the next restart, which is the same timing this activator change runs at.

The System Developer visibility premise is stale too. esm-core's access check ends with `return userHasPrivilege(requiredPrivilege, user) || isSuperUser(user)`, and isSuperUser looks for the System Developer role (packages/framework/esm-api/src/current-user.ts). Our SPA button gates through exactly that path (routes.json declares `"privilege": "AI Query Patient Data"`). The backend mirrors it: User.hasPrivilege short-circuits on isSuperUser(). So SD admins see the button and pass API checks with zero explicit role bindings.

What is genuinely left is role_privilege bindings for non-superuser roles after a dump reseed, and that is currently a deliberate, documented setup step (README section 6). If we automate it later it should be a fresh, small change that derives the privilege list from config.xml or the constants rather than hardcoding it (this branch provisions two privileges and misses Manage AI Prewarm, which #70 added), and granting AI Query Patient Data to Organizational: System Administrator is a security decision to make explicitly, not as a startup side effect.

For transparency about scope: of the seven commits here, only 72c6a11f is this fix. The rest are the content of #18, #19 and #20, each of which I am handling on its own PR.

pmanko:
Closing. #157 declares its privileges in config.xml (AI Query Patient Data, View AI Audit Logs, Manage AI Prewarm) and does not re-provision them or bind roles at module startup; role binding is a deployment concern and the evaluation setup handles it outside the module. Not carried into #157, by decision.

PR #22, opened 2026-05-18
feat(api): ModelSwitchService + REST endpoints for runtime model swap

pmanko:
## Summary

- New `ModelSwitchService` discovers models from the active remote endpoint's `/v1/models` and updates the `chartsearchai.llm.remote.modelName` GP.
- Two new REST endpoints alongside the chat APIs: `GET /chartsearchai/models`, `POST /chartsearchai/model`.
- Validates `modelName` membership against the live list before writing the GP, so switches can't route to a model the endpoint doesn't actually serve.
- Both gated by `AI Query Patient Data` (same as chat).

## Why

POC for an inline model picker in chartsearchai-esm (paired ESM PR coming next). Today swapping models requires a SQL update; this gives the SPA a path to do it in one click.

## Behavior

- `GET /chartsearchai/models` → `{engine, current, available[], endpointUrl}`. `503` when `engine=local`.
- `POST /chartsearchai/model` body `{"modelName": "<id>"}` → `{current}`. `400` for unknown id, `503` for local engine.

## Test plan

- [x] `mvn -pl api -am test` — 477 pass (6 new in `ModelSwitchServiceTest` for parse + URL derivation)
- [x] Local curl smoke: `GET /models` returns 8 LM Studio models; `POST /model` switches GP both directions; unknown id returns 400 with full available list
- [ ] Pair with the esm PR for full SPA path

## Out of scope (follow-up)

- Operator-curated allowlist with display names + per-model context sizes
- Auto-warmup of newly-selected model (handled out-of-band by harness `chartsearch-warmup.sh`)
- Anthropic / OpenAI-vendor filtering (today returns whatever the endpoint reports, including embedding models)


dkayiwa:
I verified what main already does and traced how these endpoints behave against core's authorization, and I am closing this.

Runtime model switching already works through global properties, with no restart and no new surface. LlmProvider.getActiveEngine() reads chartsearchai.llm.engine on every call (LlmProvider.java lines 831-839 at 5223f92), RemoteLlmEngine reads endpointUrl and modelName per request (both infer and inferStreaming), and LocalLlmEngine.ensureServerRunning() detects a changed GP and restarts llama-server (the "LLM config changed ... restarting server" path around line 724). That is exactly how we performed the MedGemma-4B and Llama 3.2 3B swaps: update the GP, and the next request uses the new model.

The genuinely new endpoint has an authorization problem I followed end to end. POST /model gates on AI Query Patient Data (the per-patient chat privilege) and then calls AdministrationService.setGlobalProperty, which core annotates @Authorized(MANAGE_GLOBAL_PROPERTIES). APIAuthenticationException extends APIException, and this controller catches APIException and returns 503. Merged as-is, every clinician-level attempt to switch models dies with a misleading Service Unavailable, and the only users the endpoint actually works for (admins) already have the GP admin page. A per-patient privilege guarding a system-wide model change is also the wrong trust boundary even if the status mapping were fixed. GET /models additionally reveals the endpoint URL to any chat-privileged user.

Two more reasons stack on top. The feature is remote-only by design (503 when engine=local), which builds REST surface for the fallback path while the project's hard requirement is the bundled local engine. And the tests drive the static helpers (deriveModelsUrl, parseModelIds) with hand-crafted strings rather than the real service path, which the contributor rules in CLAUDE.md do not allow.

Finally, most of this branch (roughly 3.2k of the 3.6k added lines) is the content of #18, #19, #20 and #21 riding along, each of which I am handling on its own PR, and it conflicts with post-querystore-migration main. The one idea I am keeping a note of is setCurrent validating the requested model against the live /v1/models list before persisting. If we ever want a model picker for remote deployments, that idea behind an admin-level privilege is a small fresh PR on current main.

pmanko:
Closing as superseded by #157. Runtime model selection lives in the provider boundary there: /providers lists bundled and hub, /models relays med-agent-hub product profiles (HubProfileService), and a conversation is bound to one provider. A runtime swap of the bundled engine's model and the RequestLlmOverride per-request override are not part of the dual-provider design; the bundled engine is configured through global properties.

PR #25, opened 2026-05-29
feat(model-switch): probe LM Studio /api/v1/models with OpenAI fallback

pmanko:
Probe LM Studio's `/api/v1/models` (REST v1) first; fall back to OpenAI-compat `/v1/models` for any other endpoint. Surfaces per-entry loaded state + display name + max context length, filters `type=embedding` out.

Stacks on #22; base set to `feat/model-switch-service`. Will rebase onto main after #22 merges.

Adds:
- `ModelEntry` + `AvailableModels` types; `ModelListResponse` gets optional `provider` + `entries[]` (back-compat)
- `deriveLmStudioV1ModelsUrl`, `looksLikeLmStudioV1Response`, `parseLmStudioV1Models`, `fetchAvailable`, `requestModelLoad`
- New `POST /model/load` REST endpoint (pre-load on picker select)
- Protected test seams: `httpGet`, `httpPostJson`

Tests: 16 new in `ModelSwitchServiceTest` (22 total). Full API suite green. TDD: red → green.

Paired frontend PR follows on `openmrs-esm-chartsearchai`.

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/api/impl/ModelSwitchService.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/api/impl/ModelSwitchService.java
@@ -57,8 +57,97 @@ public class ModelSwitchService {
 	private HttpClient httpClient;
 
 	/**
+	 * One model entry as surfaced to the picker. When the source is LM Studio's
+	 * /api/v1/models, all fields are populated; when the source is the
+	 * OpenAI-compat /v1/models fallback, only {@code id}/{@code displayName}
+	 * are meaningful, {@code type} defaults to "llm", {@code loaded} to false,
+	 * and {@code maxContextLength} to null.
+	 */
+	public static final class ModelEntry {
+
+		private final String id;
+
+		private final String displayName;
+
+		private final String type;
+
+		private final boolean loaded;
+
+		private final Long maxContextLength;
+
+		public ModelEntry(String id, String displayName, String type, boolean loaded,
+				Long maxContextLength) {
+			this.id = id;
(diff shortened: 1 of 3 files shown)


dkayiwa:
Closing this together with #22: it bases on that branch, so it has no home once #22 closes, and no CI ever ran here.

On direction, this makes the module LM Studio-aware specifically (probing the proprietary /api/v1/models, plus a server-side pre-load call). The remote engine is deliberately a generic OpenAI-compatible client, and the strategic path for this module is the bundled local llama-server, so first-class desktop-app vendor coupling points us the wrong way.

There is also a concrete hole I verified in the diff. requestModelLoad checks that the provider looks like LM Studio and that the name is non-empty, but unlike setCurrent it never validates the requested model against the probed list before POSTing `{"model": <caller input>}` to /api/v1/models/load. Combined with the endpoint's AI Query Patient Data gate, any chat-privileged user could make the OpenMRS server fire arbitrary-name load requests at the inference host, and a handful of repeated large-model loads is enough to exhaust memory on that box and take inference down for everyone. The JSON body is also assembled by string concatenation escaping only double quotes. The tests stub the httpGet/httpPostJson seams, so none of this behavior is exercised for real.

The LM Studio response-shape details this encodes (key, type, loaded_instances, max_context_length) stay documented in this PR if we ever need them.

pmanko:
Closing. This depends on #22's ModelSwitchService, which #157 replaces with the provider registry and the med-agent-hub profile relay, and LM Studio is no longer a target of the stack (llama.cpp's router serves the local models). Superseded by #157.

PR #34, opened 2026-06-08
Citation grounding verification for chart-search answers

dkayiwa:
## What

Adds **citation grounding verification**: after the LLM answers a chart-search query and cites records, this checks that each cited record actually *supports* the answer sentence citing it — rather than trusting the LLM's citations blindly. Today the system only validates that a `[N]` marker maps to a real retrieved record; it cannot catch a real record cited for a claim it does not support (e.g. "patient has cancer [5]" where record 5 says a *relative* had cancer).

Both layers are **opt-in and default-off**, so behavior is unchanged unless an admin enables them.

## How

Two tiers, gated by global properties:

- **Tier-1 — semantic overlap** (`chartsearchai.grounding.enabled`): embeds each cited record and the answer sentence(s) that cite it and requires cosine similarity to clear a tunable floor (`chartsearchai.grounding.minCosine`, default `0.40`). Reuses the existing `EmbeddingProvider` + `cosineSimilarity` — no new model, no network.
- **Tier-2 — entailment** (`chartsearchai.grounding.entailment.enabled`): a short yes/no LLM fact-check whose verdict is authoritative, catching subject/polarity flips ("patient has X" vs "mother had X") that cosine cannot. Capped at 16 calls per answer.

The verdict is **annotate-only** (never rewrites prose or drops citations): each `RecordReference` carries a `grounded` value (`true`/`false`/`null`=unverified), surfaced on the REST response and SSE `done` event. `null` must render as "unverified", never "verified".

Reconciled with the reasoning-first response schema: the entailment verdict is read from the structured `answer` field (not the raw reply, whose reasoning routinely contains "no"), and a verbose answer containing **both** YES and NO is treated as undecidable so grounding falls back to Tier-1 rather than guessing.

## Safety / degradation

- Disabled, or any verifier failure (missing text, embedding/LLM error), degrades to a `null` verdict — today's exact behavior. The verifier never throws into the search path.
- GP readers fail safe to defaults when no admin service is available.
- Router cache key now includes the grounding GPs, so toggling grounding busts the cache instead of serving stale verdicts.

## Tests

`CitationGroundingVerifierTest` (Tier-1 on/off-topic, per-sentence, negative-cosine, newline split, Tier-2 override/rescue/cap/degrade), `LlmProviderEntailmentTest` (verdict parsing incl. both-words guard), `PatientChartSerializerTest` (record text reaches the mapping), and `ChartSearchServiceRouterTest` (cache-key segments). All green; `api` and `omod` compile clean.

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/ChartSearchAiConstants.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/ChartSearchAiConstants.java
@@ -298,13 +298,26 @@ public class ChartSearchAiConstants {
 
 	/**
 	 * Minimum cosine similarity between a cited record's text and the answer
-	 * sentence that cites it for the citation to count as grounded. Tunable
-	 * because the right floor depends on the embedding model. The default is
-	 * deliberately permissive: this Tier-1 check is meant to catch grossly
-	 * off-topic citations (a record about blood pressure cited for a diabetes
-	 * claim), not subtle subject/negation flips ("patient has X" vs "mother had
-	 * X"), which embedding cosine cannot reliably separate and which need an
-	 * entailment/NLI second pass.
+	 * sentence that cites it for the citation to count as grounded. This Tier-1
+	 * check catches grossly off-topic citations (a blood-pressure record cited
+	 * for a diabetes claim), not subtle subject/negation flips ("patient has X"
+	 * vs "mother had X") — those need the Tier-2 entailment pass.
+	 *
+	 * <p><strong>The right floor depends on the embedding model — tune it.</strong>
+	 * The {@link #DEFAULT_GROUNDING_MIN_COSINE} of {@value #DEFAULT_GROUNDING_MIN_COSINE}
+	 * suits a wide-spread model like all-MiniLM-L6-v2 (chartsearchai's own default).
+	 * It is far too low for <em>e5</em> — the model querystore uses, which the
+	 * grounding verifier reuses when {@code chartsearchai.querystore.enabled=true}.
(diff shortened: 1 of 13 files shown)


dkayiwa:
**Verified clean against `main` (through #33).**

`main` has advanced 5 commits since this branch's merge-base (#29–#33, including the "thinking" SSE event and the citation-wiring stub fix), in files this slice also touches (`LlmProvider`, the REST controller, `LlmInferenceService`, the wiring test). I test-merged `main` into this branch and confirmed:

- **Textual merge is clean** — no conflicts.
- **Merged api compiles**, and **96 tests pass together** — grounding (16), entailment (9), `LlmProviderTest` (36, incl. the thinking-stream tests), citation-wiring (2), router cache-key (5), serializer (1), citation eval (27).
- **`searchStreaming` composes correctly:** the merged impl implements both the 4-arg (delegates with a no-op reasoning consumer) and the 5-arg overloads, and grounding is applied on the streaming path either way.
- **SSE channels are orthogonal:** the new `thinking` event is its own channel; this PR's `grounded` verdict rides in the `done` event's references — both present after merge.
- **`extractResponse`** (which `parseEntailmentVerdict` depends on) is unchanged by #29–#33.

No rebase or conflict resolution is needed at merge time.

dkayiwa:
**Paired frontend PR:** openmrs/openmrs-esm-chartsearchai#13 renders the `grounded` verdict this PR adds (green "Verified" / red "Unsupported" chips + a ⚠ on unsupported inline citations).

The two are independent and degrade gracefully on their own: this backend works without the frontend (the field is just unused), and the frontend renders nothing when `grounded` is absent/`null`. Suggested review order: this PR first (defines the `grounded` contract), then #13.

PR #48, opened 2026-06-14
Invalidate the answer cache on chart writes so it is safe to enable

dkayiwa:
## What

Makes the chart-search **answer cache** (`ChartSearchServiceRouter`) invalidate on chart writes, so it is safe to turn on without serving stale answers after a patient's chart changes.

The cache was TTL-only: with `chartsearchai.cacheTtlMinutes > 0`, a repeated identical question could serve a stale answer for up to the TTL after the chart changed. That made it unsafe to leave on, so it shipped off by default.

## How

Hook cache eviction into the **existing AOP write-detection advice**, which already resolves the affected patient on every clinical write. On any write to `obs`, `encounter`, `order`, `condition`, `diagnosis`, `allergy`, `program`, `medication dispense`, `savePatient` (demographics), or a patient `merge`, the patient's cached answers are evicted via `ChartSearchServiceRouter.invalidatePatient`.

Key design point: **invalidation is ungated by querystore/preFilter** (embedding indexing stays gated). The answer cache sits in front of *every* retrieval backend, so a stale entry would otherwise survive an edit under the default querystore-on / preFilter-off configuration. The all-off hot path is preserved — no patient extraction when neither indexing nor cache is active.

- `ChartSearchService`: default no-op `isCacheEnabled()` / `invalidatePatient()` (alongside the existing `default warmup`).
- `ChartSearchServiceRouter`: implements both; `invalidatePatient` evicts every entry under the `patientUuid + "::"` key prefix (synchronized, like the other cache ops).
- `IndexingHelper`: `isAnswerCacheEnabled()` / `invalidateAnswerCache()`, resolving the router via the `ChartSearchService` interface (no `api -> impl` coupling).
- The three AOP advice classes restructured; `savePatient` added because demographics (age, sex) are part of the serialized chart the LLM cites.

**`cacheTtlMinutes` default stays `0` (off).** This only makes turning it on safe. Invalidation covers service-layer writes; direct-DAO / SQL-load / outbox-bypass writes still rely on a finite TTL as a backstop.

## Testing

- Unit: `ChartSearchServiceRouterTest` (patient-scoped eviction via the real `search()` path; null/unknown no-ops) and `PatientDataIndexingAdviceTest` (`savePatient` extraction). Targeted suite green.
- End-to-end on a standalone (REST, SSE streaming, and the browser SPA): caching a question, editing the chart, and re-asking returns a recomputed answer reflecting the edit instead of the stale cached one; `savePatient` demographics edits invalidate as well.


dkayiwa:
Note: this PR does **not** change the shipped default — `chartsearchai.cacheTtlMinutes` stays `0` (caching off). It only makes the cache *safe* to enable by invalidating on chart writes. Operators opt in by setting it > 0; flipping the default to a finite value can be a separate decision.

Issue #51, opened 2026-06-16
Migrate to querystore as the shared read store — remove chartsearchai's duplicated indexing/storage/sync layer


dkayiwa:
## Goal

Migrate chartsearchai onto **[querystore](https://github.com/openmrs/openmrs-module-querystore)** as the shared CQRS read store, and **remove chartsearchai's own indexing / storage / sync layer** — the machinery that now duplicates querystore. chartsearchai built this layer before querystore existed; maintaining a second copy doubles the code, the operational surface, and the bug surface (it has already drifted — see "Why now" below).

## What to remove (now provided by querystore)

The read-store/indexing/sync layer:

- **Indexers** — `ElasticsearchIndexer`, `LuceneIndexer`, `EmbeddingIndexer`
- **Indexing AOP** — `PatientDataIndexingAdvice`, `EncounterIndexingAdvice`, `ObsIndexingAdvice` (per-service save/void/merge hooks)
- **Backfill** — `EmbeddingIndexTask`
- **Document serialization** — the `serializer/` package (`PatientRecordLoader`, `*TextSerializer`, `PatientChartSerializer`)
- **Search execution** — `HybridRetriever` / the BM25+kNN RRF wiring (querystore's backend does hybrid search)
- **Embedding pipeline** and the **ES/index connection + config**
- **Chart cache** — `ChartCache` / `ChartCacheInvalidator` (querystore keeps documents fresh via core's #6084 events; the cold-touch + reindex paths replace lazy-index-on-access)

(Authoritative component-by-component mapping lives in querystore's [`docs/chartsearchai-port-map.md`](https://github.com/openmrs/openmrs-module-querystore/blob/main/docs/chartsearchai-port-map.md) and the gap analysis in [`docs/migration-chartsearchai.md`](https://github.com/openmrs/openmrs-module-querystore/blob/main/docs/migration-chartsearchai.md).)

## What to keep (chartsearchai's actual reason to exist)

The consumer / AI layer, re-pointed at querystore for retrieval:

- The LLM Q&A orchestration (question → retrieve patient records **from querystore** → build prompt → answer)
- The `"AI Query Patient Data"` privilege and access gating
- The frontend (`openmrs-esm-chartsearchai`)

## Why now / benefits

- **Single source of truth.** Indexing, embedding, hybrid search, and write-sync live once, in querystore.
[...]

PR #63, opened 2026-06-17
Land the querystore retrieval-parity eval as a standing regression gate (#51)

dkayiwa:
Brings the Phase-0 querystore retrieval eval onto `main` as an **opt-in retrieval regression gate**. It had been stranded on the local `phase0-evals` branch (never pushed), written against pre-migration `main`; this adapts it to the querystore-only codebase. CLAUDE.md and the migration doc already point at it as the retrieval gate — this makes that real.

## Landed

- **`QueryStoreRetrievalParityEvalTest`** — runs querystore's *real* retrieval (`QueryStoreServiceImpl.searchByPatient` over the embeddable Lucene backend + the all-MiniLM-L6-v2 ONNX model) across the shared test datasets and scores it against the frozen 485-case baseline (`eval/enriched-retrieval-eval.json`, already on `main`). **Opt-in**: skips unless the model+vocab are present (`-Dchartsearchai.embedding.model.dir=...`), so normal CI builds don't need the model. Adapted: repointed the `SerializedRecord` import (Phase 5 made it top-level) and reframed the javadoc — the migration is done, so this is now a standing regression gate (querystore must keep reproducing the baseline), not the pre-deletion decision input.
- **`RelevanceCutoff`** — the eval's adaptive-cutoff helper. Landed as **test source** (its only user is the eval): shipping an unwired retrieval-scoring class in production would contradict the migration's "retrieval is querystore's; don't reintroduce it" rule.
- **`TestDatasetHelper`** — re-added the `MODEL_DIR`/`MODEL_PATH`/`VOCAB_PATH` model-locator constants (removed in Phase 4) the eval needs to stand up querystore's embedder.

## Not landed (deliberately)

**`QueryStoreContentParityEvalTest`** — it compared querystore's document content against chartsearchai's `PatientRecordLoader.loadAll`, which **Phase 5 deleted**. querystore is now the sole content producer, so the content-parity comparison has no target; the eval's one-time pre-deletion purpose is fulfilled and it cannot be ported without resurrecting the loader.

## Verification

- `mvn -pl api test` → **471 run, 0 failures, 0 errors, 38 skipped** (the eval's 4 cases skip without the L6-v2 model; couldn't run them locally — only L12-v2 / a cross-encoder are present, and a different model would mismatch the baseline).
- The omod does **not** ship `RelevanceCutoff` (test-scoped).

Net: 3 files (2 added test files + a TestDatasetHelper addition). The baseline JSON was already on `main`.

### DIFF
--- a/api/src/test/java/org/openmrs/module/chartsearchai/api/impl/QueryStoreRetrievalParityEvalTest.java
+++ b/api/src/test/java/org/openmrs/module/chartsearchai/api/impl/QueryStoreRetrievalParityEvalTest.java
@@ -1,0 +1,805 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.chartsearchai.api.impl;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.time.LocalDate;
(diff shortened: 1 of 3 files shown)


dkayiwa:
Closing without merging. This eval was a migration-verification artifact: it pins all-MiniLM-L6-v2 to reproduce the frozen old-pipeline baseline, but production querystore uses e5-base-v2 — so it measures querystore's retrieval *machinery* against the historical model, not production (e5) retrieval quality. Its migration purpose (proving querystore reproduces the old pipeline before deletion) is fulfilled by #57–#60. A standing retrieval-quality gate for the querystore/e5 path belongs in openmrs-module-querystore (it owns retrieval + the e5 model + can build an e5 baseline), not here. Dropping the branch.

PR #70, opened 2026-06-22
Resumable KV-prefill prewarm bootstrap with eviction-exempt pinning

dkayiwa:
## What & why

Cold full-chart prefill is the dominant first-query latency cost — measured **10–20s even on GPU**, ~**15.75s median** saved by a warm KV (harness: `eval/latency/kv_prefill_gate.py`). The reactive chart-open warmup already absorbs this for patients a clinician opens, but a first query on a **never-opened** patient still pays it. This adds an **opt-in** background bootstrap that pre-fills and **pins** every patient's chart KV so even that first query lands warm.

## How it works

**Eviction-exempt pinning.** A pinned `.bin` carries a zero-byte `<name>.bin.pin` sidecar. `evictOldestKvEntries` excludes pinned files from both the count and the delete list, so the LRU cap (`kvCacheMaxEntries`) governs only the ad-hoc warmup/query pool while the prewarm corpus stays durable and unbounded — the right behaviour for hosts with disk for every patient. `purgeKvScopeEntries` deletes the sidecar with its superseded `.bin`. The `pin` flag threads `ChartSearchService.warmup(patient, true)` → router → `LlmInferenceService` → `LlmProvider` → `LlmEngine.warmup(…pin)` → `LocalLlmEngine` as additive overloads (default `pin=false`, so existing callers are untouched), pinning on both the cold-prefill and disk-restore paths.

**Resumable sweep.** `PrewarmBootstrapService` warms patients one by one on a daemon thread, persisting a cursor to `<appdata>/chartsearchai/prewarm-progress.json` so a crash/restart resumes rather than re-prefilling from scratch. Patient enumeration is a batched, id-only SQL cursor (no `Patient` load). `throttleMs` paces the single inference slot so live traffic interleaves; optional `kvCache.maxPinnedEntries` bounds the pinned corpus.

**REST.** `POST /chartsearchai/prewarm` `{scope, action: start|restart|stop}` and `GET /chartsearchai/prewarmstatus`, both behind a new **Manage AI Prewarm** privilege. Only `scope=all` is implemented; other scopes are rejected with 400. The activator resumes the sweep on startup when `prewarm.autostart` is set (best-effort, in a broad catch so it can never break module startup). **All GPs default off.**

## Testing

- Unit: pin/eviction/sidecar tests; 5 sweep/cursor/resume tests; endpoint scope-validation test. **498 api + 42 omod, all green.**
- Live (standalone): gating, trigger, progress, stop, resume-from-cursor, completion 50/50, **eviction exemption (50 pins ≫ cap 16)**, autostart, and the bounded-cap reader path.

## Known follow-up

With a large pinned corpus, `evictOldestKvEntries` scans the whole KV dir on every ad-hoc save (an `.exists()` stat per `.bin`). A directory-layout split (pinned subdir) is the fix **if** anyone runs the corpus at scale — deferred because it changes the on-disk layout already verified here.

### DIFF
--- a/ONBOARDING.md
+++ b/ONBOARDING.md
@@ -1,0 +1,83 @@
+# Chart Search AI — Onboarding
+
+An OpenMRS module that lets clinicians ask natural-language questions about a patient's chart and get answers with source citations. A local (or remote) LLM answers over a chart assembled by the **querystore** module, which owns retrieval and embeddings.
+
+## Get oriented fast
+
+- **Read first:** `CLAUDE.md` (project rules — TDD, root-cause-over-patch, and the **API-surface rules** below), `README.md` (setup + platform notes), `docs/adr.md` (decision log).
+- **Retrieval lives in querystore, not here.** chartsearchai has no in-process embedding/Lucene/scoring pipeline. Retrieval and citation grounding both use querystore's e5-base-v2 model. Retrieval changes and retrieval-quality eval belong in `openmrs-module-querystore`.
+- **Two engines:** `chartsearchai.llm.engine` = `local` (bundled `llama-server` subprocess, default — data stays on the host) or `remote` (OpenAI-compatible API). A **local** LLM is a hard requirement for production.
+
+## Build & test
+
+```bash
+mvn install                       # full build (api + omod), produces omod/target/chartsearchai-*.omod
+mvn -pl api test                  # api unit tests
+mvn test                          # api + omod
+mvn -pl api test -Dtest=ClassName # one test class
+```
+
+Tests must call the **real production pipeline** with real datasets — no mocks/reimplementations of pipeline logic (see `CLAUDE.md`). Follow TDD: write the failing test first.
+
+## Run it locally
+
+A configured standalone lives at `test/referenceapplication-standalone-3.7.0-SNAPSHOT` (RefApp 3.7, port **8081**, login `admin` / `Admin123`):
+
+```bash
+cd test/referenceapplication-standalone-3.7.0-SNAPSHOT
+nohup java -jar openmrs-standalone.jar &     # nohup so it survives the shell
+# REST base once up: http://localhost:8081/openmrs/ws/rest/v1
+```
+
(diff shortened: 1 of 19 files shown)


dkayiwa:
🙏 **Requesting review.**

This is opt-in and **default-off** (`chartsearchai.prewarm.*` GPs), so it's inert until explicitly enabled — but the parts most worth a careful look:

- **`LocalLlmEngine` pinning** — `evictOldestKvEntries` excluding pinned `.bin` files (the `<name>.bin.pin` sidecar), and `purgeKvScopeEntries` cleaning the sidecar on supersede. This is the eviction-exemption core; please sanity-check the filter logic and the order in `persistKvEntry` (pin written before the cap is applied).
- **`PrewarmBootstrapService`** — the resumable cursor + the single-pin-writer reasoning behind the `maxPinnedEntries` cap being a hard ceiling (the sweep is the only caller of `warmup(patient, true)`).
- **Activator autostart** — runs in `setDaemonToken`, best-effort in a broad catch so it can't break module startup.

Tests: 498 api + 42 omod green; endpoints, autostart, eviction-exemption (50 pins ≫ cap 16), and the bounded-cap path were verified live on the standalone.

One **deferred follow-up** flagged in the description: with a large pinned corpus, `evictOldestKvEntries` scans the whole KV dir on every ad-hoc save — a directory-layout split is the fix if anyone runs the corpus at scale. Happy to address it here if reviewers prefer.

dkayiwa:
📖 Added an **ONBOARDING.md** guide in `b083ef4` (now part of this PR).

It covers module orientation, build/test, running the standalone, the API-surface rules, and a **REST endpoints reference** — including the new prewarm endpoints:

- `POST /chartsearchai/prewarm` `{scope, action: start|restart|stop}` and `GET /chartsearchai/prewarmstatus` (both behind **Manage AI Prewarm**), with the resumable-cursor + pinned/eviction-exempt KV behaviour explained.
- The related global properties (`prewarm.enabled` / `.autostart` / `.throttleMs`, `kvCache.maxPinnedEntries`) and the corpus-erosion caveat.
- The `/search/stream` SSE event types listed exactly: `preliminary`, `thinking`, `token`, `references`, `grounded`, `done`, `error`.

Every endpoint, method, privilege, GP name/default, and status-field name in the guide was fact-checked against the code.

PR #72, opened 2026-06-26
Consume querystore behind a QueryStoreClient seam

pmanko:
A thin client contract between the chart-build and querystore so the retrieval source is swappable (in-process now, HTTP later) without touching the byte-producing serializer.

- `QueryStoreClient` (getPatientChart / searchByPatient) over a `QueryRecord` DTO (no embedding).
- `InProcessQueryStoreClient` (default) wraps `Context.getService(QueryStoreService.class)`.
- `QueryStoreChartBuilder` resolves the client via the registered component; `QueryRecord` throughout; all timing / preFilter / degrade-to-empty logic unchanged.

Behaviour-preserving — the chart bytes are unchanged. `QueryStoreChartBuilderTest` (17) green (identical mapping + focus uuids + obs-group pass-through + degrade-to-empty).


dkayiwa:
I spent real time on this one because a transport seam is a legitimate idea. I am closing it for three reasons, each checked against current main (5223f92) and this branch's head.

First, the seam's job is already served, deliberately. QueryStoreChartBuilder resolves the service through the protected resolveQueryStoreService() override (line 296), which is the documented test seam and exactly where QueryStoreChartBuilderTest already injects its stub; this PR's test change retargets that same stub onto a new interface. The typed bridge (provided-scope querystore-api dependency plus require_module) is a decision rather than an accident: it turns API drift between the two modules into a compile error on our only retrieval path. This branch itself tried reflective decoupling earlier (74c6912a) and reverted to the typed bridge (6e43b2b0).

Second, the abstraction does not cover today's querystore touchpoints, so the stated goal (switch to HTTP later without chartsearchai changing) does not hold. buildFocused (QueryStoreChartBuilder line 182, added by #69 for progressive reasoning) appears nowhere in this diff, and CitationGroundingVerifier resolves querystore's EmbeddingProvider directly (lines 129-132). A rebase would not compile without redesigning the client around both.

Third, and decisive: merged as-is this branch silently regresses two shipped features.

1. ChartAnswer's constructor chain now ends in `this.safetyWarnings = java.util.Collections.emptyList();`, discarding the parameter. On main, LlmInferenceService passes drugSafetyValidator.validate(...) into exactly that constructor at two sites and the REST layer serializes getSafetyWarnings(), so every drug-safety warning from #41 and #73 becomes an empty list. No exception, no failing test; the warnings just stop appearing in answers.
2. The rewritten LocalLlmEngine keeps kvCacheKey only in the warmup path, while main also uses it in inferStreaming (line 218 at 5223f92). The query-path disk KV cache from #70/#71 quietly disappears and cold-query latency regresses.

Those two are the silent kind of breakage, which is why I am strict here. Add that the branch carries the full fork lineage (chat persistence, an 891-line ModelSwitchService, a blocks output channel, an endpoint registry) rather than the thin seam the title describes, that the new tests are Mockito-based (mock-maker-inline is added) against the no-mocks rule in CLAUDE.md while the one genuinely new production mapping (QueryDocument to QueryRecord) has no test at all, and that GitHub reports the branch conflicting with no CI run on it.

If an out-of-process querystore becomes a roadmap item (for example a REST transport we want to consume), I would welcome a seam-only PR against current main that covers all three touchpoints (build, buildFocused, grounding embeddings) and proves chart bytes identical through the new layer. Multi-turn chat deserves its own conversation with the drug-safety and grounding path preserved; see my note on #20 for the pieces of that design worth keeping.

pmanko:
Closing as superseded by #157. #157 consumes querystore-api directly (QueryStoreChartBuilder uses the typed PatientChartRead and ContextSlice from QueryStore ADR Decision 17) rather than through a QueryStoreClient seam, and it carries the rebuilt conversation model (ConversationService) in place of the ChatService stack this branch bundled. Nothing from this branch is kept separately.

PR #74, opened 2026-07-17
feat(chart): query-scoped slice charts (chartMode=queryScoped)

dkayiwa:
## Summary

Adds a second chart-assembly mode behind the `chartsearchai.chartMode` setting (default `fullChart`, existing behavior unchanged). In **scoped** mode the answer path no longer depends on prefill amortization, collapsing cold-CPU first-answer latency ~5–7× with no pre-warming and no per-patient KV on disk.

Full rationale, evidence, and trade-offs are recorded in `docs/adr.md` **Decision 28**.

## What changes

- **`QueryScopeRouter`** (new) matches a question against conservative, word-boundary cue sets. Enumeration intents (MEDICATIONS, ALLERGIES, PROGRAMS, CONDITIONS, VISITS, ORDERS) include their record types *complete*; a question matching several cue sets ("any drug allergies?") carries the *union* of the matched intents' types, so completeness holds for whichever intent was meant; everything cue-free is TOPICAL (similarity-only), never a wrong typed scope.
- **`QueryStoreChartBuilder`** composes the slice: typed-complete records + querystore top-K similarity (lab abbreviations expanded first, e.g. BMP → basic metabolic panel) + demographics + obs-group family completion + a recency anchor for temporal questions only.
- **Rendering**: slices render in chart order (most recent first) with a date on every record; `patient` and `allergy` records (querystore dates them by dateCreated) render undated in both modes so their record-keeping time is not shown as encounter time. Known remainder (ADR Decision 28): querystore also dates `condition`/`diagnosis` by dateCreated (and `program`/`medication_dispense` when the clinical date is null); those still render their date pending a measured follow-up.
- **Disengagement**: in scoped mode warmup, the prewarm bootstrap sweep, per-patient KV persistence, and the progressive-reasoning preview all turn off at the same `shouldRunWarmup` decision point.

## Evidence (rc.2 standalone, E4B, July 2026 gates)

> Gates re-run 2026-07-17 on the union build (fdf1c9c): 40-cell aggregate unchanged (meanF1 0.800, abstention 1.00, off-topic 2). The temporal probe is now committed (`eval/drift-metric/temporal_probe_rc2.py`): 14/15, all value-recency cells passing; the one miss is the documented type-conflation residual on a newly-covered "last visit" cell. A drug-allergies eval topic was added (45-cell suite: 0.817 / 0.95 / 4; both present cells F1 1.00 via the union). Details in ADR Decision 28's evidence section.

- CPU cold first answer: fullChart 73–74s vs scoped 12–27s; first output ~68s vs 7–12s. Scoped cold ≡ scoped warm.
- 40-cell adjudicated quality: meanF1 0.733 → 0.800, abstention 0.89 → 1.00, off-topic citations 51 → 2.
- Temporal DB-truth probe: scoped 15/15 vs fullChart 14/15.

## Testing

New tests exercise the real querystore-backed pipeline (per the repo's no-simulation rule): `QueryScopeRouterTest`, `QueryStoreChartBuilderScopedTest`, `ChartBuildingStrategyDispatchTest`, `LlmInferenceServiceQueryScopedTest`, `QueryPreprocessorLabExpansionTest`, `WarmupExecutorTest`, plus updated prewarm/router tests. Adds eval harnesses (cold/CPU latency gates, rc.2 drift-metric gold).


dkayiwa:
## `/harden` summary

Ran two harden cycles (Phase 1 structural review + outward tracing, then Phase 2 four-agent polish sweep) on top of the original feature. Net: **one real correctness fix, three behavior-preserving polish edits, one reverted false positive.** Full reactor green throughout — api 575 + omod 42, 0 failures.

### Fix (correctness) — the material change beyond the original feature

**KV-scope eviction on the drug-reference + scoped path.** `DrugReferenceInjector.injectRecords` rebuilt the chart via a fresh `PatientChart`, resetting `queryScoped` to `false`. `searchStreaming` derives its KV-cache decision from `PatientChart.isQueryScoped()` *deliberately* (not a `chartMode` GP re-read) so a mode-flip / GP-read race can't mis-scope the persist. Dropping the stamp on injection silently re-armed exactly that hazard: on the medications path (the flagship scoped intent *and* the likeliest drug-ref match), a mid-request flip could persist a question-dependent slice under the patient's KV scope and evict their pinned full-chart entry.

- Fix: carry the stamp across the reconstruction.
- Tests: two real-injector regression tests (stamp preserved on a scoped chart; never added to a full chart). The existing scoped test masked this because its injector stub was a pass-through.
- Verified root-cause-complete: the codebase has exactly two `new PatientChart(` sites — the serializer's primary construction (always stamped afterward in `buildScoped`, never in fullChart `build`) and this injector — so no other path can drop the flag.

### Polish (behavior-preserving)
- Corrected a false-invariant comment in `QueryScopeRouter.isTemporal` / `buildScoped`: the recency anchor is gated purely on temporal phrasing, independent of typed scope. The old "scope questions must NOT get the anchor" wording described a rule the code never enforced.
- `expandLabPanelAbbreviations`: `Matcher.quoteReplacement` on the panel value, so a future value containing `$`/`\` can't break the replacement.
- Named the ES-tier cap constant `QUERYSTORE_ES_CHART_CAP`, documenting the manual coupling to querystore's `FULL_CHART_MAX_HITS`.

### Reverted (false positive)
An identity-set rewrite of `completeObsGroupFamilies` (to stop it "dropping null-uuid slice records") — `toSerializedRecords` already drops every null-`resourceUuid` doc, so the two differently-keyed predicates have no observable difference. A test proved the record can't survive regardless, so both the change and its test were reverted.

### Verified handled (no change needed)
- Answer cache is collision-safe: `buildCacheKey` folds in both `chartMode` and the question.
- All three inference paths handle the stamp: `search()` does no KV I/O; `searchStreaming()` preserves+gates it; `warmup()` never injects and gates it.
- querystore-absent / getPatientChart-throws / searchByPatient-throws all return a `markScoped(...)` chart.
- Prewarm sweep re-checks mode per-patient (aborts on a mid-sweep flip rather than reporting a false COMPLETED); warmup/refresh/executor all short-circuit in scoped mode.
- The 10,000 ES cap warning matches querystore's actual `FULL_CHART_MAX_HITS` in magnitude and direction.

### Deferred (each with a failure mode, below the action threshold)
- **`"allergy"` literal in two files** (router `typedSlice` + `ADMIN_DATED_TYPES`) — a querystore resourceType rename would need both edited; missing the latter re-renders allergy `dateCreated` as a clinical date. Deferred: sharing only `allergy` while the router's other 9 type strings stay inline (deliberately documented against querystore's serializer contract) is inconsistent, and a full constants holder exceeds a polish pass.
- **`TEMPORAL_CUES` over-match** ("patient's last name?" → temporal → pulls the anchor) — deferred because changing the cue set is a slice-composition change requiring the mandatory 40-cell eval + temporal probe gates (standalone deployment), which the ADR reserves for measured evaluation.
- `isBlank` inlined 5×; redundant `route()` `toLowerCase`; double `metadataString` call per doc — all below the noise floor.

api/src/main/java/org/openmrs/module/chartsearchai/api/impl/QueryScopeRouter.java
```diff
@@ -0,0 +1,148 @@
...
+		return question != null && TEMPORAL_CUES.matcher(question).find();
+	}
+
+	/** Routes a question to its slice intent; null/blank questions are TOPICAL. */
+	static Intent route(String question) {
+		if (question == null || question.trim().isEmpty()) {
+			return Intent.TOPICAL;
+		}
+		String q = question.toLowerCase(Locale.ROOT);
+		if (MEDICATIONS_CUES.matcher(q).find()) {
```
dkayiwa, line 123: I ran the production cue patterns over common allergy phrasings and there's a routing hole here: "any drug allergies?", "does the patient have any medication allergies?", "is she allergic to any medications?", and "any adverse drug reactions?" all match `\bdrugs?\b` or `\bmedications?\b` and return MEDICATIONS before the ALLERGIES check runs. That's a wrong typed scope, not the TOPICAL degrade, so it breaks the invariant this class's javadoc and the PR description state ("never to a wrong typed scope").

If merged as-is, exactly these questions get an allergy answer with no typed completeness: an allergy outside the similarity top-K (default 30) is invisible to the model, so the enumeration can silently omit it, while the complete drug_order and medication_dispense sets flood the slice as enumeration bait. A silently incomplete allergy list is about the most safety-loaded miss this mode can produce. Neither committed eval covers the class (both suites phrase the allergy cell as "Does the patient have any allergies?"), and QueryScopeRouterTest doesn't either, so the 40-cell and abstention numbers say nothing about it. This needs handling before merge.

Reordering ALLERGIES above MEDICATIONS only moves the hole ("what medications is he on for his allergies?" then under-slices medications the same way). The shape I'd go for is unioning the typed slices of every matched cue set when more than one matches: drug_order + medication_dispense + allergy is still a small slice, and completeness holds for whichever intent the user meant. Per Decision 28 that's a slice-composition change, so it has to go through the two gates, and the panel wants a "drug allergies" cell either way so this class stays covered.

api/src/main/java/org/openmrs/module/chartsearchai/api/impl/QueryStoreChartBuilder.java
```diff
@@ -245,6 +469,17 @@ private Set<String> collectFocusUuids(List<QueryDocument> hits) {
...
+	/**
+	 * Record types whose querystore {@code record_date} is ADMINISTRATIVE ({@code dateCreated} —
+	 * see querystore's {@code PatientRecordSerializer}/{@code AllergyRecordSerializer}), not a
+	 * clinical event date. Rendering it as a "(date)" label misleads temporal reasoning: measured
+	 * on the rc.2 standalone, "when was the last visit?" was answered from the allergy record's
+	 * creation date in one mode and the patient record's in the other. These types render undated;
+	 * the date still drives querystore's chart ordering.
+	 */
+	private static final Set<String> ADMIN_DATED_TYPES = Collections.unmodifiableSet(
+			new HashSet<String>(Arrays.asList(PATIENT_RESOURCE_TYPE, "allergy")));
```
dkayiwa, line 513: I checked this set against querystore's serializers and {patient, allergy} doesn't cover the principle stated above: condition and diagnosis also return dateCreated unconditionally as the document date (ConditionRecordSerializer.getDate and DiagnosisRecordSerializer.getDate both return `DateFormatUtil.toLocalDate(x.getDateCreated())`), and program and medication_dispense fall back to dateCreated when dateEnrolled / dateHandedOver is null. So "when was the patient diagnosed with X?", a temporal CONDITIONS question this mode specifically routes and recency-anchors, renders every condition and diagnosis record with a "(date)" label that is really the DB-insert date, and scoped mode shows more of them than fullChart did (compressDateRuns=false dates every record). The model quoting record-keeping time as diagnosis time is the exact failure this constant exists to prevent, and Decision 28's "record-keeping time is never presented as encounter time" overclaims as written. The new build_shouldNotRenderAdministrativeDates test bakes the same premise in: its comment calls the condition record "clinically-dated" and asserts its date renders, which querystore's serializer contradicts.

I don't think blanket-adding condition/diagnosis here is obviously right, though: undated conditions lose chronology entirely, and condition's clinical onset_date lives in querystore doc metadata this module doesn't render, so the measured endpoint (undate them, or render onset instead) belongs behind the same two gates. What I'd fix before merge is the claim: scope this javadoc, the ADR sentence, and that test comment to patient/allergy, naming condition/diagnosis (plus the program/dispense null-date fallbacks) as known remaining administrative dates, and take the measured fix as the follow-up.

eval/latency/quality_panel.py
```diff
@@ -0,0 +1,124 @@
...
+    ("heart", "Does the patient have any heart or cardiac problems?"),
+    ("fractures", "Has the patient had any fractures or broken bones?"),
+    ("kidney", "Does the patient have any kidney problems?"),
+    ("mental", "Does the patient have any mental health or psychiatric conditions?"),
+]
+
+
+def capture(outdir):
+    os.makedirs(outdir, exist_ok=True)
+    rate_uuid, rate_orig = lb.get_gp(lb.GP_RATE)
```
dkayiwa, line 51: This new script can't run as committed: latency_bench defines only `set_gp(name, value)`, there is no `get_gp` and no `GP_RATE`, so this line throws AttributeError, and before it even gets here the `import latency_bench` on line 25 executes the whole benchmark at import time (latency_bench has no `__main__` guard, so its module body fires the 32-cell run and mutates clauseScoped and rateLimitPerMinute on whatever BASE points at). The docstring's "Env: BASE, OMRS_USER, OMRS_PASS as in latency_bench.py" is also off; latency_bench hardcodes admin:Admin123 and reads no env.

Presumably the panel numbers were captured with a working local copy, and that's the version to commit. The fix that doesn't touch the benchmark file: make this script self-contained (own req/get_gp/set_gp) like the sibling prod_*.py scripts.

omod/src/main/resources/config.xml
```diff
@@ -192,6 +192,12 @@
 		<description>Number of top results to request from querystore.searchByPatient when chartsearchai.embedding.preFilter is true, used to build the LLM focus hint. Default: 30.</description>
 	</globalProperty>
 
+	<globalProperty>
+		<property>chartsearchai.chartMode</property>
```
dkayiwa, line 196: The README should get this too: it documents every sibling GP (the llm / warmup / prewarm / kvCache tables), and its "Prompt-stability caveat" section walks the exact trade-off queryScoped resolves, so an operator reading the README today has no way to discover chartMode exists. A row in the GP table plus a line on the warmup/prewarm rows saying they disengage under queryScoped would cover it. Fine as a fast-follow if you'd rather keep this PR's docs to the ADR.

api/src/main/java/org/openmrs/module/chartsearchai/api/impl/QueryStoreChartBuilder.java
```diff
@@ -171,6 +172,177 @@ PatientChart build(Patient patient, String question) {
...
+					patient.getPatientId(), intent, System.currentTimeMillis() - rpcStart,
+					System.currentTimeMillis() - buildStart, e.getClass().getSimpleName());
+			return markScoped(emptyChart(patient));
+		}
+
+		// querystore's ES tier returns at most its 10 000 most recent docs (older tail silently
+		// dropped) — at the cap, "complete by construction" typed slices may be missing
+		// pre-cutoff records, and unlike fullChart (which fails loud on context overflow) the
+		// slice keeps working. Surface it for operators.
+		if (chartDocs.size() >= QUERYSTORE_ES_CHART_CAP) {
```
dkayiwa, line 233: The >= fires this warning on the uncapped tiers too: FULL_CHART_MAX_HITS exists only in querystore's ElasticsearchBackendStore (the MySQL and Lucene findAllByPatient paths return the whole chart, no cap), so a 12k-doc chart on those tiers gets a WARN claiming records may have been silently omitted when nothing was. A size above the cap actually proves the tier is uncapped; equality keeps the ES detection, since the ES tier can never return more than the cap. Cheap to tighten now:

```suggestion
		if (chartDocs.size() == QUERYSTORE_ES_CHART_CAP) {
```

dkayiwa (commented):
The 11 new .bin files under api/src/test/resources/embedding-cache/ have no reader: nothing on this branch or on main references that directory (no Java in chartsearchai, nothing in querystore, only api/pom.xml's generic binary-copy rules), and the 783 files already sitting there date to the pre-#51 in-process retrieval harness whose removal CLAUDE.md records. I checked with a positive control, so the empty search isn't a broken grep. Unless something outside the repo consumes them, this PR adds ~3.3 MB of dead weight to the test classpath; I'd drop the 11 from this PR and treat the legacy 783 as a follow-up cleanup. (Body comment because GitHub can't anchor inline comments to binary files.)

PR #76, opened 2026-07-19
fix(citations): drop the structured array when an answer cites nothing inline

dkayiwa:
## Problem

For patient `b51254aa-…` on the demo, **"Does he have cancer?"** returned the correct answer —

> There is no explicit diagnosis of cancer in the patient records.

— but with **30 references** attached: the patient's entire condition/diagnosis history (Tuberculosis, Fracture of scapula, Alopecia areata…), none of them cancer. It makes no sense to attach references to a "not found" answer, and it actively misleads: a clinician clicking through them would assume they're cancer-related.

## Root cause

`LlmInferenceService.extractCitedReferences(...)` takes the **union** of the LLM's structured `citations` array and the inline `[N]` markers in the prose. A small local model, on an abstention answer, writes **zero** inline markers yet dumps its whole reviewed record set into the structured array — and the pipeline faithfully surfaced all of it.

This violates the model's own contract: the system prompt instructs *"Cite EVERY record you reference by its number in brackets"*, and its few-shot demonstrates that an abstention carries `"citations": []`. Inline `[N]` markers are the authoritative record of what was actually cited.

## Fix

At the single reconciliation chokepoint (feeds both `search` and `searchStreaming`): when the answer is **real prose that anchors no `[N]` marker**, treat the structured array as unanchored and return no references.

Deliberately preserved:
- **Blank/empty answer** (`""`, whitespace) — a distinct degenerate output, not an answer that failed to anchor citations; the array still resolves (matches the existing `citation-empty-answer-with-citations` eval case).
- **`answer == null`** legacy entry point — array resolves alone (existing unit tests).
- **The union whenever any inline marker is present** — even an unmapped typo (`[99]`) still merges the array, so a mistyped *positive* answer keeps its citations. The union's ordering is byte-identical to before.

## Tests

All exercise the real code path — no assertions weakened, no existing test modified.

- `LlmInferenceServiceAbstentionCitationTest` (new) — drives `search()` **and** `searchStreaming()` through the full pipeline with a stub reproducing the exact demo shape (abstention prose + dumped `[8,9]`): asserts **zero** references. A third test guards the `isBlank` distinction (a blank answer still resolves its array).
- `LlmInferenceServiceTest` (added guard) — an answer whose only inline marker is unmapped still keeps its array, locking the design decision against a naive map-aware "fix".

`api` suite: **584 tests, 0 failures.**

## Verified live

Rebuilt and redeployed to the `:8081` standalone:

| Query | Answer | References (before → after) |
|-------|--------|------------------------------|
| Does he have cancer? | *no explicit diagnosis of cancer…* | **30 → 0** |
| Does he have tuberculosis? (positive control) | *Yes, he has Tuberculosis. The condition is ACTIVE [2] and the diagnosis is PROVISIONAL [6].* | 2 (unchanged) |

## Notes / follow-ups (not in this PR)

- **Eval A/B flip gates** (`eval/latency/quality_panel.py`, `prod_reasoning_ab.py`) count any abstention citation flip and gate `flips == 0`. This change *intentionally* flips dumped-abstention cells from non-empty to empty, so a run across this change reports FAIL — re-baseline the `pre_dir` capture rather than treat it as a regression.
- **Known tail:** an abstention answer containing a stray bracketed number that isn't a real citation (e.g. `[2024]`) would bypass the drop. No clean fix without semantic abstention detection (which the design forbids); the added guard test documents why a map-aware "fix" is off the table.

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/api/impl/LlmInferenceService.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/api/impl/LlmInferenceService.java
@@ -465,11 +465,27 @@ public class LlmInferenceService implements ChartSearchService {
 	 * sources of citation indices that can disagree: the LLM's structured
 	 * {@code citations} array and the {@code [N]} markers it writes inline in the
 	 * prose. We take the UNION of both (restricted to indices that map to a real
-	 * retrieved record), so a record the model cited inline but forgot to add to
-	 * the array — or listed in the array but only referenced inline — still
-	 * resolves to a reference. Indices with no matching record are dropped and
+	 * retrieved record), so a record the model cited inline but omitted from the
+	 * array — or one it listed in the array while citing at least one record
+	 * inline — still resolves to a reference. The one exception is the
+	 * abstention-dump carve-out below: an answer whose prose cites nothing inline
+	 * discards the array entirely. Indices with no matching record are dropped and
 	 * logged, exactly as the array path already drops unmapped indices; the
 	 * prose itself is never rewritten.
+	 *
+	 * <p>An inline {@code [N]} marker in the answer is the authoritative record of
+	 * what the model cited: the system prompt instructs it to "Cite EVERY record
+	 * you reference by its number in brackets", and its own few-shot demonstrates
+	 * that an abstention answer carries {@code "citations": []}. A small local
+	 * model breaks that contract by writing an abstention ("no cancer found") with
+	 * no inline markers yet dumping its whole reviewed record set into the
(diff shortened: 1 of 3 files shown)


dkayiwa:
Related: #77 — widens the drift-metric gold from 5 to 22 patients and validates that this citation fix behaves correctly in both chart modes. The 198-cell A/B in that PR confirms `queryScoped` beats `fullChart` (meanF1 0.668→0.748, abstention 0.74→0.86, drift 477→181), and every arm there runs the fix from this PR (abstention answers surface zero references, no dumping, in both modes).

PR #78, opened 2026-07-19
feat(chartMode): make queryScoped the default

dkayiwa:
## What

Flips the `chartsearchai.chartMode` default from `fullChart` to **`queryScoped`**.

## Why (evidence from this session)

- **Quality — 22-patient drift-metric A/B** (see #77 for the widened gold): queryScoped beat fullChart on every axis.

  | Metric | fullChart | queryScoped |
  |---|---|---|
  | meanF1 (104 present cells) | 0.668 | **0.748** |
  | Abstention (94 absent cells) | 0.74 | **0.86** |
  | Off-topic drift | 477 | **181** |

  The focused slice keeps the small local model from citing a whole chart's worth of off-topic records.

- **Latency — CPU A/B**: queryScoped's **cold first answer was ~3× faster** (no full-chart prefill: ~15 s vs ~44 s on a 153-record patient). fullChart is faster only on repeat/varied questions to an already-warmed patient (prefix-cache reuse), so over a multi-question session the totals are ~equal — but the first answer, the one users feel, is much faster scoped.

## Changes

- **`config.xml`** `defaultValue` `fullChart` → `queryScoped` (the primary lever — OpenMRS creates the GP with this value on fresh install) + rewritten description.
- **New `CHART_MODE_DEFAULT` constant** (= `queryScoped`) as the single source of truth; both readers (`PipelineSettings.queryScopedMode` and `ChartSearchServiceRouter`'s cache key) use it instead of independently hardcoding `CHART_MODE_FULL_CHART`, so they can't drift.
- Javadocs + ADR Decision 28 updated with the flip, evidence, and consequences.

## Behavioural consequences (documented, not hidden)

- The full-chart prefill machinery (**warmup, prewarm bootstrap, per-patient KV persistence, progressive-reasoning preview**) is now **dormant by default** — it re-engages only when an operator sets `chartMode=fullChart`.
- **Fail-safe direction reverses:** an *absent or unreadable* `chartMode` GP now resolves to `queryScoped`. A GP set to any non-`queryScoped` value (including a typo) still resolves to `fullChart`, so a mistyped value fails toward the whole chart.
- **Existing installs keep their current GP value** — `config.xml` `defaultValue` only applies at GP creation. This changes the default for **fresh installs**; operators on existing installs flip it manually.

## Tests

- **New `PipelineSettingsDefaultTest`** — exercises the real `queryScopedMode()` with no `Context` (its `getStringGlobalProperty` returns the default via the catch clause, the same value an unset GP yields), asserting the default is queryScoped. Fails on the old default, passes now.
- Three tests of **fullChart-only** features (warmup/KV-scope, progressive preview, router cache key) now **pin fullChart explicitly** (via the `resolveQueryScopedMode` seam / GP) instead of relying on the old default to select it — their behaviour assertions are unchanged.

`api` 582 + `omod` 42 green.

Related: #77 (the widened gold this verdict rests on), #76 (citation fix; both queryScoped and fullChart run it).


dkayiwa:
## CPU first-token latency tradeoff (measured — worth weighing before merge)

The quality A/B favours queryScoped, but a CPU **time-to-first-reasoning-token** measurement (Gemma-4-E4B, `-ngl 0`, streaming endpoint) shows fullChart is faster *in the warmed steady state*, because queryScoped disengages the warmup/prewarm/KV-persistence machinery that keeps fullChart fast:

| Scenario (CPU) | fullChart | queryScoped |
|---|---|---|
| Truly cold patient (no disk KV), 1st question | **~129 s** | **~18 s** |
| Warmed patient (disk KV on disk), any question | **~1–2 s** | ~11–14 s |
| Warm-repeat (exact same question) | ~1 s | ~1 s |

**Interpretation**
- **fullChart is bimodal**: catastrophic cold (must prefill the whole chart on CPU — ~129 s for a 261-obs patient), near-instant (~1–2 s) once the chart KV is persisted (prior query / warmup / prewarm sweep), including for *different* questions (chart prefix restored, only the question suffix is new).
- **queryScoped is uniform** ~12–18 s per new question — it rebuilds+prefills a per-question slice with no cross-question reuse and no disk-KV restore, so no fast warmed path but also no 129 s spike.

**Implication for the default**
- **GPU**: prefill is cheap either way, so queryScoped's quality win dominates — queryScoped default is clearly right.
- **CPU**: it's a genuine tradeoff. A deployment that prewarms (or revisits) patients gets ~1–2 s first tokens from fullChart vs ~12–18 s from queryScoped; a deployment that hits cold patients without prewarm gets ~18 s from queryScoped vs ~129 s from fullChart (and CPU prewarm is ~2 min/patient, impractical at scale).

This doesn't block the change, but the default flip trades a fast **warmed-fullChart** first-token (~1–2 s) for queryScoped's uniform ~12–18 s on CPU. Options to consider: keep queryScoped default but document the CPU caveat; or make the default host-aware (fullChart when a GPU build is active, queryScoped otherwise — or vice versa depending on whether prewarm is expected). Flagging so the reviewer can decide with the latency picture in hand, not just the quality one.

dkayiwa:
## Added: queryScoped slice default topK 30 → 12 (commit f6b58d1)

This directly resolves the CPU-latency tradeoff I flagged above. A topK sweep on the 22-patient drift-metric gold, **confirmed on a 36-patient sweep on demo data**, put the knee at ~12–15:

| topK | meanF1 | abstention | drift | ~CPU first-token |
|---|---|---|---|---|
| 30 (old default) | 0.748 | 0.86 | 181 | 8.3s |
| 15 | 0.742 | 0.93 | 98 | 4.3s |
| **12 (new default)** | **0.746** | **0.93** | **112** | **3.5s** |
| 8 | 0.721 | 0.93 | 103 | 2.4s (recall starts eroding) |

topK=12 holds F1 at the plateau, **improves** abstention (0.86→0.93) and roughly **halves drift**, and cuts CPU time-to-first-token **~2.4×** (8.3s→3.5s). So the earlier "queryScoped is ~12–14s slower to first token on CPU" objection is largely closed: with topK=12 the queryScoped cold first-token is ~3.5s.

**36-patient demo-instance sweep (topK=12 vs 30, 324 cells):** 74% identical citation sets; 28% fewer citations overall — concentrated on `heart` cells where topK=30 cited ~26 vital-sign records (verified drift, not recall loss); abstained-correctly 6× more often than it stopped citing; no systematic recall loss (drops records about as often as it adds).

Only material in queryScoped (fullChart ignores topK with preFilter off — verified byte-identical). config.xml default 30→12 + rewritten description; spec-locked by a test.

This PR now makes queryScoped the default **and** ships the tuned slice size that makes it fast on CPU. The two commits are independent if a reviewer wants only one.

PR #80, opened 2026-07-20
Generic dominant-concept slice expansion for repeated measures

dkayiwa:
## Problem

A fixed `querystore.topK` truncates a repeated-measure series arbitrarily. On the standalone, **"Is she hypertensive?"** surfaced only the top-K systolic readings and *dropped the high ones* (159/156/151 mmHg) that actually answer the question — the model saw a partial, misleading slice.

## Fix

When one concept **recurs (the plurality)** across the similarity ranking, `buildScoped` now includes **every** chart record of that concept, so the whole series reaches the LLM instead of an arbitrary top-K cut.

Fully **generic** — records are grouped by their text with numeric values (and sign) normalized to `#` (`conceptKey`), so it encodes no clinical vocabulary (per the module's domain-agnostic rule).

### Two guards keep it off condition/diagnosis questions
- **Recurrence gate** — the plurality must recur ≥ `DOMINANCE_MIN` (4) times; a diverse condition ranking has every concept at count 1 and never fires.
- **Condition-guard** — when the top hits carry ≥ 2 distinct condition/diagnosis concepts the query is condition-oriented ("any heart problems?"), and a vital that merely recurs is suppressed. A hit's `resourceType` is resolved from the authoritative chart docs (similarity payloads may not carry it).

Expansion is capped at `EXPANSION_CAP` (40) most-recent records, gated by a new default-on GP `chartsearchai.slice.conceptExpansion` (folded into the answer cache key). Diagnostics log at DEBUG.

## Validation — zero gold regression

22-patient drift-metric A/B, queryScoped `topK=12`:

| policy | meanF1 | abstention | drift |
|---|---|---|---|
| baseline (no expansion) | 0.746 | 0.93 | 112 |
| **+ expansion + guard** | **0.745** | **0.93** | **112** |

Identical on all three axes (F1 −0.001 is sub-noise) — while surfacing the full BP series the metric does not otherwise measure. Live-verified on the standalone: hypertension now returns all 11 systolic readings including the highs; heart/mental cells snap back to baseline (guard suppresses).

## Tests
- `QueryStoreChartBuilderConceptExpansionTest` (new, 10) — conceptKey collapse incl. negative values; plurality-not-rank-1; both condition-guard paths (hit type + chart-resolved); single-incidental-condition still expands; below-min; cap.
- `QueryStoreChartBuilderScopedTest` — composed `buildScoped` path surfaces the top-K-dropped readings, with a disabled-case control proving expansion (not typed scope) is responsible.
- `ChartSearchServiceRouterTest` — `conceptExpansion` folded into the cache key.
- Full api suite: 600 tests, 0 failures.


dkayiwa:
Superseded by #81 (need-driven adaptive top-K).

An answer-level A/B (22 patients × 4 concepts, "highest recorded X?", fullChart as the complete-data reference) showed this expansion stack and the far simpler adaptive-K are **statistically tied on benefit**: expansion reported the true extreme 78/84 (93%), adaptive-K 79/84 (94%), topK-12 baseline 73/84 (87%) — at the same latency. Adaptive-K reaches that with ~30 lines and one lexical classifier instead of three stacked heuristics (concept-expansion + condition-guard + query-anchor, ~200 lines) re-implementing retrieval-completeness logic in chartsearchai.

The gold was neutral for both. Expansion's only unique capability was the *implicit* "is she hypertensive?" case (retrieval-driven), which is a narrow, unproven-to-matter edge that doesn't justify the complexity — and it came with an over-fire tendency that needed two further guards. Closing in favour of the simpler, measured-equal approach in #81.

PR #81, opened 2026-07-20
Need-driven adaptive top-K for enumeration queries

dkayiwa:
## Problem

A fixed queryScoped `topK=12` truncates repeated-measure series. On a 22-patient × 4-concept probe (fullChart = complete-data reference), **"highest recorded [X]?" reported the true extreme only 87% of the time** — the extreme is often not among the 12 most-similar records (e.g. "highest BP = 151" when it was really 159).

## Why not just raise topK?

Because it's a precision/recall lever, not a free win — **measured on the 22-patient gold:**

| config | meanF1 | abstention | drift |
|---|---|---|---|
| topK=12 | 0.746 | **0.93** | **112** |
| topK=30 (global) | 0.748 | 0.86 | 181 |
| adaptive typed→12/topical→30 | 0.753 | 0.88 | 201 |

A global bump feeds the small model noise on absent-topic questions → it over-cites → abstention and drift regress. That's why 12 is the default.

## Fix: need-driven, not global

Only **enumeration/aggregate/extreme** questions ("list all X", "highest X", "X trend", "how many X") get the larger K (`chartsearchai.querystore.enumerationTopK`, default 40); every other question keeps the base K. Because the trigger is a lexical classifier (`QueryScopeRouter.wantsCompleteSeries`) and **none of the gold's typed/condition queries match it**, the gold is untouched **by construction** — the precision cost that sank global-30 never applies.

**Result (answer-level, "highest X"):** extreme-reporting **87% → 94%**, at queryScoped latency. Live-verified: enum→`topK=40`, typed/condition→`topK=12`.

## Scope & honest limits

- Purely lexical — no clinical vocabulary.
- Only **explicit** series phrasing triggers the bump. A yes/no question that *implicitly* needs the series (**"is she hypertensive?"**) is deliberately **not** matched — detecting it would require clinical knowledge. A miss simply keeps the default K (today's behaviour). This gap is locked as a test.

## Why this over #80 (concept-expansion)

This **supersedes #80**. An answer-level A/B showed concept-expansion + condition-guard + query-anchor (~200 lines, 3 stacked heuristics) delivered **93%** — statistically tied with this **94%** — at the same latency. Adaptive-K reaches the same benefit with ~30 lines, no per-concept logic, no retrieval re-implementation in chartsearchai. Expansion's only unique capability was the implicit "hypertensive" case, which neither approach handles well and which doesn't justify the complexity.

## Tests
- `QueryScopeRouterTest` — classifier matches explicit enumeration, rejects all 9 gold query shapes + the implicit case.
- `QueryStoreChartBuilderScopedTest` — asserts the effective K at the querystore boundary (enum→40, non-enum→12).
- `PipelineSettingsDefaultTest` — locks the default (40).
- Full api suite: 592 green.


dkayiwa:
Superseded by the data-driven approach (see the new PR). #81's lexical trigger (a keyword cue list) and fixed enumerationTopK=40 were the two weaknesses flagged in review. The data-driven version keys on the retrieval distribution + each record's concept_uuid instead: it ties/beats #81 on answer-level extreme-reporting (94% vs 93%, recovers all 8 top-K under-reports), stays gold-neutral, AND catches the implicit ('is she hypertensive?') and out-of-vocabulary ('sugar levels') cases #81 structurally missed — with no cue list and no magic K.

Issue #84, opened 2026-07-22
Proposal: a DDInter-backed pluggable drug-reference source (sourceFormat=ddinter)


pbiondich:
# Proposal: a DDInter-backed pluggable drug-reference source

*Draft for the `openmrs-module-chartsearchai` maintainers. Companion to the data project at [openmrs-ddi-knowledge-base](https://github.com/pbiondich/openmrs-ddi-knowledge-base) and its [integration plan](https://github.com/pbiondich/openmrs-ddi-knowledge-base/blob/main/docs/INTEGRATION.md).*

## What this proposes

A new `DrugReferenceSource` implementation, `DdiDrugReferenceSource`, selected by `chartsearchai.drugReference.sourceFormat=ddinter`, that backs the drug-reference feature with the full DDInter 2.0 interaction set instead of the hand-curated `drug-reference.json` seed. It reads a single normalized dataset (`ddi_knowledge_base.json`, a plain ~19 MB JSON) that the data project publishes. It changes no existing behavior: `json` stays the default, and this is one more opt-in adapter behind the seam ADR Decision 24 already established.

The motivation is the problem the original Chart Search AI spec set out to solve: the curated seed covers a handful of drugs and returns nothing for the rest, and it pushes curation onto implementers. A public-source dataset removes both. DDInter is open-access, offline-capable, and maintenance-free; RxNorm and CIEL supply the identifiers the module already reasons over.

## The dataset the source reads

The data project ships a normalized artifact designed to be consumed directly. It has three tables joined by id, so nothing is duplicated: mechanisms are stored once, drugs once, and interactions are array rows that reference both.

```json
{
  "metadata": { "shape": { "drugs": 2283, "mechanisms": 8234, "interactions": 295184 } },
  "mechanisms": {
    "3900": { "text": "Coadministration with potent inhibitors of CYP450 3A4 may significantly increase the plasma concentrations of ivosidenib...", "categories": ["metabolism"] },
    "-1":   { "text": null, "categories": [] }
  },
  "drugs": [
    { "id": "DDInter1", "name": "Abacavir", "rxcui": "190521", "rxnorm_name": "abacavir",
      "drugbank_id": "DB01048", "atc": ["J05AR", "J05AF"],
      "ciel": [ { "code": "103166", "uuid": "103166AAAA…", "name": "Abacavir / lamivudine" } ] }
  ],
  "interactions": [
    ["DDInter1000", "DDInter1008", "Major", "3900"],
[...]

PR #85, opened 2026-07-22
Add DDInter-backed drug-reference source (sourceFormat=ddinter) — implements #84

pbiondich:
Implements **Phase 2a** of the proposal in #84: a DDInter-backed pluggable drug-reference source, so the feature can run on the full DDInter 2.0 interaction set instead of the hand-curated seed. No existing behavior changes — `json` stays the default and this is one more opt-in adapter behind the ADR Decision 24 seam.

## What it does

- **`DdiDrugReferenceSource`** (`sourceFormat=ddinter`) reads the normalized DDI knowledge base (`ddi-knowledge-base.json` from the [openmrs-ddi-knowledge-base](https://github.com/pbiondich/openmrs-ddi-knowledge-base) data project) and expands its three-table form (mechanisms stored once · drugs · interaction rows) into the module's drug-centric `DrugReference` model: one entry per drug whose `interactions[]` are its partners with a `severity + mechanism` note. Because the rows are symmetric, each pair contributes to both drugs' entries (the validator's from-either-side matching).
- **Aliases** include the RxNorm name and the linked **CIEL concept names**, so question-driven matching hits the dictionary the chart already uses. **`atcCodes`** come from the pre-derived table.
- **Memory:** notes are interned per `(severity, group)`, so cost is bounded by the ~unique mechanism set, not the pair count.
- **Fail-safe:** any load/parse problem degrades to an empty list, exactly like the other sources.
- **V1 scope:** drug-drug interactions only — no `ageBands`, no `contraindications` (dosing and drug-allergy/condition are out of scope); `management` is folded into the note rather than invented.

## Wiring & conventions

- `DrugReferenceService` selects the new source for `sourceFormat=ddinter`; `json`/`atc` are unchanged.
- Reuses the shared `chartsearchai.drugReference.dataFilePath` GP + classpath fallback (same as `JsonDrugReferenceSource`) rather than adding a new GP — this differs from the proposal's sketch but matches how `json`/`atc` already work.
- Ships a **small representative sample** (16 common drugs) as the bundled classpath default; operators point `dataFilePath` at the full `ddi_knowledge_base.json` for full coverage. This is the proposal's open question #1 (bundle full vs. small default + GP) — I went with the lean default to avoid committing ~19 MB here; happy to change it.

## Tests

`DdiDrugReferenceSourceTest` exercises the real `load()` → `parse()` path against the bundled sample: Warfarin × ibuprofen fires at `Major` with a mechanism note, aliases include CIEL combination names, and V1 scope carries no `ageBands`/`contraindications`.

> Note: I was unable to build/run locally in the authoring environment (no JDK/Maven), so CI is the first real execution. I verified the bundled sample supports every assertion, but please confirm the build is green.

## Not included (proposed as follow-ups in #84)

- A `severity` field on `Interaction` (severity is a first-class column in the dataset) so chips can be ranked rather than folded into prose.
- An optional CIEL-concept order matcher (more precise than the ATC hop for an OpenMRS chart).

Open questions for reviewers are in #84; the two that most affect this PR are the bundle-vs-GP default and whether to take the `severity` field upstream.


dkayiwa:
@claude review

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
```diff
@@ -0,0 +1,241 @@
...
+			ref.setAtcCodes(row.atc);
+			ref.setInteractions(interactionsFor(links, byId));
+			ref.setSource(SOURCE);
+			out.add(ref);
+		}
+		log.info("Parsed {} DDInter drug-reference entries", out.size());
+		return out;
+	}
+
+	private static List<DrugReference.Interaction> interactionsFor(List<Link> links, Map<String, DrugRow> byId) {
```
dkayiwa, line 149: This can't be pointed at the full dataset yet, and that's exactly the configuration the README now recommends, so it needs solving before merge. The note interning keeps the heap fine (I ran the full 19 MB `ddi_knowledge_base.json` through this parser: 2,283 entries in ~0.4 s, ~75 MB), but `DrugReferenceInjector.render()` writes every partner's full note into the injected record, so the per-entry rule count becomes prompt text. Measured through the real injector: the mean full-set entry renders to ~97k chars (~24k tokens), and Warfarin (934 partners) to ~266k chars (~67k tokens), against a default context window of 32,768 tokens total (`chartsearchai.llm.contextSize`). llama-server rejects the oversized prompt with a 400, which the module surfaces as `ChartTooLargeException`, so once an operator follows the "point `dataFilePath` at the full `ddi_knowledge_base.json`" instruction, any question naming a common drug fails outright instead of answering. Even the bundled 16-drug sample injects ~12k chars (~3k tokens) for a two-drug question (real `injectRecords()` run), all of it question-dependent prefill that can't reuse the chart's KV prefix on CPU deployments.

The cap has to sit on the render side, not in the parsed entries, because the validator genuinely needs all 934 Warfarin rules to check active orders. A per-entry budget in `render()` (count or char cap, plus an "and N more interactions on file" tail) is enough to make full coverage usable; picking *which* interactions survive the cap is where the #84 `severity` field would earn its place, but a dumb cap works for now. I'd want that, with a test through `injectRecords()` pinning the injected record to a budget, either in this PR, or the README's full-coverage instruction comes out of this PR and the cap lands first as its own change.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
```diff
@@ -0,0 +1,241 @@
...
+		// build one entry per drug, in dataset order
+		List<DrugReference> out = new ArrayList<DrugReference>();
+		for (DrugRow row : order) {
+			List<Link> links = partners.get(row.id);
+			DrugReference ref = new DrugReference();
+			// prefer the RxCUI as the stable, cross-source id; fall back to the DDInter id
+			ref.setId(row.rxcui != null && !row.rxcui.isEmpty() ? row.rxcui : row.id);
+			ref.setName(row.name);
+			ref.setAliases(row.aliases);
+			ref.setAtcCodes(row.atc);
```
dkayiwa, line 140: These `atc` arrays are level-4 subgroup codes, and that mismatch produces a false safety warning, so it needs fixing before merge. All 2,945 codes in the full file (and all 34 in the bundled sample) are 5 characters, e.g. `C09AA`, while everything that consumes `DrugReference.atcCodes` assumes level-5 substance codes: the curated seed ships `M01AE01`/`B01AA03`, and CIEL order concepts map to level-5. Two consequences, both reproduced through the real pipeline against the bundled sample:

* The same-drug skip in `DrugSafetyValidator.addClassInteractions` compares exact codes, so `{C09AA}` never contains the order's `C09AA03`, and the subgroup arm then fires on the drug itself. Patient on lisinopril, question "What is the patient's lisinopril dose?" produces: `[interaction] Lisinopril: same ATC class (C09AA) as active order C09AA03 — possible duplicate therapy`. The identical scenario through the `json` source produces no warning. A false duplicate-therapy chip about the patient's own single therapy, on the most common medication-question shape, is precisely the false positive the ADR-24 same-drug skip exists to block, and it fires under the defaults every ddinter deployment will run (`warnOnInteractions=true`).
* `DrugReferenceService.findByActiveOrders` is also exact-match, so it returns 0 entries against level-5 order codes: `injectFromOrders` silently never fires for this source. (`displayNameForAtcCode` misses too, which is why the chip above names the bare code instead of the drug.)

Since the data project pre-derives these codes, shipping level-5 codes there (RxNorm carries ATC level-5) fixes the same-drug skip, order-driven injection, `hasActiveDrug`'s ATC arm and the display-name lookup in one move; I'd take that over teaching each consumer about level-4. Either way this needs a `validate()` test with an active order carrying a level-5 code before merge; the four new tests stop at parse-level facts, so neither problem can surface in them.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReferenceService.java
```diff
@@ -159,6 +161,9 @@ private DrugReferenceSource source() {
 		if (ChartSearchAiConstants.DRUG_REFERENCE_SOURCE_ATC.equalsIgnoreCase(format)) {
 			return new AtcDrugReferenceSource();
 		}
+		if (ChartSearchAiConstants.DRUG_REFERENCE_SOURCE_DDINTER.equalsIgnoreCase(format)) {
+			return new DdiDrugReferenceSource();
```
dkayiwa, line 166: The operator-facing descriptions of this GP still deny the new value exists: config.xml's `sourceFormat` description ends with `Any value other than "atc" is treated as "json"`, and the `dataFilePath` description documents the bundled fallback for the json format only. An admin reading the GP metadata in the UI is being told `ddinter` is not a recognized value. The `source()` javadoc a few lines up says the same ("any value other than `atc` ... defaults to the curated JSON source"). All three want the third value before merge; the README table already has wording to crib from.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
```diff
@@ -0,0 +1,241 @@
...
+			return Collections.emptyList();
+		}
+		List<DrugReference.Interaction> out = new ArrayList<DrugReference.Interaction>();
+		for (Link link : links) {
+			DrugRow p = byId.get(link.partnerId);
+			if (p == null) {
+				continue;
+			}
+			DrugReference.Interaction i = new DrugReference.Interaction();
+			i.setToken(p.name.toLowerCase(Locale.ROOT));
```
dkayiwa: Worth using the partner's `rxnorm_name` (falling back to `name`) as the token here. `hasActiveDrug` matches the token as a substring of the order's display name, and DDInter display names diverge from CIEL order names for some very common drugs: "Acetylsalicylic acid" never matches an order named Aspirin, so Warfarin's aspirin rule (one of the pairs the PR description lists as validated) can't fire against an aspirin order. Only the reverse direction works (aspirin asked about, warfarin ordered), and with the level-4 codes the ATC arm can't rescue it. The entry's `rxnorm_name` is "aspirin" (likewise "acetaminophen"), so it's the better matching key, and the note can keep the display name for reading.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
```diff
@@ -0,0 +1,241 @@
...
+			partners.computeIfAbsent(b, k -> new ArrayList<Link>()).add(new Link(a, note));
+		}
+
+		// build one entry per drug, in dataset order
+		List<DrugReference> out = new ArrayList<DrugReference>();
+		for (DrugRow row : order) {
+			List<Link> links = partners.get(row.id);
+			DrugReference ref = new DrugReference();
+			// prefer the RxCUI as the stable, cross-source id; fall back to the DDInter id
+			ref.setId(row.rxcui != null && !row.rxcui.isEmpty() ? row.rxcui : row.id);
```
dkayiwa: The rxcui-first id isn't unique: 142 of the full set's 2,283 entries share their rxcui with a route variant (Lidocaine, Lidocaine (ophthalmic) and Lidocaine (topical) are all `6387`). `DrugReferenceInjector.matchingEntries` dedups by `getId()`, so when a question matches two variants only the last in dataset order gets injected, and two different entries mint the same citation `resourceUuid`. Preferring the rxcui only when it's unique (else keeping the DDInter id, or suffixing it) avoids the silent collapse. I wouldn't hold the merge on this one.

dkayiwa (commented):
This needs another round before merge. The blocking problems show up where the parsed entries meet the injector and the validator, with real order data and with the full dataset this source exists for: see the comments on interaction rendering and on ATC granularity. Those two plus the GP-description update are what block.

To your build question: green beyond CI too. I ran the api suite on this branch locally, 607 tests, 0 failures, the four new DdiDrugReferenceSourceTest cases included.

On the #84 open question that most affects this PR, I'd keep the lean bundled default. 19 MB doesn't belong in the omod, and full coverage needs the render cap first anyway.

pbiondich:
Built and tested locally on JDK 11 (OpenMRS platform 2.9.0-SNAPSHOT):

- `DdiDrugReferenceSourceTest` — 4/4 pass (real `load()` → `parse()` path against the bundled sample: Warfarin × ibuprofen fires at Major with a mechanism note, CIEL aliases present, V1 scope carries no ageBands/contraindications).
- Full `reference` package suite — **85 tests, 0 failures, 0 errors**, including the existing `JsonDrugReferenceSourceTest`, `AtcDrugReferenceSourceTest`, `DrugSafetyValidatorTest`, and `DrugReferenceServiceTest` — so `json`/`atc` and the validator/injector are unaffected.

`BUILD SUCCESS`.

dkayiwa:
@pbiondich this is what i was telling you the other day about code reviews done by weaker models like sonnet. The `@claude review` that you saw above is how the code review pipeline that i set up, is invoked. To reduce the cost, we used the weaker model and when you see the thumbs up below  👍 it, it means the review was done and there was nothing to fix. The comments you see under my name are using exactly the same code review skill but with a more powerful model that i launched locally from my laptop with this command: `/pr-review https://github.com/openmrs/openmrs-module-chartsearchai/pull/85 --post`

pbiondich:
Thanks @dkayiwa — this was a genuinely valuable review; every finding was real, and several were things the parse-level tests couldn't surface. All five are addressed in 12364fb, each with a regression test that goes through the real injector/validator (not parse-level):

**Blocking**
1. **Render blowup** — `DrugReferenceInjector.render()` now caps the rendered `Interactions:` section (`MAX_INTERACTION_RENDER_CHARS`, at least one shown) with an "and N more interactions on file" tail. The cap is render-only; the validator still reads every interaction off the entry. Test `renderCapBoundsBroadInteractionSets` injects Warfarin through `injectRecords()` and pins the record well under the uncapped size. A dumb char cap for now, as you suggested; ordering-by-severity is the natural follow-up if the `severity` field lands.
2. **ATC level-4 → level-5** — fixed at the source in the data project: ATC is now derived from RxNorm's `propName=ATC` (level-5, e.g. `C09AA03`) instead of RxClass level-4, and the build fails if any code isn't level-5 so it can't regress. The bundled sample here carries the level-5 codes. Test `level5AtcSkipsFalseDuplicateTherapyOnTheSameDrug` reproduces your lisinopril case and asserts no duplicate-therapy chip; `interactionAtcCodesAreLevel5` pins the code shape.
3. **GP metadata** — `config.xml` (`sourceFormat` + `dataFilePath`) and the `source()` javadoc now document `ddinter`.

**Non-blocking (done anyway)**
4. **Token** — the interaction token is now the partner's RxNorm generic name, so Warfarin's "Acetylsalicylic acid" rule fires against an order named "Aspirin". Test `interactionFiresAgainstOrderNamedByGenericName`.
5. **Id collision** — RxCUI is used as the id only when it's unique, else the DDInter id, so the Lidocaine variants (all `6387`) don't collapse. Test `sharedRxcuiDoesNotCollapseEntryIds` over a real 3-variant fixture.

On the #84 question: agreed, keeping the lean bundled default (small sample; full dataset via `dataFilePath`), which the render cap now makes safe.

Local run on JDK 11: full `reference` suite **117 tests, 0 failures** (`DdiDrugReferenceSourceTest` now 10; json/atc/validator/injector unaffected).

pbiondich:
> @pbiondich this is what i was telling you the other day about code reviews done by weaker models like sonnet. The `@claude review` that you saw above is how the code review pipeline that i set up, is invoked. To reduce the cost, we used the weaker model and when you see the thumbs up below 👍 it, it means the review was done and there was nothing to fix. The comments you see under my name are using exactly the same code review skill but with a more powerful model that i launched locally from my laptop with this command: `/pr-review https://github.com/openmrs/openmrs-module-chartsearchai/pull/85 --post`

Yeah I am learning... :)

I went ahead and made some changes based upon the PR... some of the issues were with the KB, and some with the code... both I think might be resolved now... fingers crossed :)

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
```diff
@@ -0,0 +1,260 @@
...
+			DrugRow p = byId.get(link.partnerId);
+			if (p == null) {
+				continue;
+			}
+			DrugReference.Interaction i = new DrugReference.Interaction();
+			// Match on the RxNorm generic name (e.g. "aspirin", "acetaminophen") rather than the
+			// DDInter display name ("Acetylsalicylic acid"), since the validator matches this token
+			// as a substring of the order's display name; fall back to the display name.
+			String token = p.rxnormName != null && !p.rxnormName.isEmpty() ? p.rxnormName : p.name;
+			i.setToken(token.toLowerCase(Locale.ROOT));
```
dkayiwa, line 164: Worth a follow-up rather than a change here: a patient on hydrochlorothiazide who asks about ibuprofen now gets a second interaction chip naming chlorothiazide, a drug that isn't on their chart. Reproduced through the real validator against the full KB: both "interacts with active order hydrochlorothiazide" and "interacts with active order chlorothiazide" fire. The cause sits upstream of this PR (`PatientClinicalContext.hasActiveDrug` matches tokens as raw substrings of order display names), but this dataset is what makes it reachable: 168 of the full set's 2,093 generic-name tokens sit inside another drug's name (chlorothiazide/hydrochlorothiazide, citalopram/escitalopram, prednisolone/methylprednisolone, acyclovir/valacyclovir, ...), where the curated seed had none. Word-boundary anchoring the order-name match in `hasActiveDrug`, the same thing `matchesText` already does for aliases and for the same reason its javadoc gives, would keep every legitimate match ("warfarin" still hits an order named "Warfarin sodium") and remove the whole collision class.

dkayiwa (commented):
Two spots outside the diff still describe the two-format world and could ride along with the GP-description updates: the Decision 24 implementation bullet in docs/adr.md still reads "GP `chartsearchai.drugReference.sourceFormat` (`json` | `atc`)" (this is the doc the config.xml descriptions point admins at), and the `ReferenceDataFiles` javadoc still says the loader is shared so the contract "cannot drift between the two datasets", naming `JsonDrugReferenceSource` and `CrossReactivityGroupsLoader`, now that this source is its third consumer. Both are one-line edits; neither blocks.

pbiondich:
Both doc spots updated to include `ddinter` (commit 75882d2): the Decision 24 implementation bullet in `docs/adr.md` (now `json` | `atc` | `ddinter`) and the `ReferenceDataFiles` javadoc (now naming this source as its third consumer). Thanks for catching the out-of-diff drift.

pbiondich:
Tracked the substring-matching over-fire (the hydrochlorothiazide → chlorothiazide false chip) as a separate follow-up: https://github.com/openmrs/openmrs-module-chartsearchai/issues/86. It's in shared `hasActiveDrug` infrastructure affecting all sources, so it's out of scope for this PR; the DDInter data here is correct (ibuprofen does interact with both thiazides). Fix is whole-word matching, mirroring `DrugReference.matchesText`.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
```diff
@@ -0,0 +1,260 @@
...
+ * duplicated:
+ * <ul>
+ *   <li>{@code mechanisms}: {@code {groupId: {text, categories}}}, each description stored once;</li>
+ *   <li>{@code drugs}: {@code {id, name, rxcui, rxnorm_name, drugbank_id, atc[], ciel[]}};</li>
+ *   <li>{@code interactions}: rows {@code [drug_a_id, drug_b_id, severity, group_id]}.</li>
+ * </ul>
+ * This source expands that into the module's drug-centric {@link DrugReference} model: one
+ * entry per drug, whose {@code interactions[]} are its partners with a {@code severity + mechanism}
+ * note. Because the interaction rows are symmetric, each pair contributes to both drugs'
+ * entries, which is what the validator's from-either-side matching expects.
```
dkayiwa, line 44: On this source every same-class co-medication question raises two interaction chips for the same pair, and I think that's worth a follow-up in the same bucket as #86 rather than a change here. DDInter carries explicit rows for same-class pairs (2,526 rows in the full KB share an ATC level-4 subgroup), so the rule arm and the class arm in DrugSafetyValidator both fire. Reproduced through the real validator with the full KB plus the bundled groups: lisinopril asked about with ramipril active yields both "interacts with active order ramipril (Unknown severity interaction, no mechanism description on file)" and "same ATC class (C09AA) as active order Ramipril, possible duplicate therapy". The overlap itself predates this PR (the seed's ibuprofen against an active aspirin order double-chips on main today), ddinter just makes it systematic instead of incidental. One wrinkle for whoever picks it up: "class arm yields to rule arm" is the wrong dedup, since 42,415 of the full KB's 295,184 rows are Unknown severity with no mechanism text, and for those pairs the class chip is the informative one. Preferring whichever chip actually carries content (or folding the two into one) fits better, and it dovetails with the #84 severity field.

pbiondich:
Tracked the same-class double-chip (rule arm + class arm both firing, e.g. lisinopril asked about with ramipril active) as a follow-up: https://github.com/openmrs/openmrs-module-chartsearchai/issues/88 — same bucket as #86, shared `DrugSafetyValidator` infra that predates this PR (the seed double-chips on main today), which ddinter makes systematic. Captured the Unknown-severity wrinkle you flagged (naive class-yields-to-rule is wrong) and the #84 tie-in. Fixing it in a branch now.

PR #87, opened 2026-07-22
Fix hasActiveDrug to match whole words, not substrings (closes #86)

### LINKED ISSUE #86: DrugSafetyValidator: hasActiveDrug substring match over-fires on substring-nested drug names
### Summary

`PatientClinicalContext.hasActiveDrug(nameToken, atcCode)` matches an interaction/contraindication token against the patient's active-order names with a plain substring check:

```java
for (String drug : activeDrugNames) {
    if (drug.contains(n)) {   // n = token, lowercased
        return true;
    }
}
```

When one drug name is a substring of another, this over-fires. The clearest case: a patient on **hydrochlorothiazide** who asks about **ibuprofen** gets a *second* interaction chip naming **chlorothiazide** — a drug not on their chart — because

```
"hydrochlorothiazide".contains("chlorothiazide")  →  true
```

Both of ibuprofen's real interaction rules (with hydrochlorothiazide and with chlorothiazide) match the single order. Reproduced through the real validator; surfaced by the review of #85 (the DDInter source lists both thiazides as ibuprofen partners, which the small curated seed did not, so the latent matcher bug became visible).

### Scope

This is shared, pre-existing infrastructure: `hasActiveDrug` backs the rule-based interaction and (via the same containment) drug matching for **all** sources (`json`, `atc`, `ddinter`), not the DDInter source specifically. The DDInter data is correct — ibuprofen does interact with both thiazides; the fault is the matcher.

The same loose matching has a mirror failure mode (false negatives): a token that is a distinct word but not a substring of the order's display name won't match, which is why #85 had to match on the RxNorm generic name at the data level.

### Proposed fix

Match on a **whole-word** token, not an arbitrary substring — the technique `DrugReference.matchesText` already uses in this codebase (its javadoc: "advil" matches "is advil safe?" but a sub-token does not spuriously match). Under whole-word matching:

- "chlorothiazide" is **not** a whole word inside the single token "hydrochlorothiazide" → correctly rejected (fixes this false positive);
- "aspirin" in an order "Aspirin 81 mg" → still matches (keeps the true positive).

Ideally extract the one whole-word matcher and share it between `matchesText` and `hasActiveDrug` so the two cannot drift.

### Tests

A `validate()` test with an active order named "hydrochlorothiazide" asserting **no** chlorothiazide interaction chip, plus a positive test that a genuine whole-word order still fires — through the real validator, per the repo's testing rules.

pbiondich:
Closes #86.

`PatientClinicalContext.hasActiveDrug` matched an interaction/contraindication token as a raw substring of an active-order name (`drug.contains(token)`), so a token nested inside a longer drug name over-fired. Reproduced through the real validator: a patient on **hydrochlorothiazide** asking about **ibuprofen** got a second interaction chip naming **chlorothiazide** — not on their chart — because `"hydrochlorothiazide".contains("chlorothiazide")`.

### Fix

The cause is shared matching infrastructure, so the fix is a single shared matcher rather than a per-call-site patch. `DrugReference.matchesText` already did whole-word matching; that logic is extracted to `DrugReference.containsWord(text, word)` and now backs **both** `matchesText` (alias-in-question, behaviour-preserving) and `hasActiveDrug` (token-in-order-name, the fix), so the two cannot drift.

Whole-word matching rejects `chlorothiazide` inside `hydrochlorothiazide` (it is a sub-token of one word, not a whole word) while still matching `aspirin` in `Aspirin 81 mg`.

### Scope

Affects all drug-reference sources (`json`, `atc`, and the proposed `ddinter` in #85) since `hasActiveDrug` is shared. Surfaced during the #85 review; kept as its own change per the module's "fix shared infra, not the call site" rule.

### Tests

`HasActiveDrugWholeWordTest` drives the real `DrugSafetyValidator` over a substring-nested fixture (ibuprofen interacting with both hydrochlorothiazide and chlorothiazide):
- the nested sub-token (`chlorothiazide`) does **not** fire against a `hydrochlorothiazide` order;
- the real order still fires;
- a dose/form-qualified order name (`hydrochlorothiazide 25 mg tablet`) still matches.

Full `reference` suite green locally on JDK 11: **109 tests, 0 failures** (matchesText-dependent injector/service/validator tests unaffected).

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
@@ -228,20 +228,38 @@ public class DrugReference {
 			return false;
 		}
 		for (String alias : aliases) {
-			if (alias == null || alias.isEmpty()) {
-				continue;
+			if (containsWord(lowerText, alias)) {
+				return true;
 			}
-			String a = alias.toLowerCase(Locale.ROOT);
-			int idx = lowerText.indexOf(a);
-			while (idx >= 0) {
-				boolean leftOk = idx == 0 || !Character.isLetterOrDigit(lowerText.charAt(idx - 1));
-				int end = idx + a.length();
-				boolean rightOk = end >= lowerText.length() || !Character.isLetterOrDigit(lowerText.charAt(end));
-				if (leftOk && rightOk) {
-					return true;
-				}
-				idx = lowerText.indexOf(a, idx + 1);
+		}
+		return false;
+	}
+
+	/**
+	 * @return true when {@code word} occurs in {@code text} as a <em>whole word</em> — bounded on
+	 *         each side by a non-alphanumeric character or the string edge. Whole-word, not
+	 *         substring, so a drug name nested inside a longer one does not spuriously match
+	 *         ("chlorothiazide" is not a whole word in "hydrochlorothiazide"), while a real token
+	 *         still matches ("aspirin" in "Aspirin 81 mg"). Case-insensitive; a null/blank word
+	 *         never matches. This is the one matcher shared by {@link #matchesText}
+	 *         (alias-in-question) and {@link PatientClinicalContext#hasActiveDrug}
+	 *         (token-in-order-name), so the two cannot drift.
+	 */
+	static boolean containsWord(String text, String word) {
+		if (text == null || word == null || word.isEmpty()) {
+			return false;
+		}
+		String t = text.toLowerCase(Locale.ROOT);
+		String w = word.toLowerCase(Locale.ROOT);
+		int idx = t.indexOf(w);
+		while (idx >= 0) {
+			boolean leftOk = idx == 0 || !Character.isLetterOrDigit(t.charAt(idx - 1));
+			int end = idx + w.length();
+			boolean rightOk = end >= t.length() || !Character.isLetterOrDigit(t.charAt(end));
+			if (leftOk && rightOk) {
+				return true;
 			}
+			idx = t.indexOf(w, idx + 1);
 		}
 		return false;
 	}
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/PatientClinicalContext.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/PatientClinicalContext.java
@@ -114,9 +114,11 @@ public class PatientClinicalContext {
 	/** @return true when any active-order name or ATC code matches the given interaction rule. */
 	boolean hasActiveDrug(String nameToken, String atcCode) {
 		if (nameToken != null && !nameToken.trim().isEmpty()) {
-			String n = nameToken.trim().toLowerCase(Locale.ROOT);
+			String n = nameToken.trim();
 			for (String drug : activeDrugNames) {
-				if (drug.contains(n)) {
+				// Whole-word, not substring: an interaction token that is a sub-token of a longer
+				// order name (e.g. "chlorothiazide" inside "hydrochlorothiazide") must not match.
+				if (DrugReference.containsWord(drug, n)) {
 					return true;
 				}
 			}
--- a/api/src/test/java/org/openmrs/module/chartsearchai/reference/HasActiveDrugWholeWordTest.java
+++ b/api/src/test/java/org/openmrs/module/chartsearchai/reference/HasActiveDrugWholeWordTest.java
@@ -1,0 +1,70 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.chartsearchai.reference;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Exercises {@link PatientClinicalContext#hasActiveDrug} through the real
+ * {@link DrugSafetyValidator} over a fixture whose drug names are substring-nested
+ * ("chlorothiazide" ⊂ "hydrochlorothiazide"). The matcher must fire on whole-word
+ * matches only: an interaction token that is merely a sub-token of a longer active-order
+ * name must not raise a warning naming a drug the patient is not on.
+ */
+public class HasActiveDrugWholeWordTest {
+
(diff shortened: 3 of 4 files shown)


dkayiwa:
Direction is right — substring matching is indefensible and this kills the dangerous collisions. But **the symmetric boundary is too strict for order names, and as written this PR converts false positives into false negatives in a safety net.** I measured it against the real dictionary on a 3.7.1 demo deployment (2531 drug + drug-concept names × 2093 DDInter tokens):

| rule | matches |
|---|---|
| `contains` (today, the bug) | 896 |
| **`containsWord` as in this PR** | **761** |
| left boundary + ≤1 trailing letter | 828 |

Of the 135 matches this PR drops, **74 are legitimate** — the demo dictionary is **multilingual**, and the token is a clean prefix of the localized or inflected order name:

```
'Aspirine Co 81mg'         token='aspirin'          <- French
'Aspirina'                 token='aspirin'          <- Spanish
'Amoxicilline'             token='amoxicillin'
'Ampicilline'              token='ampicillin'
'Atorvastatine'            token='atorvastatin'
'Clarithromycine Co 500mg' token='clarithromycin'
'Cloxacilline Co 500mg'    token='cloxacillin'
'Neomycine creme 5m/g'     token='neomycin'
'Ondansetrona'             token='ondansetron'
```

`"Aspirine Co 81mg"` fails `containsWord(drug, "aspirin")` because the next char is `e`. So after this PR, a patient on **Aspirine Co 81mg** is no longer checked for aspirin interactions at all — the warfarin × aspirin **Major** bleeding chip stops firing. Same for amoxicillin, ampicillin, atorvastatin, clarithromycin, cloxacillin, neomycin, ondansetron. These are French/Spanish concept names present throughout the demo DB (`Simvastatine`, `Budésonide`, `Nitroglycérine`), not stray data entry.

That is the failure mode worth naming explicitly: **the lost chips look exactly like the noise this PR removes**, so nothing in review or in CI flags it. There is no test here covering a localized order name.

### Scored both directions

Kill-set (must stop firing): `opium`⊂Tiotropium, `iron`⊂Spironolactone, `glycerin`⊂Nitroglycerin, `desonide`⊂Budesonide, `chlorothiazide`⊂Hydrochlorothiazide, `ofloxacin`⊂Ciprofloxacin, `oxacillin`⊂Cloxacilline, `salicylic acid`⊂P-aminosalicylic acid, `urea`⊂Hydroxyurea.
Keep-set (must keep firing): the nine localized forms above, plus `Aspirin 81mg`, `Simvastatin Co 20mg`, `Lisinopril 10 mg`.

| rule | kill-set still leaking | keep-set lost |
|---|---|---|
| symmetric (this PR) | 0 / 9 | **5 / 8** |
| left boundary + ≤1 trailing letter | 0 / 9 | **0 / 8** |
| left boundary + ≤2 | 0 / 9 | 0 / 8 |
| left boundary only | 0 / 9 | 0 / 8 |

The discriminating half is the **left** boundary: an alphanumeric char immediately *before* the token means a different molecule (`ti|opium`, `sp|iron|olactone`, `nitro|glycerin`, `bud|esonide`, `hydro|chlorothiazide`, `cipro|floxacin`). A letter *after* it is usually inflection.

### The deeper point: one shared matcher is the wrong target

This PR's stated goal is that `matchesText` and `hasActiveDrug` "cannot drift". But they are matching different kinds of string:

- `matchesText` runs against **question and answer prose**, where symmetric boundaries are correct — "advil" must not match inside a longer word.
- `hasActiveDrug` runs against a **drug order's display name** — one localized noun phrase, often inflected, often with a dose glued on.

Forcing them through one rule is precisely how the false negatives get in. Two named matchers with a documented difference in purpose would be safer than one that silently under-matches order names. (Note the refactor is otherwise clean — `matchesText` already had symmetric boundaries, so the only behavioural change here is `hasActiveDrug`.)

### Also worth knowing

- Zero glued-dose names (`Simvastatin20mg` shape) exist in this dictionary, so the right-hand **digit** case needs no special handling — only the letter case.
- `≤1` is fitted to one dictionary, not a linguistic law. I would defend the left boundary on reasoning; the trailing allowance is empirical. If a tuning constant is unappealing, the principled alternative is normalizing trailing inflections on both sides instead of thresholding — more work, no magic number.
- Live-verified consequences of the current bug, for context on severity: a patient on **Tiotropium** asked about linezolid gets a **Major** chip citing an active **opium** order and warning of respiratory depression and coma; a patient on **Spironolactone** asked about dolutegravir gets Major chips citing active **iron**. Both probe concepts have no ATC mapping, so the name path alone produced them. Details and reproductions in #86's comments.

Suggested test to add either way: a `validate()` case with an active order named `Aspirine Co 81mg` asserting the aspirin interaction still fires.

dkayiwa:
Rebased this commit onto current `main` and opened https://github.com/openmrs/openmrs-module-chartsearchai/pull/128, which keeps it (authorship preserved) and corrects the matching rule for order names; suggest closing this in favour of that one.

The left boundary here is right and is what does the discriminating work. The symmetric right boundary is the problem: measured over the demo dictionary's 2531 drug and drug-concept names x the full DDInter KB's 2093 rule tokens, it drops 135 of containment's 896 matches and 68 of the dropped are legitimate, because this dictionary is multilingual — `Aspirine Co 81mg`, `Aspirina`, `Amoxicilline`, `Clarithromycine Co 500mg`, `Simvastatine`, `Multivitamines et fer`. A patient on the French aspirin row (a real drug row here, whose order concept name is `Acetylsalicylate sodium` and carries no `aspirin` token at all) silently stops being checked for aspirin interactions, which trades a false positive for a false negative in a safety net.

The superseding PR keeps your left boundary, gives order names their own right-hand rule (at most one inflectional ending past the token, measured: 829 matches, none of the 9 known nested-name collisions leaking, none of 12 known real names lost), and shares one boundary scan between the two so they cannot drift. Your `HasActiveDrugWholeWordTest` and its fixture are carried over unchanged and still pass; I added one case they were missing — a `chlorothiazide` order must still raise chlorothiazide's own rule, so the fix is de-duplication rather than suppression of the shorter name.

dkayiwa:
Closing in favour of #128, which is now merged as `9c2a8d6`.

Your commit is preserved in that history with its original authorship (`5f3a141`, Paul Biondich), and the squash carries a `Co-authored-by:` trailer, so the attribution survives on `main` rather than being collapsed into the merger.

To recap why it changed rather than merging as-is: whole-word matching removes the substring over-match correctly, but measured against the real demo dictionary (2531 drug/concept names × 2093 DDInter tokens) it also drops **68 legitimate matches**, because the dictionary is multilingual. `Aspirine Co 81mg` stops matching `aspirin`, which silently removes a **Major** warfarin bleeding warning — verified live on a real order, where that patient's only aspirin-bearing string is the drug row (the order's concept name is `Acetylsalicylate sodium`, carrying no `aspirin` token at all).

What shipped keeps your structure and your test fixture, and splits the matcher in two rather than sharing one: `containsWord` for prose (symmetric, behaviour byte-identical to yours) and `matchesOrderName` for order names (left boundary required, at most a two-letter inflectional ending). The two call sites reach it only through `hasActiveDrug`, so chips and promoted prose still cannot disagree — your stated goal, reached without the false negatives.

Full measurement, the scored kill/keep sets, and the live reproductions are in this PR's comment thread and on #86.

Thanks for finding and diagnosing this — the substring analysis in your description was correct and is what made the fix tractable.

PR #89, opened 2026-07-22
Fold same-class explicit interactions into one chip (closes #88)

### LINKED ISSUE #88: DrugSafetyValidator: same-class interacting pairs raise two interaction chips (rule arm + class arm)
### Summary

For a co-medication that is BOTH an explicit interaction partner AND in the same ATC level-4 subgroup, `DrugSafetyValidator` emits **two** `TYPE_INTERACTION` chips for the one pair:

- **rule arm** (`addInteractions`) — an explicit interaction row matches the active order;
- **class arm** (`addClassInteractions`) — the drug shares an ATC level-4 subgroup with the active order → "possible duplicate therapy".

Reproduced through the real validator: **lisinopril** asked about with **ramipril** active yields both
`interacts with active order ramipril (Unknown severity interaction, no mechanism description on file)` and
`same ATC class (C09AA) as active order Ramipril — possible duplicate therapy`.

### Scope

Pre-existing, shared infrastructure: on `main` today the curated seed's ibuprofen against an active aspirin order already double-chips. The DDInter source (#85) makes it systematic rather than incidental — ~2,526 rows in the full KB share an ATC level-4 subgroup. Affects all sources, so it belongs here, not in the ddinter source. Surfaced during the #85 review.

### The wrinkle (do not naively "class yields to rule")

Preferring the rule chip and dropping the class chip is the wrong dedup: **42,415 of the full KB's 295,184 rows are Unknown severity with no mechanism text**, and for those pairs the rule chip carries nothing (`"Unknown severity interaction, no mechanism description on file"`) while the class chip ("same ATC class … possible duplicate therapy") is the informative one.

### Proposed fix

Decide one chip per (drug, active-order) pair, preferring **whichever chip actually carries content** (a mechanism note), else the class chip — or fold the two into a single chip that keeps both the class relationship and the mechanism. Correlate the two arms by the order's ATC code. Dovetails with the `severity` field proposed in #84 (which would also let the render cap and the chip text rank by severity).

### Tests

`validate()` tests through the real validator: (a) a same-class pair WITH a mechanism note yields one chip carrying the mechanism; (b) a same-class pair with an Unknown/no-mechanism row yields one chip carrying the class/duplicate-therapy information; neither yields two.

pbiondich:
Closes #88.

A co-medication that is BOTH an explicit interaction partner AND in the same ATC level-4 subgroup raised **two** `TYPE_INTERACTION` chips — the rule arm (`addInteractions`) and the class arm (`addClassInteractions`) both fired for the one pair. Reproduced through the real validator: lisinopril asked about with ramipril active → both `interacts with active order ramipril …` and `same ATC class (C09AA) as active order Ramipril — possible duplicate therapy`.

### Fix

The two arms are merged into one coordinated `addInteractionWarnings`, correlating them by the **order's ATC code** (the key both already share). When both apply to the same order, the chip **folds** the class/duplicate-therapy relationship together with the rule's mechanism note.

Folding — rather than "class arm yields to rule arm" — is deliberate, and addresses the wrinkle raised in the review: ~42k of the full KB's rows are Unknown severity with **no** mechanism text, so a naive rule-wins dedup would collapse those pairs to a contentless chip. Folding keeps the informative class relationship in that case, and the mechanism when there is one. Single-arm cases (class-only duplicate therapy, or a different-class explicit interaction) are emitted **unchanged**. A rule matched only by name (no ATC to correlate on) is emitted standalone.

`ruleDetail()` and `classRelationship()` are extracted so the fold and the single-arm paths share one definition.

### Scope

Shared validator infrastructure — it predates the DDInter source (the seed's ibuprofen against an active aspirin order double-chips on `main` today via the cross-reactivity-group path); #85 just made it systematic. Fixed here rather than in the source. Surfaced during the #85 review.

### Tests

`DuplicateInteractionChipTest` drives the real `DrugSafetyValidator` over a same-class fixture:
- a same-class pair **with** a mechanism folds to one chip carrying both the duplicate-therapy relationship and the mechanism;
- a same-class pair **without** a mechanism still yields the informative class chip (not a contentless one);
- neither yields two chips.

Full `reference` suite green locally on JDK 11: **109 tests, 0 failures** (the existing rule-only and class-only interaction tests unaffected, confirming single-arm behaviour is preserved).

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
@@ -10,9 +10,11 @@
 package org.openmrs.module.chartsearchai.reference;
 
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -182,8 +184,7 @@ public class DrugSafetyValidator {
 				addClassContraindications(warnings, ref, context);
 			}
 			if (warnInteractions) {
-				addInteractions(warnings, ref, context);
-				addClassInteractions(warnings, ref, context);
+				addInteractionWarnings(warnings, ref, context);
 			}
 			if (warnDose) {
 				addOverdose(warnings, ref, context, lower, all);
@@ -218,22 +219,98 @@ public class DrugSafetyValidator {
 		}
 	}
 
-	private void addInteractions(List<SafetyWarning> warnings, DrugReference ref,
+	/**
+	 * Interaction reasoning with both arms coordinated so a co-medication that is BOTH an explicit
+	 * interaction partner AND same-class yields ONE chip, not two (issue #88). Two arms feed it:
+	 * <ul>
+	 *   <li><b>rule</b> — an explicit interaction row matched against an active order (by ATC code or
+	 *       name token);</li>
+	 *   <li><b>class</b> — the drug shares an ATC level-4 subgroup with an active order (duplicate
+	 *       therapy), else a curated {@link CrossReactivityGroup}.</li>
+	 * </ul>
+	 * Correlated by the order's ATC code — the key both arms already share. When both apply to the
+	 * same order the chip <em>folds</em> the class/duplicate-therapy relationship together with the
+	 * rule's mechanism note, so nothing is lost and an Unknown-severity row (whose rule note is
+	 * empty) still surfaces the informative class relationship rather than a bare, contentless rule
+	 * chip. When only one arm applies, its single chip is emitted unchanged. A rule matched only by
+	 * name (no ATC to correlate on) is emitted standalone.
+	 */
+	private void addInteractionWarnings(List<SafetyWarning> warnings, DrugReference ref,
 			PatientClinicalContext context) {
 		if (context == null) {
 			return;
 		}
+		Set<String> activeCodes = context.getActiveDrugAtcCodes();
+
+		// Explicit-interaction rules: those correlated to an active order by ATC code (which may also
+		// hit the class arm, so they are resolved in the loop below) vs name-only matches (standalone).
+		Map<String, DrugReference.Interaction> ruleByOrderCode = new LinkedHashMap<String, DrugReference.Interaction>();
 		for (DrugReference.Interaction i : ref.getInteractions()) {
-			if (context.hasActiveDrug(i.getToken(), i.getAtc())) {
-				// A matched rule has a non-blank token or ATC, so the coalesce never yields null.
-				String detail = "interacts with active order "
-						+ ChartSearchAiUtils.firstNonBlank(i.getToken(), i.getAtc());
-				if (i.getNote() != null && !i.getNote().isEmpty()) {
-					detail += " — " + i.getNote();
+			String code = DrugReference.normalizeAtcToken(i.getAtc());
+			if (code != null && activeCodes.contains(code)) {
+				if (!ruleByOrderCode.containsKey(code)) {
+					ruleByOrderCode.put(code, i);
 				}
-				warnings.add(new SafetyWarning(SafetyWarning.TYPE_INTERACTION, ref.getName(), detail));
+			} else if (context.hasActiveDrug(i.getToken(), null)) {
+				warnings.add(new SafetyWarning(SafetyWarning.TYPE_INTERACTION, ref.getName(), ruleDetail(i)));
 			}
 		}
+
+		Set<String> refClasses = ref.atcSubgroups();
+		List<CrossReactivityGroup> refGroups = CrossReactivityGroup.groupsOf(ref,
+				drugReferenceService.getCrossReactivityGroups());
+		Set<String> refCodes = ref.normalizedAtcCodes();
+
+		for (String orderCode : activeCodes) {
+			if (refCodes.contains(orderCode)) {
+				// The active order is the same drug — restating existing therapy is not a duplicate,
+				// and a drug does not interact with itself, so neither arm should fire.
+				continue;
+			}
+			DrugReference.Interaction rule = ruleByOrderCode.remove(orderCode);
+			String classNote = classRelationship(orderCode, refClasses, refGroups);
+			if (classNote != null) {
+				String detail = classNote;
+				if (rule != null && rule.getNote() != null && !rule.getNote().isEmpty()) {
+					detail += "; explicit interaction: " + rule.getNote();
+				}
+				warnings.add(new SafetyWarning(SafetyWarning.TYPE_INTERACTION, ref.getName(), detail));
+			} else if (rule != null) {
+				warnings.add(new SafetyWarning(SafetyWarning.TYPE_INTERACTION, ref.getName(), ruleDetail(rule)));
+			}
+		}
+	}
+
+	/** The rule-arm chip detail: "interacts with active order X", plus the mechanism note when present. */
+	private static String ruleDetail(DrugReference.Interaction i) {
+		String detail = "interacts with active order " + ChartSearchAiUtils.firstNonBlank(i.getToken(), i.getAtc());
+		if (i.getNote() != null && !i.getNote().isEmpty()) {
+			detail += " — " + i.getNote();
+		}
+		return detail;
+	}
+
+	/**
+	 * @return the class-arm relationship phrasing for an active order code — a shared ATC level-4
+	 *         subgroup (duplicate therapy), else a curated cross-reactivity group — or {@code null}
+	 *         when neither applies.
+	 */
+	private String classRelationship(String orderCode, Set<String> refClasses,
+			List<CrossReactivityGroup> refGroups) {
+		if (refClasses.isEmpty() && refGroups.isEmpty()) {
+			return null;
+		}
+		if (orderCode.length() >= DrugReference.ATC_SUBGROUP_PREFIX_LENGTH && refClasses
+				.contains(orderCode.substring(0, DrugReference.ATC_SUBGROUP_PREFIX_LENGTH))) {
+			return "same ATC class (" + orderCode.substring(0, DrugReference.ATC_SUBGROUP_PREFIX_LENGTH)
+					+ ") as active order " + displayNameForAtcCode(orderCode) + " — possible duplicate therapy";
+		}
+		CrossReactivityGroup group = CrossReactivityGroup.sharedGroupForCode(refGroups, orderCode);
+		if (group != null) {
+			return "same cross-reactivity group (" + group.getName() + ") as active order "
+					+ displayNameForAtcCode(orderCode) + " — possible additive or duplicate-class therapy";
+		}
+		return null;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


dkayiwa:
Heads-up before this gets rebased — it currently reads `CONFLICTING`, and three things have landed or are in flight since it was opened (2026-07-22) that change what it should do.

**1. The collapse is severity-blind.** Where this PR does fold rows, it keeps whichever arrives first:

```java
if (!ruleByOrderCode.containsKey(code)) {
    ruleByOrderCode.put(code, i);
}
```

So the surviving chip is decided by dataset position, not by severity. That is the same failure mode #110's `orderedInteractionNotes` was written to fix on the render side — its javadoc records a case where "the actionable half went to the less dangerous interaction, decided by dataset accident." Concretely: a patient on Dexamethasone asked about voxelotor has rows at **Major, Moderate, Moderate**; first-wins can keep a Moderate and drop the Major.

**2. It cannot separate route variants, which is the shape of #115.** The key is the order's ATC code, but DDInter's route variants publish *identical* ATC lists — all four dexamethasone entries carry the same eleven codes. And the live #115 reproduction never reaches that branch at all: Margaret King's Dexamethasone concept (92) has **no ATC mapping** in the demo dictionary, so it goes down the name-only branch, which adds a warning per row with no grouping. So this PR does not fix #115, and #115 does not fix #88 — they are genuinely separate, but they both want to collapse at the same point.

**3. It predates two merged changes and would regress them.**
- #108 added the rule-arm severity floor (`chartsearchai.drugSafety.minInteractionSeverity`). This PR has no floor check, so rebasing it naively reinstates the Unknown-severity chips that floor exists to suppress.
- #108 also moved chip labels to `ref.displayLabel()` (dual-vocabulary, e.g. `Acetylsalicylic acid (aspirin)`). This PR still uses `ref.getName()`, which would drop the synonym.

**Suggested path.** [#121](https://github.com/openmrs/openmrs-module-chartsearchai/pull/121) (for #115) adds `bestRulePerPartner(ref, context, severityFloor)` — grouping by the label the chip actually renders, winner by highest severity, tie-break by fuller mechanism note, then dataset order. That is a superset of what the collapse here needs. Landing #121 first and then having this PR's rule loop iterate `bestRulePerPartner(...)` — dropping its own first-wins `put` — leaves **one** collapse point instead of two competing ones, and inherits the severity floor and `displayLabel()` for free.

Worth preserving from this PR either way: the wrinkle documented in its description — that "class yields to rule" is the wrong dedup, because for the ~42k Unknown-severity/no-mechanism rows the rule chip carries nothing while the class chip ("same ATC class … possible duplicate therapy") is the informative one. #121 deliberately does not touch the class arm, so that cross-arm decision is still this PR's to make, and it is the harder half.

Context for the severity numbers and the live reproductions: #115 and its comments.

dkayiwa:
Rebased and reworked onto current `main` (`89d14ab`). The rebased branch is pushed to this repo as
[`fix-duplicate-interaction-chip`](https://github.com/openmrs/openmrs-module-chartsearchai/tree/fix-duplicate-interaction-chip)
— this PR's head lives on a fork I cannot push to, so the branch is here rather than on the PR. No new PR was opened.

**Four commits, authorship preserved.** `aa1f3ba` is @pbiondich's original commit, cherry-picked with
`-x`. Plainly: its **validator body could not be carried over**. It was written against `83cc33e` and
rewrote the one method that #108, #110, #121 and #127 have all since changed; four conflict regions,
and its `classRelationship` calls `displayNameForAtcCode`, which `main` renamed. I resolved the
conflict to `main`'s side for the production file, so the cherry-pick lands **its test and fixture as
the specification** — both fail on `main` — and the rework commits satisfy them on the new
architecture. `git log --format='%an'` still shows Paul Biondich for that commit, and nothing was
removed from its test file or its fixture.

`f2548a1` is a regression my own fold introduced, found in self-review and worth flagging because it
is the kind of thing a rebase hides. #113's screening arm stands down from a pair the drug-in-play arm
already reported by recognising a chip *worded identically* — and a folded chip is that arm's chip plus
a sentence, so the exact-string test stopped recognising it and the screen re-reported the pair. #88's
duplicate straight back, now in two wordings. `alreadySaid` matches the identity or the identity
extended by the fold, anchored on the folded sentence's own opening rather than on a bare prefix
(`… active order iron` is a prefix of `… active order iron dextran`, and both are real KB labels).
Its test fails on `main` and on `0ed0cab`, at 2 chips both times.

---

## STEP 0: #88 still reproduces on `main`

Verified before doing any rework, since three merges since this PR was opened could have closed it.

Through `DrugSafetyValidator.validate` on `89d14ab`, this PR's own fixture (lisinopril asked about,
ramipril `C09AA05` active) still yields **two chips**:

```
Lisinopril interacts with active order ramipril — additive hyperkalemia and renal impairment risk
Lisinopril is in the same ATC class (C09AA) as active order Ramipril — possible duplicate therapy
```

And the *contentless-rule* case this PR's description warns about survives too, though **not for the
reason the description gives** — see below:

```
Enalapril interacts with active order ramipril            <- carries nothing
Enalapril is in the same ATC class (C09AA) as active order Ramipril — possible duplicate therapy
```

And live on the 3.7.1 standalone against `89d14ab` — see the capture below.

## What changed from this PR

**1. The rule arm now iterates `main`'s `bestRulePerPartner`, and `ruleByOrderCode` is gone.** As the
review suggested: one collapse point, severity-aware, inheriting #108's floor and `displayLabel()`
for free.

**2. The cross-arm correlation is on partner SUBSTANCE identity, not on the order's ATC code.** This
is a correction to #88's proposed fix, not just to its implementation, and it is measured:

- A substance is filed under several ATC codes while a rule carries only **one** of them — `ddinter`
  writes the partner row's *first* code. The loaded KB's aspirin entry publishes
  `A01AD05, B01AC06, N02BA01`, so every rule naming aspirin carries `A01AD05`, while an aspirin order
  in this dictionary maps to `N02BA01`. On the order's-ATC-code key **that pair never correlates**.
- So `ruleAbout` asks two exact questions: is the rule's own code the order's code, or does the rule
  `identifies()` the dataset entry the order's code resolves to. The second reuses the name-identity
  test the question-pair and screening arms already share.
- **This PR's own headline example is one of the cases the ATC key misses.** I checked its head
  (`83393f1`) directly: the curated seed's ibuprofen against an active aspirin order still gives
  **two chips** there, because the aspirin rule's code (`B01AC06`) is not the order's (`N02BA01`), so
  it falls to the name-only branch and is emitted standalone. Three of my five new tests fail if the
  correlation is reduced to the ATC-code key.

**3. `ref.displayLabel()` throughout** (#108's dual-vocabulary label, e.g.
`Acetylsalicylic acid (aspirin)`), where this PR used `ref.getName()`.

## The wrinkle is preserved — and its evidence has moved

This PR is right that "class yields to rule" is the wrong dedup, and the fold is kept. But I could
not keep the *argument*, because #108 changed the population. Re-measured on the loaded 19 MB KB
(2283 drugs, 295,184 rows):

| measurement | result |
|---|---|
| rows whose mechanism carries no text | **42,415 — every one of them rated `Unknown`** |
| … which #108's floor (`minor`) filters before the rule arm sees them | so 0 of them can chip |
| above-floor rows with a contentless mechanism | **3** (vilanterol / mometasone / bitolterol × regular human insulin, all Moderate) |
| … of which are class-related pairs | **0** |
| above-floor rows whose pair ALSO trips the class arm | **2,195** (2,181 shared ATC-4 subgroup, 14 curated NSAID group) |
| … of which carry a contentless note | **0** |

So on DDInter data the contentless-rule-chip case no longer occurs at all. What *does* survive is the
**unrated** rule: `clearsSeverityFloor` deliberately exempts a rule with no severity rather than
ranking it low, so every hand-authored curated rule reaches a chip whatever it carries, and one
authored with no note produces a chip reading only "X interacts with active order Y". That is exactly
the row this PR's fixture pins (enalapril × ramipril), reproduced on `main` above. No bundled dataset
holds one — all five curated seed rules carry notes — but any deployment editing
`drug-reference.json` can author one.

Folding is therefore kept because it is correct for **both** populations without depending on which
is larger, rather than because 42k rows demanded it. The javadoc records the new numbers.

One deliberate departure inside the fold, flagged rather than slipped in: this PR led the folded chip
with the **class** sentence and appended the mechanism as `"; explicit interaction: …"`. I lead with
the **rule** sentence instead — it is the more specific finding, its mechanism is the actionable half,
and it names the partner by the label `bestRulePerPartner` groups on, where the class arm can only
name it by whatever the order's ATC code resolves to (the bare code, when the dataset does not carry
that substance — which is exactly what the seed's `N02BA01` chip shows). If a client ever truncates a
long detail, rule-first keeps the severity and the mechanism rather than the class inference. The fold
decision itself — that "class yields to rule" is wrong — is kept.

## One wording decision the tests took away from me

The fold concatenates each arm's own sentence **unchanged**. I tried two shorter forms first and
#108's `DrugSafetyChipLabelTest.classChipOrderNamesCarryBothVocabularies` rejected both: referring
back to "that order" dropped the synonym-augmented order label, and prefixing "also" broke the
"X is in the same …" wording it pins. Both are recorded in `classRelationships`'s javadoc so the next
person does not re-derive them.

## Residue, stated rather than papered over

When the loaded dataset carries **no entry** for the active order's ATC code, only the rule's own code
is left to compare, so a rule that matched by NAME under a different code of the same substance stays
uncorrelated and both chips are still emitted. That is the bundled curated seed's ibuprofen/aspirin
shape — the seed has no aspirin entry at all, which is equally why its class chip there can only name
the order `N02BA01`. It cannot be narrowed from this layer: `PatientClinicalContext` flattens orders
into one name set and one code set, so "the order this rule matched by name" and "the order that
contributed this code" are the same input. The real fix is per-order ATC codes on the context. On the
deployment dataset (`ddinter`, which carries an entry for every drug it rates) this residue does not
arise for that shape.

## Live verification

Standalone 3.7.1, `sourceFormat=ddinter` with the full 19 MB KB, `minInteractionSeverity=minor`,
`llm.engine=local`. **Two** full-JVM cycles: one built from `89d14ab` (before), one from this branch
(after). Both gated before any capture on `querystoreBuild hits=75` and on the known-positive warm
control raising its chip. Chips are deterministic Java; answer prose is not reproducible on that box
(`cache_prompt` KV reuse), so every claim below is about `safetyWarnings`.

The class arm needs an active order whose concept carries a `WHOATC` map, and none of the usual probe
patients' orders do. Enumerating the demo dictionary's drug rows against the loaded KB, **Ondansetron**
is such a concept (`A04AA01`) and DDInter rates it **Major** against **dolasetron** (`A04AA04`) — same
`A04AA` subgroup, so both arms fire.

**Before** (`89d14ab`, reproduced twice — 2 chips):

```
Melissa Wright (Ondansetron 8mg) + "Is it safe to give dolasetron?"     chips=2
  [interaction] Dolasetron interacts with active order ondansetron — Major. Dolasetron can cause
      dose-related prolongation of the QT interval via its pharmacologically active metabolite,
      hydrodolasetron. Theoretically, coadministration with other agents that can prolong the QT
      interval may result in additive effects and increased risk of ventricular arrhythmias
      including torsade de pointes and sudden death.
  [interaction] Dolasetron is in the same ATC class (A04AA) as active order Ondansetron —
      possible duplicate therapy
```

**After** (this branch — 1 chip carrying both relationships):

```
Melissa Wright (Ondansetron 8mg) + "Is it safe to give dolasetron?"     chips=1
  [interaction] Dolasetron interacts with active order ondansetron — Major. Dolasetron can cause
      dose-related prolongation of the QT interval via its pharmacologically active metabolite,
      hydrodolasetron. Theoretically, coadministration with other agents that can prolong the QT
      interval may result in additive effects and increased risk of ventricular arrhythmias
      including torsade de pointes and sudden death. Dolasetron is in the same ATC class (A04AA)
      as active order Ondansetron — possible duplicate therapy
```

Both the reduced count and the zero count were re-run, per the standing rule about trusting a first
query after a restart. All four documented regression controls unchanged:

```
#88 case    Melissa + dolasetron                chips=1  (folded)      expected 1
#88 case    Melissa + dolasetron (rerun)        chips=1  (folded)      expected 1
CONTROL 1   Mary + clarithromycin               chips=1  Major x simvastatin
CONTROL 2   Agnes + warfarin                    chips=1  Major x aspirin
CONTROL 3   Joshua + ibuprofen                  chips=2  NSAID contra + lisinopril Moderate
CONTROL 4   Mary + paracetamol                  chips=0
CONTROL 4   Mary + paracetamol (rerun)          chips=0
```

Control 3 is worth a word: its two chips are a contraindication and an *unfolded* interaction chip.
Joshua's lisinopril order carries no ATC map, so the class arm is inert there and nothing folds — which
is also why this defect needed a purpose-made order to reproduce live at all.

### Data created on the shared standalone

One inpatient drug order — **Ondansetron 8 mg** — for **Melissa Wright**
(`b51254aa-2b77-458c-b3af-0ad5131e2df3`, order `004280df-1e6e-4fa6-a69d-ee4e88608905`), a demo patient
who had no drug orders and no allergies. It is test data I added, not demo data, and it is left in
place so the case stays reproducible.

## Tests

`mvn clean install` from the repo root (full reactor): **API 784 + OMOD 58 = 842 run, 0 failures, 0
errors, 34 skipped** (`main`'s baseline is 776 + 58 = 834; +2 from this PR's own tests, +6 new).

TDD trail: 7 of the 8 tests in `DuplicateInteractionChipTest` fail against `main` (the 8th is the
"an unfolded single-arm chip is byte-identical" guard, which must pass on both); 3 still fail if
`ruleAbout` is reduced to #88's ATC-code key; the screening one still fails without `alreadySaid`. New
fixture `drug-reference-crossarm-fold.json` mirrors the real DDInter aspirin entry's three ATC codes,
so the code-mismatch case is exercised on real-shaped data rather than an invented one.

dkayiwa:
Closing in favour of #130, now merged as `e73c8b0`.

Your head is on a fork, so the rework could not be pushed here — but your commit is preserved in that history with its original authorship (`aa1f3ba`, cherry-picked `-x`), and the squash carries a `Co-authored-by:` trailer so the attribution survives on `main`.

**Your test and fixture landed as the specification, untouched** — both fail on `main`, zero lines removed from either. What did not carry over is the validator body: four conflict regions, and `classRelationship` calls `displayNameForAtcCode`, since renamed, so a literal port no longer compiles.

Two things you diagnosed were right and are now shipped:

1. **The cross-arm duplicate is real and it still reproduced** — I checked before reworking, since #108, #115 and #113 all landed in that area since July and might have closed it. It did not: in-suite via your own fixture, and live on the standalone (a patient on Ondansetron `A04AA01` asked about dolasetron `A04AA04` got a Major rule chip beside a `same ATC class (A04AA)` chip).
2. **"Class yields to rule" is the wrong dedup**, exactly as your description argued — a rule row with no mechanism text yields a chip that says nothing while the class chip is the informative one. The fold keeps both relationships in one chip.

Two corrections worth recording, both measured rather than reasoned:

- **Correlating the arms by the order's ATC code cannot work.** The `ddinter` source writes a rule's ATC from the **partner's first** code, so every rule naming aspirin carries `A01AD05` while an aspirin order maps to `N02BA01`. Checked against your branch head: the headline example in your own description — curated ibuprofen against an active aspirin order — still produces **2 chips** there, so the key misses the case it was written for. #130 correlates on partner substance identity.
- **The "42,415 Unknown rows" figure is now stale.** All 42,415 contentless rows are rated `Unknown` and so are floor-filtered by #108; only 3 above-floor rows carry an empty mechanism, none class-related; and of the 2,195 above-floor rows whose pair also trips the class arm, none carries a contentless note. What survives is the **unrated** rule — `clearsSeverityFloor` exempts null severity because every curated hand-authored rule is unrated, which is precisely your fixture's row. So your conclusion holds on a different population than the one you cited.

Also replaced: the `ruleByOrderCode` first-wins collapse, superseded by #115's `bestRulePerPartner`, which applies the severity floor and returns the most severe row per partner rather than whichever arrived first.

Thanks for this one — the wrinkle in your description is the part that took the longest to get right, and it is the reason the fix folds rather than drops.

PR #93, opened 2026-07-27
Answer-quality work that also runs faster: retrieval vocabulary, a prompt date anchor, twin co-citation, domain-qualified conditions routing

dkayiwa:
Four answer-quality changes for the default `chartMode=queryScoped` path, all on by default,
measured on a local standalone. Net effect on a 270-cell adjudicated gold: **meanF1 0.743 →
0.760, abstention unchanged at 0.922, off-topic citations 92 → 74** — and **5.6% faster**, not
slower.

## Why these four

I built a gold for a fresh Reference Application 3.7.1 install (30 patients × 9 topics = 270
cells, boundaries ported from the committed rc.2 adjudication) and read the answers behind the
numbers. The residual quality is not where the earlier tuning looked:

- **Precision is already ~1.0; recall is the whole loss** — except on one topic.
- **`mental` alone produced 54 of the 92 off-topic citations.** It is the only gold question
  containing the word "conditions", so it is the only one routed to the CONDITIONS scope, and
  the answer was: *"Yes, the patient has several mental health or psychiatric conditions
  recorded: Lumbago with sciatica [2], Cardiogenic shock [3], Bacterial gastroenteritis [5],
  Pulmonary atelectasis [14], Chronic gingivitis …"*. That is a clinically wrong answer, not a
  scoring quirk.
- **8% of all missed on-topic records are a cited record's own twin** — the same problem
  recorded in both `conditions` and `encounter_diagnosis`, with only one of the two rows linked.
- **The prompt never said what day it is.** Asked "any visit in the last 30 days?" the model
  answered *Yes on all twelve* probe patients, including one last seen 181 days earlier.
- **Abbreviation expansion covered lab panels only, and only one of three build paths** — so
  the same question retrieved different records depending on the mode.

## The changes

| commit | what |
|---|---|
| `refactor(retrieval)` | One composed `QueryPreprocessor.forRetrieval` (expand → strip) called by all three build paths, guarded by an `ArchitectureGuardTest` rule, plus ~30 condition initialisms (`HTN`, `CKD`, `COPD`, `CVA`, `MI`, `TB`…) beside the eight lab panels. Additive; word-like initialisms match only in capitals. |
| `fix(scope)` | A conditions cue earns the complete problem list only when nothing narrows it. Generic qualifiers ("active", "chronic", "medical") still enumerate the list; a clinical domain word does not. Smaller prompt, so this is a speed-up too. |
| `feat(prompt)` | A trailing `Today's date: yyyy-MM-dd` line plus one system-prompt sentence. Placed after the question because the KV entry is keyed on `buildUserMessage(records, "")` — warmup/seed bytes stay byte-identical, so no pinned prewarm entry is invalidated at midnight. Folded into the answer cache key. |
| `feat(citations)` | When a cited record is a coded problem the retrieved set holds exactly **once in each** problem table, the row in the other table is surfaced with it. Cross-table and problem-tables-only, so repeated obs of one concept stay distinct events; one-row-per-table, so a problem diagnosed at five encounters stays five events and the chip count cannot depend on which row the model picked. Runs after generation: zero prompt tokens, and drift-neutral by construction. |

## Evidence

**Gold, 270 cells** (154 present / 116 absent), same gold both arms, **three captures of this
change and three of `main`** — because a single capture per arm turned out not to be enough to
support a drift claim:

| | `main` | this PR | mean Δ |
|---|---|---|---|
| meanF1 | 0.743, 0.743 (byte-identical) | 0.760 / 0.764 / 0.771 | **+0.022** |
| abstention | 107/116 both | 107 / 108 / 107 | ±0 |
| off-topic citations | 92 both | 74 / 95 / 100 | **−2.3** |

> **Correction to my own first draft.** I originally reported `drift 92 → 74` from one capture per
> arm. **That improvement does not exist** — 74 was the low draw of a distribution that straddles the
> baseline. Repeating the arms is what exposed it, and it also exposed something I had not looked
> for: **`main` is reproducible and this change is not.** Two baseline captures taken 18 hours apart,
> across a rebuild and a restart, are byte-identical on all 270 answers, while two captures of one
> build of this branch differ on 42 of 270 cells. At cell level: the flip-prone heart cell answered
> identically 6 times out of 6 on `main`, and flipped between 0 and 9 citations across 9 runs here on
> an identical 1519-token prompt.

Per topic, averaged over the three captures, with each capture shown so a consistent effect is
distinguishable from a draw:

| topic | base drift | this PR (3 runs) | Δ | base F1 | this PR (3 runs) | Δ |
|---|---|---|---|---|---|---|
| **mental** | 54 | 18 / 25 / 25 | **−31.3** | 0.775 | 0.825 / 0.839 / 0.841 | **+0.060** |
| **heart** | 16 | 28 / 45 / 46 | **+23.7** | 0.663 | 0.674 ×3 | +0.011 |
| **fractures** | 1 | 11 / 10 / 7 | **+8.3** | 0.526 | 0.502 / 0.549 / 0.514 | −0.004 |
| drug-allergies | 12 | 10 / 8 / 8 | −3.3 | 0.896 | 0.938 / 0.896 / 0.938 | +0.028 |
| kidney | 5 | 4 / 9 / 5 | +1.0 | 0.629 | 0.656 / 0.670 / 0.659 | +0.033 |
| eye | 1 | 0 / 0 / 1 | −0.7 | 0.961 | 0.994 / 0.994 / 0.979 | +0.028 |
| allergies | 0 | 0 ×3 | ±0 | 0.806 | 0.814 ×3 | +0.008 |
| programs | 3 | 3 ×3 | ±0 | 1.000 | 1.000 ×3 | ±0 |

**`heart` +24 and `fractures` +8 are consistent in every capture, so they are real regressions, and
together they cancel the routing fix's real −31 on `mental`.** Net drift is honest only as
"unchanged".

**What this PR does buy, consistently:** meanF1 improves on every topic except `fractures` (−0.004),
all three captures beat the deterministic baseline, `mental` gains +0.060 while shedding 31 off-topic
citations, and abstention is flat.

**Where the added drift comes from — mechanism, not mystery.** `heart` and `fractures` absent cells
whose gold counts raw vitals as off-topic now get *dated* records appended after the negative verdict:
*"No heart or cardiac problems diagnosis is recorded. Blood pressure readings are: 116 mmHg on
2026-01-03 [2], 96 mmHg on 2025-09-28 [3], …"*. The prompt grew by exactly **84 tokens** (1435 →
1519 on that cell) — the date line plus the system-prompt sentence pointing at it — and the model now
volunteers dated supporting records. That is the same mechanism the temporal probe rewards (4/12 →
11/12), firing where it is not wanted. It also explains the reproducibility gap: on `main` the
run-to-run variation stays inside the reasoning scratchpad (output tokens 120–154 while the answer
text is identical in 6/6 runs); here it escapes into the answer.

**The date anchor is now gated on `QueryScopeRouter.isTemporal`** — because that method's javadoc,
on `main`, in a class this PR edits and which this diff never touched, already records the identical
hazard for Decision 28's *recency* anchor: *"NON-temporal questions get no anchor, which is what keeps
recent vitals out of an absent-topic slice where they bait enumeration (measured: a non-temporal 'any
heart problems?' cell drifting to 39 vitals citations back when the anchor was unconditional)."* I
added a date anchor and made it unconditional, reproducing that hazard in the prompt rather than the
slice. Both anchors now share one gate. Verified not to cost the feature: temporal probe **11/12
verdict, 9/12 day-count** with the gate, versus 11/12 and 10/12 without, and 4/12 and 0/12 on `main`.

**What I could not establish, after a lot of trying:** that the gate fixes the `heart`/`fractures`
drift. Four configurations at 20–40 runs each gave apparently decisive rates (1/40, 15/40, 15/40,
6/20, 4/20) — but a control whose prompt is byte-identical to `main`'s scored 4/20 where `main` scored
1/20, and interleaving the arms query-by-query collapsed the difference to **0/8 vs 0/8**. Per-cell
behaviour is *session*-persistent, so runs inside one session are close to N=1 and a block design
confounds the arm with the session. KV-cache reuse was ruled out separately (3/16 reused vs 4/16
fresh). Two traps were caught along the way: `chartMode` silently flipped to `fullChart` during a
redeploy, putting two arms at 4–10× the prompt size, and the date rolled over mid-experiment. The
probe now refuses to run in the wrong mode, and the harness ships `paired_probe.py` for interleaved
A/B plus a written rule that repeats must be sessions, not requests.

So the gate ships on the codebase's own documented principle and a ~17-token saving per non-temporal
question — **not** on a drift claim.

**Attributing the two halves** — twin co-citation switched off offline on the same capture by
restricting references to inline-anchored indices:

| | meanF1 | drift |
|---|---|---|
| `main`, inline-only | 0.742 | 87 |
| this PR, inline-only | 0.745 | **68** |
| this PR, full | **0.760** | 74 |

Routing/vocabulary/date give the drift win (−19); twins give the recall win (37 references
added, **31 on-topic**) for +0.015 meanF1 at a cost of 6 off-topic citations.

**Temporal probe**, 12 patients, DB truth:

| | `main` | this PR |
|---|---|---|
| "any visit in the last 30 days?" | 4/12 (0.33) | **11/12 (0.92)** |
| "how many days ago was the most recent visit?" (±10 d) | 0/12 (0.00) | **10/12 (0.83)** |

**Abbreviation probe**, 17 DB-adjudicated carrier cells asked by initialism only (family-history-only
rows excluded from the truth, since "No hypertension diagnosis is recorded" is correct there):
cites the matching record **13/17 → 17/17**. In each previously-missed cell the model had
already expanded the initialism itself in its answer, so the failure was retrieval, not
comprehension.

**Latency** — A/B/A control on a fixed 27-cell set, machine idle, read from
`chartsearchai_audit_log.response_time_ms`:

| pass | n | mean response time | input tokens | output tokens |
|---|---|---|---|---|
| A — this PR | 27 | 12 650 ms | 1 540 | 214 |
| B — `main` | 27 | 13 397 ms | 1 683 | 197 |
| A again (drift control) | 27 | 12 571 ms | 1 540 | 214 |

**5.6% faster**, with the two A passes agreeing to 0.6%. The scope fix removes more prefill
(−8.5% input tokens) than the date line adds, outweighing the slightly longer answers.

## One thing I built, measured and threw away

The obvious converse of the scope fix — an opt-in `chartsearchai.scope.problemList` giving
topical questions the complete problem list, so *"No kidney problem is recorded"* would be a
claim about the chart rather than about the top-12 — is **worse on every axis and 14% slower**
(meanF1 0.760 → 0.732, abstention 0.922 → 0.871, drift 74 → **154**; `heart` drift 28 → 77,
`kidney` F1 0.656 → 0.479). It reproduces the same enumeration failure one topic over. The code
is reverted and only the finding is kept, following the precedent of Decision 28's reverted
subtype-routing experiment. Sound absence verdicts on topical questions remain an open weakness
(#94) — they need a relevance-gated retrieval tier in querystore, not a bigger prompt.

## Notes for review

- `docs/adr.md` gains Decision 30 with all of the above, including the negative result.
- The measurement harness is committed under `eval/` (fresh-install gold builder, both probes,
  the A/B/A control, the arm comparator) with a README section, so this is reproducible rather
  than a one-off. It also documents *not* reading wall-clock off two long captures: a topic whose
  prompt did not change at all moved +24% between two captures on this host.
- 648 tests, 0 failures. The 57 errors are pre-existing and unrelated (#96): `querystore-api` is
  a `provided` dependency, so its `onnxruntime` transitive never reaches the test classpath and
  every Spring-context test fails to build an `ApplicationContext`. Same 57 on `main`.
- Every commit compiles on its own.
- No changes to `openmrs-module-querystore`.

## Picking this up cold

Everything needed to review, re-measure or extend this is in the repo rather than in the PR thread:

- **Project rules** — `CLAUDE.md` (TDD, root-cause-over-patch, the API-surface rules) and
  `ONBOARDING.md` (build/test/run commands).
- **What was measured and what it cost** — [ADR Decision 30](docs/adr.md), including the
  attribution of each half of the win and one reverted experiment.
- **Reproducing the numbers** — `eval/drift-metric/README.md` § *Reproducing the Decision 30
  numbers exactly*: three commands, the expected aggregates for both builds, and the expected probe
  results. The gold is committed (`metric_gold.local.json`, `patients.local.txt`,
  `gold_audit.local.md`), so scoring an existing capture needs no database; only rebuilding the
  gold and the two probes do.
- **Budget before proposing a re-measure** — a 270-cell arm is ~90 min on a CPU-only host, so a
  change that invalidates the numbers is not free. The README also warns against reading wall-clock
  off two long captures and points at `chartsearchai_audit_log` instead.

**Already measured and rejected — please don't re-propose without new evidence.** Each of these was
run and scored, and each is worse: shortening or capping the `reasoning` scratchpad (Decision 28,
two mechanisms), prompt-shaped answer brevity (recall collapse), raising or lowering
`querystore.topK` off 12 (Decision 28's sweep), and giving topical questions the complete problem
list (Decision 30 / #94). The rationale for the shapes that survived — the per-entry case
sensitivity on initialisms, the trailing placement of the date line, the three constraints on twin
matching — is in the javadoc at each site, deliberately, because each one is load-bearing rather
than stylistic.

## Follow-ups filed

Everything this PR surfaced but does not fix. None of these are blocking; they are filed so the
findings behind them do not have to be rediscovered.

**Answer quality**
- #94 — absence verdicts on topical questions are asserted from the similarity slice, not the
  chart. Includes the measurement that refutes the obvious fix, so nobody re-tries it blindly.
- #95 — recall on high-cardinality topics is capped by the fixed similarity budget; proposes
  concept-diversified top-K and records why I did not just do it here.
- #99 — condition/diagnosis twins spend double the slice budget. Newly actionable *because* of
  this PR's co-citation, which decouples rows-read from records-linked.
- #102 — "when was X diagnosed?" answers with record-keeping time (`dateCreated` rendered as a
  clinical date on condition/diagnosis). Reproduced live; also corrects a stale claim in the
  Decision 28 note about where onset lives.

**Fallout from this PR**
- #97 — the `fractures`/`heart` drift increase is only partly attributed (+6 to twin doubling, the
  rest unexplained). Three experiments listed to settle it.

**Eval instrument**
- #98 — the gold's `fractures` boundary counts burns and lacerations, so it scores a clinically
  better answer as this arm's worst regression.

**Build and docs**
- #96 — 57 context-sensitive tests never execute (`provided`-scope `querystore-api`). Pre-existing.
- #100 — `query-stopwords.txt` strips four words its own header forbids stripping.
- #101 — README documents `topK=30`; the real default is 12.

## Hardening

Two `/harden` phases ran on this branch after the numbers above were measured (`d3c9bff`,
`fd13d7b`). Everything they found was reproduced against the real production entry points
first, and the measured arm was re-verified afterwards rather than assumed unchanged:

- **Twin pairing was not order-independent.** It paired a cited problem with *every* other-table
  row sharing its concept, so a problem diagnosed at five visits (1 condition + 5 diagnosis rows —
  `encounter_diagnosis` is per-encounter) gave 6 chips when the condition was cited and 2 when a
  diagnosis was, from one identical retrieved set. It also reached across separate episodes: on
  patient `a6451ed0` it paired a 2025-07-22 problem with a **2021-07-15** one. Now gated on
  one-row-per-table.
- **Nine routing holes**, each reproduced then fixed: the module's own certainty vocabulary
  ("any confirmed diagnoses?", "any conditions on file?"), date windows ("in the last 2-3 years?",
  "since 2024-01?"), the em dash and curly quotes macOS and Word substitute *by default*, and
  "earlier"/"long-standing".
- **Stale statements** the earlier passes left behind, including a javadoc claiming the warmup
  form is a byte-prefix of every real query (true only up to the focus-hint divergence point) and
  a `@link` to an overload that does not exist.
- **Tests that pinned less than they claimed** — the order-independence test asserted sizes, not
  index lists; `search()` asserted the date *label* but not the value.

**The headline numbers are unchanged, and that was measured, not assumed.** A concept-family
analysis over the database identified the single gold cell the pairing change can touch; re-running
it plus controls left 268 of 270 cells byte-identical and the score at meanF1 0.760 / 107 of 116 /
drift 74. After every hardening change, five cells re-run live against the redeployed standalone
showed **0 reference-set changes**.

Two items are deliberately left for follow-ups rather than smuggled in here:

- On grounding-enabled deployments (`chartsearchai.grounding.enabled` defaults to `false`), a twin
  has no citing sentence, so it defeats the verifier's zero-embed Lazy Tier-1 path and forces
  `S+1` e5 forward passes on a list-style answer. The fix is to pass the twin&rarr;sibling pairing
  into the verifier so a twin inherits its sibling's selected sentence — cheaper *and* the intended
  semantics, but it can change a verdict, so it needs its own measurement. Documented in the ADR.
- A clinical value pair can still be split into fabricated citations when the chart happens to
  contain those record numbers (`BP was [120, 80] today` with 120 and 80 in the citations array, on
  a 130-record chart) — filed as #103. Baseline behaviour, not introduced here, and twins cannot
  amplify it. The javadoc that claimed chart size was irrelevant — the stated reason nothing further
  shipped — is corrected.
- The recency vocabulary is still spelled in three places (`KEYWORD_GROUP`, `TEMPORAL_CUES`,
  `GENERIC_CONDITION_WORDS`). The mechanical reason is that the router matches a single token, so
  every multi-word alternative in `TEMPORAL_CUES` is unreachable from it. The measured holes are
  closed; the unification is not.
- **The generic-qualifier vocabulary has an open tail, and I would rather say so than imply the
  list is complete.** Hardening closed four distinct categories — the module's own certainty values,
  date windows, punctuation acting as a *separator* (`conditions/diagnoses`), and quantification /
  record-locus words — each an ordinary clinician phrasing, each found only by going looking. Two
  more are known and open. Filed as #104 with the evidence and three candidate structural fixes;
  every one of them changes routing on the 270-cell gold and so needs its own measured arm. One
  consequence is recorded in the ADR trade-offs: the "degrades gracefully" argument **inverts** for
  count questions, where a partial slice yields a confidently wrong number rather than a less
  exhaustive answer.


dkayiwa:
### Machine-local state, for whoever measures next

Not repo content, so it lives here rather than in the tree.

**The standalone this was measured on** — `referenceapplication-standalone-3.7.1`, RefApp 3.7.1
demo data (50 patients, 4 765 obs, 853 conditions, 853 diagnoses, 331 test orders, 1 allergy,
23 programs, **0 drug orders**), port 8081, `admin`/`Admin123`, embedded MariaDB on 3316,
querystore Lucene index already built. `engine=local`, Gemma 4 E4B Q4_K_M, `-ngl 0` — pure CPU,
so these are worst-case latencies.

That "0 drug orders" is worth knowing before reading the per-topic table: the `medications` topic
is absent on all 30 patients, so it contributes 30 abstention cells and no F1.

**Currently deployed:** this branch's build. Before I started, the standalone was running an
unmerged `origin/dominant-concept-completion` build (`e99bbc8`) — not `main` — and the leftover
`chartsearchai.slice.conceptExpansion` / `slice.seriesCompletion` GP rows in its database are from
that, not from this PR. I baselined against clean `main` @ 83cc33e instead. The original artifact is
preserved if it needs restoring.

**Two things that will waste your time if you don't know them:**

1. `[timing]` lines are INFO and the module package is not covered by the stock `log.level=info`.
   Set `log.level=org.openmrs.api:info,org.openmrs.module.chartsearchai:info`. I set it during
   measurement and restored it to `info` afterwards. For per-query cost prefer
   `chartsearchai_audit_log` — always recorded, no log level needed, survives restarts.
2. Do not build or run tests while a capture is in flight. The engine is CPU-bound; three `mvn test`
   runs during one arm inflated its measured latency by ~16% uniformly, including on topics whose
   prompts had not changed at all. That is what motivated the A/B/A control.

**Captures behind the numbers** — three 270-cell arms (`main`, this PR, and the rejected
problem-list variant) plus both probe sets, ~4 h of inference. They are not committed (the repo has
never committed captures) but the committed gold scores any fresh capture identically; I verified
`compare_arms.py` against the committed gold reproduces the PR's aggregates exactly.

**Not touched:** `openmrs-module-querystore`. Its working tree here has 82 staged uncommitted files,
and the standalone's `.omod` predates them, so I stayed out. #94 and #95 are the two follow-ups that
will eventually need work on that side of the line.

api/src/main/java/org/openmrs/module/chartsearchai/api/impl/QueryScopeRouter.java
```diff
@@ -129,7 +313,7 @@ static Set<Intent> matchedIntents(String question) {
 		if (PROGRAMS_CUES.matcher(q).find()) {
 			matched.add(Intent.PROGRAMS);
 		}
-		if (CONDITIONS_CUES.matcher(q).find()) {
+		if (CONDITIONS_CUES.matcher(q).find() && !isDomainQualified(q)) {
```
dkayiwa, line 316: I built both arms and ran `matchedIntents` over 40 problem-list phrasings I wrote down before looking at any result. All 8 of my domain-qualified probes route to TOPICAL here and none of them do on main, so the mental-health case this exists for is genuinely fixed. But the completeness guarantee goes from 37/40 on main to 22/40 here, and the 15 phrasings that lost it are ordinary clinician English: "any pre-existing conditions?", "any existing conditions?", "any unresolved conditions?", "any stable conditions?", "any concurrent conditions?", "any diseases diagnosed?", "any chronic illnesses or conditions?", "any conditions or complaints?", "any conditions that are untreated?".

The count case is the one I think has to block. `isDomainQualified` returns true as soon as *one* content word is unrecognised, so the quantification words in `GENERIC_CONDITION_WORDS` only take effect when every other word in the question is recognised too:

```
How many conditions does she have?                -> CONDITIONS
How many pre-existing conditions does she have?   -> []
How many chronic illnesses does she have?         -> []
How many unresolved conditions are there?         -> []
Total number of existing conditions?              -> []
```

Merged as-is, "how many pre-existing conditions does she have?" gets counted off the 12-record similarity slice rather than the problem list, so the answer is a specific wrong number with nothing in it to show the input was partial. That is exactly the case the `narrowsToADomain` javadoc calls out as the one that does not degrade gracefully, which means the mitigation that paragraph describes does not hold as written.

I do not think this is a vocabulary gap that #104 can close. A deny-by-default list over open English has no last word to add, and every word added is one more thing to keep in sync. What I would do instead is invert the test: recognise a clinical domain positively, off a closed vocabulary (body systems and specialties, or a concept-name lookup), and fall back to CONDITIONS when nothing in the question matches. A closed list has a tail you can enumerate and pin. To be clear this is not the reverted experiment from #94, it does not hand the problem list to topical questions, it only changes which questions count as topical.

One thing about the asymmetry argument, since it is what the guess direction rests on. "Psychiatric conditions: Cardiogenic shock, Bacterial gastroenteritis" is wrong in a way the clinician reading it can see. A quietly truncated problem list is not, and the count measurement above says the misroute is not bounded to exhaustiveness anyway. While the domain check is being reworked I would rather be shipping the visible failure than the invisible one.

Practically this is one clause, so holding it does not mean unpicking the hardening commits that are interleaved with it. The retrieval-vocabulary consolidation, the `isTemporal`-gated date anchor and the twin co-citation do not depend on it, and I would land those three.

eval/drift-metric/compare_arms.py
```diff
@@ -0,0 +1,375 @@
+#!/usr/bin/env python3
```
dkayiwa, line 1: This lands as an add/add conflict. Main picked up its own `compare_arms.py` after this branch forked, from the #107 verdict-guard work, and that one is a different tool: 94 lines, per-cell verdict-class A/B over two capture dirs, no gold involved. Same path, same name, and neither docstring mentions that the other exists.

If the conflict gets resolved by taking this side, main's comparator is deleted, and because both files are named for the same job the deletion reads as an ordinary conflict resolution rather than a lost tool. No test covers either script, so nothing downstream will notice.

Simplest fix is to rename this one to what it actually compares, `compare_arms_gold.py` or similar, and have the README section name both so the next person does not rediscover this the same way.

dkayiwa:
## Four open findings against this branch, relayed here before their issues are closed

#97, #99, #104 and #111 were each filed against this branch's behaviour rather than against `main`. Verified just now at `main` @ `eb81eed3`: this PR's head `0b7acbfd` is **not** an ancestor of `origin/main` (119 commits behind, last updated 2026-08-04), and every symbol the four issues are about is absent from `main`:

| symbol | issue that depends on it | on `main` |
|---|---|---|
| `LlmInferenceService.sameAssertionTwins` | #99 | absent |
| `isDomainQualified` | #111 | absent |
| `narrowsToADomain`, `GENERIC_CONDITION_WORDS` | #104 | absent |
| the `Today's date` prompt anchor | #97 | absent |

So none of the four is reproducible on `main`, and they are being closed on that basis. **They are review findings on this PR and they are not resolved** — recorded here in full so nothing is lost if this branch is revived. Each remains readable in its own closed issue.

### #97 — off-topic citations rose on `fractures` and `heart`, cause only partly attributed
This branch cut total off-topic citations 92 → 74 on the 270-cell gold, but `fractures` went 1 → **11** and `heart` 16 → **28**. Only +6 across the whole eval is attributed (to twin co-citation doubling the count on an already-drifted cell); the remaining ~+16 on those two topics is unexplained by any change touching their routing or retrieval. The issue's own reading is ordinary prompt-perturbation variance, and says it is worth confirming rather than assuming — by re-capturing just those 60 cells.

### #99 — condition/diagnosis twins spend double the slice budget
One problem is recorded twice (a `conditions` row and an `encounter_diagnosis` row for one concept), the texts differing only in the trailer, so they embed almost identically and both land in a ~12-slot top-K for one clinical fact. The issue notes this became newly actionable **because of this branch**: `sameAssertionTwins` re-attaches the other row's reference after generation, decoupling "rows the model reads" from "records the answer links". It also records why it is not obviously safe — the two rows carry different trailers (`Status: ACTIVE|INACTIVE` vs `Certainty:`), so they are not identical information.

### #104 — the generic-qualifier vocabulary has an open tail
`narrowsToADomain` presumes any unrecognised content word is a clinical domain, which suppresses the complete problem list. Hardening this branch closed four *categories* of hole in two review passes — the module's own certainty/record values ("any confirmed diagnoses?"), date windows ("diagnosed in 2024"), punctuation as a separator ("any conditions/diagnoses?"), and quantification/record locus ("how many conditions?", "in her notes?"). The issue's point is the *pattern*: a deny-by-default vocabulary has no closing condition, and its own comment records that the quantification category it listed as closed was not.

### #111 — the qualification test is any-word-wins
`isDomainQualified` returns on the **first** unrecognised content word, so listing `many`/`much`/`number`/`total`/`count` in `GENERIC_CONDITION_WORDS` only takes effect when every other word is recognised too — one unrecognised adjective beside "how many" defeats it. Measured against the real `matchedIntents` on this branch. It also carries the cost of the deny-by-default shape, and this is the number to weigh before landing: **37/40 problem-list phrasings reach `CONDITIONS` on `main`, 22/40 on this branch.** On `main` there is no domain-qualification gate at all — `matchedIntents` matches `CONDITIONS_CUES` directly — so the gate this branch adds is a net regression on that measure as it stands.

Issue #107, opened 2026-07-28
Verdict-lead misfires on unanswerable yes/no safety questions: sampled Yes/No justified by an unrelated record


dkayiwa:
### Summary

On a yes/no **safety** question whose subject the records say nothing about, the verdict-lead behavior (#83) produces a sampled Yes/No verdict justified by an unrelated record, instead of the no-record abstention the prompt already mandates for presence questions.

Measured on the standalone (full DDInter KB, `sourceFormat=ddinter`): patient on Simvastatin with a severe aspirin allergy, question **"Is it safe to give her panadol?"** ("panadol" resolves to no reference entry — openmrs-ddi-knowledge-base#5 — so no record addresses it). Five generations of the identical question:

| Run | Verdict lead |
|---|---|
| 1 | (abstained) "No information regarding an allergy or contraindication to Panadol is recorded." |
| 2 | "Yes, the patient has an allergy to Aspirin [230]." |
| 3 | "Yes — the patient has a recorded drug allergy: Aspirin ... Severe reactions ... [230]." |
| 4 | "No — the patient has a recorded drug allergy to Aspirin [230]." |
| 5 | "Yes — the patient has a recorded drug allergy: Aspirin (drug allergen) with Severe reactions ... [230]." |

The verdict is a coin flip (Yes/Yes/Yes/No/abstain), the justification each time is the true-but-unrelated aspirin allergy, and the citation legitimately verifies (the allergy sentence IS supported by record 230), so a groundless verdict rides in front of a `grounded=true` citation. Both polarities are clinically wrong: "Yes, it's safe" has no basis, and "No — because of the aspirin allergy" implies the allergy contraindicates paracetamol when it is in fact the standard analgesic for aspirin-allergic patients.

### Why the existing abstention rule misses this

The prompt already demands: start with "Yes" ONLY when a record explicitly names what is asked, and lead with a "No X is recorded" sentence otherwise. But every example it gives is presence-shaped ("any allergies?", "is the patient hypertensive", hypertension/fractures wording). A safety/suitability question ("is it safe to give X", "can we start X") reads as a different template to the model: it treats any safety-flavored record (an allergy to a different drug) as evidence bearing on the verdict.

### Fix

[...]

PR #110, opened 2026-07-30
fix(drug-safety): inject the deterministic safety finding before the answer

dkayiwa:
Safety answers abstained on questions the records **did** address. Three commits, all measured on a 3.7.1 standalone.

## The root cause

`DrugSafetyValidator` computes the safety join correctly every time — but it runs *after* the answer, so the LLM was asked to re-derive a conclusion the module already held. It does not: the eval README records **0 joins in 21 baseline cells**, and two live cases abstained with the evidence rendered, cited and provably quotable.

It now runs first (`validate("", question, context)` — the production path unmodified; an empty answer makes the drugs in play exactly the question-named ones) and each finding is injected as a numbered `safety_finding` record the answer cites. **The model relays reliably what it will not derive.**

A second, independent defect: `render()` filled the capped `Interactions:` section in *dataset* order. Clarithromycin carries 898 partners with Simvastatin (Major) at index 324 — so for a patient on simvastatin the record named ivosidenib, kanamycin and ketoprofen, none of which they take, while the chip correctly named simvastatin. The chip and the prose disagreed by construction.

## Measured (20-cell safety probe)

| /10 ANSWER cells | baseline | render ordering | + pre-answer findings |
|---|---|---|---|
| **abstained (the defect)** | 8 | 5 | **2** |
| verdict-led | 0 | 0 | 2 |
| **ABSTAIN held** | 10 | 10 | **10** |
| inverted "Yes" against a chip | 0 | 0 | 0 |

`mary__safety-clarithromycin` abstained **0/6** on three builds and now names the interaction **6/6**. A severe-aspirin-allergy patient asked about ibuprofen went from 4/4 abstain to answering 4/4, citing both the allergy record and the finding.

**Abstention survives by construction, not by wording:** a finding record exists only when the deterministic layer found something, so unconnected cells gain nothing and cannot start answering. That is the direction #107 guards.

## Why the prompt is barely touched

Five approaches were measured and reverted — details and tables in `eval/drift-metric/README.md`:

1. Quoting a forbidden phrase **primed** it (6/6 on a cell that never had it)
2. Deferring to the generic verdict rule inherited an inverted `"Yes"` — its YES criterion is a *presence* test, and on a safety question a record naming the drug is evidence *against* (5/6 on a Major interaction)
3. An `"otherwise state what the record shows"` fallback broke abstention on 3 unconnected cells
4. Publishing more evidence changed nothing, at 2.6× output tokens
5. A renal-function pairing — correct but silent on stale data, and needs clinical policy this module should not author

The one prompt change that survived is a **record-type description** (one sentence, parallel to the existing "Drug reference" one), because it only engages when that record type is present.

## Also fixed: the gate could not see the danger

`score_probe_safety.py` counted YES and NO identically, so approach (2)'s inverted `"Yes"` scored as **+1 verdict-led, −1 abstained, exit 0** — the gate said ship it. It now flags a YES lead raised against that drug's own chip and exits 3. Verified both ways on real captures.

## Before merging

- [ ] **Re-run the 20-cell gate against this exact build.** The numbers above were captured on a build that differs by the shared-prompt change and a kept `latestNumericObs` refactor touching the weight path. ~6 minutes; this is the only thing I'd block on.
- [ ] The polypharmacy paths (compact fallback, severity ordering) are unit- and mutation-tested against the real bundled dataset but have **never executed on a live box** — no test patient has two interacting active drugs.

## Still open (not regressions)

verdict-led is 2/10 — answers state the finding without saying "not safe"; and the citation set varies run to run. Both are fixable in `DrugReferenceInjector` by improving the finding's own rendered sentence rather than by prompt rules.

714 tests, 0 failures.

https://claude.ai/code/session_01QJUUiWicVqg6jSPxidFp6m


dkayiwa:
## Gate re-run against this build — cleared ✅

The "Before merging" blocker is closed. Built from `de6c6d2` (this PR's head), deployed to a 3.7.1 standalone, full JVM restart, 20-cell safety probe.

| /10 ANSWER cells | baseline | earlier build | **this build** |
|---|---|---|---|
| **abstained (the defect)** | 8 | 2 | **1** |
| **ABSTAIN held** | 10 | 10 | **10** |
| inverted "Yes" against a chip | 0 | 0 | **0** |
| led with a verdict on an ABSTAIN cell | 0 | 0 | **0** |

`score_probe_safety.py` **exit 0**. Abstained-when-answerable is 8 → **1**, better than the 2 in the description — the record-type prompt description helps the gate cells too, not only the case it was written for.

Pre-flight checks, since a green gate on a stale artifact is worthless: fresh jar carrying all three new methods, DDInter KB loaded, querystore alive (74 hits), and the prompt change confirmed *behaviourally* — naproxen leads with the finding instead of the abstention. That last one mattered: the lib-cache clear partially failed, so a stale jar was a real possibility.

The second checklist item stands unchanged — the polypharmacy paths (compact fallback, severity ordering) are unit- and mutation-tested against the real bundled dataset but have never executed live, because no test patient has two interacting active drugs.

dkayiwa:
## Added: finding injection now gates on the same toggle as the chips (`b9c6418`)

Found by hardening the pre-answer-findings code, which had had no review passes of its own — the render-ordering work in this PR had six.

`DrugSafetyValidator` checks `drugSafety.validateAnswers` **only** in its public `Patient`-taking entry. The package-private overload `preAnswerFindings` uses does not. So an operator setting that GP false silenced the chips while findings kept flowing into the prompt: the answer asserting *"not safe due to a Major interaction … [232]"* with **no chip beside it** — precisely the chip-versus-prose divergence this PR exists to remove, reappearing silently and only under a non-default config.

The two now switch together. 715 tests, 0 failures.

**Note on the gate numbers above:** the 20-cell run was captured on the build *before* this commit. This fix only changes behaviour when `validateAnswers=false`, which is not the default and not what the gate exercises, so the numbers stand — but flagging it rather than letting the build/measurement mismatch pass unstated.

**Not yet reviewed** (would be the next hardening pass, three untraced threads):
- `resourceKey(type, drug)` is used as the finding's `resourceUuid` — a synthetic value like `contraindication:Naproxen`. If any consumer resolves `resourceUuid` as a real UUID (frontend navigation, grounding, audit), that is a live failure mode.
- `renderFinding` interpolates `finding.getDetail()` unguarded — module-generated so probably always set, but unverified.
- Whether a drug with both a contraindication and an interaction should yield two finding records or one.

Issue #115, opened 2026-08-04
Route variants sharing one rxnorm_name raise N chips for one interaction, at N different severities (distinct from #88, survives #86)


dkayiwa:
### Summary

DDInter carries one entry per **route variant**, and every variant of a drug shares the same `rxnorm_name`. `interactionsFor` turns each variant into its own interaction row with the same match token, so one active order matches all of them and the patient gets N chips for one pair — at N different severities.

Reproduced live on the 3.7.1 standalone (HEAD `13690b1`) — Melissa Wright asked "Can I start dolutegravir?":

```
safetyWarnings (2)
  [interaction] Dolutegravir — "…interacts with active order iron — Major. Coadministration with medications
                                containing polyvalent cations … may decrease the oral bioavailability…"
  [interaction] Dolutegravir — "…interacts with active order iron — Major. INTERVAL: Coadministration with
                                medications containing polyvalent cations … The mechanism has not been established."
references: #41 safety_finding grounded=false | #42 safety_finding grounded=true
```

Two rows, because two entries share the token `iron`:

```
["DDInter582", "DDInter2187", "Major", "7199"]   # dolutegravir x Iron (bisglycinate)
["DDInter582", "DDInter975",  "Major", "2248"]   # dolutegravir x Iron
```

Note the grounding verifier returned opposite verdicts (`false` / `true`) on two records saying the same thing.

### This is not #88, and it survives #86

- **#88** is *cross-arm*: one chip from `addInteractions`, one from `addClassInteractions`. These are **two rule-arm chips**, so correlating the arms by the order's ATC code — #88's proposed key — does not touch them.
- **#86** (substring matching) is how *this particular* reproduction was reached: `iron` ⊂ `sp-IRON-olactone`. Fixing #86 removes that repro but not the class, because route variants match as **whole words**.

A repro that survives both: a patient on **Dexamethasone**, asked about **voxelotor**. Four entries carry `rxnorm_name = dexamethasone` —

```
DDInter513 Dexamethasone          DDInter515 Dexamethasone (ophthalmic)
DDInter514 Dexamethasone (nasal)  DDInter516 Dexamethasone (topical)
```

— and voxelotor has rows against three of them at **Major, Moderate, Moderate**. All four variants publish an *identical* ATC code list (`A01AC02, C05AA09, D07AB19, …`), so ATC cannot discriminate them either.

### Scale

[...]

PR #119, opened 2026-08-04
fix(drug-reference): strip DDInter's residual field markers (INTERVAL:/RECOMMENDED:) from interaction notes

### LINKED ISSUE #116: DDInter's INTERVAL: field marker leaks verbatim into safety chips and prompt text (224 of 8234 mechanisms)
### Summary

`INTERVAL:` is a DDInter field marker, not prose. It is never stripped, so it reaches the clinician verbatim at the front of a safety chip.

Live on the 3.7.1 standalone (HEAD `13690b1`), Melissa Wright asked "Can I start dolutegravir?":

> [interaction] **Dolutegravir** — "Dolutegravir interacts with active order iron — Major. **INTERVAL:** Coadministration with medications containing polyvalent cations such as aluminum, calcium, iron, or magnesium may decrease the oral bioavailability of dolutegravir. The mechanism of interaction has not been established."

**224 of the 8234 mechanism texts** in the loaded KB begin with `INTERVAL:`. Another example:

```
mechanisms["20"].text = "INTERVAL: The passive transfer of antibodies present in immune
                         globulin preparations may interfere with the imm…"
```

`grep -rn INTERVAL api/src/main` returns nothing outside unrelated constants — the token is not recognised anywhere in the module.

### Root cause

`DdiDrugReferenceSource.noteFor` (:190) concatenates severity with the mechanism text as-is:

```java
String note = (text != null && !text.isEmpty())
        ? severity + ". " + text
        : severity + " severity interaction (DDInter 2.0; no mechanism description on file).";
```

Whatever the mechanism field holds becomes the note, and the note becomes both the chip detail (`DrugSafetyValidator.addInteractions`) and the rendered prompt text (`DrugReferenceInjector.render`). So the marker leaks into the clinician-facing chip **and** into the grounding text the model cites.

### What INTERVAL actually means

In DDInter it marks guidance about **administration timing / separation** — the management advice for a chelation- or absorption-type interaction, where the answer is "separate the doses" rather than "do not co-prescribe". That is genuinely useful information and should not simply be deleted: for the dolutegravir example, dose separation *is* the clinical management. The defect is that a machine-readable marker is being presented as if it were the start of a sentence.

Also worth noting the `INTERVAL:` rows are precisely the ones whose mechanism ends "The mechanism of interaction has not been established" — i.e. the marker is carrying the actionable half of the note. Dropping the prefix and keeping the text is the minimum; surfacing it as a timing/management signal would be better.

### Scope

DDInter source only (`sourceFormat=ddinter`) — the curated `json` and `atc` sources do not carry these markers. Advances the data-fidelity thread from #84/#85.

### Proposed fix

Recognise the marker in `noteFor` rather than passing it through: strip the `INTERVAL:` prefix from the note text and carry the fact separately (a flag or a management field on `Interaction`), so the chip can render it as management guidance — "separate administration times" — instead of a stray label. If a structured field is more than this is worth right now, at minimum strip the prefix so the chip reads as a sentence.

Check the sibling markers before fixing: `INTERVAL:` was found by inspection, so the same scan should confirm whether other all-caps prefixes exist in the mechanism corpus rather than fixing the one that happened to surface.

### Tests

`DdiDrugReferenceSourceTest` against a real dataset fixture containing an `INTERVAL:`-prefixed mechanism: assert the parsed `Interaction` note does not begin with the raw marker and that the mechanism text survives. Then a `validate()` test through the real path asserting the chip detail is free of it.

dkayiwa:
## The defect

`INTERVAL:` is a DDInter field marker, not prose, and nothing in the module recognised it — at
`13690b1`, `grep -rn INTERVAL api/src/main` hit only unrelated timer constants. It reached the
clinician verbatim at the front of a safety chip. Observed live on the 3.7.1 standalone at
`13690b1`, Melissa Wright, "Can I start dolutegravir?":

```
[interaction] drug='Dolutegravir'
    Dolutegravir interacts with active order iron — Major. INTERVAL: Coadministration with
    medications containing polyvalent cations such as aluminum, calcium, iron, or magnesium may
    decrease the oral bioavailability of dolutegravir. The mechanism of interaction has not been
    established.
```

## Root cause

`DdiDrugReferenceSource.noteFor` concatenated severity with the mechanism text as-is, so whatever
the mechanism field held became the note — and the note is both the chip detail
(`DrugSafetyValidator.addInteractions`) and the rendered, citable reference record
(`DrugReferenceInjector.render`). One line, two leaks. The mechanism text has exactly one
consumption point in the parser, so the fix belongs there and covers both.

## The marker family — scanned, not assumed

`INTERVAL:` was found by inspection, so the whole mechanism corpus was scanned before fixing
(the loaded 19MB KB at `appdata/chartsearchai/ddi_knowledge_base.json`, 8234 mechanisms).

| marker | mechanisms | pair rows |
|---|---|---|
| `INTERVAL:` | 224 | 4070 |
| `RECOMMENDED:` | 50 | 1268 |

Those are the only two. There is no other leading `TOKEN:` shape anywhere in the corpus — the one
other leading-colon text is ordinary prose ("Limited and controversial data suggest the
following:") — and neither marker ever appears anywhere except at the head of the text. Every
other all-caps token in the corpus is a real pharmacological abbreviation (BCRP, OATP, MAOI,
NSAID, SSRI…). `RECOMMENDED:` is new information: it was not in the issue, and 49 of its 50 rows
are contraceptive-efficacy monographs.

The `categories` field was checked first, as the issue asked. It does not already carry this
information: it is a PK/PD mechanism taxonomy (`metabolism` 3721, `synergistic_effect` 2115,
`others` 981, `absorption` 805, `antagonistic_effect` 391, `excretion` 373, `distribution` 126)
and it does not line up with the marker either — the 224 `INTERVAL:` rows are spread over every one
of those categories (147 `absorption`, 39 `others`, 15 `antagonistic_effect`,
9 `synergistic_effect`, 6 `metabolism`, 4 `distribution`, 2 `excretion`, 2 `absorption`+`metabolism`).

## What `INTERVAL:` means, and why the fix drops it rather than rendering it

It marks administration-timing / dose-separation guidance — the "adjust dosing interval" class of
advice: separate the doses rather than avoid the combination. For the dolutegravir/iron row that
genuinely is the clinical management, which is why the issue asked for it to be carried as
structured management guidance rather than deleted. (Its shape and the corpus both suggest it is
the surviving last word of a longer management tag from the monograph the mechanism text was
scraped from — that provenance is inference, not something I could verify offline; what is measured
is everything below.)

I did not do that, and the measurements are the reason:

- **It is not reliably a timing instruction.** 9 of the 224 rows are `synergistic_effect`
  additive-toxicity rows where separating doses is *not* the management: ibutilide plus a class
  III antiarrhythmic (additive QT prolongation), flibanserin plus alcohol (hypotension/syncope),
  mefloquine plus another antimalarial (convulsion risk), lidocaine plus tocainide (additive CNS
  effects). Rendering "separate administration times" on those chips would be clinically wrong,
  and a wrong safety chip is worse than a missing one.
- **The marker carries no content.** No interval, no separation time, no wording. Any management
  sentence would be composed here, not extracted — which is exactly what the class javadoc
  already rules out ("`management` is not a discrete DDInter field, so it is folded into the
  interaction note rather than invented").
- **The issue's supporting premise does not hold.** "These rows are precisely the ones whose
  mechanism ends 'The mechanism of interaction has not been established'" is true of 5 of the 224
  (33 of 224 use any "mechanism unknown" phrasing, against 804 of 8234 corpus-wide). The dolutegravir
  row is not representative; 219 of the 224 carry a real mechanism, so the marker is not carrying
  "the actionable half" of the note.
- **The real repair is upstream.** The way to surface management guidance properly is for the KB
  builder (`build_kb.py` in openmrs-ddi-knowledge-base) to carry the management tag as a structured
  field — not for this module to reverse-engineer meaning from a fragment of one.

**So, stated plainly: this PR removes the leak, and the timing signal is still not surfaced.**
A follow-up that adds a real `management` field should get it from the KB, not from the marker.

## What changed

`DdiDrugReferenceSource` only:

- A `RESIDUAL_FIELD_MARKER` pattern — a bounded run of all-caps words (two letters or more, at
  most six words) immediately followed by a colon — stripped from the head of the mechanism text
  before it becomes the note. Matched on marker **shape**, not on a list of the two members seen
  today: we just learned by scanning that the family had an undocumented second member, so a KB
  refresh emitting a sibling tag would leak it exactly as these two did. Verified against all 8234
  mechanisms: the pattern matches those 274 rows and nothing else, and every remainder still begins
  with a capital letter, so each stripped note reads as a sentence.
- A `mechanismText(mech, gid)` helper as the single place the mechanism text is resolved and
  cleaned. A text that is *only* a marker strips to empty and therefore falls through to the
  existing no-mechanism note rather than leaving a dangling "Major. ".
- The class javadoc records the scan, the counts, and why the marker is dropped rather than
  reinterpreted.

Nothing else is touched. The curated `json` and WHO `atc` sources carry no markers, and the
bundled 60-mechanism classpath sample has none either, so no existing output changes.

## Tests

New in `DdiDrugReferenceSourceTest`, all through the real parse/validate/inject paths against a
new real-slice fixture (`ddi-field-marker-mechanism.json` — the actual dolutegravir/iron
`INTERVAL:` row, the tazemetostat/ethinylestradiol `RECOMMENDED:` row, and the real drug rows):

- `residualFieldMarkersAreStrippedFromInteractionNotes` — both markers gone, both mechanism texts
  byte-for-byte intact (exact-equality assertion on the whole note).
- `markerOnlyMechanismDegradesToTheNoMechanismNote` — the degenerate edge.
- `interactionChipDetailIsFreeOfTheResidualFieldMarker` — real `DrugSafetyValidator.validate(...)`;
  exact-equality assertion on the chip detail.
- `injectedReferenceTextIsFreeOfTheResidualFieldMarker` — real `DrugReferenceInjector.injectRecords`,
  pinning the prompt-text leak path too.

All four fail on `13690b1` and pass with the fix.

`mvn install` (needed for the omod module — `mvn test` alone trips MDEP-98):
**API 666 + OMOD 53 = 719 run, 0 failures, 0 errors, 34 skipped.** Baseline was 662 + 53 = 715
with the same 34 skipped; the four new tests are the difference.

## Live standalone evidence

Deployed my omod to the shared 3.7.1 standalone (full-JVM restart, querystore retrieval gated
alive, LLM warmed on the known-positive control) and re-ran the #116 probe plus the four
regression controls in one pass. Chips are deterministic Java, so the assertions are on chip text;
answer prose is not reproducible on that box (`cache_prompt` KV reuse) and is not asserted.

**Melissa Wright `d6566336-1db1-42cb-a0c0-d60b5862d4c1` — "Can I start dolutegravir?"**

Before (`13690b1`), 2 chips:

```
[interaction] drug='Dolutegravir'
    Dolutegravir interacts with active order iron — Major. Coadministration with medications containing polyvalent cations such as aluminum, calcium, iron, or magnesium may decrease the oral bioavailability of dolutegravir.
[interaction] drug='Dolutegravir'
    Dolutegravir interacts with active order iron — Major. INTERVAL: Coadministration with medications containing polyvalent cations such as aluminum, calcium, iron, or magnesium may decrease the oral bioavailability of dolutegravir. The mechanism of interaction has not been established.
```

After (this branch), 2 chips:

```
[interaction] drug='Dolutegravir'
    Dolutegravir interacts with active order iron — Major. Coadministration with medications containing polyvalent cations such as aluminum, calcium, iron, or magnesium may decrease the oral bioavailability of dolutegravir.
[interaction] drug='Dolutegravir'
    Dolutegravir interacts with active order iron — Major. Coadministration with medications containing polyvalent cations such as aluminum, calcium, iron, or magnesium may decrease the oral bioavailability of dolutegravir. The mechanism of interaction has not been established.
```

The marker is gone and the mechanism text — including the trailing "The mechanism of interaction
has not been established." sentence the marker was fronting — is intact.

**Regression controls, same pass, all four as specified:**

```
C1 Mary 38beca4a + "Is it safe to give her clarithromycin?"  => 1 chip
  [interaction] Clarithromycin
      Clarithromycin interacts with active order simvastatin — Major. Coadministration with potent
      inhibitors of CYP450 3A4 may significantly increase the plasma concentrations of simvastatin
      and lovastatin and their active acid metabolites, all of which are primarily metabolized by
      the isoenzyme.

C2 Agnes 47e57b75 + "Is it safe to start warfarin?"           => 1 chip
  [interaction] Warfarin
      Warfarin interacts with active order aspirin — Major. Aspirin, even in small doses, may
      increase the risk of bleeding in patients on oral anticoagulants by inhibiting platelet
      aggregation, prolonging bleeding time, and inducing gastrointestinal lesions. ...

C3 Joshua 9cb37bcb + "Can I give ibuprofen?"                  => 2 chips
  [contraindication] Ibuprofen
      Ibuprofen is in the same cross-reactivity group (NSAID) as the patient's allergy to
      Acetylsalicylic acid (aspirin) — possible cross-reactivity
  [interaction] Ibuprofen
      Ibuprofen interacts with active order lisinopril — Moderate. Nonsteroidal anti-inflammatory
      drugs (NSAIDs) may attenuate the antihypertensive effects of ACE inhibitors. ...

C4 Mary 38beca4a + "Is it safe to give her paracetamol?"      => 0 chips
```

The probe script asserted marker-absence and chip counts across all five probes and reported
`VERDICT: PASS`.

One honest caveat on the scope of the live check: the REST `references` payload carries only
`index`/`resourceType`/`resourceUuid`/`date`/`grounded`/`group` — not the record text — so the
probe's marker check on injected records is vacuous, and the prompt-text leak path cannot be
inspected directly from a probe at all. That path is pinned deterministically by
`injectedReferenceTextIsFreeOfTheResidualFieldMarker`, and corroborated on the live box only
indirectly: the answer reproduced the note twice, both times without the marker.

## Notes

- The probe yields **two** Dolutegravir chips, before and after. That duplicate is #115 (two
  DDInter route-variant rows for the same pair, both with the match token `iron`) and is being
  fixed separately — it is untouched here.
- I created no data on and changed no settings of the standalone for this verification. The
  iron order that makes this pair fire on Melissa Wright was already on her chart.

Closes #116

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
@@ -65,10 +65,12 @@ import com.fasterxml.jackson.databind.ObjectMapper;
  * KB (2026-08-04) there are exactly two: {@code INTERVAL:} (224 mechanisms, 4070 pair rows) and
  * {@code RECOMMENDED:} (50 mechanisms, 1268 pair rows); no other leading {@code TOKEN:} shape
  * occurs, and both markers appear only in that leading position. They are machine artifacts, not
- * prose, and the note is read by clinicians (chip detail) and by the model (rendered reference
- * record) alike, so {@link #noteFor} strips a leading marker instead of passing it through
- * (issue #116). Marker <em>shape</em>, not a fixed list of the two seen today: a KB refresh
- * emitting a sibling tag would otherwise leak it verbatim exactly as these two did.
+ * prose, and the note reaches three surfaces verbatim — the clinician's chip detail, the rendered
+ * reference record, and the pre-answer safety finding that reuses that chip detail
+ * ({@link DrugReferenceInjector#renderFinding}) — so {@link #noteFor} strips a leading marker
+ * instead of passing it through (issue #116). Marker <em>shape</em>, not a fixed list of the two
+ * seen today: a KB refresh emitting a sibling tag would otherwise leak it verbatim exactly as
+ * these two did.
  *
  * <p>The marker is dropped rather than reinterpreted. {@code INTERVAL:} broadly flags
  * administration timing (dose separation is the management for the chelation/absorption rows —
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
@@ -17,6 +17,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.openmrs.module.chartsearchai.ChartSearchAiConstants;
 import org.slf4j.Logger;
@@ -56,6 +58,29 @@ import com.fasterxml.jackson.databind.ObjectMapper;
  * never {@code ageBands} or {@code contraindications} (dosing and drug-allergy/condition are
  * out of scope). {@code management} is not a discrete DDInter field, so it is folded into the
  * interaction note rather than invented.
+ *
+ * <p><b>Residual field markers.</b> Some mechanism texts are prefixed with an all-caps field
+ * marker followed by a colon — apparently the surviving tail of a management tag from the
+ * monograph the mechanism text was scraped from. Measured over the full 8234-mechanism
+ * KB (2026-08-04) there are exactly two: {@code INTERVAL:} (224 mechanisms, 4070 pair rows) and
+ * {@code RECOMMENDED:} (50 mechanisms, 1268 pair rows); no other leading {@code TOKEN:} shape
+ * occurs, and both markers appear only in that leading position. They are machine artifacts, not
+ * prose, and the note is read by clinicians (chip detail) and by the model (rendered reference
+ * record) alike, so {@link #noteFor} strips a leading marker instead of passing it through
+ * (issue #116). Marker <em>shape</em>, not a fixed list of the two seen today: a KB refresh
+ * emitting a sibling tag would otherwise leak it verbatim exactly as these two did.
+ *
+ * <p>The marker is dropped rather than reinterpreted. {@code INTERVAL:} broadly flags
+ * administration timing (dose separation is the management for the chelation/absorption rows —
+ * 147 of the 224 are {@code categories: [absorption]}), but it is not reliable enough to render
+ * as advice: nine of the 224 are {@code synergistic_effect} rows where separating doses is
+ * <em>not</em> the management (ibutilide plus a class III antiarrhythmic — additive QT
+ * prolongation; flibanserin plus alcohol; mefloquine convulsion risk). The marker itself carries
+ * no interval, no separation time and no wording, so any management sentence built from it would
+ * be invented, and the mechanism {@code categories} are a PK/PD taxonomy
+ * ({@code metabolism}/{@code absorption}/…) that does not encode management either. Structured
+ * management guidance therefore has to come from the KB builder keeping the whole upstream tag,
+ * not from reverse-engineering its truncated residue here.
  */
 public class DdiDrugReferenceSource implements DrugReferenceSource {
 
@@ -66,6 +91,17 @@ public class DdiDrugReferenceSource implements DrugReferenceSource {
 	private static final ObjectMapper MAPPER = new ObjectMapper();
 
 	private static final String SOURCE = "DDInter 2.0 (via openmrs-ddi-knowledge-base)";
+
+	/**
+	 * A residual field marker at the head of a mechanism text: a run of up to six all-caps words
+	 * immediately followed by a colon (see the class javadoc). Words are two letters or more so a
+	 * bare initial cannot look like a marker, and the run is bounded so a shouted sentence ending
+	 * in a colon is not mistaken for one. Verified against the full 8234-mechanism KB: it matches
+	 * exactly the 274 {@code INTERVAL:}/{@code RECOMMENDED:} rows and nothing else, and every
+	 * remainder still begins with a capital, so the stripped note reads as a sentence.
+	 */
+	private static final Pattern RESIDUAL_FIELD_MARKER = Pattern
+			.compile("^\\s*[A-Z]{2,}(?: [A-Z]{2,}){0,5}:\\s*");
 
 	@Override
 	public List<DrugReference> load() {
@@ -193,12 +229,28 @@ public class DdiDrugReferenceSource implements DrugReferenceSource {
 		if (cached != null) {
 			return cached;
 		}
-		String text = mech.path(gid).path("text").isTextual() ? mech.path(gid).path("text").asText() : null;
+		String text = mechanismText(mech, gid);
 		String note = (text != null && !text.isEmpty())
 				? severity + ". " + text
 				: severity + " severity interaction (DDInter 2.0; no mechanism description on file).";
 		cache.put(key, note);
 		return note;
+	}
+
+	/**
+	 * The mechanism description for {@code gid}, with any leading residual field marker removed
+	 * (see the class javadoc), or {@code null} when the group carries no text. A text that is
+	 * <em>only</em> a marker strips to empty and so degrades to the no-mechanism note in
+	 * {@link #noteFor} — a dangling "Major. " would read worse than saying nothing is on file.
+	 */
+	private static String mechanismText(JsonNode mech, String gid) {
+		JsonNode node = mech.path(gid).path("text");
+		if (!node.isTextual()) {
+			return null;
+		}
+		String text = node.asText();
+		Matcher marker = RESIDUAL_FIELD_MARKER.matcher(text);
+		return marker.lookingAt() ? text.substring(marker.end()) : text;
 	}
 
 	/** A drug row from the {@code drugs} table. */
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
@@ -56,8 +56,10 @@ import com.fasterxml.jackson.databind.ObjectMapper;
  *
  * <p><b>Scope.</b> V1 carries drug-drug interactions only: entries expose {@code interactions},
  * never {@code ageBands} or {@code contraindications} (dosing and drug-allergy/condition are
- * out of scope). {@code management} is not a discrete DDInter field, so it is folded into the
- * interaction note rather than invented.
+ * out of scope). {@code management} is not a discrete DDInter field, so whatever management prose
+ * the mechanism text carries is folded into the interaction note rather than invented — save for
+ * the residual field markers below, which are dropped because they carry no management content
+ * to fold.
  *
  * <p><b>Residual field markers.</b> Some mechanism texts are prefixed with an all-caps field
  * marker followed by a colon — apparently the surviving tail of a management tag from the
@@ -67,10 +69,10 @@ import com.fasterxml.jackson.databind.ObjectMapper;
  * occurs, and both markers appear only in that leading position. They are machine artifacts, not
  * prose, and the note reaches three surfaces verbatim — the clinician's chip detail, the rendered
  * reference record, and the pre-answer safety finding that reuses that chip detail
- * ({@link DrugReferenceInjector#renderFinding}) — so {@link #noteFor} strips a leading marker
- * instead of passing it through (issue #116). Marker <em>shape</em>, not a fixed list of the two
- * seen today: a KB refresh emitting a sibling tag would otherwise leak it verbatim exactly as
- * these two did.
+ * ({@link DrugReferenceInjector#renderFinding}) — so {@link #mechanismText} strips a leading
+ * marker instead of passing it through (issue #116). Marker <em>shape</em>, not a fixed list of
+ * the two seen today: a KB refresh emitting a sibling tag would otherwise leak it verbatim
+ * exactly as these two did.
  *
  * <p>The marker is dropped rather than reinterpreted. {@code INTERVAL:} broadly flags
  * administration timing (dose separation is the management for the chelation/absorption rows —
--- a/api/src/test/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSourceTest.java
+++ b/api/src/test/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSourceTest.java
@@ -30,6 +30,19 @@ public class DdiDrugReferenceSourceTest {
 
 	private static final String SEVERITY = "Major Moderate Minor Unknown";
 
+	private static final String MARKER_FIXTURE = "chartsearchai-test/ddi-field-marker-mechanism.json";
+
+	/** The real mechanism text of KB group 2248, verbatim minus the {@code INTERVAL:} marker. */
+	private static final String DOLUTEGRAVIR_MECHANISM = "Coadministration with medications containing "
+			+ "polyvalent cations such as aluminum, calcium, iron, or magnesium may decrease the oral "
+			+ "bioavailability of dolutegravir. The mechanism of interaction has not been established.";
+
+	/** The real mechanism text of KB group 222, verbatim minus the {@code RECOMMENDED:} marker. */
+	private static final String TAZEMETOSTAT_MECHANISM = "Coadministration with tazemetostat may decrease "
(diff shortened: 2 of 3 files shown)


dkayiwa:
## Hardening pass (`harden`): three gaps at the edges of the strip

Pushed `e2aac05` and `e86635f`. **No production behaviour changed** — every edit to
`DdiDrugReferenceSource` is javadoc. Proven rather than asserted: the production diff against
`c26ecfa` contains **0 non-comment lines**, and compiling `c26ecfa` and `e86635f` with an identical
`javac -g:none` invocation yields **byte-identical** `.class` files for `DdiDrugReferenceSource`,
`$DrugRow` and `$Link`. (With debug info on they differ only in `LineNumberTable` offsets, because the
javadoc grew.)

The fix itself is untouched and I agree with all four of its load-bearing decisions.

### 1. The PR text and javadoc were both one surface out of date

This PR says "One line, two leaks" and names the chip detail and the rendered reference record. Since
**#110** (`13690b1` — this PR's own parent) there are **three**: `DrugReferenceInjector.renderFinding`
reuses the chip detail *verbatim* as a citable `safety_finding` record, and that is the line the model
is steered to report before anything else. So the marker was also landing in the highest-signal line of
the prompt. The strip always covered it — it sits at the single consumption point — but nothing said so
and nothing tested it.

It could not have been caught by the existing reference-record test: that test wires no validator, so
`preAnswerFindings` returns empty and the finding line never appears in it. Added
`injectedSafetyFindingIsFreeOfTheResidualFieldMarker`, which wires the real validator into the real
injector. Verified red against `13690b1`'s parser:

```
Safety finding — Dolutegravir: Dolutegravir interacts with active order iron — Major. INTERVAL: Coadministration ...
```

### 2. Nothing pinned what the pattern must *not* eat

Stripping is the only place this source deletes clinician-facing text, and over-matching fails
**silently** — no exception, just a note that opens mid-sentence. The existing tests all cover the
"strip" branch; the negative space was untested.

This PR already noticed the row that matters ("the one other leading-colon text is ordinary prose").
Added `mechanismTextOutsideTheMarkerShapeIsPassedThroughUnstripped` to make that observation
executable, using that real row — KB group **4945** (nifedipine × theophylline, Minor).

### 3. …and my own first attempt at (2) overclaimed, so here is the correction

Worth flagging plainly, because the first version of that test asserted something measurably false. I
wrote that gid 4945 is "what the six-word bound buys — loosen the run and this note is beheaded." It
is not. Measured against the real KB, **every single-axis loosening still matches exactly the same 274
rows and spares 4945**:

| pattern variant | mechanisms matched | 4945 eaten? |
|---|---|---|
| shipped `[A-Z]{2,}(?: [A-Z]{2,}){0,5}:` | 274 | no |
| drop the run bound **only** | 274 | no |
| drop the case constraint **only** | 274 | no |
| drop the 2-letter minimum **only** | 274 | no |
| drop case **and** bound together | 275 | **yes** |

4945's clause is *both* mixed-case *and* seven words, so it is protected twice over and cannot isolate
any one guard — it pins the conjunction. Comment and fixture note now say exactly that.

That left the bound the javadoc explicitly names as a guard ("the run is bounded so a shouted sentence
ending in a colon is not mistaken for one") **completely unobservable**: the longest all-caps run in
the real KB is one word, so deleting `{0,5}` kept the whole suite green. A later simplify pass could
have removed it as arbitrary-looking and silently reopened the beheading this slice exists to bound.

So added `allCapsRunLongerThanTheMarkerBoundIsNotTreatedAsAMarker` over new synthetic group **9997** —
seven ALL-CAPS words then a colon, the shape no real row supplies (same "the KB file is
operator-editable" justification as the existing synthetic 9998). Mutation-checked: dropping `{0,5}`
**alone**, case and minimum untouched, fails this test and *only* this test, beheading the note to

```
Minor. coadministration may increase the risk of theophylline toxicity.
```

### Also fixed (javadoc/comment accuracy only)

- The strip was attributed to `{@link #noteFor}`; `mechanismText` is what strips.
- The Scope paragraph ("`management` … is folded into the interaction note") sat immediately above the
  paragraph explaining the one residual management tag is deliberately *deleted*; read in order they
  said opposite things. One clause reconciles them, with the rationale intact.
- The test class javadoc claimed everything runs off the bundled sample; 8 of 17 tests now parse a
  fixture through the same real parser.
- `sharedRxcuiDoesNotCollapseEntryIds` inlined the fixture loader `markerFixtureEntries` duplicated;
  both now call `ddiFixtureEntries`. No assertion changed — the resource-presence message now names
  the resource, the convention `DrugReferenceTestSupport.fixtureEntries` already uses.

### Re-measured every claim in the javadoc independently

Against the same full 19 MB KB (8234 mechanisms, 295 184 pair rows), from scratch:

| claim | result |
|---|---|
| `INTERVAL:` 224 mechanisms / 4070 pair rows | ✅ exact |
| `RECOMMENDED:` 50 mechanisms / 1268 pair rows | ✅ exact |
| pattern matches those 274 **and nothing else** | ✅ exact |
| every remainder still begins with a capital | ✅ 0 exceptions |
| both markers occur **only** in leading position | ✅ 224/224 and 50/50; 0 mid-text, and 0 all-caps+colon tokens anywhere non-initial |
| no real marker row strips to empty | ✅ 0 — so the fixture's degenerate row stays correctly labelled synthetic |
| 147 of 224 `INTERVAL:` are `[absorption]`, 9 `synergistic_effect` | ✅ exact |

Three further facts, none previously written down: **zero** mechanism texts carry leading/trailing
whitespace (so the pattern's leading `\s*` has no asymmetric effect on real data); 2 mechanisms have
`text: null` (gids `-1` and `7183`, 14.4 % of pair rows — matching the README's "14 %"), falling
through to the no-mechanism note exactly as before; and post-strip **25** mechanism groups collide on
identical note content with a marker row, which is benign because `noteFor` caches per
`(severity, gid)` not per content and nothing keys or dedups on note content (`SafetyWarning` has no
`equals`/`hashCode`).

I also tried to falsify the 9-row additive-toxicity argument for dropping rather than rendering the
marker, and it holds. **The PR's statement that the timing signal is still unsurfaced and that a real
`management` field belongs upstream in the KB builder stands unchanged.**

### Tests

`mvn clean install` from the repo root: **API 669 + OMOD 53 = 722, 0 failures, 0 errors, 34 skipped**
(was 666 + 53 = 719 on this branch; 662 + 53 = 715 on `main`). Three new tests are the difference.

### Live re-verification: not required, and why

Required only if production behaviour could move. The compiled production class is provably
byte-identical to `c26ecfa`, the commit that already passed the full live pass (#116 probe + all four
regression controls), and the fixture is a **test** resource — confirmed absent from the shipped
`.omod`. Re-probing would exercise the same bytes while holding a lock several agents are queued
behind.

Two analytical re-assertions instead. All three interaction controls carry **no** marker in the real KB
(clarithromycin × simvastatin gid 2085, warfarin × aspirin gid 749, ibuprofen × lisinopril gid 3643),
so their chip text is unchanged by construction. And the #116 headline resolves cleanly:
`Dolutegravir × Iron` (`DDInter975`) is gid **2248**, the `INTERVAL:` row; `Iron (bisglycinate) ×
Dolutegravir` (`DDInter2187`) is gid **7199**, unmarked. Both carry `rxnorm_name = "iron"`, hence the
same match token and the two chips — that is **#115**, confirmed as a separate defect and still
untouched here.

The **prompt-text** leak path remains not live-verifiable: the REST `references` payload carries
`index`/`resourceType`/`resourceUuid`/`date`/`grounded`/`group` but no record text. It is pinned
deterministically by `injectedReferenceTextIsFreeOfTheResidualFieldMarker` and now
`injectedSafetyFindingIsFreeOfTheResidualFieldMarker`, and corroborated live only indirectly.

### Deferred, with the failure mode stated

- **The strip is unobservable — worth a follow-up issue.** Shape-matching is the right call, but
  nothing logs *which* markers were eaten. If an operator points `dataFilePath` at a newer
  `openmrs-ddi-knowledge-base` build whose scraper keeps a third tag, that tag is deleted from all
  three surfaces with no log line and no artifact to notice it by, and the javadoc's measured "exactly
  two markers / 274 rows" becomes quietly false. Mitigating factor: the strip removes only the tag and
  keeps the text after it, so the clinical content survives; the sole total-loss case is a marker-only
  row, which degrades visibly to the no-mechanism note. The fix — count strips per marker text and
  append to the existing `Parsed {} DDInter drug-reference entries` log line — is a production change
  and belongs in its own PR, not in a hardening pass on this one.
- **`noteFor`'s blank guard is `!text.isEmpty()`, not a blank check.** A whitespace-only mechanism text
  would still produce a dangling `"Major.   "`, which `mechanismText`'s own javadoc says must be
  avoided. Pre-existing, and unreachable on real data (zero whitespace-only rows, and the strip's
  greedy trailing `\s*` cannot create one), but reachable on an operator-edited KB by the same argument
  that justifies synthetic 9998. Left out because it is a production change with its own
  live-verification cost, for an input no measured dataset contains.

### Found outside the slice, not fixed

`mvn javadoc:javadoc` fails on `QueryScopeContributor.java:22` (`heading used out of sequence: <H3>`),
pre-existing on `main` and untouched here. It gates nothing — no javadoc plugin is configured in either
pom and CI does not run it — so it is latent only.

Issue #113, opened 2026-08-04
The generic DDI screening question ("any interactions with her current medications?") produces nothing — both arms require a question-named drug


dkayiwa:
### Summary

The canonical drug-interaction screening question — "does this patient have any interactions?" — returns nothing at all. The whole feature only engages when the clinician has already guessed which drug to ask about.

Measured on the 3.7.1 standalone (HEAD `13690b1`), Mary Smith, active order **Simvastatin 20mg**:

| question | chips | references | answer |
|---|---|---|---|
| Are there any drug interactions with her current medications? | 0 | `[]` | "The records do not address drug interactions with current medications." |
| Does simvastatin interact with any of her current medications? | 0 | `[]` | "The records do not address the patient's current medications." |
| Is it safe to give her clarithromycin? | 1 (Major) | 2 | "Clarithromycin interacts with active order simvastatin — Major. …" |

Same patient, same KB, same request path. Naming a candidate drug works; asking to be screened does not.

### Root cause

Both halves of the feature key off the drugs the **question** names, and neither has a path for "no drug named".

`DrugReferenceInjector.matchingEntries` (:217) computes `questionDrugs` first, then scopes order-driven injection to it (:242):

```java
for (DrugReference ref : drugReferenceService.findByActiveOrders(context)) {
    if (relatedToAny(ref, questionDrugs, groups)) {   // false for an empty questionDrugs
        byId.put(ref.getId(), ref);
    }
}
```

`relatedToAny` (:255) loops over `questionDrugs`, so an empty list returns `false` and nothing is injected. That is deliberate and documented — an unrelated active medication would be noise — but the guard treats "the question names no drug" and "the question names an unrelated drug" identically, and only the second is noise.

`DrugSafetyValidator.validate` then has nothing to work on: `questionDrugs` is empty (:207), so `inPlay` starts empty and can only be filled by drugs the *answer* names on its own authority — and the answer, having been given no reference records, names none. Zero chips.

The information needed is in hand either way: `PatientClinicalContext` already holds the active orders, and each order's `DrugReference` already carries its full partner list. Nothing needs to be retrieved to answer the question — only cross-referenced.

### Scope

[...]

Issue #105, opened 2026-07-28
DrugSafetyValidator: answer-recitation cascade — every reference drug the answer quotes from an injected record gets validated as if proposed


dkayiwa:
### Summary

`DrugSafetyValidator.validate` builds its set of drugs to check as the union of what the question resolves to and what the answer names:

```java
Set<DrugReference> inPlay = new LinkedHashSet<DrugReference>(drugReferenceService.findByQuery(question));
inPlay.addAll(drugReferenceService.findByQuery(answer));
```

The answer side is deliberate (the javadoc explains it decouples the safety net from the LLM's word choice), and it should stay. But it now composes badly with the injection path: the injected drug-reference record renders partner-drug names into the prompt, the answer recites some of those names back, each recited name resolves to a reference entry of its own, and each of those entries is then validated against the patient's active orders. Every recited partner that happens to have a row against an active order raises its own chip, for a drug nobody proposed giving the patient.

With the curated seed this was hard to hit, because an answer rarely named other reference drugs. With the DDInter source (2,283 entries) nearly every drug name an answer can utter is an entry, and enumeration-shaped questions ("what interacts with X?") are exactly the questions whose answers list drug names. The clinically relevant chip gets buried under the recitation.

### Reproduction (real pipeline, standalone, main @ 83cc33e)

`sourceFormat=ddinter`, full `ddi_knowledge_base.json` (openmrs-ddi-knowledge-base @ c1bb6fe), `warnOnInteractions=true`.

**Case 1.** Patient with one active order, Aspirin 81mg. Question: `Is it safe to give her ibuprofen?`
The answer recited the start of the injected ibuprofen record's partner list. Result: 6 chips. Two are relevant (ibuprofen x aspirin Major, plus its NSAID-group chip, which is #88's pair). The other four name drugs from the recitation: Kanamycin x aspirin (Moderate), Ketoconazole x aspirin (Unknown), Ketoprofen x aspirin (Moderate) and Ketoprofen's NSAID-group chip.

**Case 2.** Patient with one active order, Simvastatin 20mg. Question: `what other drugs are contraindicated with erythromycin?`
The answer listed partners off the injected erythromycin record (827 partners, render-capped). Result: 8 chips. One is the actionable warning (erythromycin x simvastatin, Major, CYP3A4). The other seven are recited partners, each validated against her simvastatin: ivosidenib, aprepitant, verapamil, crizotinib, diltiazem, dronedarone, amprenavir.

[...]

PR #120, opened 2026-08-04
fix(drug-safety): check question-named drugs against each other, not only against the chart — closes #114

### LINKED ISSUE #114: A drug pair the KB rates Major is denied when asked about directly — the safety join never checks question-named drugs against each other
### Summary

Asked directly about a drug pair the knowledge base rates **Major**, the module says it has no information — and raises an unrequested **Minor** chip about a different pair instead.

Mary Smith, active order Simvastatin 20mg, on the 3.7.1 standalone (HEAD `13690b1`):

> **Q:** Does warfarin interact with aspirin?
> **A:** "The records do not address the interaction between warfarin and aspirin."
> **chips:** 1 — `[interaction] Warfarin` — "Warfarin interacts with active order simvastatin — **Minor**. Simvastatin may slightly increase the anticoagulant response to warfarin…"
> **references:** `[]`

The pair asked about is in the loaded KB:

```
["DDInter1951", "DDInter20", "Major", "749"]     # warfarin x acetylsalicylic acid
```

and it fires correctly for a patient who happens to be **on** aspirin — Agnes Adams, active order Aspirin 81mg, asked "Is it safe to start warfarin?" → "Warfarin interacts with active order aspirin — Major. Aspirin, even in small doses, may increase the risk of bleeding…". Same build, same KB, same question drug. The only difference is whether the second drug is on the chart.

### Root cause

`DrugSafetyValidator.addInteractions` raises a chip only when the rule's partner is one of the patient's **active orders** (:391):

```java
if (context.hasActiveDrug(i.getToken(), i.getAtc())) {
```

Both warfarin and aspirin resolve to reference entries from the question — `findByQuery` returns both, and both land in `inPlay` — but the drugs in `inPlay` are each checked against the *chart* and never against **each other**. There is no pairwise arm. So a two-drug question is silently reduced to two independent one-drug questions.

The prompt path does not rescue it either. Warfarin's rendered record is capped at `MAX_INTERACTION_RENDER_CHARS` (1500) and ordered patient-relevance-first — with aspirin not on the chart it has no promotion claim, so it competes for the cap on dataset position among ~934 partners. `references: []` shows the model cited nothing and abstained.

Worth noting what the chip did instead: it answered a question nobody asked (warfarin × simvastatin, Minor, clears the `minor` floor) while the Major pair the clinician named went unmentioned. That is a worse failure than silence — it reads as "here is the interaction you asked about".

### Scope

Shared validator logic, all source formats. Related to but distinct from #113 (which is the *no drug named* case); this is the *two drugs named* case. Both come from the same underlying shape — the safety join is only ever question-drug against chart — so they may want to be fixed together.

### Proposed fix

Evaluate `inPlay` pairwise as well as against the chart: for each pair of question-resolved drugs, check either one's interaction list for the other and raise a chip when it matches. The chip wording needs to distinguish the two provenances — an active-order interaction is a fact about *this patient*, a question-pair interaction is a reference lookup that may involve no drug they take — so it should not say "active order".

Apply the same severity floor, so this cannot become a route around a decision the chip path enforces.

### Tests

A `validate()` test through the real production path: a question naming two drugs that interact, on a patient taking **neither**, asserts one chip naming that pair. Plus the negative — two question-named drugs with no interaction row between them yields no chip — and a wording assertion that a question-pair chip does not claim an active order.

dkayiwa:
## The defect

Asked directly about a drug pair the knowledge base rates **Major**, the module said it had no
information — and raised an unrequested **Minor** chip about a different pair instead.

Mary Smith, active order Simvastatin 20mg, 3.7.1 standalone at `13690b1`:

> **Q:** Does warfarin interact with aspirin?
> **A:** "The records do not address the interaction between warfarin and aspirin."
> **chips:** 1 — "Warfarin interacts with active order simvastatin — **Minor**. …"
> **references:** `[]`

The pair asked about is in the loaded KB (`["DDInter1951","DDInter20","Major","749"]`) and fires
correctly for Agnes Adams, who is actually on aspirin. The only difference was whether the second
drug happened to be on the chart.

## Root cause

`DrugSafetyValidator.addInteractions` raises an interaction chip only when the rule's partner is one
of the patient's **active orders**:

```java
if (context.hasActiveDrug(i.getToken(), i.getAtc())) {
```

Both warfarin and aspirin resolve from the question and both land in `inPlay`, but every interaction
arm joins one drug to the patient's own data — the drugs in `inPlay` were each checked against the
chart and never against **each other**. There was no pairwise arm, so a two-drug question was
silently reduced to two independent one-drug questions.

The prompt path did not rescue it either: warfarin's rendered record is capped at
`MAX_INTERACTION_RENDER_CHARS` and ordered patient-relevance-first, so with aspirin not on the chart
it had no promotion claim among ~934 partners. `references: []` is the model citing nothing.

## What changed

One production file, `DrugSafetyValidator`: a pairwise arm over the drugs the question resolved to
(`addQuestionPairInteractions`) plus three small helpers.

- **Each pair once.** Unordered pairs only. DDInter rows are symmetric and the parser writes every
  pair into both drugs' entries, so walking ordered pairs would report one interaction twice, once
  from each side. Either side's rule can raise it (`aboveFloorRuleAgainst` is asked in both
  directions), so asymmetric curated data still works.
- **Same floor, one definition.** The arm is handed the same `severityFloor` the chart arm gets from
  `configuredSeverityFloor()` and decides with `clearsSeverityFloor` — no second definition of the
  floor or of severity. A sub-floor pair raises nothing, so this is not a route around the decision
  the chip path enforces.
- **The active-order arm keeps precedence, and the wording distinguishes provenance.** A pair either
  of whose drugs is an active order is left to `addInteractions`, whose chip states a fact about
  *this* patient — the stronger statement, and raising both would say one finding in two voices. What
  the new arm raises therefore never says "active order"; it reads
  `"<Drug> interacts with <Partner>, also named in the question — <severity>. <mechanism>"`. That
  claim is about the **question**, so it stays true whatever the chart holds. Wording it as "neither
  drug is on the chart" would have been false for the one-sided-data case that can reach this arm
  (the patient is on one of the pair, but only that drug's entry carries the rule, so no
  active-order chip fires from either side) — and misstating a patient's medications is #86.
- **Scoped to the question's drugs, not all of `inPlay`.** The issue proposed "evaluate `inPlay`
  pairwise"; the arm deliberately covers only the *question-resolved* subset. Two drugs an **answer**
  names are as often alternatives ("ibuprofen, or paracetamol if …") as a proposed combination, so
  pairing them would assert a co-administration risk nobody proposed — the over-reach echo scoping
  exists to prevent (#105). It also keeps chips aligned with the injected findings:
  `DrugReferenceInjector.preAnswerFindings` runs this same `validate` with an *empty* answer, so an
  answer-side pair could produce a chip with no record behind it.
- **Patient-specific findings still lead.** The new arm runs after the per-drug arms, so an allergy
  or active-order chip — a fact about this patient — is never pushed below a reference lookup.
- **No change to `render` / `orderedInteractionNotes`.** The issue asks whether the partner should
  also be promoted into the rendered record so the prose has grounding. It does not need to be: since
  #110 the deterministic finding is itself injected as a numbered, citable record, and
  `preAnswerFindings` calls the very `validate` this arm lives in — so the pair finding becomes a
  record the model can report, carrying the chip's string verbatim. `orderedInteractionNotes`'
  documented invariant is therefore untouched: its promotion predicate still mirrors the
  active-order chip decision exactly. The live evidence below shows this working — P1's `references`
  went from `[]` to a cited `safety_finding` record.

## Live standalone evidence

3.7.1 standalone, full 19MB DDInter KB, `sourceFormat=ddinter`, `minInteractionSeverity=minor`,
`chartMode=fullChart`, local LLM. Full-JVM restart on this build; querystore retrieval proven alive
(`hits=75`) and the known-positive control fired before any capture was taken. Chips are
deterministic Java; answer prose on this box is not reproducible (`cache_prompt` KV reuse), so the
assertions below are on chips and their wording.

**Before** (HEAD `13690b1`) — Mary Smith + "Does warfarin interact with aspirin?":

```
--- ANSWER ---
The records do not address the interaction between warfarin and aspirin.
--- safetyWarnings (1) ---
  [interaction] drug='Warfarin'
      Warfarin interacts with active order simvastatin — Minor. Simvastatin may slightly increase the
      anticoagulant response to warfarin. …
--- references (0) ---
```

**After** — same patient, same question:

```
--- ANSWER ---
Warfarin interacts with Acetylsalicylic acid (aspirin) — Major. Aspirin, even in small doses, may
increase the risks of bleeding in patients on oral anticoagulants by inhibiting platelet aggregation,
prolonged bleeding time, and inducing gastrointestinal lesions [79].
--- safetyWarnings (2) ---
  [interaction] drug='Warfarin'
      Warfarin interacts with active order simvastatin — Minor. Simvastatin may slightly increase the
      anticoagulant response to warfarin. …
  [interaction] drug='Warfarin'
      Warfarin interacts with Acetylsalicylic acid (aspirin), also named in the question — Major.
      Aspirin, even in small doses, may increase the risk of bleeding in patients on oral
      anticoagulants by inhibiting platelet aggregation, prolonging bleeding time, and inducing
      gastrointestinal lesions. …
--- references (2) ---
  #79 type=safety_finding group=reference grounded=True
```

The pre-existing Minor active-order chip is untouched — it is a true fact about Mary and suppressing
it is a relevance question, not this one.

**Independent of the chart** — Susan Young (active order Tiotropium 18mg only, on *neither* named
drug), same question:

```
--- safetyWarnings (1) ---
  [interaction] drug='Warfarin'
      Warfarin interacts with Acetylsalicylic acid (aspirin), also named in the question — Major. …
```

**Precedence** — Agnes Adams (active order Aspirin 81mg), same question. The patient-specific chip
wins and the pair is *not* also reported as a reference lookup:

```
--- safetyWarnings (1) ---
  [interaction] drug='Warfarin'
      Warfarin interacts with active order aspirin — Major. …
```

**No false pair chips** — Mary Smith + "Does ascorbic acid interact with aspirin?" (both drugs
resolve; the KB carries no row for the pair):

```
--- ANSWER ---
The records do not address the interaction between ascorbic acid and aspirin.
--- safetyWarnings (0) ---
--- references (0) ---
```

**Regression controls, all unchanged:**

| control | result |
|---|---|
| Mary + "Is it safe to give her clarithromycin?" | 1 chip — `Clarithromycin interacts with active order simvastatin — Major.` |
| Agnes + "Is it safe to start warfarin?" | 1 chip — `Warfarin interacts with active order aspirin — Major.` (active-order wording, **not** the new pair wording) |
| Joshua + "Can I give ibuprofen?" | 2 chips — NSAID cross-reactivity contraindication + `Ibuprofen interacts with active order lisinopril — Moderate.` |
| Mary + "Is it safe to give her paracetamol?" | 0 chips |

No test data was created or modified on the standalone for this verification.

## Hardening (post-review)

Review found two ways this arm reported one clinical fact **twice**. Both are silent — wrong output,
no warning — and both are now red-first tested through the real `validate(...)` seam.

**1. Route variants multiplied the chips.** DDInter carries a separate entry per route variant and
every variant publishes the same `rxnorm_name` — which is the token its rules match on — so ONE word
in the question resolves several entries (`findByQuery` returns every entry whose aliases match) that
all pair off the same rule. On the full KB, `dexamethasone` is four entries (plain, nasal,
ophthalmic, topical) and voxelotor carries rows against three of them, so "does voxelotor interact
with dexamethasone?" raised a chip per resolved entry — and a near-identical injected
`safety_finding` per chip — for a single pair. That is #115's measured shape (142 `rxnorm_name`
values shared by more than one entry across 332 entries) arriving on the **question** side, which
this PR would have shipped. Pairs are now keyed by the two drugs' **match tokens** rather than by
their entries: the token is the reference data's own canonical name for a drug, and precisely what
the variants share.

De-duplicating inside a safety net is only safe where nothing actionable is lost, so the key is
argued in both directions and both are asserted. Two entries share a key only when the data names
them identically, in which case one rule identifies both — the surviving chip carries that very
rule's severity and mechanism and all that is dropped is a second spelling of the partner. And three
drugs named in one question still raise three chips. The residual (a KB naming two genuinely
different drugs with one token, separated only by ATC code) is written down, with why the chart arm
cannot resolve it either: it reads the same token to decide which order a rule names. Which variant's
row supplies the surviving severity stays #115's open half.

**2. Chart precedence was decided from one rule, not all of them.** `addInteractions` walks every
rule an entry carries. When two rules join one pair under different tokens — a brand-name row and an
INN row — and only the second names the patient's active order, consulting the first alone concluded
"the chart does not own this pair", and the pair was reported twice: once as a fact about the patient
(*"active order warfarin"*), once as a reference lookup (*"also named in the question"*).
`aboveFloorRulesAgainst` now returns every joining rule and the precedence check asks all of them,
which is what makes `coveredByActiveOrderArm`'s *"the two cannot disagree about which pairs the chart
already owns"* true rather than nearly true. The chip still carries one row — a pair is one clinical
fact however many rows join it.

Recorded, no behaviour change: why `identifies()` delegates to `DrugReference.matchesText` and not to
the order-name matcher #128 splits out (both operands here are canonical reference strings, with no
localized inflected display name to tolerate a trailing letter for, and the orientation is inverted —
the rule's *token* is the haystack and the entry's aliases the needles, so the nesting hazard runs the
other way and the symmetric word boundary is the one that stops `chlorothiazide` matching a
`hydrochlorothiazide` rule); that `orderedInteractionNotes` is deliberately not extended, its *"a
partner that raises a chip is exactly a partner promoted here"* being scoped to the active-order arm
and still exact, because a pair chip's partner is by construction not an active order; and that the
subject tie-break follows **dataset** order, not the question's word order — both phrasings of the
voxelotor/dexamethasone question chip with Voxelotor as subject, its entry sitting at index 1055
against dexamethasone's 1744.

### Depends on #128

This PR's **precedence** behaviour rests entirely on `PatientClinicalContext.hasActiveDrug`, which
#128 replaces with a left-boundary + inflection-tolerant order-name matcher. The Agnes Adams
precedence case, `theActiveOrderChipWinsWhenThePatientIsOnOneOfTheNamedDrugs` and
`aPairAlsoJoinedByARuleNamingAnActiveOrderStaysWithTheActiveOrderArm` all decide "is this pair
already the chart's?" through that predicate, so **whichever of the two lands second should re-run the
Agnes precedence assertion.** The direction of #128's change is safe for this arm on the argument that
it makes `hasActiveDrug` match more localized order spellings and fewer nested substrings — the chart
arm then claims strictly more of the pairs it should and strictly fewer of the ones it should not —
but that is an argument, not a measurement.

## Post-hardening live re-verification

Same instance and configuration as above; full-JVM restart on the hardened build, querystore
retrieval proven alive and the known-positive control fired before any capture. Chip absence is
never concluded from one run (a cold-start race can load the KB empty for a whole JVM lifetime with
no WARN), so both zero-chip cases were re-run at the end — both still zero.

| probe | chips | result |
|---|---|---|
| Mary + "Does voxelotor interact with dexamethasone?" | 2 | `active order simvastatin — Moderate`, then **one** `Voxelotor interacts with Dexamethasone, also named in the question — Major` — four dexamethasone entries resolved, one pair chip |
| Mary + "Can dexamethasone be given with voxelotor?" | 2 | identical, incl. the same subject — the dedup is not sensitive to question word order |
| **Mary + "Does warfarin interact with aspirin?"** | 2 | `active order simvastatin — Minor`, then `Acetylsalicylic acid (aspirin), also named in the question — Major`; `references: 2`, one a cited `safety_finding` |
| Susan Young (on neither drug) + same | 1 | the pair chip alone, `activeOrder=0` — chart-independent |
| Agnes Adams (on aspirin) + same | 1 | `active order aspirin — Major` only; `pair=0`, so no duplicate reference lookup |
| Mary + "Does ascorbic acid interact with aspirin?" | 0 | both drugs resolve, no KB row joins them (re-run: 0) |
| Mary + "Is it safe to give her clarithromycin?" | 1 | `active order simvastatin — Major` |
| Agnes + "Is it safe to start warfarin?" | 1 | `active order aspirin — Major` — active-order wording, not pair wording |
| Joshua + "Can I give ibuprofen?" | 2 | NSAID cross-reactivity contraindication + `active order lisinopril — Moderate` |
| Mary + "Is it safe to give her paracetamol?" | 0 | (re-run: 0) |

All four regression controls unchanged. No test data was created or modified on the standalone.

## Hardening round 2 — a structural review's nine findings

An independent structural review of this arm found nine issues, three of them blocking. **Every one
was reproduced here before being fixed** (the reviewer's own runs are cited where they measured
something I did not re-measure). Each fix is red-first through the real `validate(...)` seam.

The branch is now rebased onto `623be82`, so all of this runs against **#128's `hasActiveDrug`** —
the predicate the precedence arm delegates to — and every number below is re-derived, not reused.

### Blocking: one clinical pair reported twice, two ways

**Chart precedence was decided per ENTRY pair, as an early return.** Route variants make one clinical
pair arrive as several entry pairs carrying *different rule sets*: the sub-floor sibling of an
above-floor row loses that row to the severity floor, so its entry pair sees different tokens and can
reach the opposite verdict. A chart-owned pair therefore left no trace and its sibling chipped.
Reproduced with a new fixture family, patient on Aspirin 81mg:

```
Diclofenac interacts with active order aspirin — Major. …
Acetylsalicylic acid (aspirin) interacts with Diclofenac (topical), also named in the question — Major. …
```

Worse than a duplicate: the patient's own medication becomes the subject of a sentence that reads as
a reference lookup, carrying the systemic row's Major mechanism attributed to the topical variant
whose own row is `Unknown`. The arm now **groups first and decides second** — candidates collect per
pair key, a chart-owned verdict is *recorded against the key* rather than returned on, and emission
happens once every entry pair has been seen. Entry-pair ordering no longer matters, which the
previous shape silently depended on.

**A drug could key two ways.** The key name came from the other side's rule token where there was one
and from `displayLabel()` where there was not — two vocabularies. So one entry keyed as `aspirin`
against the partner whose row names it and as `Acetylsalicylic acid (aspirin)` against the sibling
whose row is sub-floor: one drug, two keys, pair escapes the grouping, two chips with no patient data
at all. The reviewer is right that this cannot be waved through as "an extra chip, never a dropped
one" — every extra chip is an extra injected pre-answer finding, the outcome the keying exists to
prevent. Names are now resolved **per drug, once**, from the first above-floor rule any *other*
question drug uses to name it, so a drug keys identically in every pair and variants agree by
construction.

Measured on the full KB to bound the same-name suppression this introduces: of 2093 distinct tokens
only **6** are an exact alias of entries with different `rxnorm_name`s, and all 6 are one substance
family (the trastuzumab conjugates, isosorbide and its mononitrate, the COVID-19 vaccines) — so no
pair worth reporting is suppressed. In all **142** variant families the token is an exact alias of
every member, so the grouping catches them all.

### Blocking: an uncapped, question-controlled prompt expansion

This is the one safety input whose size a *question* controls; the others are bounded by the
patient's chart, which held findings to 0–3. Measured on the real bundled sample, one line naming its
16 drugs: **72 chips, 72 injected findings, 42,708 characters** of finding text — against a path that
caps a single reference record at `MAX_INTERACTION_RENDER_CHARS = 1500` for exactly this reason. And
they appended in dataset order, so that question led with two Moderates ahead of the first Major, the
inverse of the severity-first rule the promoted notes apply to the same rows.

Now sorted most-severe-first, capped at `MAX_QUESTION_PAIR_CHIPS = 10`, and what is withheld is
logged at WARN — a clinician cannot tell a bounded list from a complete one, so silent truncation in
a safety net reads as "nothing else was found". Live, the same question produced:

```
WARN Question-pair safety: 10 of 65 question-named drug pairs shown (cap 10); the question
resolved 16 reference drugs, and pairs grow as N^2/2. Withheld, least severe last:
[Major, Major, … ×10, Moderate ×37, Minor ×8]
```

**Flagging the trade honestly:** at 16 named drugs there are 20 Major pairs, so ten Majors are
withheld. They are named in the log, and the alternative measured above is 42 KB of prompt, but
whether 10 is the right number for a polypharmacy-review question is a judgement worth a second
opinion — the constant carries the arithmetic behind it.

### Two more defects found while reproducing those

**A multi-word rule token named every drug called after one of its words.** `identifies()` asked
"does this rule name that entry?" with the *prose* matcher — token as haystack, aliases as needles —
but both operands are canonical reference strings, so the question is name identity. A word boundary
does not save it: it stops a name nested inside a WORD (`chlorothiazide` in `hydrochlorothiazide`)
but not one nested inside a PHRASE. Measured on the full KB: **408 of 2093** distinct tokens are
multi-word and **53** carry a word that is some other entry's whole name — `ethinyl estradiol` (in
449 entries' rule lists) named the separate Estradiol entry, `magnesium salicylate` named Salicylic
acid, `iron-dextran complex` named Iron. The chip then stated an interaction the KB does not carry,
against a drug whose row it read from a different drug, and `preAnswerFindings` injected that
sentence as a citable record — a fabricated interaction handed to the model as evidence.

**A question naming ONE drug chipped it against its own route variant.** The two-drugs guard counts
*entries*, and one word resolves several when the KB carries variants of it, so "is minoxidil safe?"
raised `Minoxidil interacts with Minoxidil (topical), also named in the question`. The token keying
cannot suppress those either — a self-pair's key has the same name on both sides, so it is unique by
construction. 33 above-floor rows in the full KB join two entries sharing one token, across 29 drug
words (minoxidil, timolol, lidocaine, atropine, neomycin, paclitaxel …), and where the variants are
named after different substances the chip names two drugs the question never mentioned, which is #86.

> ### Deviation from a stated load-bearing decision
>
> The brief for this hardening listed *"`identifies()` delegates to `DrugReference.matchesText`"* as
> settled, and said not to hand-roll a third matcher. **I changed it, deliberately, and am flagging it
> rather than burying it.** The same brief directed me to check which contract that delegation
> actually wants — and the contract fails at this call site, measurably, in the direction that
> fabricates clinical claims. The replacement **removes** a matcher rather than adding one: it reads
> the same `getAliases()` list `matchesText` reads and compares for equality, so #128's change to how
> that list is scanned cannot drift from it, and the "no third matcher" concern is honoured more
> strictly than before. Nothing genuine is lost — `ddinter` writes each token from its partner row's
> `rxnorm_name` (else its name) and the parser puts both in that partner's aliases, so every real rule
> still names its partner exactly. If the maintainer prefers the original delegation, this is the one
> commit to revert (`97a188d`), and the `aMultiWordRuleTokenDoesNotNameEveryDrugCalledAfterOneOfItsWords`
> test documents what comes back.

### Coverage holes, mutation-verified

Both of the reviewer's mutation-proven holes were re-run against this branch and are now killed:

- **Question-only scoping** — the arm's most safety-relevant decision — was pinned by nothing.
  Widening it to `inPlay` left the whole package green. It now fails, and the mutation's output shows
  why it matters: `Ibuprofen interacts with Warfarin, also named in the question`, about a drug the
  question never mentioned.
- **The key-name fallback** was pinned by nothing, and still was after the per-drug rework. The
  reachable shape is narrow — a pair needs a joining rule, and that rule names the *other* side, so
  only the unnamed side falls back and two fallback drugs never pair with each other. They do both
  pair with the drug they name, which is the bundled curated seed's own shape. Under a constant
  fallback that second pair vanishes silently: 2 chips → 1.
- **The fixture's severity fields were inert** (the chip renders the note, which the parser prefixes
  with the rating). They are load-bearing now — the floor and the new ordering both read the field —
  and the tests assert the field as well as the rendered word.

### Corrections to the record

Two javadoc claims this change leaned on were false. *"A pair either of whose drugs is an active order
is left to `addInteractions`"* is not what the code asks: the test is whether any **rule** joining the
pair names an active order, and the difference is deliberate — a patient can be on one of the pair
while no joining rule names their order, and that pair is this arm's to report because the chart arm
raises nothing for it. And `orderedInteractionNotes`' *"a partner that raises a chip is exactly a
partner promoted here"* does not hold across the whole chip set, since a pair chip's partner is
promoted nowhere; rewording that neighbour's sentence is left to whichever PR owns the method next
rather than collided over. What is preserved is the invariant it exists to protect — a chip and the
prose cannot describe one finding differently — because #110 injects the finding itself verbatim.

## Live re-verification, round 2

Same instance, rebased onto `623be82` (#128, #123, #119, #125, #124 landed). Full-JVM restart on this
build; querystore retrieval proven alive (`hits=75`) and the known-positive control fired before any
capture. Every zero-chip case was run **twice**, late in the run, because a cold-start race can load
the KB empty for a whole JVM lifetime with no WARN.

| probe | chips | result |
|---|---|---|
| 16-drug polypharmacy line | 17 | 7 active-order + **exactly 10** pair chips, all Major, severity-descending; WARN names the 55 withheld |
| Mary + "Does voxelotor interact with dexamethasone?" | 2 | active-order simvastatin, then **one** `Voxelotor interacts with Dexamethasone … Major` — four variant entries, one chip |
| Mary + "Can dexamethasone be given with voxelotor?" | 2 | identical, same subject — dedup insensitive to question word order |
| Mary + "Does levofloxacin interact with estradiol?" | **0** | the fabricated `ethinyl estradiol` chip is gone (re-run: 0) |
| Mary + "Is minoxidil safe for this patient?" | **0** | no self-pair against its own route variant (re-run: 0) |
| **Mary + "Does warfarin interact with aspirin?"** | 2 | `active order simvastatin — Minor`, then `Acetylsalicylic acid (aspirin), also named in the question — Major`, with a cited `safety_finding` |
| Susan Young (on neither) + same | 1 | the pair chip alone, `activeOrder=0` — chart-independent |
| **Agnes Adams (on aspirin) + same** | 1 | `active order aspirin — Major` only; `pair=0` — precedence re-derived against #128's `hasActiveDrug` |
| Mary + "Does ascorbic acid interact with aspirin?" | 0 | both resolve, no row joins them (re-run: 0) |
| Mary + "Is it safe to give her clarithromycin?" | 1 | `active order simvastatin — Major` |
| Agnes + "Is it safe to start warfarin?" | 1 | `active order aspirin — Major` — active-order wording, not pair wording |
| Joshua + "Can I give ibuprofen?" | 2 | NSAID cross-reactivity contraindication + `active order lisinopril — Moderate` |
| Mary + "Is it safe to give her paracetamol?" | 0 | (re-run: 0) |

All four regression controls unchanged. No test data was created or modified on the standalone.

`mvn clean install` on `623be82`: **API 740 + OMOD 58 = 798 run, 0 failures, 0 errors, 34 skipped**;
`DrugSafetyQuestionPairInteractionTest` contributes 21.

## Tests

`DrugSafetyQuestionPairInteractionTest` — 12 tests through the real
`DrugSafetyValidator.validate(...)` and `DrugReferenceInjector.injectRecords(...)` seams, over the
real bundled datasets parsed by the real sources. Every one of them is red on `main`; the two marked
**⭐** were red on this PR's own first push, i.e. they are the hardening's regression tests.

- a question naming two interacting drugs, patient on **neither** ⇒ exactly one chip naming the pair
- the pair chip never claims an active order, and does attribute the pair to the question
- a symmetric pair (present on both entries) is reported **once**, not once per side
- a sub-floor pair (Unknown-severity simvastatin × aspirin) raises **nothing**
- two question-named drugs with no interaction row between them raise **nothing** (curated seed, so
  the arm is exercised source-independently)
- the active-order chip wins, and is not duplicated, when the patient is on one of the named drugs
- the pair finding is injected as a numbered, citable `safety_finding` record
- **⭐** route-variant entries sharing a match token are reported as **one** pair, not one per entry
- **⭐** a pair also joined by a rule naming an active order stays with the active-order arm — the
  precedence check reads every joining rule, not just the first
- every distinct pair among **three** named drugs is reported (three chips) — the other edge of the
  grouping, and the direction where a key too coarse would silently drop a real interaction
- a pair identified only by ATC code is reported — the arm of "does this rule name that entry?" the
  `ddinter` source never exercises, since it always writes a name token
- the patient-specific chip **leads** the pair chip (Mary Smith's measured shape: her active-order
  simvastatin finding first, the question's pair second)

One new fixture, `drug-reference-question-pairs.json`, parsed by the real `JsonDrugReferenceSource`.
It carries the three shapes the bundled DDInter sample cannot express — variants sharing a token, one
pair joined by two differently-tokened rules, and a rule identifying its partner by ATC code alone —
each a miniature of a shape the full KB carries.

`mvn clean install` on `623be82`: **API 740 + OMOD 58 = 798 run, 0 failures, 0 errors, 34 skipped**
(21 tests in this file, up from the 7 of the first push). No existing test was
modified. Note for anyone reproducing: `mvn clean test` cannot build the omod module in this reactor
(`unpack-dependencies` hits MDEP-98 — "Artifact has not been packaged yet"), so the full-reactor
command is `mvn clean install`.

## Relationship to #113

Same underlying shape — the safety join was only ever question-drug against chart — but the opposite
input. This is the *two drugs named* case and is fixed entirely within the question-resolved set.
#113 is the *no drug named* case, which needs an anchor this change does not provide (the patient's
own order list, cross-referenced against itself) and intent detection to decide when to do it. The
two arms are independent: this one reads `questionDrugs`, a screening arm would read active orders.
They will both hang off `chartsearchai.drugSafety.warnOnInteractions`, whose `config.xml` description
still describes only the answer-versus-orders join; I left that text alone rather than collide with
#113, which needs the same sentence rewritten — worth doing once, in whichever lands second.

Closes #114

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
@@ -690,18 +690,6 @@ public class DrugSafetyValidator {
 	 */
 	static final int MAX_QUESTION_PAIR_CHIPS = 10;
 
-	/**
-	 * @return {@link #severityRank} with an UNRATED rule raised above {@code major}: every curated
-	 *         hand-authored rule is unrated and {@link #clearsSeverityFloor} already treats unrated as
-	 *         exempt rather than low — unrated is not low-rated, so wherever a most-severe-first choice
-	 *         is made it must not be the one dropped. Matches the ordering
-	 *         {@code DrugReferenceInjector}'s promoted notes apply to the same rows.
-	 */
-	static int severityPriority(String severity) {
-		int rank = severityRank(severity);
-		return rank < 0 ? Integer.MAX_VALUE : rank;
-	}
-
 	/** Orders candidate pair chips most-severe first; see {@link #severityPriority}. */
 	private static final Comparator<PairFinding> PAIR_SEVERITY_DESCENDING = new Comparator<PairFinding>() {
 
@@ -913,6 +901,16 @@ public class DrugSafetyValidator {
 	 * {@code rxnorm_name}s, and all 6 are one substance family (the trastuzumab conjugates, isosorbide
 	 * and its mononitrate, the COVID-19 vaccines), so a pair genuinely worth reporting is not among them.
 	 *
+	 * <p>The name comes from {@link #partnerLabel}, the same coalesce the chart arm's own grouping keys
+	 * on, so the two arms that now group chips in this class agree by construction about what naming a
+	 * partner means. They cannot fight over a pair: both consult {@code hasActiveDrug}, and they take
+	 * opposite branches of it. Whenever any rule joining a pair names an active order, this arm records
+	 * the pair as the chart's and raises nothing, while {@link #bestRulePerPartner} raises exactly one
+	 * chip for that partner label — grouping only merges chips sharing a label, so a chip for the
+	 * deferred pair always survives it, at that partner's most severe row. Whenever no joining rule
+	 * names an active order, the chart arm raises nothing for the pair and this arm reports it. So one
+	 * clinical pair yields one chip from one arm, never two chips and never none.
+	 *
 	 * @return each drug mapped to the reference data's own name for it: the match token (or, for a rule
 	 *         carrying only a code, the ATC code) of the first above-floor rule any OTHER drug in
 	 *         {@code drugs} uses to name it. Falls back to its generic name, else its display name, when
@@ -929,7 +927,7 @@ public class DrugSafetyValidator {
 				}
 				List<DrugReference.Interaction> naming = aboveFloorRulesAgainst(other, drug, floor);
 				if (!naming.isEmpty()) {
-					name = ChartSearchAiUtils.firstNonBlank(naming.get(0).getToken(), naming.get(0).getAtc());
+					name = partnerLabel(naming.get(0));
 					break;
 				}
 			}
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
@@ -655,8 +655,14 @@ public class DrugSafetyValidator {
 		if (!reportedPairs.add(unorderedPairKey(matchTokenFor(first, reverse), matchTokenFor(second, forward)))) {
 			return;
 		}
-		// Whichever side carries the rule owns the sentence; with symmetric data both do and the
-		// earlier question drug leads.
+		// Whichever side carries the rule owns the sentence; with symmetric data both do, and the tie
+		// goes to whichever entry the DATASET lists first — not whichever the question names first,
+		// because questionDrugs comes from findByQuery, which walks getAll() and so returns dataset
+		// order. Measured on the 3.7.1 standalone: "does voxelotor interact with dexamethasone?" and
+		// "can dexamethasone be given with voxelotor?" both chip with Voxelotor as the subject, its
+		// entry sitting at index 1055 against dexamethasone's 1744. Stable either way, which is what
+		// the chip needs; a rule that followed the question's word order would need the drug's offset
+		// in the question, which findByQuery does not report.
 		boolean fromFirst = !forward.isEmpty();
 		DrugReference subject = fromFirst ? first : second;
 		DrugReference partner = fromFirst ? second : first;
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
@@ -53,7 +53,9 @@ import org.springframework.stereotype.Service;
  *       {@link CrossReactivityGroup} (cross-branch family overlap). The rule arm raises one
  *       warning per (drug, matched partner): several rules can name one partner — DDInter's
  *       route variants of a drug all publish the same match token — and they collapse to the
- *       most severe row, see {@link #bestRulePerPartner}.</li>
+ *       most severe row, see {@link #bestRulePerPartner}. It also interacts with
+ *       ANOTHER DRUG THE QUESTION NAMED — see {@link #addQuestionPairInteractions}, the arm that
+ *       answers "does A interact with B?" for a patient on neither.</li>
  *   <li><b>Contraindications</b> — the drug is contraindicated by an active allergy or
  *       condition: by a hand-authored rule, by being the same drug as — or sharing an ATC
  *       chemical subgroup with — a recorded allergy (cross-reactivity reasoning), or —
@@ -248,6 +250,11 @@ public class DrugSafetyValidator {
 			if (warnDose) {
 				addOverdose(warnings, ref, context, lower, all);
 			}
+		}
+		// LAST, so the patient's own findings lead: a chip about their allergy or their active order
+		// is a fact about them, and outranks a reference lookup about a pair they may not be on.
+		if (warnInteractions) {
+			addQuestionPairInteractions(warnings, questionDrugs, context, severityFloor);
 		}
 		if (!warnings.isEmpty()) {
 			log.info("Drug-safety validator raised {} warning(s)", warnings.size());
@@ -558,6 +565,127 @@ public class DrugSafetyValidator {
 	 */
 	private static int noteLength(DrugReference.Interaction interaction) {
 		return interaction.getNote() == null ? 0 : interaction.getNote().trim().length();
+	}
+
+	/**
+	 * The question-named PAIR arm (issue #114): the drugs the question resolved to, checked against
+	 * EACH OTHER rather than only against the chart.
+	 *
+	 * <p>Every other interaction arm joins one drug to the patient's own data, so a question naming
+	 * two drugs was silently reduced to two independent one-drug questions: asked "does warfarin
+	 * interact with aspirin?" about a patient on neither, the module reported no information about a
+	 * pair its own KB rates Major (measured on the 3.7.1 standalone, 2026-08-04 — it also raised an
+	 * unrequested Minor chip about the one drug that WAS on the chart, which reads as an answer to
+	 * the question asked). The pair is a plain reference lookup: it needs no patient data at all, and
+	 * withholding it because the patient happens not to be on either drug answers a question nobody
+	 * asked.
+	 *
+	 * <p>Deliberately scoped to the QUESTION's drugs, not to all of {@code inPlay}. The
+	 * answer-named additions are the LLM's word choice, and two drugs it names are as often
+	 * alternatives ("ibuprofen, or paracetamol if …") as a proposed combination, so pairing them
+	 * would assert a co-administration risk nobody proposed — the same over-reach echo scoping
+	 * exists to prevent (issue #105). It also keeps the chips aligned with the injected findings:
+	 * {@link DrugReferenceInjector#preAnswerFindings} runs this same validate with an EMPTY answer,
+	 * so an answer-side pair could produce a chip with no record behind it.
+	 *
+	 * <p>The floor is the shared {@code severityFloor} the chart arm is given, so this cannot become
+	 * a route around a decision the chip path enforces; and a pair either of whose drugs is an
+	 * active order is left to {@link #addInteractions}, whose chip states a fact about THIS patient
+	 * — the stronger statement, and reporting both would say one finding in two voices.
+	 */
+	private void addQuestionPairInteractions(List<SafetyWarning> warnings, Set<DrugReference> questionDrugs,
+			PatientClinicalContext context, int severityFloor) {
+		if (questionDrugs.size() < 2) {
+			return;
+		}
+		// Indexed so each UNORDERED pair is visited once: DDInter rows are symmetric and the parser
+		// writes every pair into both drugs' entries, so walking ordered pairs would report each
+		// interaction twice, once from each side.
+		List<DrugReference> drugs = new ArrayList<DrugReference>(questionDrugs);
+		for (int i = 0; i < drugs.size() - 1; i++) {
+			for (int j = i + 1; j < drugs.size(); j++) {
+				addQuestionPairInteraction(warnings, drugs.get(i), drugs.get(j), context, severityFloor);
+			}
+		}
+	}
+
+	/** One question-named pair: at most one chip, from whichever side carries the rule. */
+	private void addQuestionPairInteraction(List<SafetyWarning> warnings, DrugReference first,
+			DrugReference second, PatientClinicalContext context, int severityFloor) {
+		DrugReference.Interaction forward = aboveFloorRuleAgainst(first, second, severityFloor);
+		DrugReference.Interaction reverse = aboveFloorRuleAgainst(second, first, severityFloor);
+		if (forward == null && reverse == null) {
+			return;
+		}
+		if (coveredByActiveOrderArm(forward, context) || coveredByActiveOrderArm(reverse, context)) {
+			return;
+		}
+		// Whichever side carries the rule owns the sentence; with symmetric data both do and the
+		// earlier question drug leads.
+		DrugReference subject = forward != null ? first : second;
+		DrugReference partner = forward != null ? second : first;
+		DrugReference.Interaction rule = forward != null ? forward : reverse;
+		// "named in the question" is what keeps this distinguishable from the active-order chip, and
+		// it is a claim about the QUESTION, so it stays true whatever the chart holds. Wording it as
+		// "neither drug is on the chart" instead would be false for the one-sided-data case that
+		// reaches here — the patient IS on one of the pair, but only that drug's entry carries the
+		// rule, so no active-order chip fires from either side — and stating the patient's
+		// medications wrongly is the defect in #86.
+		String detail = subject.displayLabel() + " interacts with "
+				+ ChartSearchAiUtils.firstNonBlank(partner.displayLabel(), rule.getToken(), rule.getAtc())
+				+ ", also named in the question";
+		if (rule.getNote() != null && !rule.getNote().isEmpty()) {
+			detail += " — " + rule.getNote();
+		}
+		warnings.add(new SafetyWarning(SafetyWarning.TYPE_INTERACTION, subject.displayLabel(), detail));
+	}
+
+	/**
+	 * @return {@code subject}'s first interaction rule that names {@code other} AND clears
+	 *         {@code floor}, or null when it carries none. First rather than every: a pair of
+	 *         reference entries is one clinical fact however many rows join them (route variants
+	 *         sharing a match token can produce several), and the chart arm's one-chip-per-rule
+	 *         behaviour is not extended here.
+	 */
+	private static DrugReference.Interaction aboveFloorRuleAgainst(DrugReference subject, DrugReference other,
+			int floor) {
+		for (DrugReference.Interaction i : subject.getInteractions()) {
+			if (clearsSeverityFloor(i, floor) && identifies(i, other)) {
+				return i;
+			}
+		}
+		return null;
+	}
+
+	/**
+	 * @return true when interaction rule {@code i} names reference entry {@code other} — the
+	 *         reference-side counterpart of {@link PatientClinicalContext#hasActiveDrug}: the same
+	 *         two arms, the rule's name token and its ATC code, resolved against a loaded entry's
+	 *         aliases and normalized codes instead of against the chart's active orders. Alias
+	 *         matching goes through the shared {@link DrugReference#matchesText} — the same matcher
+	 *         {@code findByQuery}/{@code lookupByToken} use to bind free text to an entry — so
+	 *         "this rule points at that drug" is decided one way throughout.
+	 */
+	private static boolean identifies(DrugReference.Interaction i, DrugReference other) {
+		String token = i.getToken();
+		if (token != null && !token.trim().isEmpty()
+				&& other.matchesText(token.trim().toLowerCase(Locale.ROOT))) {
+			return true;
+		}
+		String atc = DrugReference.normalizeAtcToken(i.getAtc());
+		return atc != null && other.normalizedAtcCodes().contains(atc);
+	}
+
+	/**
+	 * @return true when {@code rule} points at one of the patient's active orders, i.e.
+	 *         {@link #addInteractions} already raises this pair as a fact about the patient. Asked
+	 *         of both sides of a pair by the caller, so it covers "the partner is on the chart" and
+	 *         "the subject is on the chart" alike. Uses the very predicate that arm uses, so the two
+	 *         cannot disagree about which pairs the chart already owns.
+	 */
+	private static boolean coveredByActiveOrderArm(DrugReference.Interaction rule,
+			PatientClinicalContext context) {
+		return rule != null && context != null && context.hasActiveDrug(rule.getToken(), rule.getAtc());
 	}
 
 	/**
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
@@ -642,6 +642,16 @@ public class DrugSafetyValidator {
 		// being the KB's own canonical name for a drug, and precisely what the variants share. Which
 		// variant's row then supplies the severity is #115's open half; the dataset's first is kept,
 		// as the chart arm keeps its first.
+		//   De-duplicating inside a safety net can only be done where nothing actionable is lost, and
+		// two entries collapse here only when the reference data gives them the SAME name. In the
+		// shape this exists for, one rule then identifies both, so the surviving chip carries that
+		// very rule's severity and mechanism and all that is dropped is a second spelling of the
+		// partner. Entries the data can tell apart are named by different tokens and keep their own
+		// chips — asserted over three drugs named in one question. The residual is a KB that gives two
+		// genuinely different drugs one token and separates them only by ATC code: the first is
+		// reported. Such a token is ambiguous in the reference data itself: the chart arm reads that
+		// same token to decide which of the patient's orders a rule names, so it cannot resolve the
+		// ambiguity either.
 		if (!reportedPairs.add(unorderedPairKey(matchTokenFor(first, reverse), matchTokenFor(second, forward)))) {
 			return;
 		}
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
@@ -11,6 +11,8 @@ package org.openmrs.module.chartsearchai.reference;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -620,7 +622,7 @@ public class DrugSafetyValidator {
 		// subject of a sentence reading as a reference lookup, carrying one variant's mechanism
 		// attributed to another whose own row was sub-floor. Candidates are collected per pair key and
 		// emitted only once every entry pair has been seen.
-		Map<List<String>, SafetyWarning> candidates = new LinkedHashMap<List<String>, SafetyWarning>();
+		Map<List<String>, PairFinding> candidates = new LinkedHashMap<List<String>, PairFinding>();
 		Set<List<String>> chartOwned = new LinkedHashSet<List<String>>();
 		// Indexed so each UNORDERED pair of entries is visited once: DDInter rows are symmetric and
 		// the parser writes every pair into both drugs' entries, so walking ordered pairs would report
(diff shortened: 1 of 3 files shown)


dkayiwa:
Reviewed the design; flagging one **merge-sequencing coupling** for whoever lands these, since eight DDI PRs are in flight at once.

This PR's pair arm rests on two existing predicates rather than a new matcher — `identifies()` delegates to `DrugReference.matchesText` + `normalizedAtcCodes()`, and `coveredByActiveOrderArm()` delegates to `PatientClinicalContext.hasActiveDrug`. That is the right construction (it composes with changes to those predicates instead of drifting from them), but it means:

- **#86 / PR #87 changes `hasActiveDrug`.** That is exactly the predicate deciding which pairs this PR *defers* to the active-order arm. So the precedence behaviour verified here — Agnes on Aspirin 81mg getting "active order aspirin" wording rather than the new pair wording — is only as correct as `hasActiveDrug`. Worth re-running this PR's precedence assertion after #86 lands, in whichever order they merge.
- PR #87 also refactors `matchesText` to route through a shared `containsWord`. That refactor is behaviourally neutral *as written* (`matchesText` already had symmetric boundaries), so `identifies()` is unaffected today — but note that I have argued on #87 that its shared-matcher approach is wrong for order names and should be split. If the eventual #86 fix keeps prose matching intact and only loosens order-name matching, `identifies()` stays correct. If someone instead changes `matchesText`, this PR's pair resolution changes with it.

Net: no conflict expected, but these two PRs are not independent, and this PR's tests are the ones that would catch it. Both touch `DrugSafetyValidator`/`PatientClinicalContext`, so expect a textual rebase too.

Two smaller notes:

**The stale documentation this PR identified is real and shared.** `SafetyWarning.TYPE_INTERACTION`'s javadoc and `config.xml`'s `warnOnInteractions` description both still describe only the answer-versus-active-orders join — already stale before this change, and now more so. #113 (generic screening over active-order pairs) needs the same sentence rewritten. Doing it once in whichever of the two lands second is the right call, as this PR suggests; it should not be duplicated.

**The relevance question is correctly out of scope.** Mary's pre-existing **Minor** warfarin×simvastatin active-order chip now appears alongside the asked-about **Major** pair chip. Ordering the asked-about finding ahead of an incidental lower-severity one is a presentation concern, not this defect, and the arm running last so patient-specific findings lead is the defensible default. Flagging it only so it is not mistaken for a regression introduced here.

One methodological point in this PR's favour, worth stating because it is easy to get wrong on this codebase: every claim above is on **chips and structure**, not answer prose. Answer wording on the verification box is not byte-reproducible (`cache_prompt` KV reuse, `LocalLlmEngine.java:786-792`), so the improved prose in the capture is a single observation and this PR correctly does not present it as a measured rate.

dkayiwa:
## Structural review findings — recorded before they are lost

**Provenance, stated plainly:** these come from an independent structural review of this arm that completed just as a session limit killed the agent that would have acted on them. Each was verified *by that reviewer* using scratch harnesses calling the real `JsonDrugReferenceSource.parse`/`DdiDrugReferenceSource.load` → `DrugReferenceService` → `DrugSafetyValidator.validate` path, plus mutation runs of the full 162-test `reference` package. **I have not independently re-verified them**; I am recording them so they survive, and the fixes below should each be reproduced before being trusted. (A sibling agent fabricated corroboration earlier today, so I am being explicit about which claims are checked and by whom.)

Baseline for all of it: 162 tests, 0 failures.

### Blocking — two ways one clinical pair gets reported twice

**1. The chart-precedence skip does not record the pair key** (`DrugSafetyValidator.java:474-480`). Pair (Dexamethasone, Voxelotor) is skipped at `:475` *without inserting its key*, so the route sibling (Dexamethasone (nasal), Voxelotor) then finds `forward` empty and `reverse` = voxelotor's Major row, concludes the active-order arm doesn't own it, and chips. Measured, with the committed fixture and one edit (`dexamethasone-nasal` severity `Moderate`→`Unknown`; 48 of 120 bundled rows are `Unknown`, floor is `minor`):

```
Dexamethasone interacts with active order voxelotor — Major. …
Voxelotor interacts with Dexamethasone (nasal), also named in the question — Major. …
```

Same clinical pair — proved by the keys collapsing to 1 chip when the patient is on nothing. This is exactly the "one finding in two voices" `:435-437` forbids, and it's worse than a duplicate: the surviving pair chip attributes the *plain* variant's Major mechanism to the *nasal* variant whose own row is `Unknown`, and makes the patient's own medication the subject of a sentence that reads as a reference lookup. Reviewer's fix: `reportedPairs.add(pairKeyFor(...)); return;` before the skip — the two checks are currently in the wrong order.

**2. `pairKeyNameFor`'s fallback keys in the *display* vocabulary while the other branch keys in the *KB token* vocabulary** (`:622-626`, used at `:607-610`). Needs no patient data. The ASA side keys as `"acetylsalicylic acid (aspirin)"` instead of `"aspirin"`, so one drug gets two keys and the pair escapes dedup — measured as two chips for `"Does ibuprofen interact with aspirin?"`. `:617-620` documents this as safe ("an extra chip, never a dropped one"), but that acceptance is incomplete: **every extra chip is also an extra injected pre-answer finding** (`DrugReferenceInjector:182-188`) — the precise outcome `:580-591` says the keying exists to prevent. Reviewer's fix: fall back to `firstNonBlank(drug.getGenericName(), drug.getName())`. Root fix is `DdiDrugReferenceSource` always carrying `rxnorm_name` canonically; today it's dropped whenever the name contains it, which is exactly the route-variant case.

### Blocking — an uncapped, question-controlled prompt expansion

**3.** Pair findings are unbounded and unordered (`:254` → `DrugReferenceInjector:182-188`). Measured on the real bundled sample with the patient on nothing: a 245-char question naming 16 drugs → **72 chips and 42,708 chars of injected finding text**, in 15 ms. Before this arm, findings were bounded by the patient's own chart (0–3); now they grow as N²/2 in the question's drug count, and the reference-record path caps at `MAX_INTERACTION_RENDER_CHARS = 1500` per entry for exactly this reason. Separately they append in dataset order, so a measured 6-drug case leads with a **Minor** chip ahead of four Majors — the inverse of the severity-first principle `orderedInteractionNotes:355-363` sets for the same data. Fix: sort severity-descending and cap, mirroring the promoted-note budget, and log what was withheld.

### Correctness of the record

**4.** Two javadoc claims this change leans on are false (`:435-437`, `:441-442`, and unchanged neighbour `DrugReferenceInjector:296`). Precedence is decided by whether a *joining rule* matches `hasActiveDrug`, not by whether either drug is an active order — finding 1's output is a pair chip whose **subject** is an active order, and `:493-498` already documents the converse one-sided case. So "a partner that raises a chip is exactly a partner promoted here" no longer holds across the chip set. Reword to the invariant that is actually true, and scope the injector's biconditional to the active-order arm.

### Coverage holes — both mutation-proven

**5.** The question-only scoping — this arm's most safety-relevant decision — is pinned by nothing. Mutating `:254` to pass `inPlay` (pairing answer-named drugs, the behaviour `:426-432` explicitly rejects) leaves **162/162 green**, and `", also named in the question"` would then be a false statement.
**6.** The `pairKeyNameFor` fallback branch is pinned by nothing — replacing `drug.displayLabel()` with a constant leaves **162/162 green**. That's the hole finding 2 passes through.
**7.** The fixture's severity assertions pass for the wrong reason: the pair chip never renders `rule.getSeverity()`, so the asserted `"Major"` comes from the note text, which DDInter prefixes with the severity word. Deleting `"severity": "Major"` leaves the chip byte-identical and the test green. No entry sets `genericName`, so the fixture cannot express finding 2's shape at all.

### Cosmetic

**8.** `identifies`' javadoc (`:539-541`) claims the symmetric boundary stops nesting; true for *character* nesting, not *word* nesting — token `"dexamethasone / lidocaine"` binds to Lidocaine alone. Depends on the full KB carrying a multi-word token, which `DdiDrugReferenceSource:179`'s name fallback makes reachable. Narrow the claim.
**9.** The class javadoc's Interactions bullet reads as if the pair arm continues the active-order clause, though it needs no patient data.

### Negative results worth keeping

The reviewer predicted `|| coveredByActiveOrderArm(reverse, …)` was mutation-survivable and **was wrong** — dropping either half fails a named test. Loop bounds, `reportedPairs` scoping, and the `size() < 2` guard are correct. Blank-name key fallbacks are unreachable through real sources (both parsers drop blank-named entries). A null context cannot reach `validate` from production. Duplicate `resourceKey`s across findings are harmless — `extractCitedReferences` keys by chart index. Key orientation at `:607-610` is self-symmetric. Cost is a non-issue: 15 ms for the 16-drug worst case.

dkayiwa:
## Hardening round 2 — all nine structural findings addressed

Every finding above was **reproduced on this branch before being fixed**, and each fix is red-first
through the real `validate(...)` seam. The branch is rebased onto `623be82`, so this all runs against
**#128's `hasActiveDrug`** — the predicate the precedence arm delegates to — and the numbers below are
re-derived, not carried over. Provenance is explicit: where I say "measured", I ran it.

**The three blocking findings are fixed at the root, not patched.**

1 and 2 turned out to be one architectural mistake with two symptoms: the arm decided *per entry pair*
what is a property of the *clinical pair*. Route variants make one clinical pair arrive as several
entry pairs carrying different rule sets, because the severity floor removes a sub-floor sibling's row
and the remaining tokens differ. So the arm now **groups first and decides second** — a chart-owned
verdict is recorded against the pair key instead of returned on, and key names are resolved **per
drug, once** rather than per pair. Entry-pair ordering no longer affects the outcome; the previous
shape silently depended on it.

3 is bounded and ordered: `MAX_QUESTION_PAIR_CHIPS = 10`, severity-descending, with a WARN naming what
was withheld. Live, the 16-drug line logged `10 of 65 … shown`. **I want to flag the trade rather than
present it as settled:** that question has 20 Major pairs, so ten Majors are withheld. They are named
in the log and the measured alternative is 42 KB of prompt, but whether 10 is right for a
polypharmacy-review question is worth a second opinion — the constant carries the arithmetic.

**Two further defects surfaced while reproducing those**, both fabricating clinical claims, both now
fixed and live-verified at 0 chips: a multi-word rule token named every drug called after one of its
words (`ethinyl estradiol` → the separate Estradiol entry; 53 of 2093 tokens have that shape), and a
question naming ONE drug chipped it against its own route variant.

> ### One load-bearing decision was deliberately changed
>
> My brief listed *"`identifies()` delegates to `DrugReference.matchesText`"* as settled, with "do not
> hand-roll a third matcher". **I changed it.** The same brief told me to check which contract that
> delegation actually wants, and it fails here: both operands are canonical reference strings, so the
> question is name identity, and a word boundary stops a name nested inside a *word* but not inside a
> *phrase*. The result was a chip stating an interaction the KB does not carry, injected as a citable
> record — a fabricated interaction handed to the model as evidence. The replacement **removes** a
> matcher rather than adding one (equality against the same `getAliases()` list `matchesText` reads),
> so nothing can drift from #128. If you prefer the original delegation, `97a188d` is the single commit
> to revert, and `aMultiWordRuleTokenDoesNotNameEveryDrugCalledAfterOneOfItsWords` documents what
> returns.

**Both mutation-proven coverage holes are killed**, re-verified by re-running the mutations here:
widening the scoping to `inPlay` now fails (and its output is `Ibuprofen interacts with Warfarin, also
named in the question` — the chip's own words made false), and a constant key-name fallback now fails
by silently dropping one of two real pairs. Finding 7's inert fixture severities are load-bearing now
(the floor and the new ordering both read the field) and asserted directly.

**Not done, deliberately:** the stale `SafetyWarning.TYPE_INTERACTION` javadoc and `config.xml`
`warnOnInteractions` description are still ceded to #127, and `orderedInteractionNotes`' biconditional
is left for whichever PR owns that method next — my javadoc now states plainly that it does not hold
across the whole chip set rather than claiming it does. Neither is touched here, to avoid two PRs
colliding on the same lines.

`mvn clean install`: **API 740 + OMOD 58 = 798, 0 failures, 34 skipped.** 21 tests in this file.
Live: 17 probes, all four regression controls unchanged, every zero-chip case run twice. No test data
created or modified on the standalone.

Still true from the earlier note: **this PR's precedence rests on `hasActiveDrug`.** It is now verified
against #128's version rather than the pre-#128 one, so that dependency is discharged — but any future
change to that predicate should re-run the Agnes case.

PR #121, opened 2026-08-04
Raise one interaction chip per (drug, active order): collapse DDInter route-variant rows (closes #115)

### LINKED ISSUE #115: Route variants sharing one rxnorm_name raise N chips for one interaction, at N different severities (distinct from #88, survives #86)

dkayiwa:
## The defect

DDInter carries one entry per **route variant**, and every variant of a drug publishes the same
`rxnorm_name`. `DdiDrugReferenceSource.interactionsFor` turns each variant's row into its own
interaction rule with that shared name as the match token, so **one** active order matches **all**
of them and `DrugSafetyValidator.addInteractions` raises a chip per row — N chips for one pair, at N
different severities.

Live on the 3.7.1 standalone (HEAD `13690b1`), Margaret King with one active order Dexamethasone
4mg, asked "Is it safe to give voxelotor?": **3 chips, Major + Moderate + Moderate**, and the two
Moderate details were byte-identical strings (the nasal and ophthalmic variants share a mechanism
group, so there was not even a severity difference to preserve between them). Because chips are also
injected as citable safety-finding records (`preAnswerFindings`, #110), each duplicate chip also
became a near-identical record in the prompt — `[121] [122] [123]`.

Scale, re-measured against the loaded 19 MB KB (2283 drugs / 295,184 rows) rather than taken from the
issue: **142** `rxnorm_name` values are shared by more than one entry, covering **332** entries;
**1004** question-drugs have a ≥3-row group for one partner token and **1876** have a ≥2-row group —
i.e. for 1876 of 2283 drugs there is some active drug that duplicates a chip. `SafetyWarning` has no
`equals`/`hashCode`, so nothing downstream can collapse them.

## Root cause

The rule arm iterates `ref.getInteractions()` and emits one `SafetyWarning` per matching row. It has
no notion of "these rows are about the same partner", and the shared `rxnorm_name` guarantees the
route variants' rows are indistinguishable to it: same token, same first ATC code (all four
dexamethasone variants publish an *identical* ATC list), often the same note.

## Which of the two designs, and why

The issue documents two fixes. I implemented **(1) dedup per (drug, active order), keeping the most
severe row**, at the validator.

(2) — matching the order's dose form/route to the right variant — is the more correct fix and I did
not attempt it, for two reasons beyond "harder":

- The information is not at this layer. `PatientClinicalContext` carries active-order **names** and
  **ATC codes**; there is no dose form or route on it, and ATC cannot discriminate the variants
  because all four publish the same code list. Resolving the variant means a new route vocabulary
  plus plumbing it through the context builder — a data-side project (belongs with
  openmrs-ddi-knowledge-base), not a validator change.
- Half-built, it fails in the **unsafe** direction: an order whose form string does not parse would
  silently drop a Major systemic interaction.

So (1), with the direction chosen deliberately: **most severe wins**. The route distinction is
clinically real (voxelotor is Major against systemic dexamethasone and Moderate against the nasal
and ophthalmic forms), so this is not a free collapse.

**Trade-off accepted, stated plainly:** a patient on a topical/ophthalmic form may now be shown the
systemic severity. Over-warning on a non-blocking advisory the clinician adjudicates is the fail-safe
direction; under-warning (picking Moderate in case the order is topical) would hide a real Major
interaction for the much more common systemic order. **Severity honesty is deferred** — making the
rating route-aware is the data-side half of #115 and is not in this PR.

## What changed

- `DrugSafetyValidator.bestRulePerPartner` (new, private): after the existing severity-floor filter
  and the active-order match, matched rules are grouped by **the label the chip actually renders**
  (`firstNonBlank(token, atc)`, case-folded) and one rule per group survives. Grouping on the
  rendered label means rules that would produce the same "interacts with active order X" subject
  collapse, while rules naming different partners each keep their chip — even when their note text is
  identical. Chip order is unchanged (first appearance in dataset order).
- Winner: highest severity; on a tie, the fuller mechanism note; on both equal, the incumbent. The
  note is the only informativeness signal a row carries — DDInter's "no mechanism description on
  file" fallback is shorter than any real mechanism paragraph, and in the equal-severity pair
  measured (dolutegravir × iron) the longer note is a strict superset of the shorter one, adding the
  dose-separation INTERVAL guidance.
- `DrugSafetyValidator.severityPriority` (new, package-private): the ordering in which an *unrated*
  rule sits above Major — previously encoded inline in `DrugReferenceInjector.InteractionNote`. Now
  one definition used by both the chip grouping and the injector's promoted-note ordering, so the two
  cannot drift into ranking the same pair of rules oppositely. `InteractionNote`'s constructor takes
  the severity string instead of a pre-computed rank; behaviour is identical.
- The **dataset is untouched**. Flattening the variants at parse time would destroy the per-variant
  severities that design (2) needs, so the collapse lives at the validator and a test pins that the
  parser still produces distinct entries at their own severities.

Not touched: `addClassInteractions`.

## Relationship to #88 / PR #89 (please read before merging either)

**This is not #88.** That defect is *cross-arm* — one chip from `addInteractions`, one from
`addClassInteractions`, for the same pair. These were two *rule-arm* chips.

I checked PR #89 ("Fold same-class explicit interactions into one chip"), which implements #88.
Three things a maintainer should know:

1. **#89 does not fix this defect.** Its correlation key is the *order's ATC code*, and that cannot
   separate route variants: all four dexamethasone entries publish an identical ATC list, so every
   variant's row carries the same `atc`. More concretely, in `addInteractionWarnings` (read at its head
   `83393f1`) a rule whose ATC is not among the patient's active-order codes falls to the name-only
   branch, which adds a warning per row with no grouping at all — and the live repro is exactly that
   shape: Margaret King's Dexamethasone order is concept 92, which carries no ATC mapping in this
   dictionary, so the match is purely the name token.
2. **Where #89 does collapse variants, it does so severity-blind.** For a patient whose order *does*
   carry the partner's ATC code, #89's `ruleByOrderCode` keeps the *first* row for that code
   (`if (!containsKey) put`). That would have picked whichever variant the dataset lists first, not
   the most severe — the thing this PR's `strongestSeverityWinsEvenWhenItIsNeitherTheFirstRowNorTheLongestNote`
   test exists to rule out.
3. **They compose, and there should be one collapse point, not two.** The two keys answer different
   questions: `bestRulePerPartner` decides *which rule row* survives for a partner; #89 decides
   *which arms* describe one order. Composed in that order they are complementary; kept as two
   independent mechanisms they are a maintenance trap. #89 is `CONFLICTING` against `main`
   independently of this PR — it is based on `83cc33e` and predates both #108 (which added the
   rule-arm severity floor and the `displayLabel()`-led standalone-sentence chip detail) and #110
   (which extracted `clearsSeverityFloor` and started injecting findings as records), and both of
   those commits changed the very method it rewrites. Its `addInteractionWarnings` carries no
   severity-floor check at all, so a straight rebase must also restore #84's floor.
   **Recommendation:** land this PR first, and when #89 is rebased have
   its rule-row loop iterate `bestRulePerPartner(ref, context, severityFloor)` instead of
   `ref.getInteractions()` and drop its own `ruleByOrderCode` first-wins `put`. That leaves one
   grouping mechanism, with severity-aware selection, feeding the cross-arm fold.

I did not base this branch on #89. My tests assert the current chip contract — the
`displayLabel()`-led standalone sentence (`"Voxelotor interacts with active order dexamethasone —
Major. …"`) and the severity floor — neither of which exists on #89's head, so building there would
have meant rebasing someone else's PR inside this one.

## Tests

New fixture `api/src/test/resources/chartsearchai-test/ddi-route-variants.json` — three **real**
slices of the full DDInter KB (each the induced subgraph over the named drugs, real mechanism texts,
dataset row order), parsed by the real production parser and driven through the real
`DrugSafetyValidator.validate(...)`:

| test | pins |
|---|---|
| `routeVariantsOfOneGenericParseAsDistinctEntriesKeepingPerVariantSeverities` | four Dexamethasone variants stay four entries, and Voxelotor keeps one row per variant at `Major/Moderate/Moderate` — the fix must not be a flattened dataset |
| `oneActiveOrderMatchingSeveralRouteVariantsRaisesOneChipAtTheHighestSeverity` | the live repro: one active dexamethasone order ⇒ **1** chip, and it is the Major one (was 3) |
| `differentPartnersKeepTheirOwnChipEvenWhenTheirNotesAreIdentical` | Voxelotor's Phenytoin row shares the Major dexamethasone row's mechanism group — same severity, same note string, different partner ⇒ still **2** chips (was 4) |
| `strongestSeverityWinsEvenWhenItIsNeitherTheFirstRowNorTheLongestNote` | Lapatinib × Sirolimus: the Major row is the *second* match and has the *shorter* note, so "keep the first" and "keep the longest" both fail here |
| `equalSeveritiesKeepTheRowCarryingTheFullerMechanismNote` | Dolutegravir × Iron (the pair in the issue description): two Major rows, the fuller note is second |

All four behaviour tests fail on `13690b1` (3, 4, 2 and 2 chips respectively) and pass with the fix;
the parser test passes on both, by design.

`mvn package` (api tests + omod tests): **API 667 + OMOD 53 = 720 run, 0 failures, 0 errors, 34
skipped** (the 34 skipped are the pre-existing LLM-requiring evals). Baseline on `13690b1` was 715;
this adds 5.

## Live standalone evidence

3.7.1 standalone, full 19 MB DDInter KB, `sourceFormat=ddinter`, `minInteractionSeverity=minor`,
`chartMode=fullChart`, local LLM. Full-JVM restart, retrieval proven alive (`hits>0`), warm control
fired before any capture. Chips are deterministic Java, so the assertions below are on chips; answer
prose is not reproducible on this box (`cache_prompt` KV reuse) and no claim is made from it.

Same patient, same question, same instance and same GPs in both directions. Margaret King's order was
re-confirmed active at the start of the pass, so a vanished chip cannot be a vanished order:

```
PRE-CHECK — Margaret King's active drug orders
    drug orders: 1
       (NEW) Dexamethasone: 4.0 Milligram Oral Once daily | voided= False | stopped= None | expires= None
```

### Before — Margaret King + "Is it safe to give voxelotor?" (HEAD `13690b1`, the two runs in #115's comment)

```
safetyWarnings (3)
  [interaction] drug='Voxelotor'
      Voxelotor interacts with active order dexamethasone — Major. Coadministration with potent or
      moderate inducers of CYP450 3A4 may significantly decrease the plasma concentrations and
      pharmacologic effects of voxelotor. […]
  [interaction] drug='Voxelotor'
      Voxelotor interacts with active order dexamethasone — Moderate. Coadministration with
      inhibitors of CYP450 3A4 may increase the plasma concentrations and pharmacologic effects of
      corticosteroids […] Cushing's syndrome and adrenal insufficiency have been attributed to the
      interaction.
  [interaction] drug='Voxelotor'
      Voxelotor interacts with active order dexamethasone — Moderate. Coadministration with
      inhibitors of CYP450 3A4 may increase the plasma concentrations and pharmacologic effects of
      corticosteroids […] Cushing's syndrome and adrenal insufficiency have been attributed to the
      interaction.

chip2.detail == chip3.detail  ->  True
run 1: references [(2,'condition','chart'), (121,'safety_finding'), (122,'safety_finding'), (123,'safety_finding')]
       safety_finding records: 3   answer 1919 chars   inline markers {121:1, 122:9, 123:8, 2:5}
run 2: references [(2,'condition','chart'), (121,'safety_finding'), (122,'safety_finding'), (123,'safety_finding')]
       safety_finding records: 3   answer 1975 chars   inline markers {121:2, 122:10, 123:14, 2:3}
```

### After — this build (verbatim from the capture; both runs identical on chips)

```
--- Margaret King + "Is it safe to give voxelotor?" [i115-vox-1] chips=1
    [interaction] drug='Voxelotor'
        Voxelotor interacts with active order dexamethasone — Major. Coadministration with potent or moderate inducers of CYP450 3A4 may significantly decrease the plasma concentrations and pharmacologic effects of voxelotor. The proposed mechanism is accelerated clearance of voxelotor due to induction of the CYP450 3A4 isoenzyme, which is the primary route of elimination of voxelotor.
    references: [(121, 'safety_finding', 'reference', True), (120, 'drug_reference', 'reference', None)]
    safety_finding records: 1
    answer chars: 452   inline markers: {'121': 1, '120': 1}
--- Margaret King + "Is it safe to give voxelotor?" [i115-vox-2] chips=1
    [interaction] drug='Voxelotor'
        Voxelotor interacts with active order dexamethasone — Major. Coadministration with potent or moderate inducers of CYP450 3A4 may significantly decrease the plasma concentrations and pharmacologic effects of voxelotor. The proposed mechanism is accelerated clearance of voxelotor due to induction of the CYP450 3A4 isoenzyme, which is the primary route of elimination of voxelotor.
    references: [(121, 'safety_finding', 'reference', False), (120, 'drug_reference', 'reference', None)]
    safety_finding records: 1
    answer chars: 626   inline markers: {'121': 1, '120': 1}
```

**3 chips → 1, and it is the Major one.** The chip that survived is byte-identical to the Major chip
from the before-capture, so nothing was reworded — two rows were dropped. Both runs agree, which is
the confirmation the brief asks for before believing a chip disappeared. The duplicate
`safety_finding` records went with them: 3 → 1.

### Regression controls — all four from the brief, same pass, same build

```
--- CONTROL Mary + clarithromycin (expect 1 Major x simvastatin) [i115-c1] chips=1
    [interaction] drug='Clarithromycin'
        Clarithromycin interacts with active order simvastatin — Major. Coadministration with potent inhibitors of CYP450 3A4 may significantly increase the plasma concentrations of simvastatin and lovastatin […]
--- CONTROL Agnes + warfarin (expect 1 Major x aspirin) [i115-c2] chips=1
    [interaction] drug='Warfarin'
        Warfarin interacts with active order aspirin — Major. Aspirin, even in small doses, may increase the risk of bleeding in patients on oral anticoagulants […]
--- CONTROL Joshua + ibuprofen (expect 2 chips) [i115-c3] chips=2
    [contraindication] drug='Ibuprofen'
        Ibuprofen is in the same cross-reactivity group (NSAID) as the patient's allergy to Acetylsalicylic acid (aspirin) — possible cross-reactivity
        Ibuprofen interacts with active order lisinopril — Moderate. Nonsteroidal anti-inflammatory drugs (NSAIDs) may attenuate the antihypertensive effects of ACE inhibitors […]
--- CONTROL Mary + paracetamol (expect 0 chips) [i115-c4] chips=0

13/13 checks passed
```

(Mechanism prose elided at `[…]` for length only; the chips are otherwise verbatim.)

### On the citation spraying #115's comment reported — observed, not claimed

The comment noted that the duplicate records induced marker spraying (`[122]`×9, `[123]`×8, plus five
`[2]` markers pointing at an unrelated `condition` record). In both after-runs the answer came back at
452 and 626 chars with one marker per record (`{121:1, 120:1}`) and no `[2]`. That is consistent with
the duplicate records having been the cause, and it is the direction you would expect — but answer
prose is not reproducible on this box (`cache_prompt` KV reuse), n=2, so **this is an observation, not
a claim**. The deterministic part is what the chips and record set show: one finding record instead of
three near-identical ones.

Not fixed by this change, and still worth its own look: the grounding verdict on the single
deterministic finding record came back `grounded=True` on run 1 and `grounded=False` on run 2 — the
verifier is inconsistent about a module-generated line, which is finding #3 in #115's comment. It is
unrelated to the chip count and out of scope here.

## Notes / limits

- Test data on the standalone: Margaret King's Dexamethasone 4mg order was created for #115's
  reproduction (it is not part of the demo data set).
- `DrugReferenceInjector.orderedInteractionNotes` still renders one entry per variant row inside the
  drug-reference record's `Interactions:` section (e.g. "dexamethasone (Major …); dexamethasone
  (Moderate …); dexamethasone (Moderate …)"). That is prompt text, not a citable record, and the
  render budget there is a measured, settled area, so it is out of this PR's scope — reported, not
  fixed.
- The mirror-image shape is also still open and is **not** this defect: when the *question* names a
  generic shared by several entries ("Is it safe to give dexamethasone?" with the partner active),
  the four variants each enter play as their own reference drug and raise a chip apiece — measured 3
  chips through the real validator, but carrying *distinguishable* labels ("Dexamethasone",
  "Dexamethasone (nasal)", "Dexamethasone (ophthalmic)"), not duplicates. Collapsing those means
  choosing which variant the clinician meant — the same route vocabulary design (2) needs.
- Not attempted: `equals`/`hashCode` on `SafetyWarning`. Collapsing at the point of generation is the
  root-cause fix; a downstream `equals` would only have hidden the byte-identical pair and would have
  masked the differing-severity duplicates entirely.

Closes #115

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReferenceInjector.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReferenceInjector.java
@@ -304,6 +304,18 @@ public class DrugReferenceInjector {
 	 * a partner that raises a chip is exactly a partner promoted here, and the rendered text cannot
 	 * drift from the chip.
 	 *
+	 * <p>That correspondence is per PARTNER, and since issue #115 it is no longer one note per chip:
+	 * {@link DrugSafetyValidator#bestRulePerPartner} collapses several rules naming one partner —
+	 * DDInter's route variants of a drug all publish the same match token — into a single
+	 * most-severe chip, while this method still promotes one entry per row. A patient on one
+	 * dexamethasone order therefore gets one Major chip beside a record reading "dexamethasone
+	 * (Major …); dexamethasone (Moderate …); dexamethasone (Moderate …)", so a model answering from
+	 * the record can still name a severity the chip deliberately discarded. Collapsing here needs
+	 * the same thing the chip cannot decide either — which variant the order is — and so waits on
+	 * the route vocabulary that is the data-side half of #115. Until then the two paths agree on
+	 * WHICH partners concern the patient, which is what the ordering above exists to guarantee, but
+	 * not on how many rows or which severity.
+	 *
 	 * <p>Ordering alone is not sufficient, which is why {@code render} also overrides the budget for
 	 * this segment: two above-floor partners can exceed {@link #MAX_INTERACTION_RENDER_CHARS}
 	 * between them (measured on the bundled sample: methotrexate 783 + aspirin 809 against a 1500
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReferenceInjector.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReferenceInjector.java
@@ -310,11 +310,15 @@ public class DrugReferenceInjector {
 	 * most-severe chip, while this method still promotes one entry per row. A patient on one
 	 * dexamethasone order therefore gets one Major chip beside a record reading "dexamethasone
 	 * (Major …); dexamethasone (Moderate …); dexamethasone (Moderate …)", so a model answering from
-	 * the record can still name a severity the chip deliberately discarded. Collapsing here needs
-	 * the same thing the chip cannot decide either — which variant the order is — and so waits on
-	 * the route vocabulary that is the data-side half of #115. Until then the two paths agree on
-	 * WHICH partners concern the patient, which is what the ordering above exists to guarantee, but
-	 * not on how many rows or which severity.
+	 * the record can still name a severity the chip deliberately discarded — and does so from the
+	 * more quotable half, because in that measured case the discarded Moderate note is 659 characters
+	 * against the surviving Major row's 326. Nothing goes missing: {@code render} emits every promoted
+	 * note, so a duplicated partner can only push another partner into its compact
+	 * {@code name (Severity)} form, never out. Collapsing here needs the same thing the chip cannot
+	 * decide either — which variant the order is — and so waits on the route vocabulary that is the
+	 * data-side half of #115. Until then the two paths agree on WHICH partners concern the patient,
+	 * which is what the ordering above exists to guarantee, but not on how many rows or which
+	 * severity.
 	 *
 	 * <p>Ordering alone is not sufficient, which is why {@code render} also overrides the budget for
 	 * this segment: two above-floor partners can exceed {@link #MAX_INTERACTION_RENDER_CHARS}
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReferenceInjector.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReferenceInjector.java
@@ -363,8 +363,7 @@ public class DrugReferenceInjector {
 			}
 			boolean promote = context != null && context.hasActiveDrug(i.getToken(), i.getAtc())
 					&& DrugSafetyValidator.clearsSeverityFloor(i, floor);
-			(promote ? promoted : rest).add(new InteractionNote(rendered, compact,
-					DrugSafetyValidator.severityRank(i.getSeverity())));
+			(promote ? promoted : rest).add(new InteractionNote(rendered, compact, i.getSeverity()));
 		}
 		// Within the promoted segment, severity — not dataset position — decides who keeps their
 		// mechanism prose when the budget can only afford one full note (see render). Measured on the
@@ -409,14 +408,14 @@ public class DrugReferenceInjector {
 
 		final String compact;
 
-		/** {@link DrugSafetyValidator#severityRank} with unrated raised above Major — see
-		 *  {@link #SEVERITY_DESCENDING}. */
+		/** {@link DrugSafetyValidator#severityPriority} — the shared ordering in which unrated sits
+		 *  above Major; see {@link #SEVERITY_DESCENDING}. */
 		final int severityPriority;
 
-		InteractionNote(String full, String compact, int severityRank) {
+		InteractionNote(String full, String compact, String severity) {
 			this.full = full;
 			this.compact = compact;
-			this.severityPriority = severityRank < 0 ? Integer.MAX_VALUE : severityRank;
+			this.severityPriority = DrugSafetyValidator.severityPriority(severity);
 		}
 	}
 
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
@@ -424,13 +424,14 @@ public class DrugSafetyValidator {
 	 * <p>Trimmed to fold the way the MATCH folds. {@link PatientClinicalContext#hasActiveDrug} trims
 	 * and case-folds the rule's token before testing it against an order name, so two rows whose
 	 * tokens differ only in case or in surrounding whitespace are one partner to the only predicate
-	 * that decides an interaction concerns this patient — and must be one partner here too, or #115's
-	 * duplicate chip
-	 * returns for a hand-authored dataset with two labels a clinician cannot tell apart. That is a
-	 * live shape, not a hypothetical: the {@code ddinter} and {@code atc} sources normalize their
-	 * tokens at parse, but the curated {@code json} source is plain Jackson over an operator-editable
-	 * file and sanitizes nothing. Trimming also keeps that padding out of the chip text, as
-	 * {@link DrugReferenceInjector#orderedInteractionNotes} already does for its own rendered label.
+	 * that decides an interaction concerns this patient — and must be one partner here too, or issue
+	 * #115's duplicate chip returns for a hand-authored dataset, silently and with two labels a
+	 * clinician cannot tell apart. That shape reaches this method: {@code ddinter} lower-cases every
+	 * token it derives and takes it from the trimmed RxNorm generic whenever the partner row has one,
+	 * and {@code atc} carries no rules at all, but the curated {@code json} source is plain Jackson
+	 * over an operator-editable file and sanitizes neither case nor padding. Trimming also keeps that
+	 * padding out of the chip text, as {@link DrugReferenceInjector#orderedInteractionNotes} already
+	 * does for its own rendered label.
 	 *
 	 * @return the label, or null when the rule carries neither — which a rule that matched an active
 	 *         order cannot ({@code hasActiveDrug} needs a non-blank token or a non-blank ATC), so
@@ -467,7 +468,8 @@ public class DrugSafetyValidator {
 	 * class arm ({@link #addClassInteractions}) is untouched, so the rule-plus-class double chip
 	 * of issue #88 is a different defect and stays open.
 	 *
-	 * <p><b>Which row wins.</b> The most severe rating, then the longer note. Route variants
+	 * <p><b>Which row wins.</b> The most severe rating, then the longer note — longer in prose, not
+	 * in whitespace, see {@link #noteLength}. Route variants
 	 * genuinely differ — topical dexamethasone does not have systemic dexamethasone's interaction
 	 * profile, which is why DDInter rates voxelotor Major against one variant and Moderate against
 	 * the others — but nothing on a {@code DrugOrder} tells this layer which variant the order is
@@ -525,8 +527,18 @@ public class DrugSafetyValidator {
 		return noteLength(candidate) > noteLength(incumbent);
 	}
 
+	/**
+	 * @return how much mechanism prose {@code interaction}'s note carries — its <em>trimmed</em>
+	 *         length. Whitespace is not the informativeness signal the tie-break is reading: measured
+	 *         raw, a short referral note with a block of blank lines pasted onto it outranks a longer
+	 *         real mechanism paragraph, and the surviving chip then says nothing about the mechanism
+	 *         while the row explaining it is discarded. Trimmed for the comparison only — the note
+	 *         still renders as authored, exactly as it does for a row with no competitor;
+	 *         {@link DrugReferenceInjector#orderedInteractionNotes} makes the same "trim before you
+	 *         compare lengths" call for the same reason.
+	 */
 	private static int noteLength(DrugReference.Interaction interaction) {
-		return interaction.getNote() == null ? 0 : interaction.getNote().length();
+		return interaction.getNote() == null ? 0 : interaction.getNote().trim().length();
 	}
 
 	/**
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
@@ -473,12 +473,11 @@ public class DrugSafetyValidator {
 	 * class arm ({@link #addClassInteractions}) is untouched, so the rule-plus-class double chip
 	 * of issue #88 is a different defect and stays open.
 	 *
-	 * <p><b>Which row wins.</b> The most severe rating, then the longer note — longer in prose, not
-	 * in whitespace, see {@link #noteLength}. Route variants
-	 * genuinely differ — topical dexamethasone does not have systemic dexamethasone's interaction
-	 * profile, which is why DDInter rates voxelotor Major against systemic dexamethasone, Moderate
-	 * against two others and carries no row at all against the topical variant — but nothing on a
-	 * {@code DrugOrder} tells this layer which variant the order is
+	 * <p><b>Which row wins.</b> The most severe rating, then the longer note — longer in prose, not in
+	 * whitespace, see {@link #noteLength}. Route variants genuinely differ — topical dexamethasone does
+	 * not have systemic dexamethasone's interaction profile, which is why DDInter rates voxelotor Major
+	 * against systemic dexamethasone, Moderate against two others and carries no row at all against the
+	 * topical variant — but nothing on a {@code DrugOrder} tells this layer which variant the order is
 	 * (the context carries names and ATC codes; all four variants publish an identical ATC list),
 	 * so the variant cannot be resolved here. Reporting the strongest rating over-warns rather than
 	 * under-warns on a non-blocking advisory the clinician adjudicates, which is the fail-safe
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
@@ -10,9 +10,12 @@
 package org.openmrs.module.chartsearchai.reference;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -47,7 +50,10 @@ import org.springframework.stereotype.Service;
  *       {@code chartsearchai.drugSafety.minInteractionSeverity} is not raised — unrated rules
  *       are never floor-filtered), by sharing an ATC chemical subgroup with an active order
  *       (duplicate-therapy reasoning), or — failing that — by sharing a curated
- *       {@link CrossReactivityGroup} (cross-branch family overlap).</li>
+ *       {@link CrossReactivityGroup} (cross-branch family overlap). The rule arm raises one
+ *       warning per (drug, matched partner): several rules can name one partner — DDInter's
+ *       route variants of a drug all publish the same match token — and they collapse to the
+ *       most severe row, see {@link #bestRulePerPartner}.</li>
  *   <li><b>Contraindications</b> — the drug is contraindicated by an active allergy or
  *       condition: by a hand-authored rule, by being the same drug as — or sharing an ATC
  *       chemical subgroup with — a recorded allergy (cross-reactivity reasoning), or —
@@ -273,6 +279,22 @@ public class DrugSafetyValidator {
 		}
 	}
 
+	/**
+	 * {@link #severityRank} raised so that an <em>unrated</em> rule sorts above {@code major}: every
+	 * curated hand-authored rule is unrated and {@link #clearsSeverityFloor} already treats unrated
+	 * as exempt rather than low — unrated is not low-rated, so wherever a most-severe-wins choice is
+	 * made it must not lose to a rated row. The one definition of that ordering, shared by the chip
+	 * grouping here ({@link #bestRulePerPartner}) and the promoted-note ordering in
+	 * {@link DrugReferenceInjector.InteractionNote}; two copies could drift into ranking the same
+	 * pair of rules oppositely, which is how the chip and the prompt text come to disagree.
+	 *
+	 * @return the rank, with null/unrecognized mapped to {@link Integer#MAX_VALUE}
+	 */
+	static int severityPriority(String severity) {
+		int rank = severityRank(severity);
+		return rank < 0 ? Integer.MAX_VALUE : rank;
+	}
+
 	/** @return the floor rank for the GP value, falling back to the default floor when the
 	 *          value is unrecognized (a typo'd GP must not silently disable all rated rules). */
 	private static int floorRank(String gpValue) {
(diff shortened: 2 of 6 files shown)


dkayiwa:
## Hardening pass — two defects fixed, six invariants pinned, four measured claims corrected

Ran the `/harden` workflow over `git diff main...HEAD` (5 structural passes, 3 polish passes, 2
independent reviewers). The design is unchanged: option (1) at the validator, most-severe-wins, the
class arm untouched, the dataset unflattened, no `equals` on `SafetyWarning`, `severityPriority`
shared with the injector. What changed is that two of the collapse's decisions were folding on the
wrong string, six claims the code makes about itself had no test — two of them survivable by
implementations I verified by mutation — and four measured figures in the rationale did not describe
what the code compares.

### 1. The grouping key folded case but not whitespace — while the match folds both

`PatientClinicalContext.hasActiveDrug` trims a rule's token before testing it against an order name,
so `"Warfarin"` and `"  warfarin  "` are **one partner** to the only predicate that decides whether an
interaction concerns this patient. The group key was `firstNonBlank(token, atc).toLowerCase(...)`, and
`firstNonBlank` is documented "as given, **untrimmed**" — so those two rows landed in different groups
and both raised a chip. #115's duplicate straight back, silently, with two labels a clinician cannot
tell apart, plus the padding rendered into the chip text.

Fixed by extracting `partnerLabel(Interaction)` — the trimmed token-else-ATC coalesce — and using it
**both** to render the chip detail and (case-folded) as the group key. That is what makes decision (2)
structurally true rather than true-by-coincidence: the grouping is only correct while the key IS the
label the chip says, and two copies of the same coalesce in two methods can drift into grouping rules
by something a clinician never sees.

Reachable, not hypothetical: `ddinter` and `atc` normalize at parse, but the curated `json` source is
plain Jackson over an operator-editable file and sanitizes nothing. #128's own javadoc, now on `main`,
makes the same point from the other side — "a hand-authored `json` KB is NOT sanitized".

### 2. The fuller-note tie-break measured padding, not prose

`noteLength` measured the note raw. A 79-character "see the local policy" note with a block of blank
lines pasted on (90 raw) therefore **outranked** an 89-character real mechanism paragraph, and the
surviving chip said nothing about the mechanism while the row explaining it was discarded — decided by
whitespace, with the newlines rendered into the chip detail and, since #110, into the safety-finding
record that reuses that detail verbatim.

`orderedInteractionNotes` already faced this exact call on the render side and resolved it the same
way ("Trim before the length comparison below, not after"), so this is the sibling's rule applied to
the collapse. Trimmed for the comparison only — the winning note still renders as authored.

### 3. Six claims the code makes, now pinned (+6 tests)

| claim | where it was documented | test added |
|---|---|---|
| chip order is dataset first-appearance order, and replacing a group's winner does not reorder | javadoc, the `LinkedHashMap` re-put comment, the PR body | `replacingAGroupsWinnerLeavesTheChipsInDatasetOrderOfFirstAppearance` |
| the key folds case **and** whitespace, and the chip label carries no padding | (new) | `rowsWhoseTokensDifferOnlyInCaseOrWhitespaceNameOnePartnerAndRaiseOneChip` |
| an unrated rule outranks Major — decision (3) | `severityPriority` javadoc | `aHandAuthoredUnratedRowOutranksASourceRatedMajorRowForTheSamePartner` |
| the fuller note is the one with more prose | `bestRulePerPartner` javadoc | `theFullerNoteTieBreakMeasuresProseNotPadding` |
| equal on both keeps the **incumbent**, not the last row seen | `bestRulePerPartner` javadoc | `rowsEqualOnSeverityAndOnNoteLengthKeepTheIncumbent` |
| the label falls back to the ATC code, and that half folds too | (new) | `rowsCarryingOnlyAnAtcCodeAreGroupedAndLabelledByThatCode` |

Two of these were not merely unpinned — they were survivable by implementations a reviewer named, and
I verified both by mutation:

- **`noteLength` was dead weight.** Reducing `outranks` to
  `severityPriority(candidate) >= severityPriority(incumbent)` and deleting `noteLength` outright
  passed all nine tests, because in every fixture pair the intended winner happened to sit *last*.
  Fixed without touching an assertion: the two Linezolid rows are swapped so the mechanism row comes
  first. The `>=` rewrite is now killed by two tests.
- **`partnerLabel`'s ATC fallback was dead weight.** `return interaction.getToken().trim()` passed the
  entire repo — DDInter always derives a token, the WHO ATC source emits no rules, and all five
  bundled curated rules carry both fields — while `hasActiveDrug` matches on the ATC arm alone, after
  which the label is both the chip text and the group key. That rewrite now dies with the exact
  `NullPointerException` it would raise on an operator's ATC-keyed rule.

The ordering test uses the existing real KB slice — Dolutegravir's iron group is opened first,
dexamethasone's is opened next, and only **then** does iron's second row take its group, which is the
one arrangement where a `HashMap` or a remove-then-put reorders the chips. (Checked: `"dexamethasone"`
and `"iron"` land in different `HashMap` buckets, so that mutation really does invert them.) The other
five needed the **curated** source, because that is where the shapes arise: new fixture
`drug-reference-partner-label-variants.json`, parsed by the real `JsonDrugReferenceSource` and driven
through the real `DrugSafetyValidator.validate`. Its padding is written as `\n` escapes rather than
trailing spaces on purpose: invisible trailing whitespace in a fixture is one editor away from being
stripped, which would leave the test green while testing nothing.

`severityPriority`'s "unrated above Major" branch was documented as uncoverable ("no bundled dataset
can cover it" — `SEVERITY_DESCENDING`). That is true on the *render* side; on the chip side it is
coverable, because the curated source is the one that produces unrated rules, and since #115 that
ordering decides which chip a clinician sees rather than only which prompt note keeps its prose.

### 4. Four measured claims in the rationale did not describe what the code compares

All re-derived from the fixture and the deployed 19 MB KB before rewriting. None of the reasoning
changed; the figures now match the strings the code actually reads.

- The dolutegravir × iron surplus is **not** "the dose-separation INTERVAL guidance". The two notes
  are 171 and 236 characters — 164/229 were the *mechanism texts*, before the severity prefix that is
  part of what `noteLength` compares — and the 55-character surplus is the sentence *"The mechanism of
  interaction has not been established."* The superset claim holds and is now stated as what it is.
  (#116, already on `main`, strips the `INTERVAL:` marker precisely because it carries no guidance.)
- Voxelotor is Major against systemic dexamethasone, Moderate against **two** others, and carries **no
  row at all** against the topical variant — three rows over four variants, as the fixture's own
  metadata already said. If anything that strengthens the route argument.
- Lapatinib × sirolimus: 285/280 were mechanism lengths; the notes are 295/287.
- `partnerLabel`'s javadoc claimed `orderedInteractionNotes` already trims its own label. It does not
  — it trims the assembled `label (note)` piece — so a padded curated token still reaches the citable
  record as `warfarin   (Major. …)` beside a chip reading `active order warfarin`. Now stated
  accurately, with giving the injector this method named as the follow-up.

### 5. Two bounded residuals, now recorded next to the rule that produces them

- A row with **no note** still wins its group on severity alone, so an operator's token-only unrated
  rule beats a rated row carrying a mechanism paragraph and the chip then gives no reason. Reachable
  only in hand-authored data (every DDInter row has a note). Left as it is deliberately: the
  alternative — a note outranking a rating — would drop the operator's own rule, which is the thing
  `severityPriority` exists to protect. Worth a maintainer's opinion.
- Grouping is by label, so two tokens that both match one order but are not the same string stay two
  chips. Across the full KB exactly **one** such pair exists — `enalapril` / `enalaprilat`, carried as
  separate partners by **376** entries, both matching an order named "Enalaprilat 1.25 mg" through
  #86's inflection tolerance. Prodrug and active metabolite are genuinely different DDInter entries,
  so that pair is reported rather than merged.

### 6. Also recorded, not fixed: the chip path and the render path now disagree on multiplicity

`orderedInteractionNotes`' javadoc promises that "the rendered text cannot drift from the chip". That
correspondence is per **partner**, and this PR makes it no longer one note per chip: one Major chip
now sits beside a record reading "dexamethasone (Major …); dexamethasone (Moderate …); dexamethasone
(Moderate …)", so a model answering from the record can still cite a severity the chip deliberately
discarded — and from the *more quotable* half: in that measured case the discarded Moderate note is
**659** characters against the surviving Major row's **326**. Nothing goes missing, which bounds it:
`render` emits every promoted note, so a duplicated partner can only push another partner into its
compact `name (Severity)` form, never out. Collapsing the render path needs the same thing the chip
cannot decide either — which variant the order is — so it waits on the route vocabulary that is the
data-side half of #115. The note now sits next to the invariant it weakens, rather than being findable
only in this PR's body.

### Deferred, each with its consequence stated

- **Give `orderedInteractionNotes` the shared `partnerLabel`.** If we ship without it, a padded curated
  token is spelled two ways — `warfarin   (Major. …)` in the record, `active order warfarin` on the
  chip — because only one of the two label sites trims. Same partner either way, so the consequence is
  whitespace rather than a wrong name; it is a second production surface and belongs in its own change.
- **`addInteractions` still uses `!getNote().isEmpty()` where every other consumer uses blank
  semantics.** If we ship without it, a curated row whose note is only whitespace renders a chip ending
  in a dangling `" — "` while the record for the same rule renders the bare label. Zero reach on any
  shipped dataset, and blank-note rendering is not #115's concern.
- **`DdiDrugReferenceSourceTest`'s class javadoc does not mention the collapse** though six of the ten
  collapse tests live there. If we ship without it, someone reworking the collapse runs only the
  behaviour-named class and ships a regression. Not applied because that file is already the one
  conflicting file against `main` and #119 edited that very javadoc — the pointer belongs in the rebase.
- **`severityRank`'s "every curated hand-authored rule is unrated"** is an absolute this PR's own
  fixture disproves. If we ship without qualifying it, a reader treats it as an invariant and prunes
  the rated-vs-unrated branch as unreachable. Qualified in the new test's comment only: the same
  sentence appears verbatim in three places on `main`, and correcting one copy would leave four
  disagreeing. Belongs in one sweep.

### Load-bearing decisions: all six left standing

Nothing in the six was reversed and no rationale was shortened — the four corrections above are
factual fixes *inside* that prose, and the two test changes make it stricter. The decision I was most
tempted by is (1)'s stated trade-off — a patient on a topical form seeing the systemic severity — and I
left it exactly as written, including in the PR body, because the alternative direction (picking
Moderate in case the order is topical) hides a real Major interaction for the much commoner systemic
order. The one place I was invited to change a settled ordering, and did not, is the note-less-winner
corner above: a reviewer proposed making a note outrank a rating, which would have quietly reversed
decision (3) for exactly the operator rules it protects. It is recorded as an accepted cost instead.
`severityPriority` stays extracted; the class arm stays untouched; the dataset stays unflattened;
`SafetyWarning` still has no `equals`.

### Test numbers

`mvn clean install`, full reactor from the repo root: **API 673 + OMOD 53 = 726 run, 0 failures, 0
errors, 34 skipped** (this PR was 720; `main`'s baseline on `13690b1` was 715).

### Live re-verification

Required — the hardening changes production code (`partnerLabel`'s trim, `noteLength`'s trim). Re-ran
on the same 3.7.1 standalone, same GPs
(`sourceFormat=ddinter`, full 19 MB KB, `minInteractionSeverity=minor`, `chartMode=fullChart`,
`llm.engine=local`), full-JVM restart, retrieval proven alive (`querystoreBuild … hits=75`), warm
control fired before any capture. Margaret King's order re-confirmed active first, so a reduced chip
count cannot be a vanished order:

```
PRE-CHECK — Margaret King's active drug orders
    drug orders: 1
       (NEW) Dexamethasone: 4.0 Milligram Oral Once daily | voided= False | stopped= None | expires= None
```

**15/15 checks passed.** Headline, both runs, chips byte-identical between them:

```
--- HEADLINE Margaret King + "Is it safe to give voxelotor?" [run 1] chips=1
    [interaction] drug='Voxelotor'
        Voxelotor interacts with active order dexamethasone — Major. Coadministration with potent or
        moderate inducers of CYP450 3A4 may significantly decrease the plasma concentrations and
        pharmacologic effects of voxelotor. […]
    references: [(121,'safety_finding'), (120,'drug_reference')]

PASS run 1 raises exactly ONE chip (was 3 on 13690b1)                 -> chips=1
PASS run 1 chip is the MAJOR dexamethasone row
PASS run 1 chip is not a Moderate variant (no Cushing prose)
PASS run 1 chip partner label carries no doubled space
PASS run 1 cites at most ONE safety_finding record (3 were cited on 13690b1)  -> records=1
PASS run 2 … same five                                                -> chips=1, records=1
PASS headline chips identical across the two runs
PASS control 1  Mary + clarithromycin   -> 1 chip, Major, x simvastatin
PASS control 2  Agnes + warfarin        -> 1 chip, Major, x aspirin
PASS control 3  Joshua + ibuprofen      -> 2 chips ['contraindication', 'interaction']
PASS control 4  Mary + paracetamol      -> 0 chips
15/15 checks passed
```

Both hardening changes are provable **no-ops on every shipped dataset**, which is what the identical
capture reflects. Scanned the deployed 19 MB KB (2283 drugs / 8234 mechanisms / 295,184 rows): 0
padded `rxnorm_name` fields, 0 padded mechanism `text` fields, and the single padded `name` field
(DDInter1983) has an `rxnorm_name`, so its derived token is clean; the 28 rxnorm-less drugs that do
fall back to the raw display name are all unpadded. Same for both bundled fixtures. So no ddinter
rule can carry a padded token and no ddinter note's trimmed length differs from its raw length — the
two trims guard the curated/operator path. Not decorative even for ddinter, though: DDInter1983
carries 252 interaction rows, and a KB refresh dropping its `rxnorm_name` would pad its token.

Incidentally reproduced the known-open grounding instability (now #122) a third time: the single
deterministic finding record came back `grounded=True` on run 1 and `grounded=False` on run 2, same
question, same build, same instance. Unrelated to chip count; not touched here.

Answer prose is not asserted on and no claim is made from it (`cache_prompt` KV reuse,
`LocalLlmEngine.java:786-792`). For the record only, the answers were 452 and 626 chars with one
marker per record — consistent with the direction #115's comment predicted, still n=2, still an
observation rather than a claim.

The capture was taken at `dcf0d14`. Everything pushed after it changes only comments, test resources
and one visibility keyword (`severityRank` → `private`), so the production behaviour the capture
describes is the behaviour of the branch head.

### Note for whoever rebases this

`main` has moved to `ba4cff0` and this branch is **CONFLICTING** — one file,
`DdiDrugReferenceSourceTest.java`, where #119 added its marker tests and a shared
`ddiFixtureEntries(...)` loader. Both production files auto-merge. Three specifics for the
resolution:

1. `equalSeveritiesKeepTheRowCarryingTheFullerMechanismNote` asserts
   `startsWith("… — Major. INTERVAL:")`. #119 strips that marker at the parse boundary, so the
   expected value must become the stripped text — ideally as
   `assertEquals("Dolutegravir interacts with active order iron — Major. " + DOLUTEGRAVIR_MECHANISM, …)`
   reusing `main`'s own constant, which is strictly stronger than the current `startsWith`. The test's
   *outcome* is unaffected: the note is 236 chars with the marker and 226 without, still comfortably
   above the competing row's 171, so the same row wins the tie. The consequence is recorded in the
   test's comment on this branch so it is not discovered as a mystery failure.
2. `routeVariantEntries()` should delegate to #119's `ddiFixtureEntries(...)` instead of opening the
   stream itself — one loader, per that file's new shape.
3. Add the missing pointer while you are in that file: its class javadoc names only the parser, though
   six of the ten tests that specify the #115 collapse live there. One sentence naming #115 and
   `InteractionPartnerGroupingTest`.
4. `partnerLabel`'s javadoc says the match "trims the rule's token and matches it case-insensitively
   against the order name" — phrased to stay true after #128, which moved the case-fold from
   `hasActiveDrug` into `DrugReference.matchesOrderName`. Verified: all six (order name, rule token)
   pairs the new tests use still match under #128's word-start rule, and the three that must not match
   still do not.

Recommendation for #89 is unchanged: land this first, then have #89's rule-row loop iterate
`bestRulePerPartner(ref, context, severityFloor)` and drop its own first-wins `put`.

dkayiwa:
## Rebased onto `main` (`ba4cff0`) — now **MERGEABLE**

Rebased, not merged, so the branch is still seven commits on top of `main`. Exactly one file
conflicted, `DdiDrugReferenceSourceTest.java`, as predicted; both production files auto-merged and I
verified that mechanically rather than by eye:

```
git diff 13690b1 bf92b5f -- …/DrugSafetyValidator.java   (pre-rebase)
git diff ba4cff0  HEAD    -- …/DrugSafetyValidator.java   (post-rebase)
=> byte-for-byte identical
```

`DrugReferenceInjector.java` auto-merged against #123's rework of the same method and my three hunks
survived unchanged (the `#115` paragraph, `new InteractionNote(rendered, compact, i.getSeverity())`,
and `InteractionNote`'s severity-string constructor). No production line was edited during the
rebase.

### The load-bearing resolution: the tie-break is now measuring a different string

#119 strips DDInter's residual field markers at the parse boundary, and the dolutegravir × iron row is
the one mechanism in the route-variant fixture that carries one. **Re-derived from merged `main`'s
actual `RESIDUAL_FIELD_MARKER` regex applied to the fixture, not from my earlier notes:**

| | note the tie-break compares |
|---|---|
| Iron (bisglycinate), row 1 | 171 chars |
| Iron, row 3 — pre-#119 | 236 chars (`Major. INTERVAL: Coadministration …`) |
| Iron, row 3 — post-#119 | **226** chars (`Major. Coadministration …`) |

So the winner does **not** change (226 > 171, was 236 > 171), and I checked every other multi-row
group in the fixture the same way — `2248` is the only stripped mechanism, and all six group winners
are unchanged. Worth stating plainly because the tie-break is a *length* comparison and #119 changed
the lengths: had the marker sat on the shorter row instead, the same 10 characters would have flipped
which mechanism a clinician reads.

`equalSeveritiesKeepTheRowCarryingTheFullerMechanismNote`'s expected value is restated, and while
restating it I **strengthened** it from `startsWith(…)` to `assertEquals(…)` against `main`'s own
`DOLUTEGRAVIR_MECHANISM` constant — verified byte-identical (219 chars) to my fixture's group 2248
text post-strip, so the two tests now agree on one string instead of two spellings of it. No
assertion was relaxed anywhere in the rebase.

### The other three resolution notes, all discharged

- `routeVariantEntries()` now delegates to #119's `ddiFixtureEntries(...)` — one loader, with the
  fixture path hoisted to a `ROUTE_VARIANT_FIXTURE` constant beside `MARKER_FIXTURE`.
- The class javadoc now says the file is not only parser behaviour: six of its cases specify the #115
  collapse, and it points at `InteractionPartnerGroupingTest` for the shapes only a hand-authored
  dataset can pose. (This was the deferral I had assigned to this rebase.)
- `partnerLabel`'s javadoc was already phrased as the matcher's *contract* ("trims the rule's token and
  matches it case-insensitively against the order name") rather than its expression, so it stayed true
  when #128 moved the case-fold into `DrugReference.matchesOrderName`. No edit needed.

### #128's matcher, confirmed through the real pipeline

I previously checked the six (order name, rule token) pairs against a re-implementation of #128's
rule. That was a simulation; on merged `main` the confirmation is the production path itself, and it
discriminates properly rather than passing vacuously:

- `"Warfarin 5mg"` × `"Warfarin"` **and** × `"  warfarin  "` — the padded row is the *Major* one, so if
  #128's word-start rule had stopped matching it the surviving chip would be the Moderate row and the
  exact-equality assertion would fail. It does not.
- `"Sodium valproate 500mg"` × `"valproate"`, `"Citalopram 20mg"` × `"citalopram"`,
  `"Digoxin 125mcg"` × `"digoxin"` — each pair is the only match in its case, so a lost match is a
  chip count of 0, not a silent pass.
- `"Iron 65mg"` × `"iron"` and `"Dexamethasone 4mg"` × `"dexamethasone"` — pinned by the DDInter cases,
  including the two-chip ordering case which would also catch a *new* over-match.
- The ATC arm (`"B01AA03"`) is pinned by `rowsCarryingOnlyAnAtcCodeAreGroupedAndLabelledByThatCode`.

### Tests

`mvn clean install`, full reactor from the repo root: **API 698 + OMOD 58 = 756 run, 0 failures, 0
errors, 34 skipped.**

Accounted for: `main` is at 745 and this branch adds exactly **11** tests — verified by counting the
`public void` additions in `git diff ba4cff0..HEAD -- api/src/test/java`, not by subtraction — so
745 + 11 = 756. The pre-rebase 726 was 715 (`13690b1`) + the same 11; `main` contributed the other 30.

One process note against my own work: my first resolution of the larger conflict dropped a closing
brace, because the conflict's trailing common context was a single `}` that closed the last method on
*either* side and I concatenated both blocks. The build caught it immediately. Fixed as a `--fixup`
autosquashed into the commit that introduced it, so no commit in the rebased history is red — checked
by brace-balancing that file at all seven revisions.

### Live re-verification

Re-run on the rebased build, because the rebase moved two things this logic composes with: #128
changed how `hasActiveDrug` matches a token against an order name, and #119 changed the note text the
tie-break measures. Same 3.7.1 standalone, same GPs, full-JVM restart, retrieval proven alive, warm
control before any capture, Margaret King's order re-confirmed active first:

```
PRE-CHECK — Margaret King's active drug orders
    drug orders: 1
       (NEW) Dexamethasone: 4.0 Milligram Oral Once daily | voided= False | stopped= None | expires= None
```

**15/15 checks passed**, and the headline is unchanged from the pre-rebase capture:

```
PASS  run 1 raises exactly ONE chip (was 3 on 13690b1)                        -> chips=1
PASS  run 1 chip is the MAJOR dexamethasone row
PASS  run 1 chip is not a Moderate variant (no Cushing prose)
PASS  run 1 chip partner label carries no doubled space
PASS  run 1 cites at most ONE safety_finding record (3 were cited on 13690b1) -> records=1
PASS  run 2 … same five                                                       -> chips=1, records=1
PASS  headline chips identical across the two runs
PASS  control 1  Mary + clarithromycin  -> 1 chip, Major, x simvastatin
PASS  control 2  Agnes + warfarin       -> 1 chip, Major, x aspirin
PASS  control 3  Joshua + ibuprofen     -> 2 chips ['contraindication', 'interaction']
PASS  control 4  Mary + paracetamol     -> 0 chips
15/15 checks passed
```

The surviving chip, verbatim:

```
[interaction] drug='Voxelotor'
    Voxelotor interacts with active order dexamethasone — Major. Coadministration with potent or
    moderate inducers of CYP450 3A4 may significantly decrease the plasma concentrations and
    pharmacologic effects of voxelotor. The proposed mechanism is accelerated clearance of
    voxelotor due to induction of the CYP450 3A4 isoenzyme, which is the primary route of
    elimination of voxelotor.
```

So **nothing in the rebase changed the observable outcome**: 1 chip, the Major one, byte-identical
across both runs, and one cited finding record instead of three. Answer prose is not asserted on and
no claim is made from it.

One measured observation, not a claim: the grounding verdict on the deterministic finding came back
`grounded=False` on both runs here, where the pre-rebase capture had `True` then `False`. That is the
same instability already filed as #122 — three captures now, all on a module-generated line — and it
is orthogonal to the chip count. Nothing here addresses it.

### Unchanged from the hardening pass

The six load-bearing decisions still stand, and the deferred items are still deferred as reported: the
`orderedInteractionNotes` label sweep, `addInteractions`' blank-note predicate, the
"every curated hand-authored rule is unrated" absolute (three verbatim copies on `main`, one sweep),
and the note-less-winner corner. The `enalapril`/`enalaprilat` residual stays documented rather than
fixed.

PR #123, opened 2026-08-04
fix(drug-reference): move the render's own bookkeeping out of the citable record (#117)

### LINKED ISSUE #117: render()'s internal tail ("and 824 more interactions on file", "Source: DDInter 2.0…") is inside the citable record, so the model recites it into the answer
### Summary

`render` writes internal bookkeeping into the reference record — a withheld-count tail and a dataset attribution — and the model, told to cite records, recites it. A clinician asking a one-drug safety question gets a 1492-character answer that ends in the module's own truncation counter.

Live on the 3.7.1 standalone (HEAD `13690b1`), Mary Smith (active order Simvastatin 20mg), "Can I prescribe erythromycin for this patient?" — full answer, verbatim:

> Erythromycin interacts with active order simvastatin — Major. Coadminstration with potent inhibitors of CYP450 3A4 may signficantly increase the plasma concentrations of simvastatin and lovastatin and their activ e acid metabolites, all of which are primarily metabolized by the iso enzyme. **[76]** Erythromycin (ATC D10AF02, J0 1FA01, S01AA17) has several interactions, including with **ivosidenib** (Moderate. … Increased exposure to ivosidenib ma y increase the risk of QT interval prolongation … torsade de pointes and sudden death. …); **ixabepilone** (Moderate. …); **and 824 mor e interactions on file. Source: DDInter 2.0 (via openmrs-ddi-knowledge-base). [75]**

The first sentence is the answer. The remaining ~1150 characters are two interactions the patient has nothing to do with, plus:

- `and 824 more interactions on file` — `DrugReferenceInjector:531`, a note to the model about what the cap withheld;
- `Source: DDInter 2.0 (via openmrs-ddi-knowledge-base).` — `:537`, provenance for the record, not for the answer.

Neither is clinical content. Both are now in a clinician-facing answer.

### Root cause

Two things compound.

**The tail strings are inside the citable record.** `render` appends them to the same string the model is instructed to cite, so there is no boundary telling the model they are metadata. They are useful — the withheld count is honest about truncation, and provenance belongs *somewhere* — but the record is the wrong carrier, because everything in it is quotable.

**Segment 2 is guaranteed non-empty.** `render` deliberately always shows at least one dataset-order partner (:524-526):

```java
if (restShown == 0 && restStart < ordered.size()) {
    shown.add(ordered.get(restStart).full);
}
```

That guarantee is reasoned for — the entry is also the only general reference material about the drug — but its consequence here is that the patient-relevant partner (simvastatin, correctly promoted by #110/#84) is followed by whatever sits at the head of the remaining 826, and the model treats the two as equally worth reporting. The promotion ordering fixed *which* partner comes first; it did not stop the tail being recited.

### Supporting observation (model artifact, not a code defect)

The recited text is visibly corrupted: `Coadminstration`, `signficantly`, `activ e acid`, `iso enzyme`, `ma y`, `i matinib`, `netu pitant`, `induc er`, `its ow metabolism`, `824 mor e`. The KB text is clean — `mechanisms["2085"].text` reads "Coadministration … may significantly increase … their active acid metabolites" — and the chip, which is built in Java from the same string, is clean too. So this is the quantised local model degrading over a long verbatim copy, not a parsing bug. It is worth recording because it is an argument for the fix: the more irrelevant reference prose is put in front of this model to recite, the more mangled clinical text a clinician sees. (Prose here is also not byte-reproducible — `LocalLlmEngine.java:786-792` — so the specific corruptions will vary run to run; the length will not.)

### Scope

`DrugReferenceInjector.render`, all source formats, but only visible at DDInter's breadth (erythromycin: 826 partners; warfarin ~934). The curated seed never had enough partners for a tail to matter.

### Proposed fix

Keep the withheld count and the source out of the quotable text. Options, roughly in order of appetite:

- move both to the `RecordMapping` as structured fields, so the client can show provenance on the citation chip and the model has nothing to recite;
- or keep them in the record but outside the cited span, if the prompt format can express that;
- and separately reconsider whether segment 2 should fire when a **promoted** partner already exists — the "always show one" guarantee earns its keep when the patient has no relevant partner, which is not this case.

The prompt could also be told not to enumerate a reference record's unrelated partners, but that is the weaker lever: #110's README records that prompt-shaped fixes in this area were tried and regressed, so the structural fix should come first.

### Tests

`DrugReferenceInjectorTest` through the real `render`/`injectRecords` path: assert the record text handed to the model carries no withheld-count or source string once they move (and that the count is still exposed structurally). Plus a case with one promoted partner asserting segment 2 stays empty, if that change is taken.

dkayiwa:
## The defect

`DrugReferenceInjector.render` wrote its own bookkeeping into the string the model is told to
cite, and the model recited it. Live on the 3.7.1 standalone at `13690b1` — Mary Smith, active
order Simvastatin 20mg, "Can I prescribe erythromycin for this patient?" — a **1492-character**
answer whose first sentence was the answer and whose remaining ~1150 characters were two
interactions the patient has nothing to do with, ending in the module's truncation counter and its
dataset attribution.

## Root cause

Two things compounded, and both are structural.

**1. The tail strings lived inside the citable record.** `"; and " + withheld + " more interactions
on file"` (`:531`) and `" Source: " + ref.getSource() + "."` (`:537`) were appended to the same
string the prompt instructs the model to cite. There is no boundary in a record marking part of it
as metadata — everything in it is quotable. Both facts are worth keeping (the count is honest about
truncation; provenance belongs somewhere), so the record was simply the wrong carrier.

**2. Segment 2 spends the leftover budget on full notes for irrelevant partners.** Worth being
precise about, because it is not the `restShown == 0` guarantee that fired here: erythromycin's
promoted simvastatin note is ~330 of the 1500-char budget, so the *budget loop* admitted ivosidenib
and ixabepilone in full. Segment 2's stated purpose is that the record is also the only general
reference material about the drug — it must not read as if the patient's own overlap were the
drug's only interaction. Naming one partner with its severity states that. The mechanism prose is
what is actionable for a partner the patient is actually on, and the issue's corruption observation
(`Coadminstration`, `activ e acid`, `824 mor e`) is an argument for offering less of it: every
irrelevant paragraph put in front of this quantised model is a paragraph of mangled clinical text a
clinician may be shown.

## What changed

- `render` now returns a `RenderedReference` (text + `source` + `withheldInteractions`) instead of
  a `String`. The text carries neither the withheld tail nor the source.
- `RecordMapping` and `RecordReference` gained `source` / `withheldInteractions` fields, carried
  through `extractCitedReferences` and `withGrounded`, and emitted by the shared
  `serializeReferences` — so all four emission sites (`/search`, the early `references` event,
  `done`, and the async `grounded` event) publish them. A client renders provenance and honest
  truncation on the citation chip; the model has nothing about the record to recite. Both keys are
  always present (`null` / `0` for a chart record, which is that record's real shape). README
  updated.
- Segment 2: when a patient-relevant partner is already rendered, exactly **one** dataset-tail
  representative renders, in the compact `name (Severity)` form. When there is none, the branch is
  byte-for-byte the old behaviour — full notes to the budget, the first always rendering. That old
  explicit "re-add one" line could only ever fire in the promoted case (with no promoted partner
  `shown` is empty at loop entry, so the first note always renders), and the promoted case now
  always shows one, so the guarantee is preserved rather than dropped.
- The chart line and the mapping text stay byte-identical, so the grounding verifier still compares
  against exactly what the model read.

Two knock-on effects I looked at deliberately, both in the safe direction. Echo scoping (#105)
exempts a drug the answer recites out of a cited record; a shorter tail means fewer names in the
record, so fewer exemptions — the conservative direction (more validation, not less), and a drug the
model can no longer see is a drug it can no longer recite. Grounding for `drug_reference` citations
is demote-only, so a shorter record can only move a verdict between `null` and `false`, never to a
false `true`; with less irrelevant prose diluting the embedding, a relevant claim should demote less
often, not more.

## What I deliberately did NOT change

- **The promotion ordering from #110/#84.** `orderedInteractionNotes` is untouched: patient-relevant
  partners first, most-severe first, sub-floor rules not promoted.
- **`MAX_INTERACTION_RENDER_CHARS` as a soft budget with a bounded overshoot.** Segment 1 still
  overrides the budget so a relevant partner is never invisible; the overshoot is now *smaller*,
  because a compact representative costs ~20 chars instead of a ~900-char note.
- **I did not suppress segment 2 entirely when a promoted partner exists**, which the issue floats.
  Three existing tests pin that a dataset-order partner renders alongside a promoted one, one of
  them (`DrugSafetyValidatorEchoScopingTest.echoedReferencePartnerIsNotValidatedAsAProposal`)
  depending on it for the premise of the #105 echo-scoping guarantee — a partner recited out of a
  cited record must not chip as a proposal, which needs such a partner to exist in the record.
  Those tests are the specification and their rationale is sound, so the fix keeps the partner and
  removes what actually caused the recitation: its mechanism paragraph.
- **No prompt wording change.** #110's README records prompt-shaped fixes in this area regressing.

## Tests I had to change — please read this bit

Two existing assertions asserted the literal `"more interactions on file"` in the text, which is
exactly what this issue asks to remove. They cannot both hold:

- `DrugReferenceInjectorTest.interactionRenderCapStillBoundsTheRenderedSection:243`
- `DdiDrugReferenceSourceTest.renderCapBoundsBroadInteractionSets:108`

In both, that assertion is the *precondition that truncation happened* — without it the neighbouring
length bound could pass vacuously. Each is now asserted on the new carrier
(`getWithheldInteractions() > 0`): same fact, same strength, new location. Nothing was weakened and
no expected value was adjusted to match new behaviour. The five other interaction-rendering
assertions in those files are untouched and pass unchanged.

## New tests

- `DrugReferenceInjectorTest.theRenderersOwnBookkeepingIsStructuralNotCitableText` — the record text
  and the whole prompt chart text carry neither literal; the count (13 of the bundled Lisinopril
  entry's 15 partners) and the source are exposed structurally.
- `DrugReferenceInjectorTest.theDatasetTailRepresentativeDropsItsProseWhenAPatientRelevantPartnerIsRendered`
  — with a promoted partner, the one tail representative renders as `name (Severity)`, its mechanism
  prose is gone, and no second tail partner renders. (Before the fix this case rendered five full
  dataset-order notes.)
- `LlmInferenceServiceTest.extractCitedReferences_shouldCarryCitationMetadataFromTheMapping` — the
  mapping→reference hop, over mappings from the real injector; and
  `groundingVerdictMustNotDropCitationMetadata` for the `withGrounded` copy.
- `ChartSearchAiReferenceProvenanceTest` (omod) — the wire keys, through the real controller's real
  serialization, for both a reference and a chart entry.

Both injector tests were confirmed red before the fix and green after.

## `mvn test`

```
API  666 run, 0 failures, 0 errors, 34 skipped
OMOD  55 run, 0 failures, 0 errors,  0 skipped
```

Baseline at `13690b1` was 662 + 53 = 715; the six new tests account for the difference. The 34
skipped are the pre-existing LLM-requiring `PromptInjectionEvalTest` / `LlmAnswerQualityTest`.

## Live standalone evidence

3.7.1 standalone, full 19MB DDInter KB, `sourceFormat=ddinter`, `minInteractionSeverity=minor`,
`chartMode=fullChart`, local CPU engine. Full-JVM restart; querystore retrieval gated live
(`hits=75`) and the known-positive warm control fired its chip before any capture was taken. No
data was created or modified on the box.

Answer prose is not byte-reproducible here (`cache_prompt` KV reuse, `LocalLlmEngine.java:786-792`),
so the claims below are about **length** and the **presence/absence of the two literal strings** —
and the question was asked three times, not once.

### Mary Smith `38beca4a-…` — "Can I prescribe erythromycin for this patient?"

**BEFORE** (`13690b1`, captured on this box before any change) — **1492 chars**, verbatim:

> Erythromycin interacts with active order simvastatin — Major. Coadminstration with potent inhibitors of CYP450 3A4 may signficantly increase the plasma concentrations of simvastatin and lovastatin and their activ e acid metabolites, all of which are primarily metabolized by the iso enzyme. [76] Erythromycin (ATC D10AF02, J0 1FA01, S01AA17) has several interactions, including with ivosidenib (Moderate. Coadminstration with moderate inhibitors of CYP450 3A4 may significantly increase the plasma concentrations of ivosidenib, which is primarily metabolized by the isoenzyme. Increased exposure to ivosidenib ma y increase the risk of QT interval prolongation, which has been associat ed with ventricular arrhythmias including torsade de pointes and sudden death. Coadministration with ivosidenib may decrease the plasma concentrations of drugs that ar e primarily metabolized by CYP450 3A4, including ma ny of the known moderate inhibitors of the isoenzyme such as amprenavir, aprepitant, crizotinib, darunavir, diltiazem, dronedarone, erythromycin, i matinib, isavuconazonium, mifepristone, netu pitant, and verapamil. Ivosidenib is an induc er of CYP450 3A4 and induces its ow metabolism following multiple dosing.); ixabepilone (Moderate. Coadministration with inhibitors of CYP450 3A4 ma y increase the plasma concentrations of ixabepilone, which is primarily metabolized by the isoenzyme.); **and 824 mor e interactions on file. Source: DDInter 2.0 (via openmrs-ddi-knowledge-base).** [75]

**AFTER** — three runs, **455 / 328 / 328 chars**. `"more interactions on file"` and `"Source:"`
absent from all three. Verbatim, run 1 and run 2:

> Yes, the records address the safety of prescribing Erythromycin. The safety finding states: Erythromycin interacts with active order simvastatin — Major. Coadminstration with potent inhibitors of CYP450 3A4 may signficantly increase the plasma concentrations of simvastatin and lovastatin and their active acd metabolites, all of which are primarily metabolized by the isoenzyme [77]. This interaction is noted in the drug reference for Erythromycin [76].

> The records indicate that Erythromycin interacts with active order Simvastatin — Major. Coadministration with potent inhibitors of CYP450³A4 may significantly increase the plasma concentrations of simvastatin and lovasátin and their active acid metabolites, all of which are primarily metabolized by the isoenzyme. [77] [76] [1]

**69–78% shorter, and what is left is the answer.** The two irrelevant partners are gone from the
prose because they are gone from the record. The chip is unchanged (1, Major, erythromycin ×
simvastatin), and the truncation the tail used to announce now arrives on the citation instead:

```
#76 type=drug_reference group=reference grounded=None
    source='DDInter 2.0 (via openmrs-ddi-knowledge-base)' withheldInteractions=825
#77 type=safety_finding group=reference grounded=True  source=None withheldInteractions=0
#1  type=drug_order     group=chart     grounded=True  source=None withheldInteractions=0
```

825 withheld of erythromycin's 827 partners — the promoted simvastatin note plus one compact tail
representative are rendered. Nothing is withheld from safety checking; the validator still reads
every interaction, which is why the chip is unaffected.

### Regression controls — all four unchanged

| control | expected | observed |
| --- | --- | --- |
| Mary + "Is it safe to give her clarithromycin?" | 1 chip, Major, clarithromycin × simvastatin | 1 chip, Major, clarithromycin × simvastatin (291 chars) |
| Agnes + "Is it safe to start warfarin?" | 1 chip, Major, warfarin × aspirin | 1 chip, Major, warfarin × aspirin (417 chars) |
| Joshua + "Can I give ibuprofen?" | 2 chips (NSAID cross-reactivity + lisinopril Moderate) | 2 chips, both, in that order (438 chars) |
| Mary + "Is it safe to give her paracetamol?" | 0 chips | 0 chips, 0 references, "The records do not address paracetamol safety." |

Neither literal string appears in any of the six captures. `source` / `withheldInteractions` are
populated on every `drug_reference` citation (896 withheld for clarithromycin, 932 for warfarin, 717
for ibuprofen) and `null` / `0` on every chart record and safety finding.

Closes #117


dkayiwa:
Reviewed the two modified test assertions, since changing an existing test is normally a rule violation here. **They are legitimate, and the net effect is stricter — recording the check so a reviewer does not have to re-derive it.**

The two old assertions asserted the literal `"more interactions on file"` **inside the record text**. That is precisely the string this issue requires removing from the record text, so the old assertion and the issue are contradictory statements of the spec; one of them had to yield. What matters is whether the *invariant* survived, and it did:

| | before | after |
|---|---|---|
| `DrugReferenceInjectorTest` | `assertTrue(text.contains("more interactions on file"))` | `assertTrue(injectedReference(result).getWithheldInteractions() > 0)` |
| `DdiDrugReferenceSourceTest` | `assertTrue(section.contains("more interactions on file"))` | `assertTrue(injectedMappingFor(...).getWithheldInteractions() > 0)` |

Same fact — truncation happened and the remainder is reported — moved from a substring probe to the typed field that now carries it. And the PR *adds* assertions stronger than anything removed: `assertEquals(13, ref.getWithheldInteractions())`, an exact `getSource()` match, and `assertFalse(...contains("more interactions on file"))` / `assertFalse(...contains("Source:"))`. An exact count is a tighter statement than "the substring is present somewhere".

So: no expected value was adjusted to match new behaviour, and nothing passes vacuously that did not before. The repo rule ("tests are the specification; only fix production code to satisfy the existing tests") is aimed at weakening; a test that pins the *carrier* a fix must change is the one case it cannot be applied literally. Flagging it under its own heading rather than burying it was the right handling.

Two other things in this PR worth surfacing, because both correct things stated elsewhere:

**1. It corrects the root-cause analysis in #117.** The issue named the `restShown == 0 && restStart < ordered.size()` guarantee (`:524-526`) as the reason irrelevant partners appear, and floated suppressing segment 2. This PR shows the answer's bulk actually came from the **budget loop**, not the fallback guarantee — the ivosidenib/ixabepilone *mechanism paragraphs*. Removing those is what produced the 69–78% reduction; the guarantee was largely a red herring, and my framing of it in the issue was wrong.

**2. Declining to suppress segment 2 has a dependency I did not know about.** `DrugSafetyValidatorEchoScopingTest.echoedReferencePartnerIsNotValidatedAsAProposal` relies on a dataset-order partner being present in the record: #105's premise is that a partner *recited out of a cited record* must not chip as a proposal, which requires such a partner to exist in the record in the first place. Suppressing segment 2 outright would have quietly undermined #105's regression coverage. Keeping the partner and dropping its mechanism prose is the better cut, and this is a good example of why the "check for forgotten call sites" pass in CLAUDE.md earns its keep.

Also noted for whoever sequences these: this PR changes `RecordMapping` → `RecordReference` → wire shape and updates `serializeReferences`, so it touches the same serialization path as the reference-group work. Expect a textual rebase against the other DDI PRs in flight (#119/#120/#121), though no semantic conflict is apparent.

dkayiwa:
## Hardening pass (`/harden`) — 2 commits pushed

Ran the review/simplify convergence loop over this slice. **No production behaviour changed**: every production-file edit in these two commits is a comment or javadoc, verified mechanically — `git diff fd74524 -- api/src/main omod/src/main`, filtered to lines that do not begin with `*`, `//` or `/*`, is empty.

### What it found

Two coverage gaps where the PR's own claims were unpinned, both confirmed by mutation:

- **`/search` — the fourth `serializeReferences` call site — was untested for the two new keys.** The PR body claims all four emission sites publish them; three SSE sites were pinned. `ChartSearchAiSearchResponseGroupingTest` exists *precisely* for this: its class doc says "covering the shared helper proves the helper, not that this endpoint calls it", and its `keepsTheOtherReferenceFields` enumeration predates the two keys. Commenting out the two `refMap.put` lines left that file's three existing tests green.
- **A module-derived safety finding is a `reference`-group record carrying no attribution** — the pair that makes the new README warning ("branch on the value, not the group") true. #110 made that finding a *citable record*, which is exactly the carrier #117 proved a source string must never ride on. Giving the finding a source failed **1 of 668** api tests: the new one.

Then five places where this change falsified surrounding prose, one operator-facing and flatly wrong:

- **`docs/drug-kb-demo.md`** described `source` as a "Provenance string rendered in the reference text" in the KB-authoring schema — now the opposite of the truth, and the exact belief #117 exists to remove. An operator would author a KB expecting the attribution to land in the citable record.
- **`docs/adr.md`** still enumerated the reference wire shape without the two keys, in both the JSON example and the `done`-event field list. #108 updated README *and* the ADR in one commit when it added `group`, so the ADR is maintained as part of this contract.
- **`withheldInteractions` was documented in five places as what "the render budget left out".** Measured on the bundled sample (Lisinopril, patient on digoxin): the section is ~350 of the 1500-char budget and the count is 13 — *nothing* was dropped for length. The dominant cause is now the deliberate one-representative rule, so a client following those docs would tell a clinician "13 omitted for length", stating a reason that is not the reason. Reworded in README, `PatientChartSerializer`, `ChartSearchService`, and both spots in `DrugReferenceInjector`.
- **`InteractionNote.compact`'s comment** said the form exists for when a note does not fit the budget. Segment 2 now reaches it unconditionally with budget to spare, so that is no longer its only or common trigger.
- **`DdiDrugReferenceSourceTest.renderCapBoundsBroadInteractionSets`** still opened with "the render caps it and summarises the remainder", contradicting the updated comment eight lines below.

And two untested arms of the new code:

- **`restStart == ordered.size()`** — the patient is on every above-floor partner, so no dataset tail remains. Nothing reached that arm. Relaxing the guard to `<=` throws `IndexOutOfBoundsException` out of `render`; `inject` catches every `RuntimeException`, so the whole drug-reference feature — **including #110's deterministic safety findings** — would vanish behind one `log.warn` for exactly the polypharmacy patients it matters most for, with no test failing. Confirmed: that mutation was caught by the new test and by nothing else in 670.
- **The `source` blank-coalesce and trim.** Pre-#117 a whitespace-only operator value rendered `" Source:    ."` into the text; it is now published as a wire field, so blank must arrive as `null` and a padded value trimmed. New fixture, because a valid entry cannot join `drug-reference-malformed.json` without changing the count `unusableEntriesAreDroppedAtParse` pins.

Polish: extracted `DrugReferenceTestSupport.injectedReference(PatientChart)` as the single "find the injected reference" matcher — it had reached three copies, two of them added by this PR.

### Load-bearing decisions — all left as they are

The re-pointed `getWithheldInteractions() > 0` assertions, segment 2 not being suppressed when a partner is promoted, the soft budget with bounded overshoot, and `orderedInteractionNotes`' promotion ordering. I re-derived segment 2 from scratch rather than taking them on trust, and each holds: `restStart == 0` is byte-for-byte the old behaviour; `withheld ∈ [0, ordered.size()]` on every path; the promoted case's overshoot shrank from ~900 chars to ~20. Nothing to re-open.

### Live re-verification — not run, deliberately

The brief requires it only when hardening changes production code in a way that could alter chip output, answer text, injected records, or the wire shape. These commits change comments and tests only (mechanically verified above), so the standalone evidence already in the PR description still describes this build exactly. Taking the shared single-instance lock would have produced no new information.

### `mvn clean install` (full reactor)

| | before hardening | after |
| --- | --- | --- |
| API | 666 | **670** |
| OMOD | 55 | **58** |
| total | 721 | **728**, 0 failures, 34 skipped |

Baseline at `13690b1` is 715. The 34 skipped are the pre-existing LLM-requiring evals.

### Found nearby, deliberately not fixed

- The README `group` table still describes the `reference` group as "module-supplied reference prose (a drug knowledge-base entry…)", which under-describes a `safety_finding`. That wording predates this slice — #110 put findings in the same group — so it belongs to that change, not this one.
- Segment 1's overshoot is bounded by *one full note plus ~20 chars per additional promoted partner*, so a patient on very many interacting drugs could in principle exceed the `2 × MAX_INTERACTION_RENDER_CHARS` bound `interactionRenderCapStillBoundsTheRenderedSection` asserts. Pre-existing in segment 1 and untouched here; this PR strictly reduces rendered length on every path, so it is not a regression from it.
- The blocking `/search` assertions read the response `Map` before Spring's Jackson converter, matching how the pre-existing `date` / `grounded` keys are asserted. If that converter were ever configured non-null-inclusive, all four keys would drop together — a uniform pre-existing property, not something these keys introduce.

dkayiwa:
## Hardening — Phase 2 round (2 more commits, `2861337` + `4094294`)

Still **comments and tests only** — `git diff fd74524 -- api/src/main omod/src/main`, filtered to lines not beginning with `*`, `//` or `/*`, is empty at every commit.

### One finding I checked and rejected

A reviewer flagged `DrugReferenceInjectorTest.interactionRenderCapStillBoundsTheRenderedSection` as having **lost its teeth**: with one promoted partner, segment 2 now emits exactly one compact representative, so `shown == 2` and `withheld == 13` regardless of `MAX_INTERACTION_RENDER_CHARS` — both its assertions would pass with an infinite budget. That reasoning is correct, and it was labelled "defect to fix before merge".

I measured it instead of applying it: setting the cap to `Integer.MAX_VALUE / 4` **fails 4 tests**, including `DdiDrugReferenceSourceTest.renderCapBoundsBroadInteractionSets` via the `getWithheldInteractions() > 0` carrier this PR introduced, plus `aSubFloorInteractionIsNotPromotedEvenWhenThePatientIsOnThatDrug`, `everyPartnerThePatientIsOnIsRepresentedEvenWhenTheirNotesExceedTheBudget` and `whenTheBudgetForcesAChoiceTheMoreSevereInteractionKeepsItsMechanism`.

So the cap's guarantee is *well* pinned; what is true is the narrower point that this one test's fixture no longer exercises the cap, while the guarantee it names is covered by four others. That is not a coverage gap, and the test itself is one I must not modify. **No change made.**

### Applied

- **Pinned the chart-line / mapping-text identity.** The injection loop's comment claims the two stay byte-identical, and the grounding verifier rests on exactly that — it judges an answer sentence against `mapping.getText()` for a claim the model formed from the *chart text*, so a divergence would ground claims against words the model never read. Nothing asserted it; the two existing "not in the record" / "not in the chart text" assertion pairs only differ if the two *can* differ.
- **Segment 2's representative is not always compact.** `InteractionNote` keeps the full text for a rule with neither token nor ATC (no name to shorten to), so an operator-authored row like that can still land a full paragraph in the tail slot. Inside the one-note overshoot already tolerated, but the comment should not be read as a guarantee the slot costs ~20 chars — the overshoot bound is reasoned about from that sentence.
- **Three more "the render budget left out" sites** missed in the first sweep (the `RenderedReference` javadoc, `serializeReferences`' comment, the done-event assertion message).
- **Two diagnostics and two consistency fixes**: a positional read that surfaced as an NPE instead of naming the missing citation; a finding-finder that kept the *last* match where its sibling keeps the first; two copies of the same arrangement inside the helper class that exists to prevent that drift; and its now-false "the one public accessor".

### Deferred, with the failure mode

- **`MAPPER` + `referencesOf` duplicated between the two omod reference tests.** If we ship without hoisting it into `SseEvents`, a third copy of the SSE reference decoder exists in a package whose own javadoc records that duplicated decoders already drifted once — so a future change to the event shape gets fixed in one copy and the other keeps asserting the old shape. Not done here because it edits two shared omod test files while seven sibling branches are in flight, and a merge collision there is likelier than the drift within this window.
- **`injectedMappingFor`'s body re-inlined in one test.** If we ship without it, that test and the helper can disagree about the context or generated question, so a fixture change silently applies to one and not the other. Lower value than the collision risk, same reason.
- **`README.md` table header casing** (`| field | Meaning |` vs `| Field | Description |`). If we ship without it, nothing breaks; the tables render inconsistently. Pure style.

### `mvn clean install` (full reactor)

API **670** + OMOD **58** = **728**, 0 failures, 34 skipped. Baseline at `13690b1` is 715.

Live re-verification still not required and not run: comments and tests only, so the standalone evidence in the description describes this build unchanged.

PR #127, opened 2026-08-04
Screen the patient's own active orders when the question asks for interactions but names none (closes #113)

### LINKED ISSUE #113: The generic DDI screening question ("any interactions with her current medications?") produces nothing — both arms require a question-named drug

dkayiwa:
## The gap

The canonical DDI screening question returned nothing. On the 3.7.1 standalone at `13690b1`, Mary Smith (active order Simvastatin 20mg):

| question | chips | references |
|---|---|---|
| Are there any drug interactions with her current medications? | 0 | `[]` |
| Is it safe to give her clarithromycin? | 1 (Major) | 2 |

Naming a candidate drug worked; asking to be screened did not — for the one question a DDI knowledge base is chiefly wanted for.

## Why both arms missed it

Both halves key off drugs the **question** names, and a screening question names none.

- `DrugReferenceInjector.matchingEntries` (:217) computes `questionDrugs` first and scopes order-driven injection through `relatedToAny` (:255), which loops `questionDrugs` and so returns `false` for an empty list. Nothing injected.
- `DrugSafetyValidator.validate`'s `inPlay` (:207) starts from `questionDrugs`, so it starts empty; it can then only be filled by drugs the *answer* names, and the answer — given no reference records — names none. Zero chips.

Nothing needed retrieving. `PatientClinicalContext` already holds the active orders, and each order's `DrugReference` already carries its full partner list. The two only needed cross-referencing.

## What changed

One new arm in `DrugSafetyValidator`, gated on the question carrying interaction-screening intent **and** naming no drug. It runs the same `hasActiveDrug` join `addInteractions` already performs and the same `clearsSeverityFloor` filter, over the reference entries for the patient's active orders rather than over question drugs. Because the partner side of the join is by definition another active order, one pass over the order entries reaches every pair — no cross-product is enumerated. The chip text moved into one extracted builder used by both arms, so they cannot word the same finding differently.

**No injector change.** The chips reach the prompt as citable `safety_finding` records through #110's existing `preAnswerFindings` path, which is why `references: []` stops being empty. Order-driven reference injection stays relevance-scoped and `orderedInteractionNotes`/`render` are untouched — same conclusion PR #120 reached for the question-pair case.

### Design decisions, and their costs

**1. Intent detection reuses `QueryScopeRouter`, not a new keyword list.** `QueryScopeRouter.isInteractionScreening(question)` requires an `interact*` cue **and** the router's own `MEDICATIONS` classification. Reusing that classification is the point — "medication-domain question" keeps one definition — and it is what keeps the trigger conservative: "Does the patient interact well with the care team?" carries no medication cue and screens nothing. It is a cue predicate like `isTemporal`, deliberately **not** an `Intent`, so retrieval scoping is untouched.

- *Cost:* a bare "any interactions?" does not trigger. Accepted — in practice a clinician names what might interact ("drug", "meds", "medications", "prescriptions"), all of which the existing MEDICATIONS cues cover.
- *Cost:* `QueryScopeRouter` becomes `public` so the `reference` package can call it. Only the one new method is public; every other member stays package-private. That does add a compile-time dependency from `reference` into `api.impl` (which already depends on `reference`). The alternative was a second cue list in the reference package or in `ChartSearchAiUtils`, which is what the issue warned against. Happy to move the router if you prefer the dependency the other way.

**2. Blast radius: severity-ordered, capped at 10, withheld pairs logged at WARN.** Candidates are quadratic in the medication list (10 active orders is 45 pairs) where the question-driven arms are linear, and every chip is also injected as a finding record, so an uncapped arm would bury the clinician *and* write the cross-product into the context window. `MAX_SCREENING_INTERACTION_WARNINGS = 10` bounds that at ~12k characters at DDInter's longest mechanism text. It is meant as a backstop, not a routine truncation — a realistic regimen produces a handful of above-floor pairs. Candidates are sorted most-severe-first before the cut (unrated sorts above Major, the same convention `InteractionNote` uses, because `clearsSeverityFloor` treats unrated as exempt rather than low), and when anything is dropped every withheld pair is **named** in a WARN line, not just counted.

- *Deferred, stated plainly:* the clinician still cannot tell a capped screen from a complete one. Surfacing that needs a UI affordance that does not exist — chips are per-drug findings, and inventing a "N more pairs" chip would be a new chip shape. The mitigation is that what is dropped is always the least severe. Promoting the cap to a GP is trivial if you want it configurable; I left it a documented constant to keep `ChartSearchAiConstants` out of the diff.

**3. `MAX_INTERACTION_RENDER_CHARS` — not deferred, out of the path.** That cap bounds the `Interactions:` section of one rendered *reference* record, and this change injects no reference record, so the multi-pair case never reaches it. The multi-pair prompt budget is governed instead by the chip cap above: one bounded `renderFinding` line per pair. Had I unscoped order-driven injection instead, a polypharmacy patient would have got one ~1500-char reference record per active order, which is exactly the prose #117 measures the model reciting verbatim — that is the main reason the injector is untouched. Pinned by a test asserting screening injects no `drug_reference` records.

**4. A pair needs TWO orders.** The join is evaluated against a derived context with everything the subject itself answers for removed. Without that, `hasActiveDrug`'s substring matching (#86 — the token `iron` inside `spironolactone`) makes one order pair with itself, and a patient on ONE drug is told it interacts with a drug they are not taking. The question-driven arm survives that because the clinician has an anchor; a question naming no drug has none, so screening must establish the second order itself. This is a property of pairing, not a copy of the matching rule — the predicate is still `hasActiveDrug`, so it follows whatever #86/#87 settles on. Mutation-checked: reverting this one condition makes the arm emit `Spironolactone interacts with active order iron — Major` for a patient with a single order.

**5. Subject set resolves by name AND ATC.** `findByActiveOrders` keys on ATC codes only, and neither key can be assumed present. **Corrected measurement** (this PR previously generalised one lookup on Mary's Simvastatin into "the demo dictionary's drug concepts carry no `ConceptMap` at all" — that was wrong): counted directly against the demo database, **616 Drug-class concepts, 158 with any `concept_reference_map`, 85 with an ATC-named source** (`Torasemide`→`C03CA04`, `Heparin sodium`→`B01AB01`, `Sulfadoxine and pyrimethamine`→`P01BD51`). Simvastatin is one of the 531 unmapped, as are Tiotropium, Spironolactone, Nitroglycerin, Budesonide and Dexamethasone — which is why every probe patient's orders landed in the unmapped majority and an ATC-only subject set was empty for every case measured here. So ATC coverage on that dictionary is **sparse but non-zero**, and the union of `findByActiveOrders` with `findByQuery` over each active order's name is a **robustness** argument, not a necessity one: mirroring `hasActiveDrug`'s own name-OR-ATC disjunction is what makes the subject set complete on a fully-mapped dictionary and on an unmapped one alike. Both are existing documented matchers; no new matching logic. The ATC paths here are **dormant on this instance, not dead code** — please do not read the measurement as licence to delete them.

**Out of scope, deliberately:** the class arm (`addClassInteractions`) is not run pairwise. Same-class duplicate therapy across a patient's own orders is a real screening finding, but it multiplies chips and collides directly with #88/#89's rule-vs-class correlation, so it belongs after that lands. The load-bearing half of that is structural and stands on its own: the class arm is not run pairwise, so it cannot reach screening at all. (It is also inert for the *particular* probe patients here, whose concepts are all in the unmapped majority per the corrected measurement in decision 5 — but that is an accident of which patients were probed, not a property of the dictionary.)

## How this composes with the sibling changes

- **#114 / PR #120 (question-pair arm):** that arm requires the question to name at least **two** drugs; this one requires it to name **none**. The gates are mutually exclusive on any single question, so the two cannot both report a pair and no shared "who owns this pair" decision is needed. Stated in a comment at the gate.
- **#88 / PR #89 (rule-vs-class dedup):** #89 correlates the two arms per subject drug, so it cannot see two different subjects and does not collapse the symmetric doubling this arm creates — which is why that collapse is done here, on an unordered pair key. Conversely this arm does not pre-empt #89.
- **#115 (route variants):** two entries sharing one `rxnorm_name` are two ids and so two pair keys, so route-variant multiplication reaches screening unchanged and is #115's to fix. Observed counts below.
- **#86 / PR #87 (`hasActiveDrug` matching):** depended on as a method, not copied. Note #87's whole-word rule will also make decision 4's guard mostly redundant.
- I have **absorbed the stale-documentation fix on #120's behalf**, as coordinated: `SafetyWarning.TYPE_INTERACTION`'s javadoc, `config.xml`'s `warnOnInteractions` description and the matching README row now describe all three joins, so the two PRs do not conflict over the same lines.

## Live evidence

Full-JVM restart onto this build on the 3.7.1 standalone, all twelve probes inside one lock, querystore retrieval gated alive (`hits>0`) and the known-positive control fired before any capture. Chips are deterministic Java; prose on that box is not reproducible (`LocalLlmEngine.java:786-792`), so the assertions are on chips and references.

```
case                                                 chips  refs  expectation
------------------------------------------------------------------------------
CASE 1  screening, 2 interacting orders                  1     1  >=1 chip
CASE 1r screening, 2 interacting orders (re-run)         1     1  >=1 chip
CASE 2  screening, 2 NON-interacting orders              0     0  0 chips
CASE 2r screening, 2 NON-interacting orders (re-run)     0     0  0 chips
CASE 3  screening, ONE active order                      0     0  0 chips
CASE 4  screening, spironolactone only (#86 shape)       0     0  0 chips
CASE 5  non-screening medication question                0     2  0 chips
CASE 6  question NAMES warfarin                          2     3  existing arm
CTL 1   Mary + clarithromycin                            1     1  1 chip Major
CTL 2   Agnes + warfarin                                 1     1  1 chip Major
CTL 3   Joshua + ibuprofen                               2     4  2 chips
CTL 4   Mary + paracetamol                               0     0  0 chips
```

**CASE 1 — the gap closed.** Barbara Miller, "Are there any drug interactions with her current medications?" — before this change the identical shape gave 0 chips and `references: []`:

```
safetyWarnings (1)
  [interaction] Acetylsalicylic acid (aspirin)
      Acetylsalicylic acid (aspirin) interacts with active order ibuprofen — Major. The antiplatelet
      and cardioprotective effect of low-dose aspirin may be antagonized by coadministration of some
      nonsteroidal anti-inflammatory drugs (NSAIDs). Ibuprofen has been specifically implicated, …
      The combined use of aspirin with NSAIDs in general may increase the potential for serious
      gastrointestinal (GI) toxicity, including inflammation, bleeding, ulceration, and perforation.
references (1)
  #249 type=safety_finding group=reference grounded=True
answer: "Yes — there is a major interaction noted: Acetylsalicylic acid (aspirin) interacts with
         active order Ibuprofen [249]."
```

One chip for the pair, not two — the symmetric row on the other entry is collapsed. The answer cites the finding record, which is why `references` is no longer empty.

**CASE 2 — the critical negative.** Mark Smith, two active orders that the KB does not rate against each other, same question, twice:

```
safetyWarnings (0)
references (0)
answer: "The records do not address drug interactions with current medications."
```

Both drugs are real KB entries with non-empty partner lists, so the lists were genuinely searched. The #107 abstention machinery is untouched — this does not become a chip generator.

**CASE 4 — no fabricated pair.** Melissa Wright (Spironolactone 25mg only), the #86 substring shape live: **0 chips**. Without the "a pair needs two orders" guard this is where screening would have told a clinician on one medication that it has a Major interaction with iron.

**CASE 3** — Mary Smith, the issue's own repro patient: still 0 chips, correctly. She has ONE active order, so there is no pair to find; her row in the issue's table is a degenerate case, and the capability needs two interacting orders to show at all.

**CASE 5** — non-screening question on the same patient, unchanged: 0 chips, and it still enumerates the orders ("The patient is ordered to take Advil 400mg [247] and Aspirin 81mg [248]").

**CASE 6** — same patient asked about a named drug: the existing arm owns it and produces two *distinct* correct pairs (warfarin × aspirin, warfarin × ibuprofen), no screening chip alongside.

**All four brief controls** behave exactly as specified: 1 / 1 / 2 / 0, with the same chip text as before.

The cap is not exercised live (no probe patient has more than one above-floor pair) — it is exercised in the unit tests, where 15 candidate pairs are cut to 10 and all 10 kept are the Major ones. No `WITHHOLDING` log line appeared, as expected.

## Duplicate chips — measured, and lower than the issue predicted

**Zero duplicate chips across all twelve cases.** The issue expected landing this before the dedup work would "multiply duplicate chips across every pair". Measured, it did not, and the reasons are specific rather than lucky:

- the **symmetric doubling** pairwise screening creates is collapsed in this arm (CASE 1: 1 chip, not 2);
- **#88 / PR #89's rule-vs-class double cannot reach screening at all**, because the class arm is deliberately not run pairwise — a structural reason, independent of data. (It was additionally inert for these probe patients, whose concepts carry no ATC map; see the corrected coverage numbers in decision 5.);
- **#115's route-variant multiplication can still reach screening** and is not fixed here: two entries sharing one `rxnorm_name` are two ids, hence two pair keys, so a screening subject with N route variants yields N chips for one clinical pair at N severities. I did probe one route-variant drug (Benadryl → 2 diphenhydramine entries) but its pair has no row, so this is predicted from the data, not observed. It is bounded by the severity floor and the cap of 10, and it is the same defect #115 already reproduces on the question-driven arm, so screening does not make it worse in kind.

**So this is ready, not draft.** #115 landing will tighten screening for route-variant drugs and #89 does not overlap it. I have not added any local dedup beyond the symmetric collapse, which neither sibling's mechanism covers.

## Tests

`mvn clean install` from the repo root (full reactor, re-run immediately before pushing): **API 684 + OMOD 53 = 737 run, 0 failures, 0 errors, 34 skipped** — baseline was 662 + 53 = 715, so +22 and nothing lost. (The arm landed at 731; the hardening pass took it to 737 — see the hardening comment below.) The 34 skipped are the usual LLM-requiring `PromptInjectionEvalTest` / `LlmAnswerQualityTest`.

New: `DrugSafetyInteractionScreeningTest` (22 cases after hardening) through the real `validate` / `injectRecords` entry points on the real bundled datasets, plus `ddi-screening-self-witness.json` and `ddi-screening-multiword-partner.json`, DDInter-shaped fixtures parsed by the real parser that reproduce the live `iron` ⊂ `spironolactone` shape and the `insulin` ⊂ `insulin glargine` partner-resolution shape. Covered: the pair fires on a question naming neither drug; **the critical negative — two real KB drugs with no row between them raise nothing**; a sub-floor pair raises nothing; a single active order has no pair; an unrated curated rule is screened and survives a partner with no entry of its own; a non-screening medication question and an "interacts with the care team" question screen nothing; a question that names a drug keeps exactly the existing single chip; the cap keeps the 10 Major pairs out of 15 candidates; the finding is injected as a citable record; and no reference record is.

## Test data created

The standalone had no patient with two interacting active orders, so I created two (developer box; **not** demo data):

- **Barbara Miller** `e30bc8f0-08bb-406c-986a-2b153a495603` — Aspirin 81mg + Advil 400mg (Ibuprofen). `aspirin x ibuprofen` is **Major** in the loaded 19MB KB (group 2136).
- **Mark Smith** `de4b0d62-8a47-4c82-8220-1f0a87eafd46` — Chloroquine Co 250mg + Benadryl 25mg. `chloroquine x diphenhydramine` has **no row** in that KB — the negative control.

Both had zero drug orders before, and neither is a probe patient for any other open issue.

Closes #113


dkayiwa:
One factual correction to the PR body, because it is stated as a measurement about the whole dictionary and it is not one. **Your operative conclusion survives; the stated reason does not.**

The body says:

> **Measured on the standalone: the demo dictionary's drug concepts carry no `ConceptMap` at all** — `GET /concept/3cba8193…?v=custom:(mappings)` on Mary's Simvastatin returns `"mappings": []`

That generalises from one concept. Counted directly against the 3.7.1 demo DB:

```
drug-class concepts (retired=0)          616
  with ANY concept_reference_map         158
  with an ATC-named source                85
```

So **85 drug concepts do carry `WHOATC` maps** — e.g. `Torasemide` → `C03CA04`, `Heparin sodium` → `B01AB01`/`C05BA03`/`S01XA14`, `Sulfadoxine and pyrimethamine` → `P01BD51`. Mary's Simvastatin happens to be one of the 531 that do not, as do Tiotropium, Spironolactone, Nitroglycerin, Budesonide and Dexamethasone — which is why every probe patient in this batch (mine included) landed in the unmapped majority.

Why the distinction matters rather than being pedantry:

1. **"Screening could never have fired" is true for your probe patients, not for the feature.** On a patient whose orders are among the mapped 85, `findByActiveOrders` returns entries and an ATC-only subject set is non-empty. Your name ∪ ATC union is still the right design — it is the disjunction `hasActiveDrug` already uses, and it is what makes the feature work on the ~86% unmapped case — but it is a *robustness* argument, not a *necessity* one.
2. **The claim about #89 is overstated for the same reason.** The body argues the class arm "could not fire on this instance either" because `getActiveDrugAtcCodes()` is empty. That holds only for unmapped concepts; on a patient ordered `Heparin sodium` the class arm can fire. The stronger and still-correct half of your argument stands on its own: the class arm is deliberately not run pairwise, so it cannot reach screening regardless of ATC coverage.
3. It is worth knowing for anyone reading this later that ATC coverage is **sparse but non-zero**, because "no ATC data" invites someone to delete the ATC paths as dead code, and they are not dead — they are dormant for most of this particular dictionary.

Suggest amending those two sentences to something like "the demo dictionary maps only 85 of 616 drug concepts to ATC, and none of the probe patients' orders are among them, so an ATC-only subject set is empty for every case exercised here."

None of this touches your measured results — 731 tests, 12 live cases, zero duplicate chips, and the CASE 2 negative (non-interacting orders ⇒ 0 chips) is the assertion that matters most and is unaffected. Worth noting the sparse-ATC observation probably does deserve its own issue as you suggest, just framed as coverage rather than absence: with 531 of 616 drug concepts unmapped, both class arms and order-driven reference injection are inert for most of this dictionary, which is a real limitation on how much of the DDI feature a default OpenMRS install actually gets.

dkayiwa:
## Hardening pass (`/harden`) — four false-output paths closed, and the docs this PR falsified corrected

Pushed `80c67c0`, `bb3b4cb`, `1e582bb`, `7689707`, `dfa604a`. **`mvn clean install` from the repo root, full reactor: API 678 + OMOD 53 = 731 → API 684 + OMOD 53 = 737, 0 failures, 0 errors, 34 skipped** (+6 tests, nothing lost). Every fix below is chip-*narrowing* or de-duplicating — none of them can add a chip — so the live risk was one-sided: that something which should fire stopped firing. It didn't (evidence at the bottom).

The two defects in the second batch came from an independent adversarial pass over the *merged* arm rather than a re-read of the individual guards, which is where all four hid.

### 1. A fixed-dose combination order self-witnessed a pair — through the ATC leg

`activeOrdersOtherThan` dropped only the *subject's* own ATC codes. But one order resolves to SEVERAL entries whenever its name whole-word-matches an alias of each constituent, and the order's single mapped code belongs to just one of them — so the co-formulated other half's code stayed in the derived context as if it were a second order. Measured on the bundled DDInter sample, one `Aspirin and omeprazole` order (Yosprala, ATC `A02BC05`):

```
[interaction] Acetylsalicylic acid (aspirin)
    Acetylsalicylic acid (aspirin) interacts with active order esomeprazole — Minor. …
```

Two halves of one tablet as an interacting pair, naming a drug the patient is on in no form at all. Test: `aCoFormulationOrderIsStillOneOrderAndNotAPair`.

### 2. An order the dataset knows only by its ATC **code** self-witnessed a pair

The same "a pair needs TWO orders" property, on the half a whole-word name filter structurally cannot see. The guard drops a name when `ref.matchesText(name)` (whole-word alias match); the join it guards, `hasActiveDrug`, matches a token as a raw **substring**. When a subject arrives through the ATC leg, *no* name resolves to it, so its own order name survives and any token sitting inside that name witnesses a pair with itself:

```
one active order: "Spironolactona 25mg", ATC C03DA01
  → Spironolactone interacts with active order iron — Major.
```

That is the #86 shape decision 4 already closes for name-resolved subjects — and this is precisely the case the ATC leg exists to serve: the same javadoc measures the demo dictionary's mapped `Torasemide` against the KB's `Torsemide`. A subject with no attributable order name can no longer pair by name at all; it may still pair by ATC code. Test: `anOrderTheDatasetOnlyKNOWSByItsAtcCodeCannotPairWithItself`.

I also wrote down the one case that **cannot** be reached from inside this arm: one order whose name matches no alias *and* which is mapped to the codes of two interacting entries still pairs them, because with orders flattened into one name set and one code set, "one order with two codes" and "two orders with one each" are the same input — any reduction strong enough to catch the first discards the genuine two-order ATC pair. The fix is per-order structure in `PatientClinicalContext`; not this PR's to change.

### 3. The partner resolved to the first matching entry, not the entry the rule names → one pair chipped twice

A multi-word RxNorm generic also whole-word-contains a shorter entry's alias — `insulin` inside `insulin glargine` — and one order name (`Insulin glargine 20U`) resolves to **both** entries, so a first-match scan handed back `Insulin`, the reverse direction keyed on `Insulin glargine`, and both survived the unordered pair key: two chips for one row, in opposite directions, with the withheld-pair WARN label naming the wrong drug for the same reason. Most specific match wins now — an entry carrying the token as an alias *outright* is preferred, which is the normal case for every DDInter rule (the token IS the partner's own RxNorm generic) and is what makes both directions resolve identically. Test: `aMultiWordPartnerTokenStillKeysThePairOnce`, with a new `ddi-screening-multiword-partner.json` whose drug order deliberately lists `Insulin` first.

This one matters because "each pair once" is the arm's own headline property, and it was already partly lost by construction on the full KB (single-word tokens are all the bundled 16-drug sample has, which is why nothing caught it).

### 4. The screen could repeat a chip the drug-in-play arm had already raised, byte for byte

The gate reads the QUESTION alone — deliberately, so the pre-answer findings pass and the post-answer chips pass gate identically — so an ANSWER-named drug can be in play beside the screen, and `addInteractions` then runs the same rule over the same orders:

```
safetyWarnings (3)
  [interaction] Simvastatin      Simvastatin interacts with active order clarithromycin — Major. …
  [interaction] Clarithromycin   Clarithromycin interacts with active order simvastatin — Major. …
  [interaction] Simvastatin      Simvastatin interacts with active order clarithromycin — Major. …  <-- exact repeat
```

…for an answer naming both drugs and citing nothing (echo scoping exempts an answer-named drug only when a CITED record names it, which is why the live path — where the answer cites the finding record — measured 1 chip and this never showed). Suppressed on the chip's own identity rather than by gating on the answer, so the two passes still gate the same way; the pair stays marked seen, so the reverse direction cannot re-report it. Collapsing two chips that state one fact in *opposite* directions is still #88/#115's and is deliberately not approximated here. Test: `screeningNeverRepeatsAChipTheDrugInPlayArmAlreadyRaised`.

### Tests for claims nothing covered

- `aMedicationQuestionSayingInteractiveScreensNothing` — the cue is word-boundary anchored, so `"Show me an interactive list of her medications."` (which carries the MEDICATIONS intent) must screen nothing. Mutation-checked: dropping the anchoring makes it screen.
- `activeOrdersWithNoAtcCodesAtAllAreStillScreened` — the shape production actually runs. On the demo dictionary every probe patient's orders are unmapped, so the name leg is the only one that fires live, yet every other case passed both a name set and an ATC set; a change to the subject-set union could have killed the live-firing leg with the suite green.

### Docs this PR falsified, now corrected

- `SafetyWarning.TYPE_INTERACTION` claimed the drug-in-play join is *"the only one whose detail says active order"*. The screen reuses that very chip builder, so **two** of three joins say it — the question-pair join (#114) is the one that does not. It also defined the type as two drugs *"rated at or above `minInteractionSeverity`"*, which over-claims in both directions: an unrated rule is exempt from the floor, and the class-based joins carry no severity at all.
- README and `config.xml` said the two new joins *"have no chart drug on both sides"* / *"do not involve the chart on both sides"*. The screen has a chart drug on **both** sides — that is the point of it. `config.xml`'s replacement then over-corrected to "Every join applies `minInteractionSeverity`", which contradicted the README row two lines above it.
- *"reports only pairs the reference data actually **rates** against each other"* (class javadoc + README) is contradicted by this arm's own `anUnratedCuratedRuleIsScreened…` test. "relates" is the true word.
- *"it stands down as soon as the question names a drug"* — the gate is dataset **recognition**, not lexical. `"Does ceftriaxone interact with any of her current medications?"` (ceftriaxone absent from the KB) still screens, and the resulting finding is injected as the only `safety_finding` record; `"Does ibuprofen interact…?"` (in the KB, interacting with neither order) suppresses the screen, so the *more specific* question yields strictly less. Both behaviours are consequences of decision 1 (no second drug vocabulary) and I have not changed them — but they are now written down instead of contradicted, and I'd flag the ceftriaxone direction as worth your judgement.
- The cap's *"what is dropped is always the least severe"* is **not** unconditional: `severityPriority` sorts unrated above Major (shared with `InteractionNote`, because unrated is not rankable below a rated tier), so a dataset mixing unrated and rated rules with >10 above-floor pairs can withhold a Major pair while reporting an unrated one — reproduced: 2 unrated + 8 Major reported, three Major QT-prolongation pairs named in the WARN. **I did not reverse the ordering** (that would discard an operator's own hand-authored rule in favour of a third-party rating, and the ordering is one of your load-bearing decisions); I documented the limit, the two conditions needed to reach it, and the WARN line that makes it recoverable. Reversing it is your call.
- README's safety-validation prose and `DrugSafetyValidator.validate`'s javadoc still said the drugs checked are only those the question or answer names; `DrugReferenceInjector.preAnswerFindings` still described itself as findings *"for the drugs this question names"* — the one place a maintainer would look to understand why a screening question produces findings at all.

### Correction to a measurement in this PR's own body and javadoc

Decision 5 generalised a single lookup on Mary's Simvastatin into **"the demo dictionary's drug concepts carry no `ConceptMap` at all"**. That is wrong. Counted directly: **616 Drug-class concepts, 158 with any `concept_reference_map`, 85 with an ATC-named source** (`Torasemide`→`C03CA04`, `Heparin sodium`→`B01AB01`, `Sulfadoxine and pyrimethamine`→`P01BD51`). Simvastatin is one of the 531 unmapped, as are Tiotropium, Spironolactone, Nitroglycerin, Budesonide and Dexamethasone — which is why every probe patient landed in the unmapped majority. `docs/drug-kb-demo.md` already records the opposite for one order: *"where the Aspirin order is `N02BA`-mapped (the live :8081 instance is)"*.

So coverage is **sparse but non-zero**, and the name ∪ ATC union is a **robustness** argument, not a necessity one. The body and the `activeOrderEntries` javadoc now say so. This matters because read as "no ATC data" it invites deleting the ATC paths as dead code — **they are dormant on this instance, not dead**, and defect 2 above is a live bug *on* that path. The stronger half of the #89 out-of-scope claim is unaffected: the class arm is not run pairwise, so it cannot reach screening at all, whatever the dictionary carries.

### Load-bearing decisions: all kept

Nothing in decisions 1–6 was reversed or trimmed. Specifically I did **not** gate the screen on the answer (that would have fixed defect 4 more cheaply while breaking the pre/post-answer gate symmetry the call-site comment exists to protect), did **not** unscope `matchingEntries`/`relatedToAny`, did **not** run the class arm pairwise, did **not** copy `hasActiveDrug`'s matching rule into the arm (defect 2's fix is a property of *pairing* — it removes an unattributable witness, it does not re-implement the predicate, so it still follows whatever #86/#87 settles on), and did **not** touch the cap value, the severity ordering, or the WARN-naming of withheld pairs. The absorbed stale-doc rewrite is kept and corrected rather than reverted.

### Live re-verification

Required — production code changed in ways that alter chip output. Two full-JVM restarts onto this build on the 3.7.1 standalone, all ten probes inside one lock each, querystore retrieval gated alive and the known-positive control chip fired before any capture. Chips only; prose on that box is not reproducible.

Both runs, same ten cases, identical results — the second run is the one that matters (it carries defects 2 and 3's fixes; the first carried only 1 and 4):

```
case                                              chips  refs  dupChips  expectation
---------------------------------------------------------------------------------------
case1   Barbara, screening, 2 interacting orders      1     1  no        >=1 chip + >=1 ref
case1r  same, re-run                                 1     1  no        >=1 chip
case2   Mark, screening, 2 NON-interacting orders     0     0  no        0 chips
case2r  same, re-run                                  0     0  no        0 chips
case3   Barbara, non-screening medication question    0     2  no        0 chips
CTL 1   Mary + clarithromycin                         1     1  no        1 chip Major x simvastatin
CTL 2   Agnes + warfarin                              1     1  no        1 chip Major x aspirin
CTL 3   Joshua + ibuprofen                            2     4  no        2 chips
CTL 4   Mary + paracetamol                            0     0  no        0 chips
CTL 4r  same, re-run                                  0     0  no        0 chips
```

CASE 1 still closes the gap, byte-identical chip text to the pre-hardening capture:

```
[interaction] Acetylsalicylic acid (aspirin) interacts with active order ibuprofen — Major. …
references (1)  #N type=safety_finding
```

CASE 2 is still silent on both runs (the negative was re-run because a cold-start race can load
the drug KB empty for a whole JVM lifetime with no WARN). All four brief controls are unchanged:
1 / 1 / 2 / 0, same chip text. `dupChips` is a per-case check that no two chips in one
`safetyWarnings` array share `(type, drug, detail)` — the defect-4 shape — and it is `no`
everywhere.

The cap path has **no live probe** — no patient on the box has more than one above-floor pair — so it is unit-tested only (15 candidates cut to 10, all 10 Major), and I am not claiming live coverage for it. Likewise defects 1–4 are all unit-reproduced rather than live-reproduced: 1 and 3 need dataset shapes no demo patient has, and 4 needs an uncited answer, which is not reproducible on demand. Nothing new was created on the box; Barbara Miller and Mark Smith from the original run were reused.

### Found, not fixed

- **The one pair-guard case the flattened context cannot reach — the most serious thing I am leaving behind, and my judgement is that it does not block this merge.** A subject with no attributable order name now loses every *name* witness (defect 2's fix), but its ATC codes can only have its OWN removed. **Failure mode:** if we ship without this, a patient on a single combination product whose display name matches no alias in the loaded dataset, but whose concept is mapped to the ATC codes of two entries the dataset relates to each other, gets one chip asserting that the two halves of that one product interact — a false positive naming a drug they are not separately taking, raised silently with no error, on a clinician-facing safety chip.

  Why it cannot be fixed here: `PatientClinicalContext` flattens active orders into one name set and one code set, so *"one order carrying two codes"* and *"two orders carrying one each"* are **literally the same input**. Every reduction strong enough to catch the first also discards the second — the genuine two-order ATC pair the ATC leg exists to find (a mapped `Torasemida` against a mapped `Digoxina`, neither name in the dataset). I tried three formulations and each traded the false positive for a false negative of equal reach. The real fix is per-order structure in `PatientClinicalContext` (a list of orders, each with its own names and codes), which is shared infrastructure two sibling branches are already editing — worth its own issue.

  Why I judge it non-blocking: it needs three coincidences at once (an unrecognised order name **and** ≥2 ATC codes on that one concept **and** those codes belonging to two entries with an above-floor row between them). Measured on the demo dictionary, only 85 of 616 Drug-class concepts carry an ATC-named map at all, multi-code combination concepts are a subset of those, and no probe patient comes close; the output when it does hit is one non-blocking additive chip, bounded by the severity floor and the cap, on an answer that names both constituents anyway. It is strictly narrower than the two false-positive paths this pass *closed*. If you would rather not carry it at all, the honest options are to file the `PatientClinicalContext` change first or to drop the ATC leg from the subject set — the latter costs the `Torasemide`-class pairs and I would not choose it.
- **`docs/adr.md` Decision 23** still reads *"a chart-sufficient answer naming no reference drug produces nothing"*, which this arm falsifies — and `README.md` points a reader there "for the full design". Left alone deliberately: the ADR is a shared file outside the doc set this PR absorbed by agreement, so a one-sentence edit there risks colliding with a sibling branch. One sentence for whoever lands last.
- **`safety_finding` is absent from `CitationGroundingVerifier`'s `demoteOnlyIndexes`**, whose own comment says a second kind of injected record must decide whether it belongs there. The gap arrived with #110, bounded then to 1–2 findings; a capped screen can inject 10 against `GROUNDING_ENTAILMENT_MAX_CHECKS = 16`, so up to 10 Tier-2 slots go to verdicts the client suppresses, evicting chart citations on exactly the polypharmacy questions that produce the most findings. Belongs to #110's design, not here.
- **The opposite-direction cross-arm double** (`A interacts with active order B` beside `B interacts with active order A`, when the answer names one side uncited) — needs #88/#89's rule-versus-class correlation; approximating it on a text key would be guesswork.
- **`severityPriority` exists twice** — here and in `DrugReferenceInjector.InteractionNote`'s constructor — cross-referenced in javadoc but not shared. Unifying it means editing `DrugReferenceInjector`, which #123 is rewriting.
- **`validate` runs twice per query** (pre-answer findings + post-answer chips), so the quadratic screen and its withheld-pairs WARN both run twice. Pre-existing to the whole validator, not introduced here.
- **`docs/drug-kb-demo.md`** has no screening scenario in its cheat-sheet.

dkayiwa:
### Note for whoever rebases this — #124 partly supersedes one of my guards

`main` moved to `623be82` (#124) after I banked the state above. Measured just now, not predicted:

- **The only conflict is `README.md`** (one content hunk). `DrugSafetyValidator.java` and
  `DrugReferenceInjector.java` both auto-merge against #124 — `git merge-tree origin/main HEAD`
  reports the README hunk and nothing else. (Against `84a434c`, the `main` before #124, this branch
  was `MERGEABLE` / `CLEAN` with a conflict-free `merge-tree`. The conflicts with #120 and #121 in
  `DrugSafetyValidator.java` are still ahead of this branch and unaffected by the above.)

- **#124 added `PatientClinicalContext.ActiveDrugOrder` (uuid, display, names)** — per-order
  structure, which is precisely what the deferred item above says is needed. It carries **names
  only, no ATC codes**, so the split is now:

  - The *name* half of my guard should be **replaced** on rebase. `activeOrdersOtherThan` currently
    clears the whole name set for a subject no name resolves to, because a flattened set cannot say
    which name came from which order. With `getActiveDrugOrders()` that becomes attributable: drop
    the orders whose own `names` the subject matches and keep the rest, which recovers the name
    witnesses my blunt clear discards (the documented cost of defect 2's fix) while still closing
    the self-witness. That is a strictly better guard, and it needs the rebase to exist —
    `anOrderTheDatasetOnlyKNOWSByItsAtcCodeCannotPairWithItself` and
    `theSecondOrderBeingRealMakesItAPairAgain` are the two tests that pin both directions of it.
  - The *ATC* half is **unchanged**: codes are still flattened in `activeDrugAtcCodes`, so
    "one order carrying two codes" and "two orders carrying one each" remain the same input and the
    residual false-positive path in the deferred item above survives. Closing it needs codes on
    `ActiveDrugOrder` too.

dkayiwa:
## Rebased onto `b03aa62` — and the batch changed three of my answers

Force-pushed `f2355d0` (was `dfa604a`). **`MERGEABLE` / `merge-tree` clean against `b03aa62`.** Full root reactor: **API 776 + OMOD 58 = 834 run, 0 failures, 0 errors, 34 skipped**. Main's baseline is 751 + 58 = 809 (derived, not measured on main: the 25 tests in `DrugSafetyInteractionScreeningTest` are the only test methods this branch adds, and it touches no other test file) — so +25.

The five hardening commits are squashed into one, because each of them patched a file shape the batch has since replaced; the narrative is in the two comments above and in the commit body.

### Conflicts, measured

`README.md` (one hunk: #124 inserted its active-order-reconciliation bullet immediately before the safety-validation bullet we had both edited — kept both) and `DrugSafetyValidator.java` (three hunks: the import block, the call site, and `bestRulePerPartner`'s body where my old `addInteractions` used to be). Nothing else.

### This arm is now built on the merged arms, not beside them

The rebase turned out to be mostly *deletion*, because #120 and #121 landed the helpers my arm had been carrying its own copies of. Every one of these was a duplicate as of `b03aa62`:

| mine | replaced by | why it matters |
|---|---|---|
| `severityPriority` | #120/#121's `severityPriority` | **Same signature — the file would not have compiled.** |
| `hasAliasExactly` | #120's `namesEntry` | Same case-folded exact-alias test. |
| `pairKey` / `pairSide` | #120's `unorderedPairKey` | Mine joined on `"\|"`; #120's returns a two-element list precisely because *any* separator occurs inside some drug name. Mine was a latent collision. |
| my own floor + `hasActiveDrug` loop | #121's `bestRulePerPartner` | See below — this one changed behaviour. |

Two of those are behaviour changes, not tidying:

- **`bestRulePerPartner(ref, others, severityFloor)`** — I now ask #121's grouping the same question about the *other* orders that the drug-in-play arm asks about all of them. My own loop kept the **first** matching rule per pair; #121's keeps the **most severe** (then the fuller note). So before this rebase, a pair whose partner has route variants could have been chipped by my arm as `Moderate` while the other arm reported the same pair as `Major`, and my cap would have sorted it on the wrong severity. That divergence is gone by construction now.
- **`identifies()` for the partner side** — replaces my exact-alias-then-whole-word fallback. The fallback was the very thing that mis-resolved `insulin` inside `insulin glargine`; #120 having already moved this off `matchesText` to name identity, the fallback had no remaining job and is deleted.

`addInteractions` now emits through `interactionWarning` too, so the chip string has one definition across all three arms rather than two spellings of the same coalesce (`partnerLabel` vs a second `firstNonBlank`) — which is what `partnerLabel`'s own javadoc warns against, since #121 *groups* on that label.

### Your question 2: can one arm report a pair the other suppresses?

**No, and not because of a precedence rule — because the two arms cannot both run.** The gates are mutually exclusive: #120's needs `questionDrugs.size() >= 2`, mine needs `questionDrugs.isEmpty()`. So on any single question at most one of them executes at all.

That answers the caps too: **the two 10s never apply to overlapping sets**, because only one cap is reachable per question. There is no combined 20, and no set a pair could be counted into twice.

**Pinned, not reasoned:** `aQuestionNamingTwoDrugsGoesToThePairArmAndNotToTheScreen` puts a single question through both gates at once — it names warfarin and aspirin (a real Major row, neither on the chart) *and* carries the screening cues. Mutation-checked by removing the `questionDrugs.isEmpty()` gate, which makes it fail with `Simvastatin interacts with active order clarithromycin` beside the pair chip: one clinical pair, two arms, exactly the double-report you asked about.

Writing that test also corrected a wrong assertion of mine. My first version asserted "no chip mentions active order clarithromycin", which failed — on `Warfarin interacts with active order clarithromycin`, the **drug-in-play** arm doing its own job on a question-named drug. The precise statement is that every chip must be *subjected* on a drug the question named, since a screening chip's subject is one of the patient's own orders.

It also explains why this arm needs no analogue of #120's `coveredByActiveOrderArm`. That check exists because *its* pairs can also be the chart arm's; mine are active orders on both sides, so the only arm that could restate one of my pairs is `addInteractions` acting on an **answer**-named drug — which is exactly what the chip-identity suppression already handles, and which the gate cannot handle without breaking pre/post-answer symmetry.

What the two arms do now share is machinery: the same `bestRulePerPartner`, `partnerLabel`, `pairKeyNames`, `unorderedPairKey`, `severityPriority`, `interactionWarning`. They cannot drift on what a pair is or which of its rows is worth chipping.

### Your question 1: the supersession — and a correction to my own claim about it

Folded in. `activeOrdersOtherThan` now attributes per **order** via #124's `ActiveDrugOrder`: an order is the subject's own when the subject resolves from any of *its* names, and then everything that order contributes goes — its other names (a brand the aliases do not cover is still that order) *and* the codes of any other entry it resolves to. Every other order stays whole, so a subject can pair by name even when nothing names the subject itself, which is what the blunt `names.clear()` had to give up. New test: `perOrderStructureRecoversTheNameWitnessTheFlatFallbackHasToDiscard`.

The flattened fallback is **kept** for contexts built without per-order structure (the builder's own null-patient early return, and every hand-built context). Failure mode if it were dropped: with no orders on the context the per-order loop contributes no names at all, so the arm would silently stop finding name-witnessed pairs — no exception, no log, just fewer chips.

**And a correction.** I told you both directions were pinned by existing tests. I checked instead of assuming, and **they were not.** #128 replaced `hasActiveDrug`'s raw `contains` with word-start matching, so `iron` no longer matches inside `spironolactona` — the shape my guard was written against is now closed *at the matcher*, and my two guard tests had started passing for #128's reason rather than mine. Deleting the whole guard broke nothing.

What the guard still prevents is the **word-start** shape, which #128 does not close: a combination order's name contains its second constituent *at* a word start, so `"Aspirin and omeprazole 325/40mg"` still answers `hasActiveDrug("aspirin")` for the omeprazole subject. New test `aCoFormulationOrderWithNoAtcCodesIsStillOneOrder` isolates it (no ATC codes, so the name leg is the only possible witness), and mutation-checked: with the guard removed, two tests fail with

```
[interaction] Omeprazole interacts with active order aspirin — Minor. …
```

for a patient on **one tablet**. So the guard is load-bearing again, and now for a reason a test actually holds it to.

### Your question 3: `severityPriority`

It was a third copy — and worse than a duplicate: identical signature in the same class, so the merged file would not have compiled. Deleted; main's is the only one. `partnerLabel`, `namesEntry` and `unorderedPairKey` were the same story and went the same way.

### The residual limitation, restated against the new code

Unchanged and still not fixable here: one order whose display name matches no alias in the dataset **and** which is mapped to the ATC codes of two entries the data relates still pairs those two. #124 gave `ActiveDrugOrder` uuid, display and **names** — not codes — so a code still cannot be attributed to the order that contributed it, and "one order with two codes" remains indistinguishable from "two orders with one each". **Failure mode:** such a patient gets one chip asserting the two halves of a single product interact, silently, on a clinician-facing safety chip. My judgement is unchanged: not a merge blocker — it needs three coincidences at once, only 85 of 616 Drug-class concepts carry any ATC map, and it is strictly narrower than the paths this work closed. Carrying codes on `ActiveDrugOrder` is the fix and is a small follow-up on top of #124.

### Live re-verification on the rebased build

Full-JVM restart onto the rebased build, all ten probes inside one lock, querystore gated alive
and the known-positive control chip fired before capture (`warm chips=1`). Chips only — prose on
that box is not reproducible.

```
case                                              chips  refs  dupChips  expectation
---------------------------------------------------------------------------------------
case1   Barbara, screening, 2 interacting orders      1     2  no        >=1 chip + >=1 ref
case1r  same, re-run                                 1     2  no        >=1 chip
case2   Mark, screening, 2 NON-interacting orders     0     0  no        0 chips
case2r  same, re-run                                  0     0  no        0 chips
case3   Barbara, non-screening medication question    0     2  no        0 chips
CTL 1   Mary + clarithromycin                         1     2  no        1 chip Major x simvastatin
CTL 2   Agnes + warfarin                              1     2  no        1 chip Major x aspirin
CTL 3   Joshua + ibuprofen                            2     2  no        2 chips
CTL 4   Mary + paracetamol                            0     0  no        0 chips
CTL 4r  same, re-run                                  0     0  no        0 chips
```

**Every chip count is identical to the pre-rebase capture, including the same chip text.** The
`refs` counts moved (case1 1→2, CTL 3 4→2) and that is the batch, not this arm: #124 injects
`active_drug_order` records and #123 took the render's bookkeeping out of the citable record.
Chips are what this arm governs, and they did not move.

The cap still has **no live probe** — no patient on the box has more than one above-floor pair —
so it stays unit-tested only (15 candidates cut to 10, all ten Major) and I am not claiming live
coverage for it. Nothing new was created on the box; Barbara Miller `e30bc8f0-…` and Mark Smith
`de4b0d62-…` from the original capture were reused.

PR #128, opened 2026-08-04
fix(drug-safety): match interaction tokens by word start in order names (supersedes #87)

### LINKED ISSUE #86: DrugSafetyValidator: hasActiveDrug substring match over-fires on substring-nested drug names

dkayiwa:
Closes #86. **Supersedes #87** — this branch contains #87's commit unchanged (cherry-picked onto
current `main`, authorship preserved) plus one commit correcting the matching rule and adding the
coverage that correction needs. If this lands, #87 should be closed in favour of it.

## The defect

`PatientClinicalContext.hasActiveDrug` matched an interaction rule's drug token against the
patient's active-order display names with `drug.contains(token)`. Drug names nest, so the check
reported drugs the patient has never taken — at Major severity, mechanism text and all. Both of
these were reproduced live on the 3.7.1 standalone against `13690b1`, twice each, with neither
concept carrying an ATC mapping (so the ATC arm cannot be involved):

**Susan Young, active order `Tiotropium` (inhaled anticholinergic), "Is it safe to give linezolid?"**

```
--- safetyWarnings (1) ---
  [interaction] drug='Linezolid'
      Linezolid interacts with active order opium — Major. Coadministration of narcotic analgesics
      with monoamine oxidase inhibitors (MAOIs) has been associated with rare reports of anxiety,
      confusion, hypotension, respiratory depression, cyanosis, and coma. …
```

```
"tiotropium".contains("opium")  →  true
```

**Melissa Wright, active order `Spironolactone`, "Can I start dolutegravir?"**

```
--- safetyWarnings (2) ---
  [interaction] drug='Dolutegravir'
      Dolutegravir interacts with active order iron — Major. Coadministration with medications
      containing polyvalent cations such as aluminum, calcium, iron, or magnesium may decrease the
      oral bioavailability of dolutegravir.
  [interaction] drug='Dolutegravir'
      Dolutegravir interacts with active order iron — Major. INTERVAL: Coadministration with …
```

```
"spironolactone".contains("iron")  →  true      # sp-IRON-olactone
```

## Root cause, and why not the whole-word fix in #87

The cause is one matcher, so #87 is right that the fix belongs in shared infrastructure and right
about the half that discriminates: the **left** boundary. An alphanumeric character immediately
before the token means the token sits inside a longer word — a different molecule
(`tiotr|opium`, `sp|iron|olactone`, `nitro|glycerin`, `bud|esonide`, `hydro|chlorothiazide`,
`cipr|ofloxacin`).

Where #87 goes wrong is applying prose's **symmetric** boundary to an order name. Measured over
this deployment's real dictionary — 2531 drug and drug-concept names × the full DDInter KB's 2093
rule tokens, run through the shipped matchers themselves:

| rule | matches | known collisions leaking | known real names lost |
|---|---|---|---|
| `contains` (the defect) | 896 | 9 of 9 | 0 of 12 |
| symmetric word boundary (#87) | 761 | 0 of 9 | **9 of 12** |
| left boundary + ≤1 trailing letter | 828 | 0 of 9 | 0 of 12 |
| **left boundary + ≤2 trailing letters (this PR)** | **829** | **0 of 9** | **0 of 12** |
| left boundary + ≤3 | 829 | 0 of 9 | 0 of 12 |
| left boundary + ≤4 | 834 | 0 of 9 | 0 of 12 |
| left boundary only (unbounded tail) | 835 | 0 of 9 | 0 of 12 |

Symmetric boundaries drop 135 of containment's 896 matches, and **68 of the dropped are
legitimate**, because this dictionary is multilingual:

```
'Aspirine Co 81mg'         token='aspirin'          <- French, a real drug row here
'Aspirina'                 token='aspirin'          <- Spanish
'Amoxicilline'             token='amoxicillin'
'Clarithromycine Co 500mg' token='clarithromycin'
'Simvastatine'             token='simvastatin'
'Ondansetrona'             token='ondansetron'
'Multivitamines et fer'    token='multivitamin'
```

Under #87 a patient on `Aspirine Co 81mg` stops being checked for aspirin interactions entirely.
That trades a false positive for a false **negative** in a safety net, and it would pass review
because the removed chips look exactly like the noise we set out to remove. The two new
localized-name tests in this PR fail against #87's rule — `[]`, no chip at all — and pass here.

**Why ≤2 and not ≤1.** The tail is an empirical accommodation for inflected display names, so it
was scored across the whole corpus rather than a sample. What enters at each step: ≤1 admits 67
localized spellings (all legitimate); ≤2 admits exactly one more, the plural `Multivitamines et
fer`, whose reference entry carries 2 Major and 8 Moderate rules that would otherwise stop being
checked; ≤3 admits nothing; **≤4 is where the first false positives appear** (`Heparinoids` ~
heparin, `Multi-Vitamin Adult` ~ vitamin a). Two is therefore the far edge of the plateau on which
every legitimate name matches and no false positive has appeared — one inflectional ending, which
is what a Romance localization of an INN stem is, singular (`-e`, `-a`, `-o`) or plural (`-es`).

The alternative to a bound — normalizing inflections away on both sides — was measured on the same
corpus at 826 matches, a strict subset of this rule, and it replaces one bound with a per-language
suffix whitelist that a differently-localized deployment falls off silently. The two matches it
additionally rejects are `Aspirinn` and `Metforming`, real typo'd rows in this dictionary that a
safety net should still be checking.

Residual imprecision this rule keeps, for the record: 2 of the 829 are a nitroglycerin order
matching the token `glycerin` through its own parenthetical synonym ("glycerine trinitrate") — a
mislabel of a drug the patient is on, not a fabricated one, and shared by every rule that tolerates
an inflectional tail.

## What changed

- `DrugReference` — #87's `containsWord` (prose, symmetric) and a new `matchesOrderName`
  (active-order names, left boundary + bounded tail) now both delegate to one private
  boundary-aware scan, so the boundary rule cannot drift between them. Prose matching is
  behaviourally unchanged: 761 matches over the corpus before and after.
- `PatientClinicalContext.hasActiveDrug` — the name arm calls `matchesOrderName`. Both of its
  callers (the chip decision in `DrugSafetyValidator:391` and the prompt-promotion predicate in
  `DrugReferenceInjector:350`) reach the rule only through this method, so chips and promoted
  prose still agree by construction.
- `DdiDrugReferenceSource` — one stale comment that described the matching as substring-based.
- `HasActiveDrugWholeWordTest` — #87's file, kept as-is, plus one case it was missing: a
  `chlorothiazide` order must still raise chlorothiazide's own rule, so the fix is de-duplication
  and not suppression of the shorter name. A matcher that simply never matched the shorter token
  passed #87's two tests.
- `DrugSafetyOrderNameMatchingTest` + `ddi-order-name-collisions.json` (new) — the fabrications
  above and the localized names, over a verbatim slice of the full KB (the bundled 16-drug sample
  contains none of linezolid, opium, dolutegravir, iron or multivitamin). Every negative is paired
  with the positive that proves the rule it must not fire on is live, so none of the
  "no chip" assertions can pass by resolving nothing.

## Live verification

Standalone 3.7.1, `sourceFormat=ddinter` with the full 19MB KB, `minInteractionSeverity=minor`,
full-JVM restart onto this build, querystore retrieval confirmed alive (`hits>0`) and the
known-positive control confirmed before any absence was recorded. Chips are deterministic Java;
answer prose is not reproducible on that box (`cache_prompt` KV reuse), so every claim below is
about `safetyWarnings`.

Setup, from the capture: my omod (36,158,822 bytes) deployed at 15:04, JVM pid 94318 started
15:04:16 (fresh process holding 8081, not a `moduleaction` half-restart), `sourceFormat=ddinter`,
`drugSafety.minInteractionSeverity=minor`, `drugSafety.validateAnswers=true`, `llm.engine=local`.
Gates before any absence was recorded: `last querystoreBuild hits=75` (so charts are not empty) and
the known-positive warm control raised its chip (`warm chips=1`).

**Before** (live on `13690b1`, the pre-fix build, each reproduced twice):

```
Susan Young (Tiotropium) + "Is it safe to give linezolid?"
--- safetyWarnings (1) ---
  [interaction] drug='Linezolid'
      Linezolid interacts with active order opium — Major. Coadministration of narcotic analgesics
      with monoamine oxidase inhibitors (MAOIs) has been associated with rare reports of anxiety,
      confusion, hypotension, respiratory depression, cyanosis, and coma. …

Melissa Wright (Spironolactone) + "Can I start dolutegravir?"
--- safetyWarnings (2) ---
  [interaction] drug='Dolutegravir'
      Dolutegravir interacts with active order iron — Major. Coadministration with medications
      containing polyvalent cations such as aluminum, calcium, iron, or magnesium may decrease the
      oral bioavailability of dolutegravir.
  [interaction] drug='Dolutegravir'
      Dolutegravir interacts with active order iron — Major. INTERVAL: Coadministration with … 
```

**After** (this build, verbatim chip summary; absences confirmed on a second run each):

```
Susan/Tiotropium + linezolid                   chips=0 expected=0
Susan/Tiotropium + linezolid (rerun)           chips=0 expected=0
Melissa/Spironolactone + dolutegravir          chips=0 expected=0
Melissa/Spironolactone + dolutegravir (rerun)  chips=0 expected=0
Dorothy/Aspirine Co 81mg + warfarin            chips=1 expected=>=1 (warfarin x aspirin Major)
      [interaction] Warfarin interacts with active order aspirin — Major. Aspirin, even in small
      doses, may increase the risk of bleeding in patients on oral anticoagulants by inhibiting
      platelet aggregation, prolonging bleeding time, and inducing gastrointestinal lesions. …
Dorothy/Aspirine Co 81mg + warfarin (rerun)    chips=1 expected=>=1 (warfarin x aspirin Major)
      [interaction] Warfarin interacts with active order aspirin — Major. …
CONTROL Mary + clarithromycin                  chips=1 expected=1
      [interaction] Clarithromycin interacts with active order simvastatin — Major. …
CONTROL Agnes + warfarin                       chips=1 expected=1
      [interaction] Warfarin interacts with active order aspirin — Major. …
CONTROL Joshua + ibuprofen                     chips=2 expected=2
      [contraindication] Ibuprofen is in the same cross-reactivity group (NSAID) as the patient's
      allergy to Acetylsalicylic acid (aspirin) — possible cross-reactivity
      [interaction] Ibuprofen interacts with active order lisinopril — Moderate. …
CONTROL Mary + paracetamol                     chips=0 expected=0
```

The fifth and sixth rows are the ones that distinguish this from #87. That patient's only
aspirin-bearing string is the drug row name **`Aspirine Co 81mg`** — the order's concept name is
`Acetylsalicylate sodium`, which contains no `aspirin` token at all — so under a symmetric word
boundary the Major warfarin×aspirin chip disappears entirely. The created order was confirmed active
before the query:

```
POST /order -> created order uuid=eab3e8a9-0301-4e11-a007-74f556fd8cbf
              display='(NEW) Aspirine Co 81mg: 81.0 Milligram Oral Once daily'
GET  /order  -> count=1  {'drug': 'Aspirine Co 81mg', 'dateActivated': '2026-08-04T15:07:10.000+0300',
                          'dateStopped': None, 'autoExpireDate': None, 'voided': False}
```

All four known-good regression controls are unchanged from the documented baseline.

## Tests

`mvn clean package` (API + OMOD): **API 670 + OMOD 53 = 723 run, 0 failures, 0 errors, 34
skipped** (the 34 skipped are the LLM-requiring `PromptInjectionEvalTest`/`LlmAnswerQualityTest`, as
on `main`). Baseline `main` is 662 + 53 = 715; this adds #87's 2 tests, 5 new ones here, and 1 to
#87's file.

TDD trail, for reviewers who want to see each assertion fail first:

- against `main`'s `contains`: the two fabrication tests fail with exactly the live chips quoted
  above, and the thiazide test with 2 chips instead of 1;
- against #87's symmetric rule (`maxTrailingLetters = 0`): `HasActiveDrugWholeWordTest` passes,
  and both localized-name tests fail with `[]`;
- against ≤1: only the localized-plural test fails.

## Data created on the shared standalone

The multilingual check needs an order whose display name is localized, and no test patient had one,
so I created one inpatient drug order — `Aspirine Co 81mg`, an existing drug row in the demo
dictionary — for **Dorothy Odinga** (`71c5a6eb-b47c-4ec1-93fd-9d4a568ac724`), a spare demo patient
with no prior drug orders and no allergies. It is test data I added, not demo data.

## Not verified live

The ≤2 step (`Multivitamines et fer`) is covered by unit test only. That spelling exists in this
dictionary as a French *concept* name, not a drug row, and the REST path resolves concept names in
the English locale, so an order cannot surface it on this box without inventing dictionary rows. A
French-locale deployment produces exactly the context the unit test feeds.

## Two adjacent gaps found while measuring, deliberately not fixed here

- **Accented localizations are still unmatched, before and after this change.** The same corpus scan
  with diacritics folded out finds **78 additional matches** the matcher misses today —
  `budésonide` ~ budesonide, `dexaméthasone` ~ dexamethasone, `héparine` ~ heparin, `furosémide` ~
  furosemide, `glycérine` ~ glycerin and 73 more (224 of the 2531 names carry diacritics). This is
  pre-existing (`contains` misses them too, so nothing regresses), but it is the same class of
  false negative as the one this PR fixes and deserves its own change: folding also creates new
  collision opportunities, so it needs the kill direction measured too.
- **`hasAllergyToken`/`hasConditionToken` still use bare containment.** They match free-text allergy
  and condition text rather than order names, so they want a rule of their own, and #86 is about
  order names — flagging rather than changing it here.

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
@@ -175,7 +175,8 @@ public class DdiDrugReferenceSource implements DrugReferenceSource {
 			DrugReference.Interaction i = new DrugReference.Interaction();
 			// Match on the RxNorm generic name (e.g. "aspirin", "acetaminophen") rather than the
 			// DDInter display name ("Acetylsalicylic acid"), since the validator matches this token
-			// as a substring of the order's display name; fall back to the display name.
+			// against the order's display name (by DrugReference.matchesOrderName, which needs the
+			// token to start a word of that name); fall back to the display name.
 			String token = p.rxnormName != null && !p.rxnormName.isEmpty() ? p.rxnormName : p.name;
 			i.setToken(token.toLowerCase(Locale.ROOT));
 			i.setAtc(p.atc.isEmpty() ? null : p.atc.get(0));
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
@@ -278,7 +278,7 @@ public class DrugReference {
 	 *         each side by a non-alphanumeric character or the string edge. Whole-word, not
 	 *         substring, so a drug name nested inside a longer one does not spuriously match
 	 *         ("chlorothiazide" is not a whole word in "hydrochlorothiazide"), while a real token
-	 *         still matches ("aspirin" in "Aspirin 81 mg"). Case-insensitive; a null/blank word
+	 *         still matches ("aspirin" in "Aspirin 81 mg"). Case-insensitive; a null or empty word
 	 *         never matches. Backs {@link #matchesText} (alias-in-prose); the active-order
 	 *         counterpart is {@link #matchesOrderName}, which shares this rule's left boundary but
 	 *         not its right one — see there for why one matcher cannot serve both.
@@ -353,9 +353,19 @@ public class DrugReference {
 	 * The one boundary-aware containment scan, shared by prose matching ({@link #containsWord}) and
 	 * order-name matching ({@link #matchesOrderName}) so the boundary rule cannot drift between
 	 * them. A match needs {@code token} to start at a word boundary in {@code text} and to end at
-	 * one, give or take up to {@code maxTrailingLetters} letters — never digits, which in a drug
-	 * name are a strength rather than part of the name. Case-insensitive; a null/blank token never
-	 * matches.
+	 * one, give or take up to {@code maxTrailingLetters} letters. Letters only: a digit is never an
+	 * inflection, so a digit sitting against the token is neither stepped over nor treated as the
+	 * end of the name, and a display name that glues its strength straight onto the drug name
+	 * ({@code Aspirin81mg}) therefore does not match. That shape does not occur in the measured
+	 * dictionary — of the 67 matches this rule drops relative to bare containment, 61 are a token
+	 * inside a longer word and 6 are tails longer than two letters, none is a glued digit — and
+	 * treating a digit as the end of the name instead scores identically over that corpus (829
+	 * either way), so the two are indistinguishable on real data and this is the conservative one.
+	 * Case-insensitive; a null or empty token never matches. Whitespace-only is the caller's
+	 * business, deliberately not this method's: {@link PatientClinicalContext#hasActiveDrug} trims
+	 * its token, and the {@code ddinter} and {@code atc} sources drop blank aliases at parse. A
+	 * hand-authored {@code json} KB is NOT sanitized, so a blank alias there still matches any text
+	 * carrying two adjacent spaces — pre-existing, and an authoring guard belongs in that parser.
 	 */
 	private static boolean containsBoundedToken(String text, String token, int maxTrailingLetters) {
 		if (text == null || token == null || token.isEmpty()) {
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
@@ -279,23 +279,104 @@ public class DrugReference {
 	 *         substring, so a drug name nested inside a longer one does not spuriously match
 	 *         ("chlorothiazide" is not a whole word in "hydrochlorothiazide"), while a real token
 	 *         still matches ("aspirin" in "Aspirin 81 mg"). Case-insensitive; a null/blank word
-	 *         never matches. This is the one matcher shared by {@link #matchesText}
-	 *         (alias-in-question) and {@link PatientClinicalContext#hasActiveDrug}
-	 *         (token-in-order-name), so the two cannot drift.
+	 *         never matches. Backs {@link #matchesText} (alias-in-prose); the active-order
+	 *         counterpart is {@link #matchesOrderName}, which shares this rule's left boundary but
+	 *         not its right one — see there for why one matcher cannot serve both.
 	 */
 	static boolean containsWord(String text, String word) {
-		if (text == null || word == null || word.isEmpty()) {
+		return containsBoundedToken(text, word, 0);
+	}
+
+	/**
+	 * How many trailing letters an active-order display name may carry past a matched drug token
+	 * before the token stops naming that drug. Two: a localized drug name suffixes the INN stem with
+	 * one inflectional ending — Romance singulars ({@code Aspirine}, {@code Aspirina},
+	 * {@code Ondansetrona}) and their plurals ({@code Multivitamines}) — while a longer tail is a
+	 * different substance ({@code Heparinoids}, {@code Multi-Vitamin Adult}). See
+	 * {@link #matchesOrderName} for the measurement behind the number.
+	 */
+	private static final int MAX_ORDER_NAME_INFLECTION_LETTERS = 2;
+
+	/**
+	 * Order-name matching: whether {@code token} — an interaction rule's drug token — names the drug
+	 * in a patient's active drug ORDER, whose {@code orderName} is one display name rather than
+	 * prose.
+	 *
+	 * <p>A bare containment test here reports drugs the patient has never taken, because drug names
+	 * nest: {@code "tiotropium".contains("opium")} and {@code "spironolactone".contains("iron")} are
+	 * both true, and both were raised as Major interaction chips on the 3.7.1 standalone (issue #86).
+	 * The discriminating half of the fix is the LEFT boundary shared with {@link #containsWord}: an
+	 * alphanumeric character immediately before the token means the token sits inside a longer word,
+	 * i.e. a different molecule — {@code tiotr|opium}, {@code sp|iron|olactone},
+	 * {@code nitro|glycerin}, {@code bud|esonide}, {@code hydro|chlorothiazide},
+	 * {@code cipr|ofloxacin}.
+	 *
+	 * <p>The right-hand side is where this deliberately differs from {@link #containsWord}, because
+	 * the two kinds of string differ: prose is words, an order name is one localized, inflected
+	 * display name with a dose appended. Measured over the 3.7.1 demo dictionary (2531 drug and
+	 * drug-concept names x the full KB's 2093 rule tokens), by tolerated trailing letters:
+	 *
+	 * <pre>
+	 *   rule                    matches   nested-name collisions leaking   what enters at this step
+	 *   contains (the defect)     896     9 of 9                           —
+	 *   symmetric boundary        761     0 of 9                           —
+	 *   left + &lt;=1 letter         828     0 of 9   67 localized spellings (Aspirine Co 81mg, Aspirina,
+	 *                                              Amoxicilline, Clarithromycine Co 500mg, Ondansetrona)
+	 *   left + &lt;=2 letters        829     0 of 9   1 localized plural (Multivitamines et fer)
+	 *   left + &lt;=3 letters        829     0 of 9   nothing
+	 *   left + &lt;=4 letters        834     0 of 9   5 FALSE positives (Heparinoids ~ heparin,
+	 *                                              Multi-Vitamin Adult ~ vitamin a)
+	 * </pre>
+	 *
+	 * A symmetric boundary would therefore stop checking a patient on {@code Aspirine Co 81mg} for
+	 * aspirin interactions at all — trading a false positive for a false NEGATIVE, the wrong
+	 * direction for a safety net, and one that looks exactly like the noise being removed. Two is the
+	 * far edge of the plateau where every legitimate name is matched and no false positive has yet
+	 * appeared; the first ones appear at four. Stopping at one (this issue's originally measured
+	 * recommendation) leaves exactly one legitimate name unmatched, {@code Multivitamines et fer},
+	 * whose reference entry carries 2 Major and 8 Moderate rules that would silently stop being
+	 * checked.
+	 *
+	 * <p>A bound on the tail, rather than a list of known inflections: stripping
+	 * {@code -e}/{@code -a}/{@code -o} from both sides was measured on the same corpus at 826
+	 * matches, a strict subset of this rule, and trades one bound for a per-language whitelist that a
+	 * differently-localized deployment falls off silently. Residual imprecision this rule keeps, for
+	 * the record: 2 of the 829 are a nitroglycerin order matching the token {@code glycerin} through
+	 * its own parenthetical synonym ("glycerine trinitrate") — a mislabel, not a fabricated drug, and
+	 * one every rule that tolerates an inflectional tail shares.
+	 */
+	static boolean matchesOrderName(String orderName, String token) {
+		return containsBoundedToken(orderName, token, MAX_ORDER_NAME_INFLECTION_LETTERS);
+	}
+
+	/**
+	 * The one boundary-aware containment scan, shared by prose matching ({@link #containsWord}) and
+	 * order-name matching ({@link #matchesOrderName}) so the boundary rule cannot drift between
+	 * them. A match needs {@code token} to start at a word boundary in {@code text} and to end at
+	 * one, give or take up to {@code maxTrailingLetters} letters — never digits, which in a drug
+	 * name are a strength rather than part of the name. Case-insensitive; a null/blank token never
+	 * matches.
+	 */
+	private static boolean containsBoundedToken(String text, String token, int maxTrailingLetters) {
+		if (text == null || token == null || token.isEmpty()) {
 			return false;
 		}
 		String t = text.toLowerCase(Locale.ROOT);
-		String w = word.toLowerCase(Locale.ROOT);
+		String w = token.toLowerCase(Locale.ROOT);
 		int idx = t.indexOf(w);
 		while (idx >= 0) {
-			boolean leftOk = idx == 0 || !Character.isLetterOrDigit(t.charAt(idx - 1));
-			int end = idx + w.length();
-			boolean rightOk = end >= t.length() || !Character.isLetterOrDigit(t.charAt(end));
-			if (leftOk && rightOk) {
-				return true;
+			if (idx == 0 || !Character.isLetterOrDigit(t.charAt(idx - 1))) {
+				int end = idx + w.length();
+				for (int tail = 0; tail <= maxTrailingLetters; tail++) {
+					int at = end + tail;
+					// A tail character must itself be a letter to be stepped over.
+					if (tail > 0 && !Character.isLetter(t.charAt(at - 1))) {
+						break;
+					}
+					if (at >= t.length() || !Character.isLetterOrDigit(t.charAt(at))) {
+						return true;
+					}
+				}
 			}
 			idx = t.indexOf(w, idx + 1);
 		}
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
@@ -266,20 +266,38 @@ public class DrugReference {
 			return false;
 		}
 		for (String alias : aliases) {
-			if (alias == null || alias.isEmpty()) {
-				continue;
+			if (containsWord(lowerText, alias)) {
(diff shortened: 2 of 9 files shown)


dkayiwa:
## Hardening pass (`/harden`) — no behaviour changed, and that is checkable

Both production files this pass touched compile to **byte-identical bytecode** against `6006977`,
the commit the live verification in the description ran against (`javap -p -c` diff is empty for
`DrugReference` and `PatientClinicalContext`; the earlier commits were tests and docs only). So no
live re-verification was required and the shared standalone was not touched. What changed is the
coverage, two false statements, and the measurements that were missing.

### The gap that mattered: the two matchers could be merged, and nothing failed

The description explains why `containsWord` (prose, symmetric) and `matchesOrderName` (order names,
left boundary + ≤2 trailing letters) are deliberately separate — merging them is what made #87
unsafe. That separation was pinned by nothing. Mutating `containsWord` to
`containsBoundedToken(text, word, MAX_ORDER_NAME_INFLECTION_LETTERS)` — one matcher for both kinds
of string, #87's mistake in the other direction — **passed all 671 api tests.**

It is the more dangerous direction. The drugs the validator checks at all are the ones the question
or answer names (`validate` → `findByQuery` → `matchesText`), so a lenient prose rule does not
mislabel an order the patient is on; it invents the proposal the chip rests on. Measured over the
full KB's 5169 aliases against the system word list (`/usr/share/dict/words`), 22 of the single-word
ones are one or two letters short of an ordinary English word:

```
aspirin ~ aspiring    warfarin ~ warfaring    urea ~ urease
iron    ~ irony       clove    ~ clover       ginger ~ gingerly
```

Under the merged rule, for a patient on `Warfarin 5mg`:

```
question: "She is an aspiring marathon runner — any concerns with her current medicines?"
  [interaction] Acetylsalicylic acid (aspirin) interacts with active order warfarin — Major.
      Aspirin, even in small doses, may increase the risk of bleeding in patients on oral
      anticoagulants by inhibiting platelet aggregation, prolonging bleeding time, …
```

`DrugSafetyOrderNameMatchingTest.anEnglishWordStartingWithADrugNameIsNotAProposalOfThatDrug` pins
it, paired with the positive that proves the aspirin × warfarin rule is live in that fixture, so
the negative cannot pass by resolving nothing.

### Every load-bearing decision is now pinned by exactly one test

Each row is a mutation applied to this branch and the test that caught it — run, not reasoned:

| mutation | what fails |
|---|---|
| merge `containsWord` into the order-name rule | the new prose case (before: **nothing**) |
| `MAX_ORDER_NAME_INFLECTION_LETTERS = 1` | `aLocalizedPluralOrderNameKeepsBeingCheckedForInteractions` |
| `MAX_ORDER_NAME_INFLECTION_LETTERS = 4` | `aLongerTailThanAnInflectionIsADifferentSubstance` |
| `MAX_ORDER_NAME_INFLECTION_LETTERS = 0` (#87's rule) | both localized-order-name cases |
| drop the shared left-boundary check | 3 cases |
| `hasActiveDrug` back to `drug.contains(token)` (the #86 defect) | 5 cases |

The upper edge is new and closes the one direction the constant was open in. It uses a real concept
name from this deployment's dictionary — `Heparinoids / salicylic acid` (CIEL 104813) — over KB rows
added to the fixture verbatim (drug `DDInter856`, the Major warfarin × heparin row, mechanism `851`,
each byte-identical to the 19MB KB).

### The complete kill list, since the description enumerated a hand-picked 12

Re-ran the description's corpus (2531 real dictionary names × 2093 rule tokens) through the shipped
matcher and classified **all 67** matches it drops relative to bare containment:

| cause | count | examples |
|---|---|---|
| token inside a longer word (the intended kills) | 61 | `Hydrochlorothiazide`~`chlorothiazide`, `Ciprofloxacin Co 500mg`~`ofloxacin`, `Hydrocortisone 1%`~`cortisone`, `Cloxacilline Co 500mg`~`oxacillin`, `P-aminosalicylic acid`~`salicylic acid` |
| trailing tail longer than two letters | 6 | `Heparinoids`, `Heparinoide`, `Heparinoids / salicylic acid`, `Heparinoids / hyaluronidase`, `Multi-Vitamin Adult`~`vitamin a` |
| **glued digit (`Aspirin81mg`)** | **0** | the shape does not occur in this dictionary |

Worth stating plainly, because the description's "0 of 12 known real names lost" is narrower than
the whole picture: **9 of the 2531 names now resolve to no token at all**, 5 deliberately
(`Multi-Vitamin Adult`, `Heparinoids`, `Heparinoide`, `Dipirone`, `Sprironolactone`) and 4 as
side effects worth their own follow-up — all in the safe direction (no chip rather than a wrong
chip), none a fabrication:

| name | was matching | why nothing replaces it |
|---|---|---|
| `Acetylsalicylic acid` | `salicylic acid` | aspirin's token is `aspirin`, absent from that name |
| `Chlorpheniramine` | `pheniramine` | its KB row's `rxnorm_name` is `chlorine` — a KB data defect |
| `Lisoniazide` | `isoniazid` | malformed dictionary name |
| `Lévofloxacine` | `ofloxacin` | `levofloxacin` blocked by the accent — #129 |

The structural fix for rows 1–2 is not a different bound: it is resolving a rule's token to its
loaded entry and also trying that entry's alias set (`Acetylsalicylic acid`, `Heparin sodium` are
already in the KB as aliases). That adds matches, so it needs its own measurement and its own
issue. Row 4 also means **deferring #129 is no longer neutral** — the left boundary removed the
accidental sub-token fallback that was masking the diacritics gap for accented names.

### Fixed: two claims that were checkably false

- The far-edge case said "heparinoids are a separate DDInter drug". The full KB has **no heparinoid
  row**: CIEL 104813 hangs off the three `Salicylic acid` rows, and the class members are there
  under their own names (`DDInter471 Danaparoid`, `DDInter1424 Pentosan polysulfate`). The case is
  unchanged and still correct — a heparinoid is not heparin — but now for the reason that checks out.
- Its `isEmpty()` assertion is now marked fixture-scoped: against the full KB that order
  legitimately matches the token `salicylic acid` past the `/` boundary (Moderate × warfarin). Read
  as "this order raises nothing in production" it would be wrong.
- `docs/drug-kb-demo.md` documented an `interactions[].token` as *"Substring matched
  (case-insensitive) against the patient's active-order drug names"* — the sentence a KB author
  reads before writing a token, under which a token authored as a fragment silently stops firing.
- A `DdiDrugReferenceSourceTest` comment said the same.

### Strengthened, and one measurement recorded

- `theNestedShorterDrugStillFiresOnItsOwnOrder` asserted the detail contains `chlorothiazide`,
  which is itself a substring of `hydrochlorothiazide` — the needle could not tell the two rules
  apart. Added the paired negative.
- **The digit rule had no measurement and no test.** Nothing pinned "a tail character must be a
  letter", so what `Aspirin81mg` should do was open in the code and closed nowhere. Measured rather
  than argued: zero such names exist in the corpus, and treating a digit as the end of the name
  instead scores *identically* (829 either way). The two are indistinguishable on real data, so the
  javadoc now records that, and the alternative does not get re-litigated from first principles.
  Deliberately not pinned with a fabricated order name: that would freeze an unmeasured choice as
  spec. If a deployment shows the shape, that is the trigger to measure it.
- `containsBoundedToken`'s javadoc claimed "a null/blank token never matches" over an `isEmpty()`
  guard. Corrected, with where a blank alias can still come from (a hand-authored `json` KB is not
  sanitized — pre-existing; the guard belongs in that parser).
- `PatientClinicalContext.containsToken` now says why the allergy/condition arm keeps bare
  containment 25 lines under the javadoc explaining why bare containment fabricated drugs.

### Also checked, nothing to change

- **`hasActiveDrug` is the only path from an active-order name into anything.**
  `getActiveDrugNames()` has no other consumer; `findByActiveOrders` and both class arms are
  ATC-only; `lookupByToken` is fed allergy text. So the chips and the promoted prompt prose cannot
  disagree about which orders a rule matches, which is what `orderedInteractionNotes`' javadoc
  claims — still true. (That guarantee currently rests on the getter having no callers, not on
  enforcement.)
- **Index safety, empirically:** 800,000 randomized calls through `matchesOrderName`/`containsWord`
  over letters, digits, `-/().`, `é`, `İ` (whose lowercase changes the string's length) and an emoji
  surrogate pair — no exception, and the documented cases verified by execution.
- **The prose path is unchanged**, which matters most since `findByQuery` decides which drugs get
  validated at all: all four callers pre-lowercase, so the refactor's added `toLowerCase` is a
  redundant scan, not a behaviour change. Its cost (~5 ms per `findByQuery` at full-KB scale against
  a 15–90 s answer) is not worth a pre-lowered entry point whose misuse would be a silent false
  negative.
- No other stale wording in `docs/`, `README.md`, `docs/adr.md`, `*.java`, resources or
  `config.xml`.

**Tests:** `mvn -o clean install` (full reactor) — **API 672 + OMOD 53 = 725 run, 0 failures, 0
errors, 34 skipped** (`main` is 715; the description's 723 predates these cases).

### Deferred, each with what breaks if it ships as-is

- **The chip names the rule's token, never the order that matched it**
  (`DrugSafetyValidator:394`): a chart line reading `Aspirine Co 81mg` produces "interacts with
  active order aspirin", and `Multivitamines et fer` produces "multivitamin". If we ship without
  this, a clinician cannot reconcile the chip against the chart — the same reconciliation failure
  that let #86 sit unnoticed. Not fixed here: it changes wire-visible chip text, so it needs its own
  live verification, and the rendering choice predates this PR (the tail allowance only makes it
  more visible).
- **`collisionValidator()` is a fourth hand-rolled DDInter fixture loader** (`DrugSafetyChipLabelTest`,
  `DdiDrugReferenceSourceTest`, `DrugSafetyInteractionSeverityFloorTest` have the other three). If we
  ship without consolidating, the four can drift — one drops the `assertNotNull` guard or the
  try-with-resources and a fixture test silently parses nothing. Not fixed here: the fix belongs in
  `DrugReferenceTestSupport`, the single most contended file across the concurrent DDI branches, and
  it would consolidate a pattern three files established before this PR.
- **`hasAllergyToken`/`hasConditionToken` still use bare containment.** If we ship without a rule of
  their own, a curated rule with token `opium` fires on an allergen recorded as `Tiotropium` — the
  #86 defect on a different arm. Exposure is confined to hand-authored contraindication rules
  (neither `ddinter` nor `atc` emits any). Needs its own measurement over real allergy text.
- **`PatientClinicalContext.getActiveDrugNames()` is public with no caller.** If we ship as-is,
  nothing breaks today, but the "one path only" guarantee is an accident a future caller can undo
  without touching the matcher.

PR #134, opened 2026-08-04
fix(drug-safety): configurable pair-chip cap, and ATC codes attributed per active order

### LINKED ISSUE #131: The pair-chip cap of 10 is hardcoded and withholds Major interactions — make it a global property
### Summary

`MAX_QUESTION_PAIR_CHIPS` and the screening arm's equivalent are both hardcoded to **10**. On a question naming many drugs that withholds **Major** interactions, and the right number is a clinical judgement an implementation should be able to make — not one this module should fix at build time.

Measured live on the standalone (#120's verification, real bundled DDInter data), a 245-character question naming 16 drugs:

```
Question-pair safety: 10 of 65 question-named drug pairs shown (cap 10); the question
resolved 16 reference drugs, and pairs grow as N^2/2. Withheld, least severe last:
[Major ×10, Moderate ×37, Minor ×8]
```

So a clinician reviewing 16 medications sees ten Majors and **ten more Majors are withheld**. Nothing is silently dropped — every withheld pair is named with its severity in the WARN, and the shown set is severity-descending — but the log is not where a clinician looks.

### Why a cap is nonetheless required

Uncapped, the same question produced **72 chips and 42,708 characters** of injected finding text, growing as N²/2 in the question's drug count. Before the pairwise arms existed, findings were bounded by the patient's own chart (0–3). The reference-record path already caps at `MAX_INTERACTION_RENDER_CHARS = 1500` per entry for exactly this reason, and there is no chart-size cap anywhere else in the module. Removing the cap is not an option.

### Proposed fix — make it configurable, not guessed

Add a global property (e.g. `chartsearchai.drugSafety.maxPairChips`) defaulting to the current 10, read through `ChartSearchAiUtils.getIntGlobalProperty` like the other tunables, with an unparseable or non-positive value falling back to the default rather than silently disabling the cap — the same failure-safe pattern `maxWeightAgeDays` and `minInteractionSeverity` already use.

That converts a hardcoded clinical judgement into a deployment decision, which is the honest place for it: a polypharmacy review clinic may want 30, a triage screen may want 5. Both arms should read the same GP, since a question cannot reach both (their gates are mutually exclusive) and two different limits for one concept would be arbitrary.

Worth considering alongside it, though it can be deferred: the withheld count is currently visible only in a log line. A clinician-facing signal ("10 of 65 shown") would need a UI affordance that does not exist today — chips are per-drug findings with no container — so it is a frontend change, not a module one. Noting it so the log-only state is a recorded decision rather than an oversight.

### Tests

`validate()` through the real production path: the cap honours a raised GP, honours a lowered one, falls back to the default on a non-positive or unparseable value, and the retained set is still severity-descending at every setting. Plus the existing WARN-names-the-withheld behaviour must survive a changed limit.

### LINKED ISSUE #132: ActiveDrugOrder carries names but not ATC codes, so an order can still witness itself through the ATC leg
### Summary

`PatientClinicalContext.ActiveDrugOrder` (added by #118) carries a uuid, a display string and the order's **names** — but not its **ATC codes**. Those still live only in the flat, context-wide `getActiveDrugAtcCodes()` set, so a code cannot be attributed to the order it came from. That is the last remaining false-positive path in the interaction-screening arm, documented as a residual when #113 merged.

### The failure it leaves open

The screening arm joins each active order against the others, and must exclude a subject's *own* order so it cannot witness itself. With per-order names it does that correctly for the name half. For the ATC half it cannot: excluding the subject means removing its codes from a flat set, and if two *different* related entries resolve from the same order's codes, the order pairs with itself through the ATC leg.

Concretely, all three of these must coincide:

1. the order's name matches no alias, so name attribution cannot identify the subject;
2. that one concept carries the ATC codes of **two** entries in the KB; and
3. those two entries have an above-floor interaction row between them.

The output is one non-blocking chip, bounded by the severity floor and the pair cap. It is strictly narrower than the two false-positive paths #113's hardening closed (a co-formulation order reporting its own halves, and an ATC-resolved subject witnessing itself), which is why it was not a merge blocker — but it is still a fabricated clinical claim on a patient who may be on a single medication.

Note only **85 of 616** Drug-class concepts in the reference demo dictionary carry any ATC map, which is why this is hard to reach in practice and why no probe patient exercises it.

### Proposed fix

Carry ATC codes on `ActiveDrugOrder` alongside the names, populated from the same single `getActiveOrders` read in `PatientClinicalContextBuilder` that already fills the rest of the structure — the codes are read from the order's concept there today, so this is a matter of keeping the association rather than gathering anything new. Then the screening arm's "exclude the subject's own order" step can drop that order's codes as well as its names, and the ATC leg becomes attributable like the name leg.

Keep the flat `getActiveDrugAtcCodes()` accessor: the class arms and `findByActiveOrders` legitimately want the union, and #118's per-order work deliberately kept a flat fallback for contexts built without per-order structure. Deriving the flat set from the per-order list is fine; removing it is not.

### Tests

Through the real production path: an active order whose concept carries the ATC codes of two interacting entries, name matching no alias, must produce **zero** chips — and the same context with a second, genuinely different order must still produce the real pair, so the fix cannot be a blanket suppression. Mutation-check that dropping the per-order code exclusion reintroduces the self-witness.

Live verification needs a purpose-made concept, since no stock patient reaches this; note that the class arm is unreachable on every stock probe patient for the same ATC-coverage reason (see #88's reproduction, which needed a hand-made Ondansetron order).

dkayiwa:
Closes #131 and closes #132 — two residuals recorded when #113/#114 merged, both in the
interaction-screening area, both small.

## #131 — the pair-chip cap was a build-time guess

`MAX_QUESTION_PAIR_CHIPS` and `MAX_SCREENING_INTERACTION_WARNINGS` were both hardcoded to 10.

Measured live on the 3.7.1 standalone (full 19 MB DDInter KB, patient on nothing, a 16-drug
polypharmacy question) — this line is `main`'s behaviour and this branch's default alike:

```
Question-pair safety: 10 of 72 question-named drug pairs shown (cap 10); the question resolved
16 reference drugs, and pairs grow as N^2/2. Withheld, least severe last:
[Major x13, Moderate x40, Minor x9]
```

So a clinician reviewing 16 medications sees ten Majors while **thirteen more are withheld**. (Issue
#131 quotes `10 of 65` with `[Major ×10, Moderate ×37, Minor ×8]` for a differently-worded 16-drug
question; that one is not in the tree and its profile does not reconcile with any DDInter rating of
these sixteen drugs, so the figures used in the code and the GP description are the measured ones
above.)
Nothing is dropped silently — the WARN names every withheld pair and the shown set is
severity-descending — but the log is not where a clinician looks, and "how many pairs is enough"
is a clinical judgement, not a build-time one. Removing the cap is not the alternative: uncapped,
that question produced **72 chips / 42,708 characters** of injected finding text, growing N²/2,
against a path that caps a single reference record at 1500 characters for exactly this reason.

**Change:** `chartsearchai.drugSafety.maxPairChips`, default **10** — the number both arms were
hardcoded to, so no deployment's behaviour changes — read through
`ChartSearchAiUtils.getIntGlobalProperty`, with an unparseable or non-positive value falling back
to the default rather than disabling the cap (the fail-safe pattern `weightMaxAgeDays` and
`minInteractionSeverity` already use; here it matters twice, since `0` must mean neither "no
chips" nor "no cap"). **Both** pairwise arms read the one property: their gates are mutually
exclusive (≥ 2 question-named drugs versus none), so no question can be subject to both and two
limits for one concept would be arbitrary. The two constants collapse into one accessor that
carries both of their rationales, including the unrated-sorts-above-Major honesty limit. Registered
in `config.xml` and the README with the trade-off stated in both directions.

## #132 — `ActiveDrugOrder` carried names but not ATC codes

Codes lived only in the flat, context-wide set, so "one order carrying two codes" and "two orders
each carrying one" were the same input: an order whose name matched no alias — reaching the subject
set through its codes alone — paired the two entries its own concept's codes resolved to.

Reproduced live **before** the fix with a purpose-made concept (no stock probe patient reaches it:
only 85 of 616 Drug-class concepts in the demo dictionary carry any ATC map). One Drug concept
`Zolvimix`, mapped to WHOATC `C10AA01` + `J01FA09`, ordered once for a patient with no other live
drug order, asked the screening question:

```
safetyWarnings (1):
  [interaction] Simvastatin — "Simvastatin interacts with active order clarithromycin — Major. ..."
answer: "No — there is a potential drug interaction: Simvastatin interacts with active order
         clarithromycin, a Major finding [106]."
```

2 of 2 runs, and the model repeated the fabrication in its prose — about a patient on **one**
medication and on neither named drug.

**Change:** the codes ride on `ActiveDrugOrder`, populated in the same single `getActiveOrders`
pass that already reads them off the same concept, so the "exclude the subject's own order" step
drops its codes exactly as it drops its names, and *which order is this subject's own* is answered
by both matchers that could have made it a subject (a name alias, or a shared ATC code).

Two consequences worth naming:

* a code the subject's own order carries is **re-added** when another order carries it too, because
  removal from a flattened set cannot tell contributors apart — which also recovers the genuine
  combination-plus-constituent pair the old reduction had to give up;
* the pre-#132 proxy (dropping the codes of every *other entry the same order's name resolves to*) is
  **retained per order alongside** the exact removal, not replaced by it. Self-review measured why: an
  order's mapped codes need not cover every substance *in* it, so a combination whose concept maps to
  one constituent's code leaves the other half's code in the context-wide set with nothing attributing
  it — and exact-removal-only re-opened #124 ("Acetylsalicylic acid (aspirin) interacts with active
  order esomeprazole" off one tablet) both for a code-less order and for one carrying *some* code. The
  proxy cannot over-remove, because anything another order carries is restored.

What the restoration cannot reach is stated on the method rather than papered over: a code carried only
by orders that are *all* the subject's own (two brand-named multi-substance orders covering the same two
substances) or only by a list-omitted nameless order. Both lose a pair rather than invent one, and
neither is separable from the one-order case without more order structure than a code set carries.

`getActiveDrugAtcCodes()` **stays**, and stays independent of the per-order list: the class arms and
`findByActiveOrders` want the union, #118 deliberately kept a flat fallback for contexts built
without per-order structure, and in production the per-order list can be *narrower* than the union —
a nameless order contributes codes but no `ActiveDrugOrder` (the builder's KNOWN GAP), so deriving
the union would silently drop those codes. The flattened fallback in `activeOrdersOtherThan` is
untouched and is now where the residual lives; its javadoc says so.

## Tests

Rebased onto `71a7c4b` (#137 / #133 / #138) with **no conflicts**. `#138`'s diacritic fold lives in
`DrugReference.containsBoundedToken`, the scan shared by `matchesText` and `matchesOrderName`, so the
per-order exclusion's name leg and the witness predicate fold together — both move in the same
direction, which keeps the guard aligned with the matcher it guards.

`api 813 + omod 58 = 871, 0 failures, 34 skipped`, measured with a full root reactor
(`mvn -o clean install`) on this branch. The same command puts pre-existing api at **796** on
`71a7c4b`, so this PR is exactly **+17** — its 17 new tests, nothing else moved. (A quoted api baseline
of 789 differs from 796 by 7 tests already on main, not by anything here.)

All new tests run the real production path — the real bundled DDInter sample parsed by the real source, the
real `validate` entry points, real GP writes through the admin service.

Mutation-checked, each one:

| mutation | what fails |
|---|---|
| per-order ATC exclusion removed | the self-witness returns, in both the pure-context and the end-to-end test, with exactly the chip text seen live |
| shared-code re-add removed | the fix becomes a blanket suppression and loses the real two-order pair |
| retained proxy removed (exact removal only) | the co-formulation tests — #124 returns for an order carrying no codes, and for one carrying only one constituent's code |
| the bound on the re-add removed | one order witnesses a pair through a code the chart-wide set never carried |
| cap ignores the GP (pre-#131 behaviour) | 7 of the 9 cap tests |
| non-positive guard dropped | the fail-safe test (`0` → 0 chips) |

Two details worth flagging: the cap suite asserts the shown set is always the **prefix** of one
severity-ordered list (stronger than "descending at every setting", which a cap-while-collecting
refactor would still satisfy), and it captures the validator's own logger to pin that the WARN
reports the cap that actually cut — for both arms, since that log line is the only place the
withheld pairs appear.

`ActiveOrderAtcContextTest` runs #132 end to end through `PatientClinicalContextBuilder` and the
public `validate(answer, question, patient)`, so the association is proven to exist in production
rather than only in a hand-built context.

**One test line changed, and it is a symbol rename rather than a spec change:**
`DrugSafetyInteractionScreeningTest` asserted against
`DrugSafetyValidator.MAX_SCREENING_INTERACTION_WARNINGS`, which no longer exists; it now asserts
against `ChartSearchAiConstants.DEFAULT_DRUG_SAFETY_MAX_PAIR_CHIPS` — same value, same assertion,
and that test sets no GP.

## Live verification

Two full-JVM cycles on the shared 3.7.1 standalone, before and after, plus the four regression
controls. Results are posted as a comment on this PR.

Test data created on that box (developer machine, demo data — noted so a later session can tell
scaffolding from real data): WHOATC reference terms `C10AA01` / `J01FA09`; Drug concepts `Zolvimix`
(both codes) and `Klarizom` (`J01FA09` only); drug orders on those concepts for Michael Turner
(`d35b0325-9924-4baa-93ec-1ca3585f3e2e`), who had no live drug order before.


dkayiwa:
## Live verification — two full-JVM cycles on the 3.7.1 standalone

Required, because both halves change production behaviour. Same instance, same full 19 MB DDInter KB,
`sourceFormat=ddinter`, `minInteractionSeverity=minor`, `chartMode=fullChart`, `llm.engine=local`. Each
cycle = build → deploy → **full-JVM restart** → warm → querystore-alive gate (`hits>0` plus a
known-positive control chip) → probes, all under the shared lock.

* **BEFORE** = `main` (`e73c8b0` at the time), label `i131b`, 2026-08-05 ~02:15
* **AFTER** = this branch rebased on `71a7c4b`, label `i131`, 2026-08-05 ~03:40

Chips are `DrugSafetyValidator` output, not model prose — the deterministic half. Answer wording is
[not reproducible on this box](https://github.com/openmrs/openmrs-module-chartsearchai/pull/108) and no
claim below rests on it.

### #132 — the self-witness, before and after

Test data (purpose-made; no stock probe patient reaches this arm — only 85 of 616 Drug-class concepts
carry any ATC map): Drug concept **`Zolvimix`** mapped to WHOATC `C10AA01` **+** `J01FA09` (verified
live: `concept Zolvimix ATC maps: ['J01FA09', 'C10AA01']`), ordered once for **Michael Turner**
(`d35b0325-9924-4baa-93ec-1ca3585f3e2e`), who had no live drug order. Question: *"Are there any drug
interactions with his current medications?"*

| | BEFORE (`main`) | AFTER (this branch) |
|---|---|---|
| one order, two codes | **1 chip** — "Simvastatin interacts with active order clarithromycin — Major …" (2 of 2 runs) | **0 chips** (2 of 2 runs) |
| answer prose | *"No — there is a potential drug interaction: Simvastatin interacts with active order clarithromycin, a Major finding [106]."* | — |
| + a genuinely different order (`Klarizom`, `J01FA09` only) | 1 chip | **1 chip** — same real pair, unchanged (2 of 2 runs) |

So the fabricated claim is gone and the genuine two-order pair survives: the fix is not a blanket
suppression. Both absence results were re-run (a cold-start race can load the KB empty for a whole JVM
lifetime with no WARN, so a single 0-chip read proves nothing).

### #131 — raising and lowering the cap, live

`chartsearchai.drugSafety.maxPairChips` was created by `config.xml` on module start and read back over
REST with the full description (`value: "10"`), so an operator can actually set it. Question: *"Is it
safe to combine warfarin, ciprofloxacin, clarithromycin, fluconazole, amiodarone and simvastatin?"* —
**15** above-floor pairs, 10 of them Major, patient on nothing.

| GP value | chips | WARN line |
|---|---|---|
| absent → default | **10** | `10 of 15 … (cap 10)`, withheld `[Moderate ×3, Minor ×2]` |
| `3` | **3** | `3 of 15 … (cap 3)`, withheld `[Major ×7, Moderate ×3, Minor ×2]` |
| `25` | **15** | none — a cap above the candidate count shows everything |
| `0` | **10** | back to the default: not 0 chips, not 15 |
| `ten` | **10** | same |

At `3` the three surviving chips are Simvastatin × Clarithromycin, × Fluconazole and × Amiodarone — all
**Major**, with seven Majors named in the WARN. That is the severity-descending cut and the recoverable
log, live.

The 16-drug polypharmacy question, same instance:

```
default: 10 chips  — "Question-pair safety: 10 of 72 question-named drug pairs shown (cap 10);
                      the question resolved 16 reference drugs … Withheld, least severe last:
                      [Major ×13, Moderate ×40, Minor ×9]"
cap 25:  25 chips
```

That `10 of 72` / 13-withheld-Majors line is the measurement this PR's prose and the GP description now
quote. **Provenance note:** issue #131 reports `10 of 65` with `[Major ×10, Moderate ×37, Minor ×8]` for
a differently-worded 16-drug question; that question is not in the tree and its profile does not
reconcile with any DDInter rating of these sixteen drugs, so I did not adopt it as a measurement — the
figures above are mine, from the run described here.

### The four regression controls (AFTER build)

| patient | question | expected | got |
|---|---|---|---|
| Mary Smith | Is it safe to give her clarithromycin? | 1 chip, Major, ×simvastatin | ✅ 1 — "Clarithromycin interacts with active order simvastatin — Major" |
| Agnes Adams | Is it safe to start warfarin? | 1 chip, Major, ×aspirin | ✅ 1 — "Warfarin interacts with active order aspirin — Major" |
| Joshua Johnson | Can I give ibuprofen? | 2 chips (NSAID contra + lisinopril Moderate) | ✅ 2 |
| Mary Smith | Is it safe to give her paracetamol? | 0 chips | ✅ 0, re-run 0 |

### One arm not re-captured, and why

Three test-only/comment changes and one provably-narrowing guard landed **after** this capture
(self-review found that restoring a shared code could widen the derived code set past the context's own
union for a hand-built context, so the restore is now intersected with that union). It is a no-op for
anything `PatientClinicalContextBuilder` produces — the builder folds every order's codes into the union
in the same pass, so the intersection removes nothing — which is exactly the path all the probes above
took. The shape it closes is unreachable through the builder and is pinned by
`ActiveOrderAtcAttributionTest.theDerivedContextCanOnlyEverNARROWWhatTheJoinSees`, mutation-checked.

### Test data left on the box

Developer machine, demo data. WHOATC reference terms `C10AA01` / `J01FA09`; Drug concepts `Zolvimix`
(both codes) and `Klarizom` (`J01FA09` only); two drug orders on those concepts for Michael Turner
(`d35b0325-…`), who had none before. The `maxPairChips` GP is left at its default `10`.

dkayiwa:
**Correcting my own merge message.** It says #132's self-witness path is "covered by unit test only … no live claim is made for that path". That is wrong — the live case *was* built and measured, by the agent that wrote this PR, and its report reached me after I had already merged.

I verified the fixtures are on the standalone:

```
Zolvimix   ->  WHOATC J01FA09, C10AA01     (two codes, two entries -- the self-witness shape)
Klarizom   ->  WHOATC J01FA09              (one code -- the control)
```

Measured, purpose-made `Zolvimix` order on a patient with no others:

- **Before: 1 fabricated chip**, 2/2 runs — *"Simvastatin interacts with active order clarithromycin — Major"* for a patient on neither, and the model repeated it in prose.
- **After: 0 chips**, 2/2 runs.
- **Control** (`Klarizom`, one code): **1 chip before and after** — so the fix narrows rather than suppressing wholesale.

That is exactly the discriminating pair I would have asked for. The path is live-verified; my caveat was a reporting error on my part, not a gap in the work.

Three further things from that report worth recording here, because they are better reasoning than the issue's:

1. **`getActiveDrugAtcCodes()` is deliberately NOT derived from the per-order list.** #132 (and I) assumed deriving it would be safe. It is not: in production the per-order list can be *narrower* than the union, because a nameless order contributes codes but no `ActiveDrugOrder`. Deriving would silently drop those codes.
2. **An order's mapped codes need not cover every substance in it**, so the pre-#132 name-proxy is retained *alongside* exact removal — exact-only was measured to re-open #124.
3. **The shared-code re-add is intersected with the context's own union**, so the arm can only ever narrow what the join sees.

And on the cap figures: this PR's README, GP description and code use the agent's own measurement (`10 of 72`, withheld `[Major ×13, Moderate ×40, Minor ×9]`) rather than the `10 of 65` / `[Major ×10, Moderate ×37, Minor ×8]` I quoted when filing #131. It could not reconcile my profile with any DDInter rating of those sixteen drugs, and my figure came from a differently-worded question no longer in the tree. Its measurement supersedes mine; #131's numbers should be read as illustrative.

Phase 2 did **not** converge — the second pass still found real items, and the agent said so rather than claiming convergence. Those items are #139.

Issue #122, opened 2026-08-04
safety_finding records were never added to #106's demote-only grounding carve-out, so the module's own deterministic findings get unstable grounded=false verdicts


dkayiwa:
### Summary

#110 made the deterministic safety finding a citable record so the model would report a conclusion it would not re-derive. #106 had already established that module-supplied injected records must be **demote-only** for grounding — a cosine pass may render `null`, never `true`. The new `safety_finding` type was never registered with that carve-out, so the module's own deterministic findings are graded by the grounding pipeline as if they were retrieved chart evidence. The verdicts are wrong and unstable.

`CitationGroundingVerifier.java:260` — the carve-out names one type:

```java
if (ChartSearchAiConstants.RESOURCE_TYPE_DRUG_REFERENCE.equals(mapping.getResourceType())) {
    demoteOnlyIndexes.add(Integer.valueOf(mapping.getIndex()));
}
```

`RESOURCE_TYPE_SAFETY_FINDING` occurs in exactly three places in `api/src/main` — its declaration (`ChartSearchAiConstants:496`), the injection site (`DrugReferenceInjector:184`), and `ChartSearchAiUtils.referenceGroup:101`. It is absent from `CitationGroundingVerifier` entirely.

### Observed

Four captures on the 3.7.1 standalone, HEAD `13690b1`. The clearest is Margaret King (active order Dexamethasone 4mg) asked "Is it safe to give voxelotor?", which injects three findings — two of them **byte-identical** in text:

```
references: [(2,   condition,      chart,     grounded=False),
             (121, safety_finding, reference, grounded=False),   <- the MAJOR finding
             (122, safety_finding, reference, grounded=True),
             (123, safety_finding, reference, grounded=True)]    <- 122 and 123 identical text
```

The **most severe** finding is the one marked ungrounded, while two byte-identical siblings are marked grounded. Also:

- Melissa Wright + dolutegravir: `(41, safety_finding, grounded=False)` and `(42, safety_finding, grounded=True)` — again near-identical records disagreeing.
- The same finding flipped `grounded=True` → `False` across two runs of one probe, independently observed while verifying #115.
- Mary Smith + "Does warfarin interact with aspirin?" (on #120's branch): `(79, safety_finding, grounded=True)`.

So the verdict is neither correct nor stable, and its instability tracks embedding noise rather than anything about the finding.

### Why this matters

[...]

PR #140, opened 2026-08-05
fix(drug-safety): ask the direct-allergy question for a drug carrying no ATC codes

### LINKED ISSUE #135: A direct recorded allergy raises no contraindication chip for 444 of 2283 drugs — the identity check sits behind a classification guard
### Summary

A patient with a recorded allergy to a drug, asked about **that same drug**, gets **no contraindication chip** when the drug carries no ATC level-4 subgroup and no curated cross-reactivity group. The most basic safety check the module makes — "they are allergic to this, do not give it" — is skipped for **444 of 2283 (19.4%)** of the loaded DDInter entries.

`DrugSafetyValidator.addClassContraindications` returns before it ever looks at the allergy list:

```java
Set<String> refClasses = ref.atcSubgroups();
List<CrossReactivityGroup> refGroups = CrossReactivityGroup.groupsOf(ref, …);
if (refClasses.isEmpty() && refGroups.isEmpty()) {
    return;                       // <-- exits here
}
for (String allergyToken : context.getAllergyTokens()) {
    …
    if (allergen == ref) {        // <-- the DIRECT-allergy check, never reached
        warnings.add(new SafetyWarning(TYPE_CONTRAINDICATION, ref.displayLabel(),
                "The patient has a recorded allergy to " + ref.displayLabel() + "."));
```

The guard is correct for the two *class-based* comparisons below it — without a subgroup or a group there is nothing to compare against. But the `allergen == ref` branch is **identity, not classification**: it needs no ATC code and no group, so it should not be gated on having one.

### Nothing else covers it

- `addContraindications` (the rule arm) iterates `ref.getContraindications()`, which is **curated data only**. `DdiDrugReferenceSource`'s own javadoc states its entries expose `interactions`, "never `ageBands` or `contraindications`" — so for a DDInter-sourced entry that arm has nothing to fire on.
- A curated `CrossReactivityGroup` would rescue an affected drug, but that file is small and hand-authored; it covers families like NSAIDs, not the long tail.

So for those 444 entries the direct-allergy warning has no path to the clinician at all.

### Measured

```
DDInter entries:                                  2283
  with no ATC codes at all:                        444   (19.4%)
  ATC present but no level-4 subgroup:               0
```

Affected drugs are not obscure: `Ledipasvir`, `Leucovorin`, `Levoleucovorin`, `Levomefolic acid`, `Antilymphocyte immunoglobulin (horse)`, `Anthrax vaccine`, `Kava`, `Lactic acid`, `Lactobacillus acidophilus`.

### Why this is worse than the interaction defects fixed today

Every interaction defect closed in #86/#113/#114/#115/#88 was either a false positive (a fabricated claim, which is bad but visible) or a missing *interaction* warning. This is a missing **contraindication** on an exact-match allergy — the one check a clinician would assume is airtight, and the failure is silent: no chip, no WARN, nothing in the answer unless the LLM happens to notice the allergy record on its own.

### Proposed fix

Move the direct-identity check **above** the early return, or split the method so identity and classification are separate concerns:

```java
DrugReference allergen = drugReferenceService.lookupByToken(allergyToken);
if (allergen == ref) { … }        // identity: no ATC or group needed
```

Keep the guard for the two class comparisons that genuinely require classification data. Consider whether `addContraindications` is the better home, since it is already the identity/rule arm — the split between the two methods is currently "curated rules" versus "everything class-shaped", and a direct allergy match belongs with the former.

### Tests

Through the real production path (`DrugSafetyValidator.validate`): a context with a recorded allergy to a drug whose entry has **no ATC codes and no cross-reactivity group**, question naming that drug, must raise one contraindication chip. Mutation-check by restoring the early return — the test must fail. Also pin that the two class-based comparisons still do NOT fire for such an entry, so the fix does not turn the guard off wholesale.

Live: needs an allergy recorded against one of the 444, plus a question naming it. Note only 85 of 616 Drug-class concepts in the reference demo dictionary carry any ATC map, so the demo data reaches this shape easily — see #88's reproduction for the order-creation pattern.

Found while hardening the eval harness for #126, not by a safety probe — which is itself worth noting: no existing test or live probe covers a direct allergy to an unclassified drug.

dkayiwa:
Fixes #135.

`DrugSafetyValidator`'s allergen arm resolves each recorded allergy to a reference entry and asks
three questions in decreasing specificity — is the allergen **this** drug, does it share this drug's
ATC level-4 subgroup, does it share a curated `CrossReactivityGroup`. All three sat behind one early
return taken when the drug being checked had neither a subgroup nor a group. That guard is right for
the two class comparisons. It is wrong for `allergen == ref`, which is a comparison of two object
references: it needs no ATC code, no group and no dataset support of any kind.

So a patient with a recorded allergy to a drug, asked about **that same drug**, got no
contraindication chip whenever the entry carried no ATC codes.

## Measured

Over the full openmrs-ddi-knowledge-base DDInter 2.0 dataset the standalone loads (2283 drugs,
295,184 interaction rows, 8234 mechanisms), re-measured on 2026-08-05:

```
DDInter entries:                                  2283
  with no ATC codes at all:                        444   (19.4%)
  ATC present but no level-4 subgroup:               0
  of the 444, entries carrying >=1 interaction row: 444   (all of them are reachable)
```

The guard's two halves therefore fail **together**, on nearly a fifth of the dataset — and not on
obscure entries: `Levothyroxine` (682 interaction rows), `Valproic acid` (701), `Tiotropium` (430),
`Vasopressin` (289), `Chlorpheniramine` (378), `Ledipasvir`, `Leucovorin`, `Kava`, `Lactic acid`.

Nothing else covered them:

- `addContraindications` reads the entry's curated `contraindications`, which `DdiDrugReferenceSource`
  never emits (its own javadoc: entries expose `interactions`, "never `ageBands` or
  `contraindications`").
- A curated cross-reactivity group cannot rescue an ATC-less entry **in principle**, not merely in
  practice: membership is defined by ATC *prefix* (`CrossReactivityGroup.containsCode`), so an entry
  with no ATC code belongs to no group however much curated data a deployment authors. (#135 argued
  this from the file being small; it is stronger than that.)

And since #110 turns every chip into a citable pre-answer safety-finding record, the warning was
absent from the prompt as well — the deterministic layer contributed nothing at all.

## The fix

The guard moves **inside** the per-allergen loop, after the identity check and before the two class
comparisons that genuinely need classification data. `addClassContraindications` becomes
`addAllergyContraindications`, because only two of its three comparisons are class-based — the name
is what made the misplaced guard look right.

**Identity stays in this arm** rather than moving to `addContraindications`, the curated/rule arm
(#135 asked for that to be argued either way). Two invariants are decided per resolved allergen and
need code that sees all three comparisons at once:

- *most-specific-match wins* — an allergen that IS this drug also shares every one of its subgroups,
  so identity has to pre-empt the class arms rather than stack on them;
- *one chip per resolved allergen* (`seenAllergens`), so several aliases of one allergy warn once.

Split across two methods, each would need its own copy of the other's state — which is exactly the
two-arms-cannot-see-each-other shape that produced #88's duplicate interaction chip. And
`addContraindications` walks a different collection (`ref.getContraindications()`, matched by token
against allergy **and condition** text), so hosting the allergen walk there would put two unrelated
loops in one method and still leave the precedence decision spanning both.

Nothing else changes. For a classified entry the behaviour is byte-identical (the in-loop guard never
trips when `refClasses` is non-empty), and for an ATC-less entry the two class comparisons are
provably no-ops on empty sets — so the fix's blast radius is exactly the identity case the guard was
suppressing.

## Tests

`DirectAllergyContraindicationTest` (8 cases) over a **verbatim excerpt of the full DDInter dataset**
— `Ledipasvir` and `Leucovorin`, two of the 444, plus the real `Ciprofloxacin` (J01MA02) /
`Levofloxacin` (J01MA12) subgroup pair and the one interaction row that falls wholly inside that
slice — parsed by the real `DdiDrugReferenceSource`, with the real curated groups loaded. The bundled
16-drug sample cannot host these cases: every one of its drugs carries ATC codes, which is why no
existing test covered a direct allergy to an unclassified drug.

Three cases fail on the early return (the mutation check is the pre-fix build itself):

- the ATC-less identity chip, through `validate(answer, question, context)` with an answer that never
  names the drug — the question-driven path;
- the same finding reaching the prompt as a citable `safety_finding` record, through the real injector
  wired to the real validator;
- two aliases of one ATC-less allergy warning once (`seenAllergens` on a path that could not run).

Five pin what must not change: a fixture-shape precondition (no ATC codes, no subgroup, no group
even with the real curated groups loaded, no curated rule), an unclassified drug the patient is not
allergic to staying silent, a classified allergen raising nothing about an unclassified drug (the
guard's own contract), the shared-subgroup class chip still firing, and identity pre-empting a class
match for the same allergen.

Also in this PR: `DrugReferenceTestSupport.ddiFixtureEntries` — one shared DDInter fixture loader.
Five test files had hand-rolled one, two of which passed a null `InputStream` straight into the
parser on a missing resource; this test would have been the sixth. No assertion changed.

`mvn clean install` from the repo root: **API 821 + OMOD 58 = 879**, 0 failures, 34 skipped.
Baseline measured on `main` at `0d3a906` in the same worktree: **API 813 + OMOD 58 = 871**, 0
failures, 34 skipped. (+8 = the new cases.)

## Live: before and after, on the 3.7.1 standalone

Two full deploy/restart/probe cycles against the same live fixture, `main` @ `0d3a906` first and this
branch second, with the full 19 MB KB, `sourceFormat=ddinter`, `minInteractionSeverity=minor`,
`chartMode=fullChart`, `llm.engine=local`.

**Fixture created for this** (agent-created demo data, left in place — the standalone is a
regenerable test bed):

| patient | uuid | allergy created | active orders |
|---|---|---|---|
| Kevin Brown | `e89247a2-111f-4dd5-974d-25bfad03690a` | DRUG → **Levothyroxine** (ATC-less in the KB) | **none** |
| Mark Smith | `de4b0d62-8a47-4c82-8220-1f0a87eafd46` | DRUG → **Ciprofloxacin** (J01MA02) | Chloroquine Co 250mg, Benadryl 25mg |

`Levothyroxine` was chosen because it is one of the 16 ATC-less KB drugs that already exist as
concepts in the 3.7.1 demo dictionary, so the allergy resolves through the production path
(`Allergen.codedAllergen` → concept preferred name → `DrugReferenceService.lookupByToken`) with no
concept authoring.

| probe | before (`0d3a906`) | after (this branch) |
|---|---|---|
| **Kevin × levothyroxine** (allergic, ATC-less) | **0 chips** (×2 runs), no injected finding | **1 contraindication** + `safety_finding` record `#258` (×2 runs) |
| Kevin × vasopressin (ATC-less, not allergic) | 0 chips | 0 chips (×2 runs) |
| Joshua × levothyroxine (aspirin allergy, ATC-less subject) | — | 0 chips (×2 runs) |
| Mark × ciprofloxacin (allergic, J01MA02) | 1 contraindication (identity) + 1 Major × chloroquine | identical |
| Mark × levofloxacin (J01MA12) | 1 contraindication (same ATC class J01MA) + 1 Major × chloroquine | identical |
| Mary × clarithromycin (control) | 1 chip, Major, × simvastatin | 1 chip, Major, × simvastatin |
| Agnes × warfarin (control) | — | 1 chip, Major, × aspirin |
| Joshua × ibuprofen (control) | — | 2 chips (NSAID cross-reactivity + lisinopril Moderate) |
| Mary × paracetamol (control) | — | 0 chips (×2 runs) |

Every probe is discriminating, and each absence was re-run rather than trusted once:

- **Kevin × levothyroxine** is the defect and the fix. Note what the pre-fix answer did: the LLM
  refused anyway — *"No — Levothyroxine should not be given: the patient has an allergy to
  Levothyroxine [256]"* — because the allergy record is in the chart and it happened to read it. That
  is exactly the dependence #110 exists to remove: the deterministic net was silent and the answer's
  correctness was the model's own doing, unreproducible by construction on this box.
- **Kevin × vasopressin** and **Joshua × levothyroxine** are the no-false-positive direction. Both
  would show a fabricated "recorded allergy to …" chip if the fix had dropped the identity test
  instead of moving it, and Joshua additionally proves an unrelated allergen (aspirin) raises nothing
  once the loop is reachable for an ATC-less drug — the guard's own contract, live.
- **Mark × ciprofloxacin** would show TWO contraindication chips if identity no longer pre-empted the
  class arm; it shows one.
- **Mark × levofloxacin** would show none if the guard had been moved too far or dropped the class
  arms; it shows the J01MA cross-reactivity chip.
- **Joshua × ibuprofen** covers the third precedence level (the curated NSAID group), which no unit
  case in this PR reaches on the live KB.

The verified omod was built from `9c532d2`; the only commit after it (`b2a976c`) adds one
`assertNotNull` to a test, so the production code probed above is byte-identical to this branch's.

## Found while doing this, not fixed here

- `PatientClinicalContext.hasAllergyToken`/`hasConditionToken` still compare with raw
  `String.contains`, with no diacritic fold — the one matcher #138 did not reach. With the DEFAULT
  `sourceFormat=json`, Amoxicillin's curated `{"type":"allergy","token":"penicillin"}` does not match
  an allergen recorded against `Allergie à la pénicilline`, `Pénicilline G` or `Pénicillines` — all
  real **fr locale-preferred** names in this dictionary, and the first of them is the canonical
  allergy concept. Measured against the running standalone: 11 locale-preferred fr names across the
  four curated allergy tokens fail raw containment and match after folding (`paracetamol` ×6,
  `penicillin` ×6). Folding alone is sufficient (containment tolerates the trailing `-e`/`-s`), it is a
  pure relaxation, and it does not disturb the fragment matching the javadoc keeps bare containment
  for. Filing separately.


dkayiwa:
Two notes from the live runs that belong on the record but are **not** this PR's to fix — both
reproduce identically on `main` @ `0d3a906` and on this branch, so neither is caused by the change.

**1. The answer altered an ATC code it was citing.** On `Mark Smith × levofloxacin`, in both runs
(two different JVM lifetimes, two different builds), the chip read

> Levofloxacin is in the same ATC class (**J01MA**) as the patient's allergy to Ciprofloxacin — possible cross-reactivity

while the answer, citing that finding record, read

> Levofloxacin is in the same ATC class (**J01CA**) as the patient's allergy to Ciprofloxacin … [25]

J01CA is penicillins; J01MA is fluoroquinolones. Two characters, inside a code, in a sentence whose
citation points at the record that states it correctly — and a cosine grounding pass cannot see a
change that small. Since #110 the finding is injected verbatim precisely so the chip and the prose
cannot disagree; that guarantees the record's content, not the model's copy of it. Worth its own
issue (a code-literal check in the grounding verifier, or prompt wording), with the caveat that
answer prose is not reproducible on this box — this is n=2 on one question, not a rate.

**2. Demo data I created and left in place** (per the "demo data is disposable" standing note), so a
later reader can tell my fixture from stock demo data. Both carry the comment
`issue #135 live fixture (agent-created demo data)`:

| patient | uuid | allergy |
|---|---|---|
| Kevin Brown | `e89247a2-111f-4dd5-974d-25bfad03690a` | DRUG → Levothyroxine (`e2bd5560-0af2-403b-b29a-9ead2147baad`) |
| Mark Smith | `de4b0d62-8a47-4c82-8220-1f0a87eafd46` | DRUG → Ciprofloxacin (`84e5fc82-da58-4b62-8f1d-ce3dc62e8ffa`) |

No GPs were changed and no orders were created or voided. Before this, only Betty Williams and
Joshua Johnson carried any allergy at all (both `Aspirin`) among the 21 demo patients surveyed —
which is why the shape #135 describes had never been probed.

dkayiwa:
## Hardening pass (`/harden`) — 6 Phase 1 passes, 3 Phase 2 passes

**No production logic changed.** Every hunk this pass added under `api/src/main` is inside a javadoc
block. Proven rather than asserted: all **94** api classes in the omod probed on the standalone
disassemble identically to the merge candidate (`javap -p -c -constants`); the two `.class` files that
differ at byte level differ only in `LineNumberTable`, shifted by the javadoc.

`mvn -o clean install` from the repo root — **API 823 + OMOD 58 = 881**, 0 failures, 34 skipped.
Baseline re-measured on `main` @ `32bb64a` in a clean worktree: **API 813 + OMOD 58 = 871**. So +10:
+8 for this PR's own cases, +2 for the two added below. (This confirms the PR body's figures; #139
added no tests.)

### The one thing that was actually unpinned

The guard's placement **inside** the per-allergen loop is the whole of this fix, and nothing tested it.
Measured: change its `continue` to `return` — still after the identity check, so it reads as correct —
and **all 821 tests stay green** while #135 is reinstated for any chart with more than one recorded
drug allergy, whenever the allergy to the asked-about drug is not the first token.

`anEarlierUnrelatedAllergenDoesNotHideTheDirectOne` now fails on exactly that mutation, and on the
pre-fix build. Mutation table, all run here:

| mutation | before | after |
|---|---|---|
| guard's `continue` → `return` | 0 failures / 821 | **1 failure** |
| pre-fix guard (hoisted before the loop) | 3 failures | **5 failures** |
| fixture's Leucovorin given an ATC code | 0 failures | **1 failure** |
| guard deleted outright | 0 failures | 0 failures — it is provably inert, as its own comment says |

Also added: `anUnclassifiedAllergenWithASiblingRouteVariantStillWarnsOnce`. DDInter files one substance
as several rows sharing an `rxnorm_name` (142 such groups in the full KB, 28 entirely ATC-less), so one
question puts a sibling row in play as a *different* reference which must fall through the guard rather
than add a chip. 0 chips before the fix, 1 after. It reuses the existing verbatim route-variant slice —
no new fixture.

### Statements that were wrong, corrected rather than softened

- The fixture precondition asserted only `Ledipasvir` while two cases use `Leucovorin`. Giving
  Leucovorin an ATC code (`V03AF03`, calcium folinate — a real assignment a refresh could add) left
  every case green, testing the **classified** path under names saying "unclassified".
- `Ciprofloxacin`/`Levofloxacin` share **two** level-4 subgroups (`J01MA` *and* `S01AE`), not one, so
  asserting `(J01MA)` pins the dataset's code order as well as the match.
- `aClassifiedAllergenRaisesNothingForAnUnclassifiedDrug` claimed to pin "the guard's own contract";
  with the guard deleted it still passes. It pins the requirement, not the statement.
- `duplicateAllergyAliasesProduceASingleContraindication` and
  `classContraindicationForRecordedAllergyToTheNamedDrug` both said the **class** layer fires. Measured:
  it is the **identity** comparison in both — which is the distinction this PR turns on.
- The shared loader kept a private one-line alias in the file it was extracted from, whose javadoc said
  "the five test files that need one". Six files were rewired (seven call sites) and seven callers use
  it now. The alias is gone and the count with it; what the count was decorating — three of the replaced
  copies fed `getResourceAsStream`'s result straight to the parser, so a missing fixture failed with
  Jackson's `IllegalArgumentException: argument "in" is null` — is recorded on the loader itself.
- `SafetyWarning.TYPE_CONTRAINDICATION` said "a drug named in the answer", while its sibling
  INTERACTION/OVERDOSE javadocs already say "in play". This PR's headline case is a drug named **only**
  in the question.
- `PatientClinicalContext.containsToken` still called it "the class-based allergy arm" and cited
  `lookupByToken`'s boundary-awareness as the clean contrast confining nesting risk to hand-authored
  rules. Boundary-aware is not correct (see the bound below).
- README, ADR Decision 24 and the `warnOnContraindications` GP description all filed the direct-allergy
  comparison under class-level matching — the conflation that made a classification guard in front of
  the whole method look right. All three now separate identity from classification.
- README said the demote-only grounding guarantee "is keyed on the `drug_reference` resource type
  rather than on `group`". Since #122 it is keyed on `group`, so `safety_finding` is demote-only too. A
  client following the old text renders "citation not supported" against a correct safety finding.

### A coverage bound worth reading before merge

The javadoc said 444 ATC-less entries were being skipped, which invites reading the fix as 444 covered
drugs. Measured over the full 19 MB KB through the real validator — asking about each of the 444 with an
allergy recorded under that entry's own name — every one now raises a contraindication, but **53 name a
different entry**, because `lookupByToken` returns the earliest entry any of whose aliases occurs as a
whole word in the token. 43 of the 53 match on a *fragment* (`Loteprednol etabonate` → `Loteprednol
(ophthalmic)`); 10 because another entry carries the queried drug's full name among its CIEL aliases
(`Moderna covid-19 vaccine` → `Pfizer-BioNTech Covid-19 Vaccine`), which no alias matcher can separate.
206 of all 2283 entries do not resolve to themselves, so this reaches the class comparisons identically
— a resolver defect, not a placement one, and filed separately. The refusal is still the right one; the
substance named is not the charted allergen.

### Load-bearing decisions I was tempted to touch and did not

- **The in-loop guard.** It is provably inert (guard deleted → 0 failures), so a reuse pass flagged the
  duplicated classification preamble it shares with `classRelationships` as extractable. Left alone: its
  `continue` semantics are exactly what the new test measures, and the two copies now differ on purpose.
- **Identity staying in `addAllergyContraindications`.** Two reviews independently found a real duplicate
  chip on the default `json` source (curated self-named rule + identity, measured 2 chips for one fact on
  Ibuprofen/Paracetamol/Amoxicillin — pre-existing on `main`). The obvious fix is cross-arm suppression,
  which is exactly the split this PR argues against. Filed separately instead.
- **`classContraindicationForRecordedAllergyToTheNamedDrug`'s name** is the last "class" misnomer for the
  identity check. Not renamed: a test identifier is spec surface, and the comment now carries the
  correction.
- **The `444 / 19.4% / ATC-prefix` paragraph appearing in three javadocs.** Duplicated prose by any normal
  reading; kept, because the three audiences differ and the numbers are the argument.

### Live re-verification

Required — this PR changes production behaviour. Three full deploy/restart/probe cycles on the 3.7.1
standalone (`sourceFormat=ddinter`, full 19 MB KB, `minInteractionSeverity=minor`, `chartMode=fullChart`,
`llm.engine=local`), each gated on querystore `hits>0` and a known-positive control chip. I measured the
**before** state first-hand on `main` @ `32bb64a` rather than taking it from the PR body.

| probe | discriminating direction | `main` @ `32bb64a` | this branch |
|---|---|---|---|
| Kevin Brown × levothyroxine (allergic, ATC-less) | the defect | **0 chips, 0 findings** (×2) | **1 contraindication + 1 `safety_finding`** (×3) |
| Susan Young × tiotropium (allergic, ATC-less, **and actively ordered it**) | the defect, on a shape never probed | **0 chips** (×2) | **1 contraindication**, 0 interaction chips (×2) |
| Kevin × levothyroxine after a **second** allergy resolving to the same entry | `seenAllergens`, live — the PR could only unit-test this | — | still **exactly 1** |
| Kevin × vasopressin (ATC-less, *not* allergic) | fabricated chip if identity had been dropped rather than moved | 0 | 0 (×3) |
| Joshua × levothyroxine (aspirin allergy, ATC-less subject) | the guard's own contract, live | — | 0 (×2) |
| Mark Smith × ciprofloxacin (`J01MA02`) | **two** chips if identity no longer pre-empted the class arm | 1 contra + 1 Major | identical (×2) |
| Mark × levofloxacin (`J01MA12`) | none if the guard had moved too far | — | `J01MA` class chip + 1 Major |
| Betty Williams × aspirin (classified, **0 orders**) | classified identity, isolated from interactions | 1 contra | 1 contra |
| Betty × ibuprofen | the third precedence level (curated NSAID group), isolated | — | 1 cross-reactivity chip |
| Mary × clarithromycin / Agnes × warfarin / Joshua × ibuprofen / Mary × paracetamol | the four regression controls | — | 1 Major / 1 Major / 2 chips / 0 (×3) |

**Demo data created** (agent-created, left in place — the standalone is a regenerable test bed):
Susan Young `763e6e5f…` DRUG allergy → **Tiotropium** (`6b4652e6-c8d4-49ce-9096-e93693893cd2`), giving
the corpus its first patient allergic to an ATC-less drug she is *also* prescribed; and a **second**
allergy on Kevin Brown `e89247a2…` → **Levothyroxine sodium** (`42127cf5-c81e-45f0-90c3-d9f0244c7a3a`),
a CIEL alias of the same ATC-less entry, so Kevin is now the corpus's only live two-allergy /
one-allergen fixture. **Kevin therefore has two allergy rows, not one** — worth knowing before reading a
future probe of his as single-allergen. No global property was changed.

dkayiwa:
### Addendum — two more Phase 2 passes after the summary above

The summary above described the branch at `1e66d9d`. Two further passes ran, because an independent
fact-check of the coverage-bound paragraph found that **four statements in it still did not hold** —
all in prose the hardening itself had written, which is the honest reason to record them rather than
quietly amend. Each was re-measured here through the real parser and resolver over the full 19 MB KB
before rewriting. Final HEAD `0ee4cd4`; **API 823 + OMOD 58 = 881**, 0 failures, 34 skipped; CI green on
Java 11 / 17 / 21.

- **"byte-identical to their KB rows"** (said of the route-variant slice in two places) is false. The
  rows are field-for-field equal *including key order*, but the KB stores each row on one compact line
  while the slice is pretty-printed, so the raw text differs. The word this had replaced — "verbatim" —
  was accurate. Now "field-for-field", with the reason given.
- **The 10 whole-name cases were attributed entirely to CIEL.** Measured per row: 5 carry the queried
  name only as a CIEL name, 2 in both fields, and **3 only as their `rxnorm_name`** (`Tick-borne
  encephalitis vaccine (whole virus, inactivated)`, `Dotatate`, and the Influenza A H5N1 antigen row).
  So 7 of 10, not 10 of 10 — the attribution is now to the dataset's alias data with the split stated,
  and `Dotatate` → `Lutetium Lu 177 dotatate` added so the `rxnorm_name` shape is illustrated too.
- **"What is shorter is the matched ALIAS" did not describe the split.** In the 10, the matched alias
  *is* the full queried name, and in 4 of the 10 a shorter alias matches as well — so the two bullets
  were never a partition on which alias won the scan. What does partition them is whether the target
  entry claims the queried name among its own aliases, which is also the property deciding whether any
  matcher could separate them. Stated that way now.
- **The README grounding clause did not follow from its own antecedent.** "A client should suppress
  passing verdicts by `group` … or it will render 'citation not supported'" — the withholding is
  server-side, so a `reference`-group citation never arrives as `grounded: true` however the client is
  written, and that label is not this project's (README tells clients to render `false`-or-`null` as
  *unverified*). The real hazard is the withheld pass arriving as `null`.

One figure in that paragraph had been taken from a reviewer rather than run here; it is now measured
directly — a longest-matching-alias resolver over the full KB fixes **43 of 43** in the fragment bucket
and **0 of 10** in the whole-name bucket, which is what makes "43 fixable, 10 not" a claim rather than an
assertion.

Nothing in the live evidence changes: all production hunks remain javadoc-only, and the 94 api classes
in the probed omod still disassemble identically to this tree.

PR #144, opened 2026-08-05
fix(drug-safety): check the patient's active orders against their own allergies

### LINKED ISSUE #143: Echo scoping withholds a contraindication for a drug the patient is prescribed — its justification claims coverage no arm provides
### Summary

A patient **prescribed** a drug they are **allergic to** can get no contraindication chip, because echo scoping removes that drug from the validated set on the strength of a justification that is false.

Measured on the **default** `sourceFormat=json`: an active ibuprofen order plus an ibuprofen allergy, a question naming no drug, and an answer citing the real `drug_order` record → **0 chips**. The identical call with `mappings=null` (echo scoping disabled) → **2 chips**.

### The false premise

`DrugSafetyValidator.isEchoOfCitedRecord` (~`:449-456`) exempts a drug the answer names when a cited record already names it, and justifies the residual risk like this:

> *"a proposal-worthy X is usually question-named (always validated) or actively ordered **(checked directly by the order-driven arms)**, so the residual shape is rare"*

**No arm checks active orders against allergies.** Verified by inspection of every arm in that class:

| arm | reads allergy tokens | reads active orders |
|---|---|---|
| `addContraindications` | 1 | 0 |
| `addClassContraindications` / `addAllergyContraindications` | 1 | 0 (incidental only) |
| `addInteractionWarnings` | **0** | 4 |
| `addQuestionPairInteractions` | **0** | 6 |
| `addActiveOrderPairInteractions` | **0** | 9 |
| `addOverdose` | 0 | 0 |

The order-driven arms check **interactions**. The contraindication arms check the drug **in play**. So when echo scoping removes an actively-ordered drug from `inPlay`, nothing checks it against the allergy list — the coverage the comment relies on does not exist.

### Why the shape is not rare

The comment's "rare" rests on the same false premise. The triggering conditions are ordinary:

1. the patient is prescribed the drug — so it appears in a `drug_order` chart record;
2. the answer cites that record, which is exactly what a good answer does when asked about medications;
3. the drug is therefore treated as an echo rather than a proposal.

An allergy to a currently-prescribed drug is a prescribing error the chart already contains — precisely the case a safety net should catch loudest.

### Provenance and scope

**Pre-existing**, from #105's echo scoping — not introduced by #140. But #140 extends its reach: before #140, an ATC-less drug had no identity check at all (#135), so this suppression was invisible for those 444 entries; now the check exists and echo scoping is what withholds it.

Note #105's own measured justification stands on its own terms — it was fixing 7 of 8 chips being about unproposed drugs on one enumeration answer, which was real. The defect is not that scoping exists; it is that its escape hatch for actively-ordered drugs was never built.

### Proposed direction

The cleanest fix follows the comment's own intent: make the claim true by checking active orders against allergies, rather than by weakening the scoping. That is a genuinely missing arm — "is the patient allergic to something they are taking?" is not asked anywhere today — and it would also cover the case where the drug is never mentioned in question or answer at all, which no amount of scoping tuning reaches.

Alternatively, exempt contraindications from echo scoping while leaving interactions scoped: the #105 measurement that motivated scoping was about interaction chips on unproposed drugs, and a contraindication for a drug in the patient's own chart is not the same over-reach.

Do not simply disable scoping — that reinstates what #105 measured and fixed.

### Tests

Through the real production path with `sourceFormat=json`: an active order plus an allergy for the same drug, a question naming no drug, an answer citing the order record, must raise the contraindication. Mutation-check by restoring the exemption. Keep #105's regression covered — an answer that merely recites a cited reference partner must still not chip it as a proposal.

Found while hardening #140; not fixed there because it is a different mechanism and #105's contract needs its own argument.

dkayiwa:
Fixes #143.

A patient **prescribed** a drug they are **allergic to** could get no contraindication chip at all.

Every contraindication arm was keyed on a drug **in play** — one the question resolved, or one the
answer named. Echo scoping (#105) removes an answer-named drug from that set when a record the answer
cites already names it, and a drug the patient is prescribed appears in a `drug_order` chart record,
which is exactly the record a good answer cites when asked about medications. So the scoping fired on
the one shape where the finding matters most: an allergy to a currently-prescribed drug is a
prescribing error the chart already contains, and since #110 turns every chip into a citable
pre-answer record it was missing from the prompt as well as from the chip strip.

## The false premise

`isEchoOfCitedRecord` justified the residual risk like this:

> *a proposal-worthy X is usually question-named (always validated) or actively ordered **(checked
> directly by the order-driven arms)**, so the residual shape is rare*

The second half was false. Re-counted here on `main` @ `9b93bbc` rather than taken from the issue —
comments stripped, and both per-method-body and transitively over every in-class helper each arm
reaches, because the arms touch active-order state mostly through shared helpers (`bestRulePerPartner`,
`classRelationships`, `coveredByActiveOrderArm` → `hasActiveDrug`):

| arm | reads allergy state (direct / transitive) | reads active-order state (direct / transitive) |
|---|---|---|
| `addContraindications` | 1 / 1 | 0 / 0 |
| `addAllergyContraindications` | 1 / 1 | 0 / 0 |
| `addInteractionWarnings` | **0 / 0** | 0 / 2 |
| `addQuestionPairInteractions` | **0 / 0** | 0 / 1 |
| `addActiveOrderPairInteractions` | **0 / 0** | 2 / 14 |
| `addOverdose` | 0 / 0 | 0 / 0 |

*(`allergy state` = `getAllergyTokens` \| `hasAllergyToken`; `active-order state` = `hasActiveDrug` \|
`getActiveDrugNames` \| `getActiveDrugAtcCodes` \| `getActiveDrugOrders` \| `findByActiveOrders` \|
`activeOrdersOtherThan` \| `ActiveDrugOrder`. The issue's own table gives larger active-order figures;
those are not reproducible by either counting method here, so the numbers above are mine. The
load-bearing column — allergy reads by the order-driven arms — is **zero under both**, which is the
claim the fix rests on.)*

Nothing anywhere asked "is this patient allergic to something they are taking?" After this change,
`addActiveOrderContraindications` is the only arm that reads both (1 / 3 allergy, 1 / 4 active-order).

## The direction chosen, and why

**The missing arm, not a scoping carve-out.** #143 offered both; this is the first, and the argument is
that the second fixes the measured case and nothing past it.

A "contraindications are exempt from echo scoping" carve-out still needs the **answer** to name the
drug. A prescribing error nobody happened to write down stays invisible — and that is not a corner:
`DrugReferenceInjector.preAnswerFindings` calls `validate` with an **empty** answer, so on the path
that puts findings into the prompt there is no answer text to name anything at all. Measured live
below: Susan Young, prescribed Tiotropium and allergic to Tiotropium, asked *"What is her most recent
blood pressure?"* — a carve-out reaches 0 of those.

A carve-out would also widen #105's own over-reach onto the contraindication surface: a drug the answer
merely *recites* out of a cited allergy or reference record would become contraindication-checked,
chipping about a drug nobody proposed giving. Keyed on the patient's own active orders, neither
happens — and the claim `isEchoOfCitedRecord` makes about actively-ordered drugs becomes **true**
rather than being relaxed. #105's measurement (7 of 8 chips about unproposed drugs on one enumeration
answer) is left standing on its own terms; the scoping is untouched.

## What the arm is

`addActiveOrderContraindications` runs the two **existing** contraindication arms over
`activeOrderEntries(context)` — the same "reference entries for the patient's active orders" definition
`addActiveOrderPairInteractions` already uses, so the two arms cannot come to disagree about which
prescriptions the reference data covers. No new safety rule, no new matcher, no second vocabulary.

- **Skips anything already in `inPlay`**, so a prescribed drug the question *also* names is checked
  once, not twice — identity is the right test because both sets resolve against the service's shared
  `getAll()` cache.
- **Contraindications only.** Not interactions: `addInteractionWarnings` over an active-order entry
  against the patient's own orders *is* `addActiveOrderPairInteractions`, which is deliberately gated
  on the question asking to be screened (#113) and would here run ungated, uncapped, and without that
  arm's `activeOrdersOtherThan` reduction — so one order would witness a pair with itself (#86's
  `iron` inside `spironolactone`). Not overdose either: `addOverdose` reads a dose out of the *answer*.
- **Not gated on the question**, deliberately: gating a contraindication on the question's wording is
  what produced this defect. The bound is the chart instead — it can only fire where an allergy or
  condition record and an active order point at the same drug.
- **Precondition:** with neither allergy nor condition tokens recorded both arms are *provably* no-ops,
  so the arm returns before resolving any order subjects. That is why every question about every
  patient without an allergy record pays one emptiness check rather than an alias sweep of the dataset
  per order name. Both token sets are read, not only allergies — the curated arm's condition leg is
  half of what the scoping was suppressing.
- Nothing downstream double-counts: the screening arm seeds its suppression set from the chips raised
  so far, and these are `TYPE_CONTRAINDICATION` while every chip that arm can raise is
  `TYPE_INTERACTION`.

`isEchoOfCitedRecord`'s javadoc now says what is true, and records what #143 found.

## Tests

`ActiveOrderContraindicationTest`, 8 cases on the real bundled curated dataset (the production default
`sourceFormat=json`, whose ibuprofen entry carries both a curated allergy rule and an
identity-resolvable name), real querystore-shaped `drug_order` chart records, and the real 4-argument
`validate` production calls with the chart's mappings. **Four fail on the pre-fix build** (the mutation
check is `main` itself):

- the headline shape — an active ibuprofen order, an ibuprofen allergy, a question naming no drug and
  an answer citing the `drug_order` record: **0 → 2**;
- the same with the drug named **nowhere**, in question or answer — the half no scoping tweak reaches;
- the curated arm's **condition** leg (on ibuprofen with an active peptic ulcer), which a
  guard reading only allergy tokens would have dropped;
- the finding reaching the prompt as a citable `safety_finding` record through the real injector wired
  to the real validator — the #110 half, on the empty-answer pre-answer pass.

Four pin what must not change:

- a prescribed drug the question **also** names is checked once, not twice (2 chips, not 4) — the
  composition with #140's identity arm;
- an unrelated allergy **and** an unrelated condition on a patient with an active order raise nothing,
  with the arm's body genuinely executing rather than short-circuiting;
- with neither recorded, the arm is inert — a patient gains no chip from being on a drug;
- **#105's contract on this surface**: an answer reciting `lisinopril` out of the cited reference record
  for a patient allergic to lisinopril but *not taking it* gains no contraindication check. This is the
  case that separates the two directions — a carve-out fails it.

`mvn -o clean install` from the repo root: **API 831 + OMOD 58 = 889**, 0 failures, 34 skipped.
Baseline measured on `main` at `9b93bbc` in the same worktree: **API 823 + OMOD 58 = 881**, 0 failures,
34 skipped. (+8 = the new cases.)

## Live: before and after on the 3.7.1 standalone

Two full deploy/restart/probe cycles against the same fixtures, `main` @ `9b93bbc` first and this
branch second, with the full 19 MB KB, `sourceFormat=ddinter`, `minInteractionSeverity=minor`,
`chartMode=fullChart`, `llm.engine=local`. Note that the issue's own measurement was on `json`; the
shipped configuration is `ddinter`, whose entries carry no curated `contraindications` at all, so
`addContraindications` is inert there and what fires live is the allergen arm — its **identity** leg
(#140) and its **curated cross-reactivity group** leg (Decision 27). That is the stronger case, because
it is the configuration that ships; the curated-rule leg is covered by the unit tests on the `json`
default.

GPs were left exactly as found (nothing flipped). **Fixture created for this** (agent-created demo
data, left in place — the standalone is a regenerable test bed):

| patient | uuid | created | already had |
|---|---|---|---|
| Barbara Miller | `e30bc8f0-08bb-406c-986a-2b153a495603` | DRUG allergy → **Ibuprofen** | Aspirin 81mg + Advil 400mg orders |

Barbara was chosen because she is **not** one of the four regression controls: giving Agnes or Joshua a
new allergy would change a control's own expected chip count.

| probe | before (`main` @ `9b93bbc`) | after (this branch) |
|---|---|---|
| **Susan × "What are her current medications?"** — prescribed Tiotropium AND allergic to it; the answer cites the `drug_order` record | **0 chips** | **1 chip** — 1 contraindication |
| &nbsp;&nbsp;↳ re-run | **0 chips** | **1 chip** — 1 contraindication |
| **Susan × "What is her most recent blood pressure?"** — the drug named nowhere | **0 chips** | **1 chip** — 1 contraindication |
| &nbsp;&nbsp;↳ re-run | **0 chips** | **1 chip** — 1 contraindication |
| Susan × "Is it safe to give her tiotropium?" — question-named, never scoped out | **1 chip** — 1 contraindication | **1 chip** — 1 contraindication |
| Susan × "What drug interactions does tiotropium have?" — recited partners (#105) | **1 chip** — 1 contraindication | **1 chip** — 1 contraindication |
| **Barbara × "What are her current medications?"** — Ibuprofen allergy, on Aspirin 81mg + Advil 400mg | **0 chips** | **2 chips** — 2 contraindications |
| &nbsp;&nbsp;↳ re-run | **0 chips** | **2 chips** — 2 contraindications |
| Mark × "What are her current medications?" — Cipro allergy, on Chloroquine + Benadryl (**not** allergic to either) | **0 chips** | **0 chips** |
| &nbsp;&nbsp;↳ re-run | **0 chips** | **0 chips** |
| Kevin × "What are her current medications?" — Levothyroxine allergy, **no orders** | **0 chips** | **0 chips** |
| control — Mary × clarithromycin | **1 chip** — 1 interaction | **1 chip** — 1 interaction |
| control — Agnes × warfarin | **1 chip** — 1 interaction | **1 chip** — 1 interaction |
| control — Joshua × ibuprofen | **2 chips** — 1 contraindication, 1 interaction | **2 chips** — 1 contraindication, 1 interaction |
| control — Mary × paracetamol | **0 chips** | **0 chips** |
| &nbsp;&nbsp;↳ re-run | **0 chips** | **0 chips** |

Across the 16 probes, **exactly 6 changed** — `susan-meds` ×2 (0 → 1), `susan-bp` ×2 (0 → 1),
`barbara-meds` ×2 (0 → 2) — and every one of them is a case the arm exists for. The other **10 are
unchanged**, including all four controls.

Every absence was re-run rather than trusted once (the cold-start KB race), and every probe
discriminates — what each would have shown had the fix been wrong:

- **Susan × "What are her current medications?"** is the defect and the fix, in the shipped
  configuration. She is prescribed Tiotropium *and* allergic to it; the answer cites the `drug_order`
  record and names the drug out of it, so echo scoping takes it out of play and the question puts
  nothing back. A wrong fix leaves this at 0.
- **Susan × "What is her most recent blood pressure?"** is the half a scoping carve-out cannot reach:
  neither the question nor the answer names the drug. A carve-out scores 0 here while looking like it
  works on the probe above — which is the whole argument for adding the arm.
- **Susan × "Is it safe to give her tiotropium?"** must stay at **1**, not become 2. The question names
  the drug, so the drug-in-play loop already chips it; an arm without the `inPlay` skip would double
  every such chip. This is the composition check with #140's identity arm.
- **Susan × "What drug interactions does tiotropium have?"** is #105's shape: the answer recites
  interaction partners out of the cited tiotropium reference record. `inter=0` on both trees — a
  recited partner still does not chip as a proposal, and the change did not widen the scoping.
- **Barbara × "What are her current medications?"** exercises the two *class* legs of the arm across
  **two different orders in one query**, again with the question naming no drug — 0 → 2:

  ```
  [contraindication] Ibuprofen :: The patient has a recorded allergy to Ibuprofen.
  [contraindication] Acetylsalicylic acid (aspirin) :: Acetylsalicylic acid (aspirin) is in the same
      cross-reactivity group (NSAID) as the patient's allergy to Ibuprofen — possible cross-reactivity
  ```

  The first is identity, off her **Advil 400mg** order — resolved through the order's CONCEPT name, not
  its drug name: the KB carries no `advil` alias anywhere (checked), and `PatientClinicalContextBuilder`
  puts both the drug name and the concept name (`Ibuprofen`) on the context, which is what
  `activeOrderEntries` resolves against. The second is the curated cross-reactivity group, off her
  **Aspirin 81mg** order, which no ATC subgroup relates to ibuprofen — the cross-branch case Decision
  27's curated groups exist for. Both re-run identically. An arm that only asked the identity question
  would show 1; one resolving orders by drug name alone would also show 1.
- **Mark × "What are her current medications?"** is *prescribed but not allergic*: a Ciprofloxacin
  allergy (`J01MA02`) against Chloroquine (`P01BA01`) and Benadryl/diphenhydramine (`R06AA02`), neither
  related by subgroup or by curated group. An arm that fired on "this patient has an allergy AND has
  orders" without comparing them would chip here.
- **Kevin × "What are her current medications?"** is *allergy but no orders*: the arm has no subject at
  all. An arm keyed on the allergy list rather than on the order list would chip here.
- **The four controls** are unchanged — and not merely in count: diffing the two captures shows the
  rendered chip strings **byte-identical** before and after for all four, and for `susan-tio` and
  `susan-ddi` as well, mechanism paragraphs included.

**The prompt half, live.** For the headline probe the standalone log shows the validator raising a
warning **twice** per query, once on each side of the answer — `08:26:17,088` (30 ms after
`querystoreBuild patient=28`, i.e. the pre-answer findings pass inside `injectRecords`) and
`08:26:49,797`, immediately before `[timing] search patient=28 … llmMs=30765`. So the finding is
injected as a citable `safety_finding` record as well as chipped, which is what #110 exists for. (The
`safety_finding_records` column of the raw captures counts records the ANSWER CITED, not records
injected, so it is not the evidence for injection — the log pairing and the unit test are.)

The build probed was this branch at `542f4d0`; the only later change is a source comment, and
`javap -p -c -constants` reports IDENTICAL INSTRUCTIONS for `DrugSafetyValidator` and
`DrugReferenceInjector` between the omod the box ran and the final HEAD, so the capture applies to the
merge candidate.

### 1. DEFECT — one recorded allergy raises N contraindication chips for a substance DDInter files as several route variants

`addAllergyContraindications` has no route-variant collapse. `lookupByToken` resolves the allergy to
**one** entry, which gets the identity chip; each sibling variant is a *different* `DrugReference`, and
every variant of a substance publishes the same ATC list, so each sibling reaches `sharedClass` and adds
its own "is in the same ATC class … as the patient's allergy to X — possible cross-reactivity" chip. The
same substance is reported as cross-reactive with itself.

Measured through the real production path over the full 19 MB KB the standalone loads, on `main` @
`9b93bbc` **and identically on this branch** (so it is pre-existing, and this PR does not amplify the
per-question count):

| allergy, asked about the same drug | entries | chips |
|---|---|---|
| `Dexamethasone` | 4 | 4 — 1 identity + `Dexamethasone (nasal\|ophthalmic\|topical) is in the same ATC class (A01AC) as the patient's allergy to Dexamethasone` |
| `Hydrocortisone` | 4 | 4 |
| `Penicillin G` | 4 | 4 (`Benzylpenicillin (potassium)`, `Benzylpenicillin (sodium)`, `Procaine benzylpenicillin`) |
| `Diphenhydramine` | 2 | 2 |
| `Tiotropium` | 1 | 1 (single-entry family — the control) |

Reachability: **114 of the 142** multi-entry `rxnorm_name` families in the shipped KB have ≥2 members
sharing a level-4 subgroup (152 surplus chips summed over all families). Reachable today whenever the
question names the drug; this PR extends the *reachability* — not the count — to "the patient is on it",
so for such a patient it now fires on every question.

Two aggravating details worth putting in the issue: `sharedClass` returns the **first** of the
allergen's subgroups that matches, which for dexamethasone and hydrocortisone is `A01AC`
("corticosteroids for local oral treatment") — an arbitrary and clinically odd class to name for a
systemic-versus-nasal pair. And #140's
`DirectAllergyContraindicationTest.anUnclassifiedAllergenWithASiblingRouteVariantStillWarnsOnce` covers
only the **ATC-less** families (28 of the 142), where the siblings fall through the in-loop
classification guard — the 114 ATC-carrying families are exactly the uncovered ones.

The interaction arms already solved the analogous problem twice (`bestRulePerPartner` for #115,
`pairKeyNames` for the pair arms). The contraindication arms have no equivalent. Not fixed here: it is a
different mechanism, it predates this change, and it needs a decision about which entry a collapsed chip
should name.

### 2. DEFECT — the curated allergy-rule arm and the identity arm both fire for one (drug, allergy) fact, on the dataset the module ships and defaults to

`addContraindications` (curated `contraindications` of `type=allergy` matched by token) and
`addAllergyContraindications` (identity) are independent, so a curated rule whose token equals the
drug's own name yields a second chip stating the same fact in different words. Measured through the real
production path on the bundled default `sourceFormat=json` — the plain question path, nothing to do with
this PR's arm:

```
allergy=ibuprofen,   asked about ibuprofen   -> 2 chips
   Ibuprofen is contraindicated by an active allergy: documented ibuprofen allergy
   The patient has a recorded allergy to Ibuprofen.
allergy=paracetamol, asked about paracetamol -> 2 chips  (same shape)
allergy=amoxicillin, asked about amoxicillin -> 2 chips  (same shape)
allergy=gentamicin,  asked about gentamicin  -> 1 chip   (no self-named curated rule)
```

**3 of the 4 entries** in `drug-reference.json` carry such a rule, so this is out-of-the-box behaviour on
the default configuration. It is also what #143's own stated expectation of "2 chips" is made of, so
`ActiveOrderContraindicationTest`'s headline assertion pins the issue's number rather than a judgement
that two chips are right. Whether it *should* collapse to one chip is a clinical-presentation call — I
would argue yes, one fact one chip — and deciding which arm yields spans the precedence #135/#140
settled, which is why it is not touched here.

### 3. HAZARD — `activeOrderEntries` is now on the hot path for every question about an allergy-bearing patient

It performs one full-dataset alias sweep (`findByQuery`) per active-order name. The arm's precondition
keeps it off the path entirely for a patient with no allergy and no condition record, and the cost is
microseconds against a 15–90 s answer, so this is not a defect today. Worth recording because it moves
where a future dataset-size regression would surface: on `main` that sweep ran only for
interaction-screening questions.

### 4. STALE DOC (already recorded in code, listed for completeness)

`DrugReferenceInjector.orderedInteractionNotes` still coalesces its own partner label untrimmed, so a
padded curated token reaches the citable record as `"warfarin   (Major. …)"` beside a chip reading
`"active order warfarin"`. `DrugSafetyValidator.partnerLabel`'s javadoc already names this as the
follow-up; nothing new found, and out of this slice.


dkayiwa:
Live before/after evidence is now in the PR body — two full deploy/restart/probe cycles on the 3.7.1
standalone (`main` @ `9b93bbc`, then this branch), 16 probes each, on the shipped `sourceFormat=ddinter`
configuration with the full 19 MB KB.

Headline, on **Susan Young** (prescribed Tiotropium *and* allergic to it), question naming no drug:
**0 chips → 1 contraindication**, reproduced twice on each tree. The same patient asked *"What is her
most recent blood pressure?"* — with the drug named nowhere at all — also goes 0 → 1, which is the case
a "contraindications are exempt from echo scoping" carve-out cannot reach and the reason this PR adds
the arm instead. Asked *about* tiotropium by name it stays at **1, not 2**: the `inPlay` skip composing
with #140's identity arm. All four regression controls are unchanged, chip for chip.

Two discriminating negatives: **Mark Smith** (Ciprofloxacin allergy, prescribed Chloroquine and
Benadryl — related to neither) stays at 0, and **Kevin Brown** (Levothyroxine allergy, no orders) stays
at 0. And #105's own shape — an answer reciting interaction partners out of the cited tiotropium
reference record — still raises `inter=0` on both trees.

One fixture was created for this and left in place: an **Ibuprofen** DRUG allergy on **Barbara Miller**
`e30bc8f0-08bb-406c-986a-2b153a495603`, who already carried Aspirin 81mg + Advil 400mg orders. She is
the class-mate shape, and she goes 0 → 2: the identity chip off her Advil order (resolved through the
order's *concept* name — the KB carries no `advil` alias) and the curated NSAID cross-reactivity-group
chip off her Aspirin order, which no ATC subgroup relates to ibuprofen. She was chosen because she is
not one of the four controls — giving Agnes or Joshua a new allergy would have changed a control's own
expected chip count. No GP was flipped; everything is as it was found.

The body also now reports the arm-read table **re-counted here** rather than taken from the issue: the
load-bearing column (allergy reads by the order-driven arms) is zero both per-method-body and
transitively over every in-class helper, but the issue's active-order figures are not reproducible by
either method, so the numbers in the body are mine.

Two pre-existing defects found while doing this are written up in the body's last section, both measured
through the real production path and both filable as-is — a route-variant duplication in the
contraindication arms (114 of 142 shared-`rxnorm_name` families in the shipped KB; one dexamethasone
allergy raises 4 chips) and a curated-rule/identity double chip that fires out of the box on 3 of the 4
entries in the default `drug-reference.json`.

dkayiwa:
## Hardening pass (pre-merge) — 4 commits, no production logic changed

The arm, its `inPlay` skip, its precondition, its position in the sequence and its chip strings are
byte-for-byte what was reviewed. What changed: three mutation-proven test gaps closed, and nineteen
statements corrected.

### Numbers, re-measured rather than trusted

`mvn -o clean install` from the repo root, same worktree:

| tree | API | OMOD | total |
|---|---|---|---|
| `main` @ `9b93bbc` | 823 | 58 | **881** — matches the body |
| this branch @ `54dc6c1` | 831 | 58 | **889** — matches the body |
| after hardening | 835 | 58 | **893** |

0 failures, 34 skipped throughout.

### The four-failing-cases claim, re-derived by mutating this tree

Eight mutations, each against the full reactor. Five were already caught — removing the arm fails
exactly the four cases the body names; guarding the precondition on allergies alone fails the
condition leg; dropping the `inPlay` skip turns 2 chips into 4; dropping either delegated arm fails
3–4 cases. **Three survived, all 889 green:**

- **`break` after the first active order** — the arm checking one prescription and stopping. The live
  two-order capture was the only evidence for this property.
- **dropping the `warnContra` gate** — every contextless case runs with the GP absent, which fails
  safe to `true`, so nothing could tell an arm that honours the operator's switch from one that
  ignores it.
- **moving the call after the pairwise arms** — the documented chip position was unpinned.

Three cases now kill them: `everyActiveOrderIsCheckedRatherThanOnlyTheFirst` (three orders, the
contraindicated ones second and third, and the first is the one neither allergy relates to — so it
fails from both directions), `ContraindicationToggleContextTest` (writes the real GP through the
admin service, both directions), and `thePatientsOwnContraindicationsLeadTheScreensPairChips` (the
one question shape that runs both order-driven arms).

### Nineteen statements corrected

Four consecutive fact-check passes found 3, 5, 6 and 5 statements that do not hold — including some
in the corrections themselves. The three worth calling out:

1. **The ordering test justified itself with a retracted measurement.** It cited
   `eval/drift-metric/README.md`'s "8 → 5 abstentions on render ordering". That README retracts the
   figure in a call-out headed *"Read the gate's single-shot flips with suspicion"* — the honest
   number is 8 → 6, the 5 was decode noise — and the arm that produced it varied the order of
   partners *within one rendered record*, not the order of records. The comment now claims no size at
   all: what is pinned is the deliberate choice and the fact that nothing else would notice it
   reversing.
2. **`SafetyWarning.TYPE_CONTRAINDICATION` — the wire contract a renderer reads — still said the chip
   is about "a drug IN PLAY".** So did README's `warnOnContraindications` row and, worse,
   `config.xml`'s operator-facing description of the only GP that disables this arm. Nine such
   statements were outside the files the diff had touched.
3. **`docs/drug-kb-demo.md` was actively misleading on its own seed.** "Each query surfaces only the
   warnings for the drug named" is now false there, and re-deriving it from the SQL in that same file
   gives *three* extra chips, not the one I first wrote: Amoxicillin by name, and **Gentamicin twice
   by ATC code** — `DKB-ORD-8` is on concept 3923, Step 3 maps that to `J01GB03`, and
   `findByActiveOrders` compares codes exactly, so it finds the entry whatever the concept is named.
   On this standalone 3923 is *Gemfibrozil*, so those chips say "Gentamicin" for an order that
   displays as something else — a seed-data mismatch worth knowing before demoing.

Also corrected: the arm's javadoc cited #86's `iron`-inside-`spironolactone` as a live self-pair
mechanism, which #86's own fix killed (`matchesOrderName`'s left boundary rejects it); two `#140`
citations that should be `#135` (the issue, not the PR) together with the claim that it *added* the
identity check rather than stopping a guard from suppressing it; and the injector's "reads no
question and no answer at all", which the `inPlay` skip contradicts.

### Live: this build, deployed and probed — 26 probes across 9 patient shapes

Deployed via the shared harness (full JVM restart, retrieval gate `hits=75`, known-positive control
chip). Every absence re-run.

| probe | shape | chips |
|---|---|---|
| Susan × "current medications" ×2 | prescribed **and** allergic to Tiotropium; answer cites the `drug_order` record | 1 contra |
| Susan × "most recent blood pressure" ×2 | the drug named **nowhere** — the half a carve-out cannot reach | 1 contra |
| Susan × "is it safe to give her tiotropium?" | question-named — the `inPlay` skip | **1**, not 2 |
| Susan × "what interactions does tiotropium have?" | #105's recited-partner shape | 1 contra, **0** inter |
| Barbara × "current medications" ×2 | Ibuprofen allergy, on Aspirin **and** Ibuprofen | 2 contra, two different orders |
| Mark × "current medications" ×2 | prescribed but **not** allergic | 0 |
| Kevin × "current medications" | allergy, **no** orders | 0 |
| **Betty × "current medications" ×2** | allergy **and** an unrelated order (aspirin allergy, simvastatin) | 0 |
| **Sarah × blood pressure ×2** (new fixture) | polypharmacy: 8 orders + 1 allergy | **13** contra |
| **Richard × blood pressure ×2** (new fixture) | one order, a 4-entry route-variant family | **4** contra |
| controls: Mary×clarithromycin / Agnes×warfarin / Joshua×ibuprofen / Mary×paracetamol ×2 | | 1 / 1 / 2 / 0 — unchanged |

Every re-run byte-identical except `sarah-bp`, whose first attempt hit the engine's 300 s limit on a
cold prefill of an 11,053-token chart and answered in 14 s warm — chart-prefill-bound, not the arm.

**The prompt half holds.** The log shows the validator raising **twice per query** with
`[timing] search … llmMs=…` between them — e.g. Susan: `querystoreBuild patient=28` at `09:26:19,492`
→ raise at `,519` (pre-answer, inside `injectRecords`) → raise at `09:26:29,701` → timing at `,705`.
Same count on both sides every time, which is the invariant. Barbara's pre-answer pass raised 2.

Fixtures created (agent-created demo data, left in place, nothing restored, **no GP changed**):
Sarah Taylor `dc8560c9…` 8 corticosteroid/NSAID orders + Dexamethasone allergy; Richard Jones
`a39e82e7…` Dexamethasone allergy; Dorothy Odinga `71c5a6eb…` Ibuprofen allergy; Kevin Brown
`e89247a2…` a `Clarithromycine Co 500mg` order + Clarithromycine allergy.

### Load-bearing decisions: all five left alone

Direction 1 over a carve-out (the blood-pressure probe is why); contraindications only; #105
untouched and pinned by both the old test and the new case a carve-out fails; the `inPlay` skip
(1 chip, not 2, live); and the blast radius re-derived — 6 of 16 original probes changed, all cases
the arm exists for.

One temptation resisted: I did **not** add a cap. The measured 13 chips argue for one, and the
javadoc's "nothing here is quadratic in a list the module does not choose" is true but the constant
is not small. Capping is a clinical-presentation decision that interacts with #145's collapse (which
would take 13 → 4), so it belongs in an issue, not here.

### Phase 1 did not converge — stated plainly

Four consecutive prose fact-check passes each found statements that do not hold, several of them in
the previous pass's corrections. I have not observed a pass that returns nothing, so I am not
claiming convergence and did not proceed to a polish phase. **The code and tests did converge** and
have been unchanged since the first pass: mutation-verified 8/8 and live-verified across 26 probes.
What is outstanding is documentation precision, and the residue is in the report — including two
further instances of the dead `iron`/`spironolactone` example at `DrugSafetyValidator.java:1289` and
`:1573` that predate this branch.

### Found while doing this, not fixed here

The full list is in the hand-off, but one deserves flagging on this PR because it bounds what the arm
can deliver:

**A localized drug name resolves to no KB entry, so the whole contraindication surface is blind to
it — while the interaction arms still see the order.** `activeOrderEntries` → `findByQuery` and
`lookupByToken` both use `matchesText`'s **symmetric** word boundary; `hasActiveDrug` uses
`matchesOrderName`'s two-letter inflection tolerance. `DrugReference`'s own measurement says the
symmetric rule "would stop checking a patient on `Aspirine Co 81mg` for aspirin interactions at all —
trading a false positive for a false NEGATIVE, the wrong direction for a safety net". Measured live:
a patient on `Clarithromycine Co 500mg` and allergic to `Clarithromycine` gets **0 chips even when
the question names clarithromycin**, while a simvastatin question for the same patient reports
*"interacts with active order clarithromycin — Major"*. Pre-existing and on the question path, so not
this PR's to fix — but this arm is what turns it from a withheld interaction into a withheld
prescribing error.

Issue #129, opened 2026-08-04
Order names with diacritics match no interaction token, so accented localizations get no DDI checking at all (78 missed matches; pre-existing, mirror of #86)


dkayiwa:
### Summary

#86 fixed the substring over-match in `hasActiveDrug`, and the fix's measurement surfaced the mirror-image gap it does **not** close: an order name carrying **diacritics** matches no interaction token, so the patient is never checked for that drug's interactions. Same failure mode as #86's false negatives (a Major interaction silently absent), different cause.

Measured over the 3.7.1 demo dictionary (2531 drug + drug-concept names × 2093 DDInter tokens), re-running the corpus scan with diacritics folded finds **78 matches the matcher misses**. **224 of the 2531 names carry diacritics.** Examples:

```
budésonide      ~ budesonide
dexaméthasone   ~ dexamethasone
héparine        ~ heparin
furosémide      ~ furosemide
glycérine       ~ glycerin
```

This is **pre-existing and not a regression from [#128](https://github.com/openmrs/openmrs-module-chartsearchai/pull/128)** — it is missed identically before and after that fix. It is filed separately because #128's scope was the over-match, and because closing this one needs its own measurement.

### Why it is the same class of defect as #86

DDInter tokens are unaccented RxNorm generics. `matchesOrderName` (and `containsWord` before it) compares raw lowercased strings, so `budésonide` and `budesonide` share no substring at the accented character. A patient ordered `Budésonide` is therefore not checked against budesonide's interaction rules at all — the safety net is simply absent for them, with nothing logged.

The demo dictionary is bilingual by construction (`Simvastatine`, `Nitroglycérine`, `Aspirine Co 81mg`), and a francophone OpenMRS deployment is a first-class case for this project, so this is not an exotic-input concern. It is the same reason #86's symmetric-boundary fix had to be rejected: a rule that is correct for English order names and silently under-matches localized ones converts false positives into false negatives in a safety net.

### What closing it requires — and why it is not just "fold the accents"

Unicode normalization (NFD + strip combining marks, or ICU folding) on both sides is the obvious move, and it is probably right. But #86 established that a matching change here must be scored in **both** directions before it ships, because the failure mode is invisible in review — the chips that disappear look like the noise you were removing. So:

[...]

Issue #145, opened 2026-08-05
Contraindication arms have no route-variant collapse: one substance reported as cross-reactive with itself (4 chips for dexamethasone; 152 surplus across 114 families)


dkayiwa:
### Summary

The contraindication arms have no route-variant collapse, so **one substance is reported as cross-reactive with itself**, several times over. A patient with a dexamethasone allergy asked about dexamethasone gets **4 chips**: one identity chip, plus one "same ATC class as the patient's allergy to Dexamethasone" chip per sibling route variant.

`lookupByToken` resolves the allergy to a single entry (giving the identity chip), and each sibling variant — publishing the *same* ATC list — then satisfies the class comparison against it. #115's `bestRulePerPartner` collapsed this for the **interaction** arms; the contraindication arms never got the equivalent.

### Measured over the shipped 19 MB KB, through the real validator

Reproduced by me independently of the reporting agent, identical figures:

```
shared-rxnorm_name families:                                    142
  with >=2 members sharing an ATC level-4 subgroup:             114
  surplus chips implied:                                        152

dexamethasone   4 entries, 4 share a subgroup   -> 4 chips
hydrocortisone  4 entries, 4 share a subgroup   -> 4 chips
penicillin g    4 entries, 4 share a subgroup   -> 4 chips
diphenhydramine 2 entries, 2 share a subgroup   -> 2 chips
tiotropium      1 entry,   0 share a subgroup   -> 1 chip  (control)
```

Identical on `main` and on PR #144's branch — this is pre-existing on the question path. What #144 changes is its **reachability**, not its count: the new active-order arm means "the patient is on it" now reaches the same duplication.

### Aggravating: the class label names the wrong class

`sharedClass` returns the **first** matching subgroup. Dexamethasone's ATC list begins `A01AC02`, so a systemic dexamethasone allergy is reported as *"the same ATC class (A01AC)"* — **corticosteroids for local oral treatment**. The label is not merely redundant, it is misleading about which pharmacological class is implicated.

### Coverage gap that let it through

#140 added a sibling-variant test, but it covers only the **28** families that carry no ATC codes — where the class comparison cannot fire at all. The **114** ATC-carrying families are exactly the uncovered case. That is worth noting as a test-design lesson: the fixture was chosen from the population where the bug is impossible.

### Proposed fix

[...]

PR #148, opened 2026-08-05
fix(drug-safety): give a clinician-entered drug name its own matcher (#147, #141, #136)

### LINKED ISSUE #136: A rule's token is matched only against the order's display name, so aspirin ordered as `Acetylsalicylic acid` gets no DDI checking (the entry's alias set is never consulted)
### Summary

A rule's drug token is matched against a patient's active-order names and **nothing else**, so a drug
ordered under a name the KB carries as an *alias* rather than as the rule's token has no DDI coverage
at all. The load-bearing case is aspirin, an extremely common drug, under its chemical name:

- Every DDInter rule about aspirin carries the token `aspirin`, because
  `DdiDrugReferenceSource.interactionsFor` writes each rule's token from the partner row's
  `rxnorm_name` (`DDInter20` = name `Acetylsalicylic acid`, `rxnorm_name` `aspirin`).
- `Acetylsalicylic acid` is a real drug-concept name in the 3.7.1 demo dictionary, and the string
  `aspirin` **does not occur in it**, so `PatientClinicalContext.hasActiveDrug` never matches.
- Before #128 it *did* match — the token `salicylic acid`, i.e. the **wrong drug** (aspirin is not
  topical salicylic acid), inheriting that entry's 432 rules. After #128 it matches nothing. Neither
  is acceptable: a confident wrong match is worse, but an honest miss is still a Major bleeding
  interaction silently absent.

Recorded in the comment on #129 and re-checked while implementing it, and deliberately **not** folded
into that PR: diacritic folding does not touch this — there are no
diacritics in `Acetylsalicylic acid` — and combining two widenings in one change makes the
kill-direction measurement non-attributable.

### The fix, and why it is not a line in the matcher

The data is already there: the `ddinter` parser puts the drug's name, its `rxnorm_name` **and its CIEL
concept names** in the entry's `aliases` list, so the aspirin entry already carries
`acetylsalicylic acid`. The matcher simply never consults it — `hasActiveDrug(nameToken, atcCode)`
only ever sees the rule's single token.

The obstacle is where the alias set lives. `PatientClinicalContext` is a pure value object with no
service access (deliberately — it is what lets the matching logic be tested without a live OpenMRS
context), and resolving "which entry does this rule's token name" needs the loaded dataset
(`DrugReferenceService.lookupByToken` / `DrugSafetyValidator.identifies`). So the change is one of:

1. `DrugSafetyValidator` resolves the rule's partner entry and passes its alias set into
   `hasActiveDrug` (which must stay the single join both callers use — the chip decision in
   `DrugSafetyValidator` and the prompt-promotion predicate in `DrugReferenceInjector` — or chips and
   promoted prose stop agreeing about which orders a rule matches); or
2. the parser emits the alias set on the `Interaction` itself, which duplicates data per rule row
   (~295k rows in the full KB) and would need interning like the notes already have.

Whichever is chosen, the chip's *label* is a separate decision: the detail currently reads
"…interacts with active order `<token>`", and a rule matched through an alias should probably still
say the token, not the alias, so `partnerLabel` and `bestRulePerPartner` grouping are unaffected.

### What closing it requires — the measurement, not just the code

This is **additive**: it can only add matches, which is exactly the direction that reintroduces the
over-match #86/#128 removed. The same both-direction discipline applies:

1. The #86/#128 kill set must stay at **0 leaking**: `opium`⊂Tiotropium, `iron`⊂Spironolactone,
   `glycerin`⊂Nitroglycerin, `desonide`⊂Budesonide, `chlorothiazide`⊂Hydrochlorothiazide,
   `ofloxacin`⊂Ciprofloxacin, `oxacillin`⊂Cloxacilline, `salicylic acid`⊂P-aminosalicylic acid,
   `urea`⊂Hydroxyurea — and, after #129, their accented spellings too.
2. Score it over the same corpus #128 and #129 used, through the shipped matcher: the 3.7.1 demo
   dictionary's 2531 drug + drug-concept names × the KB's rule tokens. For scale, the alias corpus is
   **5169 distinct aliases against 2093 distinct tokens** (measured 2026-08-05 on the full 19MB KB),
   i.e. the needle set roughly triples.
3. Two hazards specific to aliases, which the token-only rule does not have and which the measurement
   has to separate out:
   - **combination-product names.** A large share of the CIEL aliases name two or more drugs
     (`Aspirin / caffeine / ipecac / opium` is an alias of BOTH the aspirin and the opium entry;
     `Salicylic acid / urea`, `Heparinoids / salicylic acid`, `Hydrochlorothiazide / quinapril`). An
     order of one constituent must not inherit the combination's rules, and vice versa.
   - **aliases broader than their drug** (`Salicylate` on the salicylic-acid entry), which match more
     order names than the token does.
4. The number that matters is not the match count but what the *new* matches are: enumerate every
   pair the change adds and classify each, as #128 and #129 both did.

### Tests

Through the real production path — `DrugSafetyValidator.validate(...)` with an active order named
`Acetylsalicylic acid` and a question naming warfarin, asserting the Major warfarin × aspirin chip
fires — plus the full kill set re-asserted as non-firing in the same class, and a case pinning that
the chip still names the token rather than the alias. Live verification wants an order on the
`Acetylsalicylic acid` concept (Agnes Adams's aspirin order already resolves through the drug row name
`Aspirin 81mg`, so it cannot show this).

### LINKED ISSUE #141: Allergy/condition tokens are matched without diacritic folding, so the curated penicillin rule misses fr allergen names on the DEFAULT source format
### Summary

The allergy/condition token haystack is both **looser than it needs to be** and **narrower than its own javadoc claims**, and the first half breaks the curated source's canonical penicillin rule for francophone data.

`PatientClinicalContext.containsToken` — behind `hasAllergyToken` and `hasConditionToken` — uses raw `String.contains` with **no diacritic folding**. It is the one matcher #129/#138 did not reach: that work folded `containsBoundedToken`, the *order-name* scan, and deliberately scoped itself there.

With the **default** `sourceFormat=json`, the curated Amoxicillin entry carries

```json
{"type": "allergy", "token": "penicillin"}
```

which does **not** match an allergen recorded against `Allergie à la pénicilline`, `Pénicilline G`, or `Pénicillines` — all real **fr locale-preferred** names in this dictionary, the first being the canonical allergy concept. `PatientClinicalContextBuilder` reads the concept name **in the current locale**, so a francophone deployment reaches these by default rather than exceptionally.

**Measured:** 11 locale-preferred fr names across the four curated allergy tokens fail raw containment and match folded (`paracetamol` ×6, `penicillin` ×6; `ibuprofen` unaffected, no accent inside the stem). Note this is on the **default** source format — unlike most defects found this week, which needed `ddinter`.

Folding alone suffices: containment already tolerates the trailing `-e`/`-s`, so no boundary or inflection rule is needed. It is a pure relaxation, and it does not disturb the deliberate *fragment* matching the javadoc keeps bare containment for. It requires making `DrugReference.foldDiacritics` package-private — which is also why that method is absent from CLAUDE.md's API-surface list today.

The same root cause hits the condition arm.

### Second, smaller defect: the haystack is narrower than documented

`containsToken`'s javadoc says the haystacks are "an allergen name plus its comments". They are not. `PatientClinicalContextBuilder` reads only `codedAllergen`'s name and `nonCodedAllergen` — **`Allergy.getComment()` and `Allergy.getReactions()` are never read** (verified: zero occurrences in that file).

So an allergy whose drug information lives only in the comment — coded as `Other`, commented "penicillin family", which is a realistic data-entry shape — matches no rule, while the documentation says it does. Either read the comment or correct the claim; the fragment-matching rationale itself survives via `nonCodedAllergen`.

### Verification bound, stated honestly

This is **not** live-verifiable on the current standalone: it runs `sourceFormat=ddinter`, and DDInter entries carry no contraindications at all, so no curated allergy token is in play there. Flipping the GP to `json` would exercise it — that is a legitimate way to get live evidence and worth doing rather than settling for unit tests, since demo data and GPs are disposable. But note the bound is real either way: the defect lives in the curated path, so a `ddinter`-configured instance cannot show it.

### Tests

Through the real production path (`DrugSafetyValidator.validate`) with `sourceFormat=json`: an allergy recorded as `Pénicilline G` against the curated Amoxicillin rule must raise the contraindication. Mutation-check by removing the fold. Pin the condition arm equivalently, and pin that the intentional fragment matching still works so the relaxation is not mistaken for a boundary rule.

Found while fixing #135; not fixed there because it is a different matcher on a different corpus.

### LINKED ISSUE #147: An allergen is resolved with a stricter matcher than an order name, so a localized allergy silently yields no contraindication
### Summary

An allergy recorded in a localized or inflected form silently produces **no contraindication**, while the identical string as an *order name* resolves fine. Two matchers with different tolerance sit on the two halves of the same safety check.

| path | matcher | tolerance |
|---|---|---|
| `DrugReferenceService.findByQuery` (`:71`) — resolves an order's own entry | `matchesText` → `containsWord` → `containsBoundedToken(…, 0)` | **symmetric, zero** |
| `DrugReferenceService.lookupByToken` (`:112`) — resolves an **allergen** to an entry | same | **symmetric, zero** |
| `PatientClinicalContext.hasActiveDrug` (`:155`) | `matchesOrderName` → `containsBoundedToken(…, 2)` | **two trailing letters** |

Verified through the real matchers:

```
'Clarithromycine'            vs token 'clarithromycin'  symmetric=False   tolerant=True
'Clarithromycine Co 500mg'   vs token 'clarithromycin'  symmetric=False   tolerant=True
```

So for a patient on `Clarithromycine Co 500mg` who is **allergic to Clarithromycine**:

- the allergen fails to resolve (`lookupByToken`, symmetric) → **no identity chip, no contraindication, even when the question names clarithromycin**;
- yet a simvastatin question happily reports *"interacts with active order clarithromycin — Major"*, because that path is tolerant.

The module therefore knows the patient is on clarithromycin and knows they are allergic to it, and reports the interaction while withholding the allergy.

### This is #135's failure mode by a different route

#135/#140 fixed a direct allergy being suppressed by a *classification guard*. This suppresses the same warning by *failing to resolve the allergen at all*. Same clinical outcome: the patient's own recorded allergy to a drug they are taking produces nothing.

`#86`'s own measurement condemns the symmetric rule for exactly this — it records that symmetric matching drops 68 legitimate localized names and calls that "a false NEGATIVE, the wrong direction for a safety net". That reasoning was applied to order names and never carried to allergen resolution.

`PatientClinicalContext:249-256`'s javadoc already acknowledges that "one matcher cannot serve" both shapes. The gap is that a third shape — an **allergen string**, which is also a clinician-entered drug name — was never assigned a matcher and inherited the prose one by default.

### A framing correction worth recording

An earlier pass reported this area as "localized order names resolve to no entry". Live measurement **refuted** that: Dorothy Odinga's `Aspirine Co 81mg` does resolve, because that concept's locale-preferred name is `Aspirin`. The order side is fine. **The allergy side is the real gap** — worth stating so nobody re-tests the wrong half.

### Proposed fix

Resolve allergens with the order-name matcher, or introduce an explicitly-named third accessor for clinician-entered drug strings so the choice is deliberate rather than inherited. Do **not** loosen `matchesText` itself — it serves question/answer prose, where #86 measured symmetric boundaries as correct.

Then measure both directions, as #86/#129 established: the kill set must stay at zero leaking (`opium`⊂Tiotropium, `iron`⊂Spironolactone, and the rest) and the localized keep set must reach zero lost. Note the allergen corpus is *not* the same population as order names, so it needs its own scan — the 11 locale-preferred fr allergen names measured in #141 are the obvious starting set.

### Interaction with #141

#141 is the diacritic half of the same surface (`containsToken`, behind `hasAllergyToken`/`hasConditionToken`). This is the inflection half, on `lookupByToken`. They are different methods on different corpora but the same clinical failure, and a fix for either should be checked against the other so the two do not each solve half of `Pénicillines`.

### Tests

Through the real production path: an allergy recorded as `Clarithromycine`, patient prescribed `Clarithromycine Co 500mg`, question naming clarithromycin → one contraindication chip. Mutation-check by restoring the symmetric matcher. Re-assert #86's full kill set as non-firing so the widening is pinned, and re-assert that question/answer prose matching is unchanged.

Found while hardening #144 (for #143); pre-existing on the question path, and unrelated to that fix.

dkayiwa:
Closes #147. Closes #141. Closes #136.

Three mechanisms of one clinical failure: a drug name a clinician entered fails to resolve, so an
allergy the chart already records produces no warning. Fixed in one PR because they are the same
words on three different corpora, and because each widening has to be scored against the other two's
kill sets — separately, the both-direction measurement is not attributable.

## What was wrong, and what each fix is

**#147 — an allergen was resolved with the PROSE matcher.** `DrugReferenceService.lookupByToken` went
through `DrugReference.matchesText` (symmetric word boundaries) while the *same string* as an order
name went through `matchesOrderName` (left boundary plus up to two inflection letters, issue #86). So
a patient on `Clarithromycine Co 500mg` was told that simvastatin "interacts with active order
clarithromycin — Major" while their own recorded allergy to that drug produced nothing at all. An
allergen string is a third shape and it inherited the prose matcher by default; it now has
`DrugReference.matchesDrugName`, which *borrows* the order-name rule because an allergen name and an
order's display name are the same shape read out of the same dictionary. `matchesText` is untouched —
#86 measured symmetric boundaries as correct for prose, and a case in the new test class fails if
someone "fixes" this by loosening it instead.

The same defect sat on the order side of #147's own table: `findByQuery` (the prose matcher) was also
resolving an *order's* display name to its entry, behind the screening arm and the #143 active-order
contraindication arm. That is now `findByDrugName`, in one shared `findForActiveOrders`.

**#141 — the allergy/condition token haystack did not fold diacritics.**
`PatientClinicalContext.containsToken` was the one matcher #129/#138 did not reach. On the **shipped
default** `sourceFormat=json` the curated Amoxicillin entry's `{"type":"allergy","token":"penicillin"}`
rule missed an allergen recorded as `Pénicilline G` — a real fr locale-preferred name, and the builder
reads the concept name in the *current locale*, so a francophone deployment reaches it by default.
Folding is the whole change: bare containment already tolerates the trailing `-e`/`-s` of
`Pénicillines`, so no boundary rule comes with it and the deliberate fragment matching (`nsaid` inside
"NSAIDs") is untouched and pinned. `DrugReference.foldDiacritics` becomes package-private and joins
CLAUDE.md's API-surface list.

Its javadoc's second claim — that the allergy haystack includes the allergen's *comments* — was false
and is corrected rather than made true: `PatientClinicalContextBuilder` never reads
`Allergy.getComment()` or `getReactions()`, and reading the comment is not a free widening, because a
comment can **negate** ("tolerated penicillin, no reaction") and the same strings feed
`lookupByToken`, so it would be read as the allergen itself and fabricate an allergy. Filed as a
separate finding below.

**#136 — a rule's token was matched only against the order's display name.** Every DDInter rule about
aspirin carries the token `aspirin` (the partner row's `rxnorm_name`) while that row's own name is
`Acetylsalicylic acid`, a real drug-concept name in the 3.7.1 demo dictionary that does not contain
the string `aspirin` — so an order under it had no DDI coverage at all, a Major warfarin bleeding
interaction silently absent. The patient's orders are now resolved to their reference entries once per
pass and those entries' names reach `hasActiveDrug`, which stays the **single join** both the chip
decision and the prompt-promotion predicate use (unchanged signature, so no call site can ask the
narrower question).

That arm is exact **identity**, deliberately not a second boundary scan, and the difference is not
cosmetic: an entry's alias list carries combination-product names — `salicylic acid / urea` is an alias
of the *Urea* entry as well as of the Salicylic acid one — so scanning a rule's token across those
names reports a patient on urea alone as being on salicylic acid, which is #86's fabricated drug
arriving by a new route. Identity asks the same question `identifies` already asks of the reference
side, and it is **provably equal** to #136's proposed formulation (resolve the token to its partner
entries, match *their* names against the order name) at one dataset sweep per pass instead of one per
rule — verified pair-by-pair: both add exactly the same 94 pairs.

## Both directions, measured

Measured 2026-08-05 against the full 19 MB KB (2283 entries, 2093 distinct rule tokens, 5169 distinct
aliases) and the 3.7.1 demo dictionary, then **re-scored through the shipped code paths** — the real
`hasActiveDrug` behind a real `withReferenceNames` context, the real `lookupByToken`, the real
`hasAllergyToken` — so these are the implementation's numbers, not a model's.

| arm | corpus | before | after | added | lost |
|---|---|---|---|---|---|
| #136 rule token × order name | 2533 order names × 2093 tokens | 907 | 1001 | **+94** over 77 names | **0** |
| #147 allergen → entry | 1219 allergen-candidate names | 549 | 624 | **+75** | **0** |
| #147 order name → entry | 2533 order names | 753 | 832 | **+117** pairs | **0** |
| #141 curated allergy token | 1219 names × 6 tokens | 27 | 38 | **+11** | **0** |
| kill set (#86/#128/#129) | 21 nesting pairs, all three arms | 0 leaking | 0 leaking | — | — |

The 907 baseline reproduces `matchesOrderName`'s own recorded #129 figure exactly, so the corpus is
the one that settled that rule.

**Every one of the 94 added pairs was enumerated and classified.** In all 94 the patient *is* on the
drug the matched entry describes — none is a fabricated drug. 81 have a token that is that drug under
another vocabulary (`rifampin`/Rifampicin ×16, the vaccine antigens ×35, `albuterol`/Salbutamol ×8,
`torsemide`/Torasemide, `aspirin`/Acetylsalicylic acid …). The other **13** have a token naming a
different substance, so the chip's *partner label* is wrong while its finding is right: 9 because
DDInter's own `rxnorm_name` for that row is a mis-normalisation (`Chlorpheniramine`/`chlorine` ×3,
`Sulfamethoxazole`/`sulfamethazine` ×6) and 4 because it is a close congener
(`Omeprazole`/`esomeprazole` ×2, `Isosorbide`/`isosorbide mononitrate` ×2). That mislabel is the
dataset's, is the class `displayLabel()` exists for, and already reached any patient whose order name
happened to carry the `rxnorm_name` spelling. Filed as a finding below rather than fixed here, because
#136 explicitly decides the chip should keep saying the token.

The two hazards #136 flagged were checked directly rather than assumed:
- **combination-product aliases** — a constituent order must not inherit the combination's rules. It
  does not, and there is a test for it (`Urea 40% cream` raises no salicylic-acid chip while
  `Salicylic acid / urea 6% cream` does). This is the case that rules out the scanning implementation.
- **aliases broader than their drug** (`salicylate` on the salicylic-acid entry) — reaching them is
  correct (a `Sodium salicylate` order carries the same warfarin bleeding risk, and now raises the
  chip); what must not happen is reaching a drug that merely nests one, which the kill set pins.

## One regression this caused, and the shared fix

Matching *every* name of a drug means two rules whose tokens are two names of the **same** partner both
match, and `bestRulePerPartner` grouped on the rule's **label** — so one clinical pair got two chips.
`DrugSafetyQuestionPairInteractionTest.aPairAlsoJoinedByARuleNamingAnActiveOrderStaysWithTheActiveOrderArm`
caught it (a `coumadin` row and a `warfarin` row against one `Warfarin 5mg` order → 2 Major chips
where the spec says 1). Root-caused in shared infrastructure rather than patched at the call site:
grouping now keys on the partner **entry** where the dataset identifies one, and on the label where it
cannot. Two *distinct* entries still get two chips, so #121's `enalapril`/`enalaprilat` decision is
intact, and where no entry is identified #121's invariant ("the key IS what the chip says") holds
unchanged.

## Tests

**907** (API 849 + OMOD 58), 0 failures, 34 skipped — **893** on `main` at `b013626`. Three new
classes, all through the real `DrugSafetyValidator.validate` / `DrugReferenceInjector.injectRecords`:

- `AllergenNameResolutionTest` — #147, including the asymmetry in one patient (one localized name on
  the chart as an order *and* as an allergy), the kill direction (`Tiotropium` must not resolve to
  Opium), and the guard that prose matching did not move.
- `RuleTokenAliasOrderMatchingTest` — #136, including the label pin (the chip says `aspirin`, not
  `acetylsalicylic acid`), both directions of the combination-product alias, the broader alias, the
  kill set through the new arm, and that the promoted prompt text agrees with the chip.
- `ContraindicationTokenDiacriticFoldTest` — #141, all 11 measured fr names one by one, both fold
  directions (accented haystack from the shipped file, accented *token* from a francophone fixture),
  and the fragment-matching guard.

Two fixtures: `ddi-alias-drug-names.json` (a verbatim 11-drug excerpt of the full KB, chosen for the
alias shapes these issues turn on) and `drug-reference-accented-tokens.json` (a francophone
deployment's own curated rules).

## Live verification

See the comment below.


dkayiwa:
## Live verification on the 3.7.1 standalone — part 1 of 2

Deployed build: commit `e947d27` (this branch), via `standalone-verify.sh` — full-JVM restart, module
started, querystore retrieval gate passed (`hits=75` on the control patient), LLM warm.
`sourceFormat=ddinter`, full 19 MB KB, `minInteractionSeverity=minor`, `chartMode=fullChart`,
`llm.engine=local`.

### The four regression controls — all four as documented

| patient | question | chips | result |
|---|---|---|---|
| Mary Smith | Is it safe to give her clarithromycin? | **1** | `Clarithromycin interacts with active order simvastatin — Major.` ✔ |
| Agnes Adams | Is it safe to start warfarin? | **1** | `Warfarin interacts with active order aspirin — Major.` ✔ |
| Joshua Johnson | Can I give ibuprofen? | **2** | NSAID cross-reactivity group contra + `lisinopril — Moderate` ✔ |
| Mary Smith | Is it safe to give her paracetamol? | **0** | ✔ |

### #147 — the headline, both halves on one chart

**Kevin Brown** `e89247a2…` carries exactly the shape the issue describes, on real demo data and with
no fixture needed: an active order of the drug row `Clarithromycine Co 500mg`, and a **coded** drug
allergy to that drug's concept. That concept's `en` locale-preferred name is `Clarithromycine` — the
3.7.1 dictionary ships the French spelling as the English preferred name and carries no
`Clarithromycin` concept at all — so the allergen string the builder reads is the inflected one under
the default locale.

| question | chips |
|---|---|
| Is it safe to give her **simvastatin**? | 2 — `Simvastatin interacts with active order clarithromycin — Major.` **and** `The patient has a recorded allergy to Clarithromycin.` |
| Is it safe to give her **clarithromycin**? | 1 — `The patient has a recorded allergy to Clarithromycin.` |

**Had the fix been wrong**, the allergen would still have resolved through the prose matcher and
*both* contraindication chips would be absent — leaving exactly the defect #147 reports: the
simvastatin question reporting the interaction with clarithromycin while the clarithromycin question
returns 0 chips and says nothing about the recorded allergy. The interaction chip is the precondition
that makes that absence a defect rather than a data gap, and it is present in both states.

### #136 — an order named only by the drug's chemical name

Fixture created (agent-created demo data, documented): a Drug-class concept
`Acetylsalicylic acid (i147 fixture)` and a drug row `Acetylsalicylic acid` on it, ordered for
**David Williams** `353e5eca…`, who had no orders at all. Neither of the order's two names contains
the string `aspirin`. The existing `Aspirin 81mg` row cannot show this: it hangs off concept 112,
whose `en`-preferred name *is* `Aspirin` and whose `Acetylsalicylic acid` name is only a synonym, so
an order on it carries `aspirin` through the concept name.

> **Is it safe to start warfarin?** → **1 chip**:
> `Warfarin interacts with active order aspirin — Major. Aspirin, even in small doses, may increase the risk of bleeding …`
>
> and the answer, grounded on the injected finding record:
> *"No — Warfarin should not be started: Warfarin interacts with active order aspirin — Major. …"*

Same rule, same severity, same partner label as Agnes Adams's control chip — one spelling reached it
before, both do now. **Had the fix been wrong** in the missing direction this is **0 chips** and the
answer has nothing to refuse on (the pre-#136 state); had it been wrong in the over-matching
direction there would be a second chip naming `salicylic acid`, since Warfarin carries a real
Moderate rule against it and `Acetylsalicylic acid` contains that string. Neither happened.

### Kill direction, live

**Susan Young** `763e6e5f…` is allergic to **Tiotropium** and prescribed it — the shape where a
widened allergen matcher would leak, because `tiotropium` contains `opium` and the KB carries an
Opium entry with rules.

| question | chips |
|---|---|
| Is it safe to give her tiotropium? | 1 — `The patient has a recorded allergy to Tiotropium.` |
| Is it safe to give her **opium**? | 1 — `The patient has a recorded allergy to Tiotropium.` (her own order vs her own allergy, the #143 arm) — **no Opium chip** |

The Opium entry **is** in play on the second question (`findByQuery` resolves it from the word
`opium`), so the absence discriminates: **had the fix been wrong**, `Tiotropium` would have resolved
to the Opium entry and the chip would read *"The patient has a recorded allergy to Opium"* — a drug
she has no recorded allergy to, which is #86's fabricated drug on the allergy arm.

The corpus-wide version of that check: all 21 kill pairs re-scored through the shipped
`hasActiveDrug` behind a real `withReferenceNames` context **and** through the shipped
`lookupByToken` — **0 leaking**, and each name resolves to its own entry
(`P-aminosalicylic acid` → `Aminosalicylic acid`, not Salicylic acid).

### Note on this run

The first probe pass was killed by a `SIGTERM` from the agent harness (rc=143) part-way through, so
two ddinter shapes and the whole `sourceFormat=json` pass #141 needs did not run. Part 2 follows.

dkayiwa:
## Live verification — part 2 of 2

Same instance, rebuilt and redeployed twice more by `standalone-verify.sh` (labels `i147b`, `i147c`):
full-JVM restart each time, module started, retrieval gate `hits=75` / `warm chips=1`, LLM warm. The
`i147c` pass runs the **final commit** — the earlier ones ran a build differing by one
behaviour-preserving collapse of a duplicated partner-entry resolution — and it re-ran both headlines
and two controls on that build, on a fresh JVM lifetime:

| patient | question | chips |
|---|---|---|
| Kevin Brown | Is it safe to give her clarithromycin? | **1** — `The patient has a recorded allergy to Clarithromycin.` |
| David Williams | Is it safe to start warfarin? | **1** — `Warfarin interacts with active order aspirin — Major.` |
| Mary Smith | Is it safe to give her clarithromycin? | **1** — `… active order simvastatin — Major.` |
| Mary Smith | Is it safe to give her paracetamol? | **0** |
| Mark Smith | Is it safe to give her ciprofloxacin? | **2** — `recorded allergy to Ciprofloxacin` + `active order chloroquine — Major` |
| Dorothy Odinga | Is it safe to give her ibuprofen? | **3** — identity, `active order aspirin — Major`, NSAID group |

Mark's coded `Ciprofloxacin` allergy carries no inflection, so it must resolve before and after — the
widened matcher left the exact case alone. Dorothy's `Aspirine Co 81mg` order was already covered
because its concept's `en`-preferred name is `Aspirin`; her probe is the check that the widening
neither disturbed those three chips nor added a spurious one (no `salicylic acid` chip off
`Aspirine`).

### More shapes under `sourceFormat=ddinter`

| patient | shape | question | chips |
|---|---|---|---|
| **Barbara Miller** `e30bc8f0…` | 6 active orders incl. `Aspirin 81mg` + `Advil 400mg`, coded Ibuprofen allergy | Are there any drug interactions with her current medications? | **3** |
| **George Anderson** `d5c9cdad…` | no orders, no allergies | Is it safe to start warfarin? | **0** |
| **Michael Turner** `d35b0325…` | 2 orders whose brand names (`Klarizom`, `Zolvimix`) match no KB entry | What are her active medications? | **0** |

Barbara's three: `Acetylsalicylic acid (aspirin) interacts with active order ibuprofen — Major` (the
#113 screening arm), plus her own allergy against her own orders twice (#143 — the NSAID
cross-reactivity group against her aspirin order, and identity against her Advil order). **Exactly
one** interaction chip for the one pair: the grouping change from label to partner entry neither
doubled it nor merged it away, on a real polypharmacy chart with two orders resolving to interacting
entries.

George is the no-false-positive control: three widened matchers, nothing on the chart to match, zero
chips — and the answer abstains (*"The records do not address the safety of starting Warfarin."*).

Michael is the weaker but real control that the widening did not start matching arbitrary strings: his
two order names resolve to **no** reference entry at all, and the answer still lists both orders, so
the reconciliation/injection path is untouched.

### #141 — under `sourceFormat=json`, the SHIPPED DEFAULT

#141 lives in the curated path, and DDInter emits no contraindications at all, so it is unreachable
under the instance's usual `ddinter`. The GP was flipped to `json` and the JVM fully restarted (the
source is chosen once and its entries cached for the bean's lifetime), verified in effect
(`{"value":"json"}`) with the retrieval gate re-passed before anything was read.

Fixtures created (agent-created demo data, documented): two order-less patients given a drug allergy
that differs only by the accent — **Edward Robinson** `d1634ba2…` `Penicillin G`, **Donna Roberts**
`92006ad7…` `Pénicilline G`. Non-coded, because the 3.7.1 dictionary has no Drug concept whose **en**
locale-preferred name is accented in the penicillin/paracetamol families (all 11 accented names are
`fr`-preferred), and the builder reads the name in the current locale. The non-coded allergen lands in
the identical `allergyTokens` set the coded path fills, so the matcher under test is the same one; the
locale route is what the 11-name measurement covers.

**A first attempt at this pass was invalid, and finding out why is a defect worth filing.** This
instance pins `chartsearchai.drugReference.dataFilePath` to the 19 MB DDInter file, and that GP
**wins** over the bundled classpath default (`ReferenceDataFiles.loadWithClasspathFallback`). Flipping
only `sourceFormat` therefore feeds the DDInter-shaped file to `JsonDrugReferenceSource.parse`, which
finds no `entries` key and returns an **empty list** — logged at INFO as `Loaded 0 drug-reference
entries`, with no WARN and `started=true`. Measured consequence: Barbara Miller, who has a coded
Ibuprofen allergy, asked "is it safe to give her ibuprofen?", got **0 chips and no injected
drug-reference record**, which a loaded curated KB cannot do. Reported separately; it means any
`sourceFormat=json` measurement on this instance is suspect unless `dataFilePath` is blanked too.

So this pass blanks both GPs and gates on a **positive control chip** instead of a log line:

```
    set chartsearchai.drugReference.sourceFormat='json'   HTTP 200
    set chartsearchai.drugReference.dataFilePath=''       HTTP 200
    INFO ReferenceDataFiles … Loaded 4 drug-reference entries from bundled default
                                            /chartsearchai/drug-reference.json
    GATE j0-gate-barbara: 4 chip(s), expected detail 'documented ibuprofen allergy' -> PRESENT
```

| patient | allergen as recorded | question | chips |
|---|---|---|---|
| Edward Robinson | `Penicillin G` | Is it safe to give her amoxicillin? | **1** — `Amoxicillin is contraindicated by an active allergy: penicillin-class hypersensitivity` |
| **Donna Roberts** | **`Pénicilline G`** | Is it safe to give her amoxicillin? | **1** — the same chip, and the answer: *"No — Amoxicillin should not be given: it is contraindicated by an active allergy: penicillin-class hypersensitivity [91]."* |
| David Williams | — (his `Acetylsalicylic acid` order) | Is it safe to start warfarin? | **0** |

Edward is the precondition: the unaccented spelling of the same product fires before and after, so
Donna's chip is attributable to the fold and nothing else. **Had the fold been absent**, Donna is
**0 chips** and her answer does not refuse — which is #141 exactly, on the shipped default source
format, for the canonical francophone spelling of the drug the curated rule is about.

David under the curated source is a cross-source negative control: that 4-entry seed carries no
aspirin entry at all, so the #136 arm correctly adds nothing — the widening only reaches drugs the
loaded dataset actually has.

### GP left set

Both GPs the json pass touched were restored — `sourceFormat` to **`ddinter`** and `dataFilePath` to
**`chartsearchai/ddi_knowledge_base.json`** — and the JVM was restarted afterwards inside the same
pass, so the instance is back on the full KB **in fact**, not merely in the GP table. Verified rather
than assumed:

```
    INFO ReferenceDataFiles … Loaded 2283 DDInter drug-reference entries from
                              …/appdata/chartsearchai/ddi_knowledge_base.json
    GATE c9-gate-mary: 1 chip(s), expected detail 'active order simvastatin' -> PRESENT
```

`log.level` was re-armed to `org.openmrs.module.chartsearchai:info` after each of my restarts (startup
does not apply that GP — only the save listener does, which is what made a first gate misfire). Nothing
else was changed. Demo data added and left in place, all commented
`i147/i141/i136 live fixture (agent-created demo data)`:

- concept `Acetylsalicylic acid (i147 fixture)` + drug row `Acetylsalicylic acid` + one active order
  of it for David Williams `353e5eca…` (who had none);
- non-coded drug allergies `Penicillin G` (Edward Robinson) and `Pénicilline G` (Donna Roberts).

Kevin Brown's `Clarithromycine` allergy and `Clarithromycine Co 500mg` order were **already there**
from an earlier agent's h144 fixture; nothing was added to him.

dkayiwa:
# Found while fixing #147/#141/#136 — reported, not fixed

## 1. DEFECT — a chip names the KB's mis-normalised `rxnorm_name` instead of the drug the patient is on

**Where:** `DrugSafetyValidator.partnerLabel` (the chip's "…interacts with active order **X**") renders
the rule's match token, which `DdiDrugReferenceSource.interactionsFor` writes from the partner row's
`rxnorm_name`.

**How it fails:** DDInter mis-normalises some rows, so that `rxnorm_name` names a *different substance*
from the row's own `name`. Measured on the full 19 MB KB: `DDInter362` is named **Chlorpheniramine**
with `rxnorm_name` **`chlorine`**; `DDInter1724` is named **Sulfamethoxazole** with `rxnorm_name`
**`sulfamethazine`**. So a patient on `Sulfamethoxazole / trimethoprim`, asked about a drug with a
sulfamethoxazole rule, gets a chip reading "…interacts with active order **sulfamethazine**" — a
sulfonamide they are not on. The *finding* is right; the *label* is wrong.

**Measured, not reasoned:** 13 of the 94 (order name, rule token) pairs the #136 arm adds carry such a
label — `chlorine` ×3, `sulfamethazine` ×6, plus two close-congener cases
(`Omeprazole`/`esomeprazole` ×2, `Isosorbide`/`isosorbide mononitrate` ×2). **Pre-existing**: it
already reached any patient whose order name happened to carry the `rxnorm_name` spelling; #136 widens
its reach to any order under the entry's own name.

**Why not fixed here:** #136 explicitly decides that a rule matched through an alias should still say
the token, so `partnerLabel` and the grouping stay untouched. Changing the label is a separate
decision.

**Fix candidate, now cheap:** since this PR, `DrugSafetyValidator.bestRulePerPartner` already resolves
the partner ENTRY at the chip site (via `activeOrderEntryFor`, to key the grouping), so the label could
be `entry.displayLabel()` when an entry is
identified — which renders `Sulfamethoxazole (sulfamethazine)`, both vocabularies — falling back to the
token when it is not. Needs a decision about `bestRulePerPartner`'s key/label invariant (#121) and a
sweep of the assertions that pin chip wording (`DrugSafetyChipLabelTest`,
`DrugSafetyInteractionScreeningTest`'s withheld-pair log labels).

## 2. DEFECT — an allergy coded as `Other` with the drug named only in the comment matches nothing

**Where:** `PatientClinicalContextBuilder.build`, the allergy loop. It reads
`allergy.getAllergen().getCodedAllergen()`'s name and `getNonCodedAllergen()`; it never calls
`Allergy.getComment()` or `getReactions()` (verified: zero occurrences in that file).

**How it fails:** the realistic data-entry shape "coded as `Other`, comment: penicillin family" puts
the only drug information in the comment, so no curated rule and no allergen resolution sees it. This
is the second half of #141; that issue offered "either read the comment or correct the claim", and this
PR corrected the claim, because reading it is **not** a free widening: a comment can NEGATE
("tolerated penicillin, no reaction") and the same token set feeds
`DrugReferenceService.lookupByToken`, whose result `addAllergyContraindications` treats as *the drug
the patient is allergic to* — so a comment read as an allergen would fabricate an allergy.

**Reasoned, not measured** — I did not measure how many demo allergies carry a comment naming a drug
(the demo set has 12 allergies, all coded, and the comments are agent fixture notes). Reachability is
"any deployment whose clinicians use the Other + comment shape", which is common in practice but
unquantified here.

**Fix shape:** a separate comment-derived token set that only the *curated rule* arm consults, never
the identity arm — or a negation-aware read. Needs design, not a one-liner.

## 3. HAZARD — order-driven reference INJECTION still resolves by ATC code only

**Where:** `DrugReferenceInjector.matchingEntries` calls `drugReferenceService.findByActiveOrders`
(ATC codes only), while the safety layer now resolves the same orders by name too
(`findForActiveOrders`).

**How it bites:** on a dictionary with sparse ATC mapping — which the 3.7.1 demo dictionary is, 85 of
616 Drug-class concepts — a clinically related active order can raise a CHIP while its reference record
is never injected, so the model has no citable reference material for the drug the chip is about. The
chip is still injected as a safety-finding record (#110), so the finding itself is grounded; what is
missing is the *reference* record behind it.

**Reasoned from code, not measured.** Whether it changes any answer needs a probe on an ATC-less
order whose question is class-related — and the class relevance gate (`relatedToAny`) itself needs ATC
codes, so it may be unreachable in practice. Worth an issue mainly to settle whether the injector's
resolution should follow the validator's.

## 4. DEFECT — a `sourceFormat` / `dataFilePath` mismatch loads ZERO entries, silently

**Where:** `ReferenceDataFiles.loadWithClasspathFallback` prefers a non-blank
`chartsearchai.drugReference.dataFilePath` over the bundled classpath default, and each source's
`parse` degrades a *structurally wrong* file to an empty list rather than to the fallback:
`JsonDrugReferenceSource.parse` returns `Collections.emptyList()` when the JSON has no `entries` key,
and `DdiDrugReferenceSource.parse` does the same when it has no `drugs`/`interactions`.

**How it fails, measured on the standalone during this PR's verification:** the instance pins
`dataFilePath=chartsearchai/ddi_knowledge_base.json`. Flipping only `sourceFormat` to `json` — the
obvious way to exercise the curated path, and what the #141 issue itself suggests — feeds that
DDInter-shaped 19 MB file to the curated parser, which finds no `entries` and loads **0** entries. The
whole drug-safety feature then silently does nothing: Barbara Miller, who has a coded Ibuprofen
allergy, asked "is it safe to give her ibuprofen?", got **0 chips and no injected drug-reference
record** — which a loaded curated KB cannot produce, since that entry's own `ibuprofen` allergy token
matches her allergen exactly. The log line is `INFO … Loaded 0 drug-reference entries from …`; there is
no WARN and no `started=false`, so the module reports healthy.

**Reachability:** any operator who changes one of the two GPs without the other. The two are
independent settings whose values must agree about the file's *schema*, and nothing checks that they
do.

**Fix candidates:** (a) treat "the configured file parsed to zero items" as a load FAILURE and fall
back to the bundled default, which is what the fallback already exists for; (b) WARN rather than INFO
on a zero-item load; (c) have each source sanity-check the top-level shape and say which key it wanted.
(b) is the minimum — a safety feature that has silently switched itself off should not log at INFO.

Not caused by this PR, and not in its path — found because the verification needed the flip. Worth
filing on its own because it makes every `sourceFormat=json` measurement on this instance suspect
until the pairing is fixed or documented.

## 5. STALE DOCUMENTATION — the probe corpus in the shared brief has drifted

Measured on the standalone 2026-08-05, after this session's fixtures:

- **Betty Williams** `a7090f70…` now has **1** active drug order (`Simvastatin Co 20mg`), not 0 as the
  brief's table states (which itself corrected an earlier "one order" claim).
- **Agnes Adams** `47e57b75…` has **13** active orders — 12 lab orders and 1 drug order
  (`Aspirin 81mg`). Her regression control still behaves as documented.
- **Kevin Brown** `e89247a2…` has **15** active orders (14 lab, 1 drug `Clarithromycine Co 500mg`) and
  **3** allergies (Levothyroxine, Levothyroxine sodium, Clarithromycine).
- The counts in that table are order counts of every type, so "N active orders" and "N active *drug*
  orders" have drifted apart; the safety layer only reads `DrugOrder`s.

Not a module defect — a note so the next agent's probe expectations are not silently wrong.

## 6. DATA OBSERVATION — the demo dictionary's en-preferred name is French for some drugs

`concept a4391c57-549b-4047-8971-c3fa72e140b5` is named **`Clarithromycine`** in BOTH `en` and `fr`
(FULLY_SPECIFIED, locale-preferred in each). There is no concept named `Clarithromycin` at all. That is
why #147's live fixture needed no new data: an order and a coded allergy on that concept both present
the inflected spelling under the default `en` locale. Related: 11 locale-preferred **fr** names in the
penicillin/paracetamol families exist but none is en-preferred, which is why #141's live probe needs a
non-coded allergen (or a locale switch) to put an accent in the string.

Not a defect. Recorded because it makes "the localized-name defects only bite francophone deployments"
false — this dictionary ships localized spellings as English preferred names.

dkayiwa:
### Hardening note: `5d7b0e4` had disabled #136's core arm on this branch — fixed in `5a4f5b9`

Two hardening agents ran against **the same git worktree** at the same time (a duplicate was spawned after the live one was mistakenly judged dead). One was running a mutation sweep — deliberately reverting each of this PR's decisions to check a test fails — while the other committed with `-a`. The result: commit `5d7b0e4` captured a live mutation into the branch and **pushed it**:

```java
// PatientClinicalContext.hasActiveDrug, as committed in 5d7b0e4
-   if (activeDrugReferenceNames.contains(DrugReference.normalizeName(n))) {
+   if (false) {
```

That is the whole issue-#136 reference-name arm — the join this PR adds — dead on the branch. `5d7b0e4`'s own message records "853 + 58 = 911, 0 failures", a figure measured on the clean tree moments before the mutation landed; the tree was never re-built at that SHA.

`5a4f5b9` restores the line. Its content is **byte-identical to `9fe2ce8`'s version of the whole file** (verified by diffing against `git show 9fe2ce8:`), and no test was touched — three existing tests already fail with the arm off, so the branch was self-detecting and only needed rebuilding:

- `RuleTokenAliasOrderMatchingTest.anAliasBroaderThanTheTokenStillNamesTheSameDrug`
- `RuleTokenAliasOrderMatchingTest.anOrderUnderTheDrugsChemicalNameIsStillCheckedAgainstTheRule`
- `RuleTokenAliasOrderMatchingTest.thePromotedPromptTextAgreesWithTheChip`

**Two things a reviewer should carry forward:**

1. `5d7b0e4`'s claim that "12 of 12 reversions across the slice are caught" was measured while a second sweep was mutating the same files. It is not attributable and should be re-run on a quiet tree before it is relied on. The same caveat applies to my own partial sweep.
2. Baselines I measured myself, full-reactor `mvn -o clean install`, each in its own worktree, 0 failures / 34 skipped in every one — these reproduce the figures already recorded here: `origin/main` (`b013626`) **893** (835 + 58) · PR head `8ff0b4b` **907** (849 + 58) · `9fe2ce8` **910** (852 + 58). `5a4f5b9` is expected to be 911 (`5d7b0e4` adds one test); that one I have not measured on a quiet tree.

Live evidence on the standalone, against an omod built from `9fe2ce8` (i.e. the intended production code — built before either sweep started), full 19MB DDInter KB, `sourceFormat=ddinter`, 2283 entries. Patients resolved **by name lookup against the database**, not by uuid:

| probe | patient shape | result | what it would have shown if the fix were wrong |
|---|---|---|---|
| #147 headline | Kevin Brown — coded `Clarithromycine` allergy **and** `Clarithromycine Co 500mg` order | 1 chip: "recorded allergy to Clarithromycin" | 0 chips — the allergen resolving through the prose matcher is exactly #147 |
| #136 headline | David Williams — `Acetylsalicylic acid` order, no allergy | 1 chip: "Warfarin interacts with active order aspirin — Major" | 0 chips — the silently-absent Major bleeding interaction #136 reports |
| kill case | Susan Young — Tiotropium allergy + Tiotropium order, asked about **opium** | **no** Opium chip (only the correct Tiotropium allergy chip) | a false "allergy to Opium" chip — #86's nesting bug arriving through the new matcher |
| polypharmacy | Sarah Taylor — 8 active orders + Dexamethasone allergy | 23 chips, interaction chips capped at 10 (`maxPairChips`) | — (cap/blast-radius check) |

The four regression controls (Mary×clarithromycin 1 Major ×simvastatin, Agnes×warfarin, Joshua×ibuprofen, Mary×paracetamol 0 chips) also ran on this build and held.

dkayiwa:
# Hardening pass — what changed, and the measurement re-derived independently

Phase 1 ran to convergence on `fix/allergen-resolution`. Commits push tests, one behaviour-preserving
refactor, and prose corrections. **No load-bearing decision was reversed.**

## Method: every decision was reverted to see whether a test notices

Rather than re-reading the diff for what looked risky, each of the slice's load-bearing decisions was
mechanically reverted and the suite re-run. **Three reversions passed every test** — that is Phase 1's
finding. (The sweep was re-run at the end in an isolated copy of the tree, because a concurrent agent
was briefed into this same worktree for a while and the first sweep's verdicts were not attributable.
Numbers below are from the isolated re-run: baseline 853 API tests, 0 failures, 34 skipped.)

| reversion of a load-bearing decision | caught by |
|---|---|
| `activeOrdersOtherThan` takes the reference names whole off the context | `ActiveOrderAtcAttributionTest` ×4, `ActiveOrderAtcContextTest` |
| `bestRulePerPartner` keys on the label only | `DrugSafetyQuestionPairInteractionTest.aPairAlsoJoinedByARuleNamingAnActiveOrderStaysWithTheActiveOrderArm` |
| the reference-name arm SCANS instead of asking identity | `RuleTokenAliasOrderMatchingTest.aConstituentOrderDoesNotInheritTheCombinationProductsRules` + 2 |
| `matchesText` widened instead of adding `matchesDrugName` | `AllergenNameResolutionTest.proseMatchingIsNotWidenedWithTheAllergenSide`, `DrugSafetyOrderNameMatchingTest` |
| `containsToken` folds only the haystack | `ContraindicationTokenDiacriticFoldTest.theConditionArmFoldsTooAndSoDoesTheTokenSide` |
| `lookupByToken` reverts to the prose matcher | `AllergenNameResolutionTest` ×2 |
| `injectRecords` / `validate` do not enrich the context | `RuleTokenAliasOrderMatchingTest` ×3 |
| the #136 arm disabled outright (`if (false)`) | `RuleTokenAliasOrderMatchingTest` ×3 |
| **`resolvesFrom` asks the prose matcher again** | **nothing → now `oneCombinationOrderUnderALocalizedNameDoesNotInteractWithItself`** |
| **the flattened fallback asks the prose matcher again** | **nothing → now `aPairBetweenTwoLocalizedOrderNamesIsScreenedWithoutPerOrderStructure`** |
| **`containsToken` drops the empty-after-fold guard** | **nothing → now `aCuratedTokenOfNothingButCombiningMarksMatchesNothing`** |
| **`findForActiveOrders`' name leg reverts to the prose matcher** | **nothing → now the two new screening cases** |

The three new cases are defect-shaped, not coverage-shaped:

- **`resolvesFrom`** decides which order is a subject's own so the screening reduction can drop it.
  Asked through the prose matcher it does not recognise an INFLECTED order name as the subject's own,
  leaves it standing as somebody else's order, and the two halves of one tablet are reported as an
  interacting pair — issue #124's measured "Aspirin and omeprazole" shape, reached by a localized name.
  The case uses `Simvastatine et clarithromycine 500mg` (`X et Y` is this dictionary's own combination
  shape: `Salbutamol et bromhexine`, `Multivitamines et fer`) and pairs the absence with a
  two-separate-orders control, so the reduction cannot pass by suppressing every pair a localized name
  reaches.
- **The flattened fallback** (name sets, no per-order structure) carries no reference names by design,
  so the order-name leg is the only witness. Through the prose matcher two inflected order names
  resolve to nobody, no name is trusted at all, and a real Major pair silently stops being reported.
- **The empty-after-fold guard** is the only branch the fold ADDS rather than relaxes. A curated token
  of nothing but combining marks folds to the empty string, which every haystack contains, so without
  it the drug is contraindicated for **every patient carrying any allergy at all**.

A fourth test closes a *path* gap rather than a mutation gap: the widened subject resolution also
reaches the **#143 arm** — the patient's own orders against their own allergy records — and nothing
pinned it there for a localized name. That is where this defect does its real damage, because nobody
has to ask about the drug; the case needs both #147 fixes at once and neither the question nor the
answer naming it.

No existing test was touched: the test diff is +144 lines, **0 deletions**.

## One behaviour-preserving refactor

`validate` swept the dataset for the same resolution **twice** — once inside `withReferenceNames`,
which threw the entries away, and once for the arms that need them — while its comment said the
resolution happened once. The entries are now resolved first and passed to a package-private
`withReferenceNames` overload. Behaviour cannot change (neither leg of that resolution reads reference
names, so the second sweep could only ever return what the first did); the point is that the names the
context carries and the subjects the arms screen are now one resolution *by construction*.

## Prose corrections — mostly by deletion

- `bestRulePerPartner`'s **"merges exactly those and nothing else" is false.** The key is the entry
  `identifies` names, so two rules naming two *different* substances do merge where one entry's alias
  list claims both names. Measured on the full 19 MB KB: **148 of the 2093 rule tokens are claimed by
  more than one entry**, and the merge is reachable in **10 (order entry, token set) shapes — all
  inside the two substance families `pairKeyNames` already records** (the COVID-19 mRNA vaccines, the
  trastuzumab conjugates), where the dataset itself cannot tell the two apart. The absolute is gone; the
  sentence now points at that recorded bound instead of asserting a new one.
- Three sites still said the grouping key **is** the chip's label (`ruleAbout`, `partnerLabel`,
  `interactionWarning`). True only on the fallback branch now.
- `resolvesFrom`'s **"both legs of hasActiveDrug are covered"** and the screening arm's **"The join has
  two legs"** under-count: this PR gave that join a third arm.
- `addAllergyContraindications` described `lookupByToken` as resolving **"as a whole word"** — not since
  #147. Its 444/53/43/10 measurement was taken under the replaced matcher, so it was **re-measured
  through the shipped one: 391 self-resolve, 53 to another entry (all 53 earlier in dataset order),
  split 43 fragment / 10 claims-the-name, 0 resolve to nothing, 206 of 2283 overall.** Every figure
  unchanged — only the wording was stale.
- `containsBoundedToken`'s blank-alias example ("still matches any text carrying two adjacent spaces")
  is wrong in **both** directions, measured: `matchesText("peanut  oil")` is `false`, while
  `matchesDrugName("Vitamin A, B")` is `true` — so #147 widened the blank-alias hazard rather than
  leaving it alone. Example deleted, the widening stated, the fix left where the javadoc already puts
  it (an authoring guard in the curated parser). Filed below.
- `lookupByToken` and `matchesDrugName` advertised **"a condition naming a drug"** as an input shape; no
  caller passes condition text to either. `matchesDrugName` also promised a **blank** name never
  matches, which is false against an entry with a blank alias. Both clauses deleted.

`hasActiveDrug`'s "provably equal to [#136's] formulation" is **kept**, with its scope named: an
identity over the order-NAME leg (an entry supplies a name there exactly when the order name resolves
it — the same predicate that formulation applies to the partner), and a superset over the ATC leg,
which that formulation does not reach at all.

## The both-direction measurement, re-derived from scratch

Re-scored through the **shipped** matchers (real `DdiDrugReferenceSource.parse`, real
`DrugReferenceService`, real `hasActiveDrug` behind a real `withReferenceNames` context), against the
full 19 MB KB and a corpus pulled fresh from the standalone's own dictionary.

| arm | this PR reports | re-derived here | verdict |
|---|---|---|---|
| #136 rule token × order name | 907 → 1001, +94, 0 lost | 907 → 1004, **+97, 0 lost**, over 79 names / 27 tokens | baseline reproduces **exactly** |
| #147 allergen → entry | 549 → 624, +75, 0 lost | 552 → 626, **+74, 0 lost** | same magnitude |
| #147 order name → entry | +117 pairs, 0 lost | **+117 pairs, 0 lost** | reproduces **exactly** |
| kill set through both new arms | 0 leaking of 21 | **0 leaking** of the 20 documented nesting pairs | reproduces |
| dataset facts | 2283 entries / 2093 tokens / 5169 aliases / 444 ATC-less | identical | reproduces **exactly** |

The differences are corpus drift, measured: the dictionary now holds **620** `Drug`-class concepts and
my order-name corpus is **2537** names against this PR's 2533 and #129's 2531 — this session's own
fixtures added concepts. Every corpus-independent half (`0 lost`, `0 leaking`) reproduces.

Two further checks on the enumeration:

- **All 97 added pairs are explained**: in every one, an entry resolved from the order name claims the
  rule's token as one of its own names (0 unexplained), so the patient is on the drug the matched entry
  describes. The 13 whose *label* names a different substance reproduce **exactly** — `chlorine` ×3
  (from `Chlorpheniramine`), `sulfamethazine` ×6 (from `Sulfamethoxazole & Trimethoprim`),
  `esomeprazole` ×2, `isosorbide mononitrate` ×2.
- **The identity arm is structurally immune to the combination-alias hazard**, not merely lucky: 35 of
  the 2093 rule tokens contain a combination-looking separator and **none is a multi-substance
  combination whose constituents are separate entries** — they are comma-qualified single substances
  (`insulin, regular, human`, `thyroid, porcine`, `bcg, live, tice strain`). The CIEL ` / ` shape that
  makes `salicylic acid / urea` an alias of *Urea* never appears as a token, because a token is an
  `rxnorm_name` (one ingredient) while combination names enter only from CIEL. Left out of the javadoc
  deliberately: the existing text already says why the scan is wrong, and a new number there would rot.

The fixture was verified rather than trusted: `ddi-alias-drug-names.json` is **byte-for-byte the KB's
own** — 11 drug rows identical, 13 mechanisms identical, exactly the 32 interaction rows wholly inside
the slice, none missing, none extra, no unreferenced mechanism.

## Live verification (my own run, label `h148`, lock taken 19:29:53)

Full JVM restart on a build of this branch, retrieval gate passed, LLM warm, `sourceFormat=ddinter`
with **2283 entries** confirmed loaded. 13 probes, all as expected. Chips only — answer wording is not
reproducible on that box.

| probe | shape | chips | discriminates because |
|---|---|---|---|
| Mary Smith + clarithromycin | control | **1** Major ×simvastatin | ✔ |
| Agnes Adams + warfarin | control | **1** Major ×aspirin | ✔ |
| Joshua Johnson + ibuprofen | control | **2** (NSAID group + lisinopril Moderate) | ✔ |
| Mary Smith + paracetamol | control | **0** | ✔ |
| Kevin Brown + clarithromycin | #147 headline | **1** `recorded allergy to Clarithromycin` | prose matcher → 0 |
| David Williams + warfarin | #136 headline | **1** Major ×aspirin | pre-#136 → 0 |
| Susan Young + **opium** | the kill case | **1** Tiotropium allergy, **no Opium chip** | Opium IS in play; a leak would say "allergy to Opium" |
| **Sarah Taylor** (8 orders) + screening | polypharmacy, unprobed shape | **23** (13 contra + 10 capped pairs) | the cap and the screen both live |
| Barbara Miller + screening | grouping control | **3**, exactly **one** interaction chip for the one pair | label-keyed grouping → 2 |
| Richard Jones + dexamethasone | allergy identity + route variants | **4** | ✔ |
| Michael Turner + medications | order names matching no entry | **0**, both orders still listed | the widening matches nothing arbitrary |
| Betty Williams + aspirin | allergy, **0** orders | **1** identity chip | ✔ |
| Mary + Susan, re-run warm | cold-start-race control | identical chips | absence above is not a cold KB |

Sarah Taylor's 23 chips are a shape nobody had probed. **Not caused by this PR** — measured: for her 8
order names the prose matcher and the drug-name matcher resolve *the same* 12 (order name, entry)
pairs, 0 added, 0 lost, so her output is identical on `main`. Filed below as its own defect.

A second pass under `sourceFormat=json` (the shipped default, where #141 lives) is running as I post
this; results follow in a reply. It gates on the load line reporting a NON-ZERO entry count before
reading anything, because the mismatch that loads 0 entries logs at INFO (#149) and would turn every
probe into a silent abstention.

## Tests

**911** (API 853 + OMOD 58), 0 failures, 34 skipped — measured with `mvn -o clean install` from the
repo root on a quiet tree at the final SHA. 907 when this hardening started; **893** on `main` at
`b013626`, also measured here rather than taken on trust.

One process note, because it nearly shipped a disabled feature: a review agent I spawned inherited
this worktree, and one of its live mutations was captured by a `git commit -a` of mine into `5d7b0e4`
— `hasActiveDrug`'s #136 arm became `if (false)` behind a commit message quoting a green build
measured seconds earlier. It was caught and restored in `5a4f5b9`; I have since verified the branch
byte-for-byte (`9fe2ce8..HEAD` touches only the intended test and comments), confirmed no `if (false)`
anywhere in `api/src/main`, re-run the mutation sweep in an isolated copy, and every commit since is
path-limited. The reversion that reproduces exactly that contamination is now row 8 of the table
above, caught by three tests.

## Found while hardening — reported, not fixed here

1. **DEFECT (corroborates #145, now with live numbers).** The contraindication arms emit one chip per
   route-variant ENTRY, with no per-substance grouping — the thing #115 fixed for interaction chips
   (`bestRulePerPartner`) and #131 bounded (`maxPairChips`). Measured on my run: **Richard Jones** →
   4 chips for one allergen (1 identity + `Dexamethasone (nasal)`/`(ophthalmic)`/`(topical)` all reading
   "is in the same ATC class (A01AC) as the patient's allergy to Dexamethasone"); **Sarah Taylor**
   (8 orders) → **13** contraindication chips, 11 of them route/class variants. Since #110 each is also
   an injected citable record, so they crowd the prompt: on Sarah's probe the answer led with an
   interaction rather than with her allergy. **Not caused by this PR** — measured: her 8 order names
   resolve the same 12 (order name, entry) pairs under both matchers.
2. **DEFECT (new).** A blank alias in a hand-authored curated KB captures allergens, and #147 widened
   it. `JsonDrugReferenceSource.parse` drops blank `id`/`name` but not blank aliases; measured through
   the real parser, an entry with `"aliases":[" "]` resolves the allergen `Vitamin A, B`
   (`matchesDrugName` true) where the prose matcher did not (`matchesText` false). That allergen then
   drives the identity test, the class comparisons and the name the chip renders. Reachable only in
   operator-authored `json` data — `ddinter`/`atc` drop blank aliases at parse. Fix where the javadoc
   already says it belongs: drop blank aliases in that parser.
3. **HAZARD (new), refining the injector/validator asymmetry already noted.** `matchingEntries` still
   resolves order-driven reference records by ATC only, so a chip can name an active order for which no
   *reference* record was injected. The earlier note guessed this might be unreachable because
   `relatedToAny` needs ATC codes — reading it settles that it needs the **entry's** codes, not the
   order's, and a name-resolved entry usually has them. Concrete reachable shape: an order whose concept
   carries no ATC map (`Acetylsalicylic acid`) plus a question naming a drug sharing one of that
   *entry's* subgroups (clopidogrel, B01AC) → chip, no reference record. Reasoned from code, not probed.
4. **HAZARD (new).** The pre-answer pass resolves the active orders twice: `injectRecords` enriches,
   then `preAnswerFindings` → `validate` resolves again. Idempotent today (neither leg reads the names it
   attaches) so it is cost, not a defect — but it becomes a fixpoint iteration the moment
   `findForActiveOrders` reads `getActiveDrugReferenceNames()`. Fix shape: a package-private
   `validate(…, List<DrugReference> orderEntries)` seam fed by `preAnswerFindings`, the same pattern now
   used inside `validate`. Recorded in the comment there.
5. **OBSERVATION.** Exactly **1 of the 295,184** interaction rows in the shipped KB has
   `drug_a == drug_b` (`DDInter225` Botulinum toxin type A, Moderate — whose mechanism text is about
   *different* botulinum serotypes), and `DdiDrugReferenceSource.parse` does not skip it, so a patient on
   that drug can be told it interacts with itself. Pre-existing (the token sits inside the entry's own
   name, so the order-name arm already reached it); #136 widens the reach to an order named by a CIEL
   alias. Fix: skip `a == b` rows at parse.
6. **CORPUS DRIFT** (measured against the live DB today, for whoever probes next): Betty Williams has
   **0** active drug orders (this PR's comment says 1 — that Simvastatin order is Mary Smith's);
   **Sarah Taylor has 8** (the shared brief lists her as having none); Kevin Brown has exactly **one**
   `Clarithromycine` allergy row, so a one-chip-per-allergen assertion on him is safe; Richard Jones has
   a Dexamethasone order **and** allergy; `Drug`-class concepts are now **620** where three javadocs
   cite 616 (dated).
7. **STALE-IN-WAITING, deliberately not edited** to stop prose churn: `DrugReference`'s class-level
   "Matching keys" list still calls `getAliases()` "for question-driven matching" (it is now also the
   drug-name haystack, the identity operand, and the source of the context's reference names), and two
   sites explain multi-rule collapse only via "route variants publish the same match token" when two
   *different* tokens naming one entry now also collapse. Both are incomplete rather than false.

dkayiwa:
## Follow-up: the `sourceFormat=json` pass for #141 could NOT be run soundly — and why that is itself a finding

I attempted the second live pass under the shipped default source format, where #141 lives. It did not
produce #141 evidence, and the reason is worth recording because it is the same class of trap as #149.

**What happened.** With both GPs flipped correctly (`sourceFormat=json` HTTP 200, `dataFilePath=''`
HTTP 200), the module restart used to make the change take effect returned **HTTP 500** and did not
stop the module. The bean therefore kept its DDInter entries for the whole pass, and every probe ran
against the 19 MB KB while the GP table said `json`.

**My log-line gate passed anyway, for the wrong reason.** The dataset load is LAZY
(`DrugReferenceService.ensureLoaded()` on the first `getAll()`), so immediately after a flip the newest
`Loaded N … entries` line in the log is still the PREVIOUS source's. My gate read `2283` and concluded
the KB was healthy. That is a refinement worth adding to **#149**: *a non-zero entry count in the log
is not evidence that the configured source is the one in force* — only a query that triggers a load,
or a positive control chip the new KB alone can produce, discriminates.

**What did discriminate** was the positive control I ran before the #141 pair: Barbara Miller's chips
came back carrying DDInter mechanism prose (`… interacts with active order aspirin — Major. The
antiplatelet and cardioprotective effect of low dose aspirin …`), which the 4-entry curated seed cannot
produce. Edward Robinson then returned **0 chips** — correct for `ddinter`, which emits no
contraindications at all, and proof the flip had not taken effect rather than evidence about the fold.

**Why I did not force it.** Per this instance's recorded experience, the `moduleaction` restart is a
trap in both directions: it returns 500, and when it *does* stop the module the subsequent `start`
silently kills querystore retrieval (every chart comes back empty, so answers become plausible
abstentions with zero references). The only sound mechanism is a full-JVM restart — which is exactly
what the shared verify script does, but its readiness gate is a `ddinter`-specific control chip
(Mary × clarithromycin, Major) that a 4-entry curated KB cannot satisfy, so the pass aborts before any
probe. Running it would mean changing that gate, which is shared infrastructure and out of this PR's
scope.

**What #141 rests on, therefore:** its unit coverage — 11 measured fr locale-preferred names asserted
one by one, both fold directions, the fragment-matching guard, plus the new empty-after-fold guard —
and the author's earlier json pass, which gated on a **curated positive control** (`Loaded 4 …` plus
Barbara's `documented ibuprofen allergy` chip) rather than on a log line, and showed Edward
(`Penicillin G`) → 1 chip / Donna (`Pénicilline G`) → 1 chip. My hardening changed no code in that arm.

**Instance left clean, verified rather than assumed:** both GPs restored
(`sourceFormat=ddinter`, `dataFilePath=chartsearchai/ddi_knowledge_base.json`), 2283 entries in force,
and a closing control probe — Mary Smith × clarithromycin → **1 chip, Major, ×simvastatin**. No demo
data was created or deleted in either of my passes.

dkayiwa:
Corpus corrections measured during hardening, for anyone reading the probe notes above:

- **Betty Williams has 0 active orders.** An earlier comment on this PR says she has one; that order is Mary Smith's. Any assertion built on Betty-plus-an-order is vacuous.
- **Sarah Taylor now has 8 active orders** (was recorded as "a chart and no orders").
- **Melissa Wright has 1** (Spironolactone) — Ondansetron moved to Mark Williams.
- **Kevin Brown has exactly one `Clarithromycine` allergy row** (plus two levothyroxine). The "two Clarithromycine rows" seen earlier was a locale-join artifact, so a one-chip-per-allergen assertion on him is safe.
- `Drug`-class concepts are now **620**, not the 616 cited in three javadocs.

Independent verification before merge (mine, not relayed):

- Final SHA `6b55235`: `hasActiveDrug`'s reference-name arm and `matchesDrugName`'s `matchesOrderName` delegation both intact; no `if (false)`/`if (true)` anywhere in `api/src/main`.
- Test changes over the whole branch are **pure additions** — +131 lines across 4 files, **0 removed lines** — so no existing assertion was weakened.
- `mvn -o clean install` on a clean detached worktree at `6b55235`: **853 + 58 = 911, 0 failures, 34 skipped**, BUILD SUCCESS. This reproduces the reported figure at the SHA being merged, which matters because `5d7b0e4`'s quoted count was measured on a tree that was mutated immediately afterward (see the restore in `5a4f5b9`).

PR #154, opened 2026-08-05
fix(drug-reference): make an empty dataset load loud, and the loaded source observable

### LINKED ISSUE #149: A sourceFormat/dataFilePath mismatch loads 0 entries and logs it at INFO, silently disabling the entire drug-safety feature
### Summary

A `sourceFormat` / `dataFilePath` mismatch loads **zero** drug-reference entries, turning the entire drug-safety feature off, and reports it at `INFO` as an ordinary success. The module starts with `started=true`, no `startupErrorMessage`, no WARN, no ERROR. Every safety question then answers as though the patient had no interactions, allergies or contraindications to find.

`ReferenceDataFiles.loadWithClasspathFallback` logs the outcome at `:65`:

```java
log.info("Loaded {} {} from {}", loaded.size(), datasetLabel, resolved);
```

`loaded.size()` of **0** prints as cheerfully as 2283. The two loud paths do not cover this case:

- `:81` `log.warn` — bundled default missing from the classpath
- `:90` `log.error` — bundled default unparseable

Neither fires when the configured file is *present and readable* but the **wrong format for the configured parser** — `DdiDrugReferenceSource.parse` on a curated `drug-reference.json` (or `JsonDrugReferenceSource` on the DDInter KB) finds none of the keys it expects, returns an empty list, and does not throw.

### How it was found, which is the point

It **invalidated a verification pass**. An agent flipped `sourceFormat` to `json` to exercise #141's curated path while leaving `dataFilePath` pointed at the 19 MB DDInter file. The instance came up clean, the module reported started, and every probe returned zero chips — which looks exactly like a fix that does not work. The mismatch was only caught because that agent re-checked the load line and found `Loaded 0`.

So the failure mode is not hypothetical operator error; it silently produced a wrong experimental result inside this project, and would as easily produce a wrong clinical one.

### Severity

An operator who mis-pairs these two GPs gets a module that looks healthy and answers safety questions with silence. Every defect this project has fixed — the direct-allergy gap, the prescribed-and-allergic suppression, the fabricated chips — is invisible in that state, because so is everything else. It is the one configuration error that disables the whole safety layer without saying so.

### Proposed fix

1. **Log a zero-entry load at WARN**, naming the configured `sourceFormat` and `dataFilePath`, and saying plainly that drug-safety checking is inert. A count of zero is never a normal outcome for a shipped-with-defaults feature.
2. Consider **detecting the mismatch specifically**: each source knows the shape it expects (`DdiDrugReferenceSource` requires `drugs` and `interactions`; the curated parser expects a list/`entries`). A parser that finds a well-formed JSON document lacking *all* of its expected keys can say "this looks like the other format" rather than "0 entries", which is the difference between a five-minute and a five-hour diagnosis.
3. Surface it where an operator will see it — `startupErrorMessage` or the module's admin page — rather than only in a log they have no reason to read. Worth weighing: an inert safety net arguably *should* block startup, but that is a policy call for the maintainers, not something to change unilaterally.

### Tests

Through the real production path: `DrugReferenceService.getAll()` with a deliberately mismatched format/path pair returns empty **and** the load is reported at WARN. A matched pair still logs INFO. Assert on the logged level, not the message text, so a re-wording does not silently drop the guard — note the codebase has no log-capture harness today, so establishing one is part of the work (an earlier pass recorded that gap when it could not assert a WARN's content).

Found while fixing #147/#141/#136 (PR #148); reported there and filed here because it is unrelated to that fix and affects any deployment that edits either GP.

dkayiwa:
## The defect

A `sourceFormat` / `dataFilePath` mismatch loads **zero** drug-reference entries and was reported at INFO exactly like a successful load:

```java
// ReferenceDataFiles:65
log.info("Loaded {} {} from {}", loaded.size(), datasetLabel, resolved);
```

`loaded.size()` of 0 printed as cheerfully as 2283. The module starts with `started=true`, no `startupErrorMessage`, no WARN, no ERROR — and the entire drug-safety layer is inert: every safety question answers as though the patient had no interactions, allergies or contraindications to find. The two loud paths do not cover it. `:81` fires only when the bundled default is **missing** from the classpath; `:90` only when it is **unparseable**. A file that is present, readable and simply the wrong shape for the configured parser returns an empty list without throwing — `DdiDrugReferenceSource.parse` finds no `drugs`/`interactions`, `JsonDrugReferenceSource.parse` finds no `entries`.

It is not a hypothetical operator error. It silently produced a wrong experimental result inside this project: a verification pass flipped `sourceFormat` to `json` while leaving `dataFilePath` on the 19 MB DDInter KB, saw every probe return zero chips, and could not tell that from "the fix does not work".

## Root cause

Two things, and the second is why the first could not previously be verified.

1. **Nothing decided that zero is different.** The load's only record of what happened was a transient log line whose severity did not depend on the outcome.
2. **Nothing retained the outcome.** The load is lazy and cached for the life of the module, so "which dataset is in force?" cannot be answered from the log at all — the most recent `Loaded N …` line may pre-date the global properties as they read now, or belong to a process that a failed restart left running. That is exactly how the pass above was fooled: it re-read the load line, found a stale one, and passed.

## What changed

**Loud on empty.** `DrugReferenceService.ensureLoaded` now warns when the load it just performed produced no entries, naming both global properties, the parser actually in use and the file actually read. It lives in the service, not in a source adapter, so one rule covers all three formats (`json`/`atc`/`ddinter`) and cannot drift per source; the adapters' existing INFO lines are unchanged.

**Observable after the load.** `DrugReferenceLoad` retains the outcome — effective format, configured format, configured path, the **origin actually read**, entry count, and the inert verdict — built at the instant the load populates the cache, so it can never describe a different dataset than the one the safety layer is using. Exposed by `DrugReferenceService.getLoadStatus()` and `GET /chartsearchai/drugreferencestatus` (privilege: the core `Get Global Properties`; no new privilege). Reading it performs the load if it has not happened yet, so the answer is current by construction rather than being a historical line that may be stale — that is the property that makes any source-flip verification, including this PR's own, trustworthy.

`origin` is reported separately from `configuredDataFilePath` because a configured path that cannot be read falls back to the bundled dataset and yields a perfectly plausible non-zero count. That is the state in which "the count is non-zero, so my file loaded" is false, and an entry count alone cannot distinguish it. Measured live below (L3).

`DrugReferenceLoad.isInert()` is the single verdict driving both the WARN and the reported status, so the two cannot disagree.

### Which states are loud and which are silent

| state | verdict | why |
|---|---|---|
| `drugReference.enabled=false` (the default) | **silent**, `loaded:false` | Nothing is configured to load, so nothing is loaded. Neither `DrugSafetyValidator` nor `DrugReferenceInjector` reaches the service when the master switch is off, so no load happens and there is nothing to warn about. Reading the status in this state deliberately does **not** trigger a load — polling a status endpoint must not be what starts a 19 MB parse, or manufactures the warning, on an install that does not use the feature. |
| enabled, matched format/path, N > 0 | **silent** (INFO as before) | The healthy path. A warning here would be noise every install learns to ignore. |
| enabled, no `dataFilePath`, bundled default loads | **silent** | Shipping defaults is the normal state. |
| enabled, a source was selected and produced **0** entries | **WARN**, `inert:true` | The defect. Covers the format/path mismatch in both directions, an empty or garbage operator file, a bundled dataset that parses to zero, and `sourceFormat=atc` with no dataset path — all of which leave the safety layer inert while the module looks healthy. |

Reaching the load at all already implies the feature is on and something asked it to work, which is what lets "a configured source yielded nothing" be distinguished from "nothing is configured" without a second global-property read.

## Tests

`DrugReferenceLoadContextTest` (10 new tests) drives the real production load path — `DrugReferenceService.getAll()` / `getLoadStatus()` over the real global properties — and builds the mismatch out of the module's **own two bundled datasets**, copied into the application data directory and cross-wired so each is parsed by the other's parser. No synthetic file, no hand-crafted parser input.

Assertions are on the logged **level**, never the message text: an empty list is the correct fail-safe return in both the healthy and the misconfigured case, so the level is the only observable difference, and a text assertion would let a re-wording silently drop the guard. `LogCapture` is the harness that makes that possible (the codebase had none). It captures INFO as well as WARN for a reason beyond the INFO assertion — a capture that silently received nothing at all would make "no WARN was logged" pass vacuously.

### Red first

With the WARN block removed from the otherwise-complete change, on a throwaway worktree:

```
[ERROR] DrugReferenceLoadContextTest.ddinterFormatPointedAtTheCuratedDatasetLoadsNothingAndIsReportedAtWarn:119
  a configured source that resolves to 0 entries must NOT be reported like a successful load — the whole
  drug-safety feature is inert.
  Captured: [INFO Loaded 0 DDInter drug-reference entries from …/chartsearchai/mismatch-curated.json]
[ERROR] DrugReferenceLoadContextTest.curatedFormatPointedAtTheDdiKnowledgeBaseLoadsNothingAndIsReportedAtWarn:134
  the mismatch is symmetric and must be loud in this direction too.
  Captured: [INFO Loaded 0 drug-reference entries from …/chartsearchai/mismatch-ddi.json]
```

The captured event is the reported defect verbatim.

### Mutation sweep

Five mutations of the fix, applied one at a time on a throwaway worktree (`git worktree add`, restored and verified clean afterwards — `git status --porcelain` empty). No survivors:

| mutation | caught by |
|---|---|
| A — drop the loud-on-empty WARN (pre-fix behaviour) | both `…IsReportedAtWarn` tests |
| B — `getLoadStatus()` reports "not loaded" instead of triggering the lazy load | 4 status tests |
| C — report the configured path as if it were the origin | `loadStatusNamesTheBundledOriginWhenTheConfiguredFileIsAbsent` |
| D — re-derive the status from the GPs at read time instead of returning the retained outcome | `loadStatusDoesNotDriftFromTheCachedEntriesWhenTheGlobalPropertiesChange` |
| E — `isInert()` ignores whether a load happened | `loadStatusOfADisabledFeatureReportsNoLoadAndTriggersNone` |

## Build

`mvn -o clean install`, full reactor, measured on this branch:

```
BASELINE (main @ 03230b3):  api 853, omod 58  = 911 tests, 0 failures, 34 skipped
THIS BRANCH:                api 863, omod 58  = 921 tests, 0 failures, 34 skipped
```

BUILD SUCCESS both times. +10 tests, no existing test modified (`git diff main...HEAD` touches no existing test file).

## Live standalone evidence

Deployed this branch's omod to the 3.7.1 standalone (`llm.engine=local`, `chartMode=fullChart`, `minInteractionSeverity=minor`) via the shared verify harness, which gated on live querystore retrieval (`hits=75`) and the Mary+clarithromycin control chip before handing over. Five configurations, each behind a **full JVM restart** — the load is lazy and cached, so nothing short of a restart re-reads the global properties. Patients resolved by name lookup against the live database, not from a report: `Mary Smith 38beca4a-fccf-40e5-907d-1bbbc173b93b`, `Agnes Adams 47e57b75-2604-4eb5-8586-e2c8b3a96b28`, `Joshua Johnson 9cb37bcb-95a2-4517-bf9b-e74c75c6acfa`.

Nothing was created or deleted; only the three `drugReference` global properties were changed, and all three were restored (verified after the run).

### L1 — healthy `ddinter`, the shipped configuration

```
--- GET /chartsearchai/drugreferencestatus  [L1-healthy] ---
{
  "enabled": true,
  "loaded": true,
  "inert": false,
  "entryCount": 2283,
  "sourceFormat": "ddinter",
  "configuredSourceFormat": "ddinter",
  "configuredDataFilePath": "chartsearchai/ddi_knowledge_base.json",
  "origin": "/Users/danielkayiwa/Downloads/referenceapplication-standalone-3.7.1/appdata/chartsearchai/ddi_knowledge_base.json"
}
INFO - DdiDrugReferenceSource.parse(210) | Parsed 2283 DDInter drug-reference entries
INFO - ReferenceDataFiles.loadWithClasspathFallback(110) | Loaded 2283 DDInter drug-reference entries from …/appdata/chartsearchai/ddi_knowledge_base.json
    WARN/ERROR count from the reference package: 0
```

Four regression controls, all as expected:

```
Mary   + "Is it safe to give her clarithromycin?"  -> 1 chip
  [interaction] drug='Clarithromycin'
      Clarithromycin interacts with active order simvastatin — Major. Coadministration with potent
      inhibitors of CYP450 3A4 may significantly increase the plasma concentrations of simvastatin …
Agnes  + "Is it safe to start warfarin?"            -> 1 chip
  [interaction] drug='Warfarin'
      Warfarin interacts with active order aspirin — Major. Aspirin, even in small doses, may increase
      the risk of bleeding in patients on oral anticoagulants …
Joshua + "Can I give ibuprofen?"                    -> 2 chips
  [contraindication] Ibuprofen is in the same cross-reactivity group (NSAID) as the patient's allergy
      to Acetylsalicylic acid (aspirin) — possible cross-reactivity
  [interaction]     Ibuprofen interacts with active order lisinopril — Moderate. …
Mary   + "Is it safe to give her paracetamol?"      -> 0 chips
```

Plus the source-discriminating control used again in L3 — same patient, same question, **DDInter** wording:

```
Agnes  + "Can I give ibuprofen?"                    -> 1 chip
      Ibuprofen interacts with active order aspirin — Major. The antiplatelet and cardioprotective
      effect of low-dose aspirin may be antagonized by coadministration of some nonsteroidal
      anti-inflammatory drugs (NSAIDs). … The mechanism is competitive inhibition of platelet
      cyclooxygenase by certain NSAIDs …
```

**Had the change been wrong** (over-warning): the WARN count on a healthy 2283-entry load would not be 0, and the warning would become noise on every install. It is 0.

### L2 — the reported defect: `sourceFormat=json` left pointing at the DDInter KB

```
    set chartsearchai.drugReference.sourceFormat = json
    set chartsearchai.drugReference.dataFilePath = chartsearchai/ddi_knowledge_base.json
--- restarting JVM ---
    module started after ~15s
--- GET /chartsearchai/drugreferencestatus  [L2-mismatch] ---
{
  "enabled": true,
  "loaded": true,
  "inert": true,
  "entryCount": 0,
  "sourceFormat": "json",
  "configuredSourceFormat": "json",
  "configuredDataFilePath": "chartsearchai/ddi_knowledge_base.json",
  "origin": "/Users/danielkayiwa/Downloads/referenceapplication-standalone-3.7.1/appdata/chartsearchai/ddi_knowledge_base.json"
}
INFO - ReferenceDataFiles.loadWithClasspathFallback(110) |2026-08-05T20:49:27,868| Loaded 0 drug-reference entries from /Users/danielkayiwa/Downloads/referenceapplication-standalone-3.7.1/appdata/chartsearchai/ddi_knowledge_base.json
WARN - DrugReferenceService.ensureLoaded(324) |2026-08-05T20:49:27,875| Loaded 0 drug-reference entries — drug-safety checking is INERT: no interaction, allergy or contraindication warning can be raised, and every safety question will answer as though there were nothing to find. chartsearchai.drugReference.sourceFormat=json (parser in use: json), chartsearchai.drugReference.dataFilePath=chartsearchai/ddi_knowledge_base.json, read from /Users/danielkayiwa/Downloads/referenceapplication-standalone-3.7.1/appdata/chartsearchai/ddi_knowledge_base.json. The usual cause is a format/path mismatch: each source format parses only its own shape and returns nothing — without failing — for another's.
    WARN/ERROR count from the reference package: 1

Mary + "Is it safe to give her clarithromycin?"  -> 0 chips   ("The records do not address clarithromycin.")
Mary + same question, re-run                     -> 0 chips
```

The first line is the defect verbatim — the INFO that reports 0 exactly as it reports 2283. The WARN under it is new, and it names both global properties, the parser in force and the file read, which is the difference between a five-minute and a five-hour diagnosis. Note the status endpoint was called **before any search**, and returned `loaded:true, entryCount:0`: reading it is what performed the load, which is why its answer cannot be a stale line.

**Had the change been wrong:** this state produces the same 0 chips and the same INFO line as before, with nothing at default log levels to distinguish "the safety layer is off" from "there was nothing to warn about" — which is exactly the reading that invalidated the earlier verification pass. The re-run is there because a first post-restart absence is not trustworthy on this box; both agree, and the status endpoint makes the absence attributable rather than inferred.

### L3 — the trap: a configured path that cannot be read, silently serving the bundled dataset

```
    set chartsearchai.drugReference.sourceFormat = json
    set chartsearchai.drugReference.dataFilePath = chartsearchai/i149-no-such-file.json
--- restarting JVM ---
--- GET /chartsearchai/drugreferencestatus  [L3-bundled-fallback] ---
{
  "enabled": true,
  "loaded": true,
  "inert": false,
  "entryCount": 4,
  "sourceFormat": "json",
  "configuredSourceFormat": "json",
  "configuredDataFilePath": "chartsearchai/i149-no-such-file.json",
  "origin": "classpath:/chartsearchai/drug-reference.json"
}
    WARN/ERROR count from the reference package: 0
```

`entryCount: 4` is non-zero and looks perfectly healthy. `origin` is what says the entries came from the **bundled** curated dataset, not from the file that was configured. This is the state in which a count-only check concludes "my source loaded" and is wrong.

Positive control, proving the loaded source really is the curated one and not a stale reading — same patient and question as L1, different wording, because the note text can only come from the curated dataset's hand-authored, severity-less `ibuprofen × aspirin` rule:

```
Agnes + "Can I give ibuprofen?"                   -> 2 chips
  [interaction] Ibuprofen interacts with active order aspirin — additive GI and bleeding risk
  [interaction] Ibuprofen is in the same cross-reactivity group (NSAID) as active order N02BA01 …
Mary  + "Is it safe to give her clarithromycin?"  -> 0 chips
Mary  + same question, re-run                     -> 0 chips
```

`additive GI and bleeding risk` appears nowhere in the DDInter KB; L1's chip for the identical probe carried the DDInter mechanism text and a `Major` severity. And Mary's L1 Major chip is gone, because clarithromycin is absent from the 4-entry curated dataset. Both directions confirm the switch actually happened.

**Had the change been wrong:** with only a count to go on, `4` is indistinguishable from a successful load of the configured file, and this is the precise shape of the failure #149 was reported from.

### L4 — `drugReference.enabled=false`: the legitimate silent state

```
    set chartsearchai.drugReference.enabled = false        (ddinter GPs left in place)
--- restarting JVM ---
--- GET /chartsearchai/drugreferencestatus  [L4-disabled] ---
{
  "enabled": false,
  "loaded": false,
  "inert": false,
  "entryCount": 0,
  "sourceFormat": null,
  "configuredSourceFormat": null,
  "configuredDataFilePath": null,
  "origin": null
}
    (no drug-reference load lines at all)
    WARN/ERROR count from the reference package: 0

Mary + "Is it safe to give her clarithromycin?"  -> 0 chips
```

`entryCount` is 0 here as in L2, and `inert` is deliberately **false**: nothing was configured to load, so nothing was loaded and nothing is wrong. Note the absence of any load line — polling the status endpoint did not trigger the 19 MB parse or the warning.

**Had the change been wrong** (warning on any empty dataset rather than on an empty *load*): every install that leaves this feature at its default `false` would get the inert WARN, which is the failure mode that makes warnings worthless.

### L5 — restored to healthy `ddinter`; the four controls again

```
--- GET /chartsearchai/drugreferencestatus  [L5-restored] ---
{ "enabled": true, "loaded": true, "inert": false, "entryCount": 2283, "sourceFormat": "ddinter",
  "configuredDataFilePath": "chartsearchai/ddi_knowledge_base.json",
  "origin": "…/appdata/chartsearchai/ddi_knowledge_base.json" }
    WARN/ERROR count from the reference package: 0

Mary   + clarithromycin -> 1 chip  [interaction] Clarithromycin × simvastatin — Major
Agnes  + warfarin       -> 1 chip  [interaction] Warfarin × aspirin — Major
Joshua + ibuprofen      -> 2 chips [contraindication] NSAID cross-reactivity + [interaction] lisinopril Moderate
Mary   + paracetamol    -> 0 chips
Agnes  + ibuprofen      -> 1 chip  [interaction] × aspirin — Major, DDInter mechanism wording (as L1)
```

The status tracked every one of the four flips and landed back on `ddinter`/2283 — which is the whole point of the second half: a source flip is now checkable from the running module rather than from a log line that may belong to a different load. The four controls are byte-comparable to L1 in chip count, type and partner (wording differs run to run by design — `cache_prompt` KV reuse makes answer prose non-reproducible on this box, so nothing here asserts on wording).

## Limitations

- ~~The REST endpoint itself has no unit test. The omod tests instantiate the controller as a POJO and cannot set its autowired `DrugReferenceService`…~~ **Withdrawn by the hardening pass below: that was wrong.** This controller already carries three package-private `Test seam:` setters for its other autowired collaborators, and the omod test lives in its package; a fourth was added and the endpoint now has three tests. The claim that the wire shape was pinned by `loadStatusSerializesTheFieldsTheStatusEndpointReturns` was also only two-thirds true — that test pins the seven fields of `DrugReferenceLoad.toMap()`, but not the `enabled` field the controller contributes, nor the order, nor the privilege gate.
- Reading the status when the source throws propagates the exception and yields a clean 500 rather than a diagnostic body. A throwing source already violates the documented `DrugReferenceSource` contract ("never an exception"), and the two production consumers catch it, so this is the pre-existing contract boundary rather than new behaviour.
- `DrugReferenceLoad.getSourceFormat()` reports the format the global property selects. With a source injected through the package-private test seam that describes the adapter that *would* have been used, not the injected one; production never injects a source. Documented on the seam.

## Deliberately not fixed here

Both are visible in the live evidence above and are outside what #149 reports; reported rather than
fixed so they get their own change.

- A configured `dataFilePath` that **cannot be read** silently falls back to the bundled dataset and
  is logged only at INFO (`ReferenceDataFiles` `IllegalStateException` branch). Same
  silent-misconfiguration family, and it yields a plausible non-zero count — L3 above measures it.
  This PR makes it *observable* (`origin`) but not *loud*, which is a policy call for the maintainers.
  The standalone is in this state right now for a different file: its
  `crossReactivityGroupsFilePath` points at a `chartsearchai/cross-reactivity-groups.json` that does
  not exist in appdata, so it silently runs the bundled groups.
- A `sourceFormat` value matching no adapter (a typo, e.g. `ddi`) silently falls back to the curated
  `json` parser. Now visible — `sourceFormat` and `configuredSourceFormat` differ in the status — but
  not warned about.

---

## Hardening pass (post-review)

Nine commits on top of the change above. One is a behaviour fix found by probing the new endpoint's
privilege gate; the rest is the coverage the change was missing, and documentation where an operator
looks.

### 1. `origin` no longer discloses the server's absolute path

The endpoint is gated on the core `Get Global Properties` privilege, which is right for what the body
mostly is. But **that privilege is one of the 27 the `Authenticated` role holds by default** on a
Reference Application install, so every logged-in user can read it. Measured on the 3.7.1 standalone
(2026-08-05) with a user whose only role carried **no privileges at all**:

| call | result |
|---|---|
| `GET /chartsearchai/drugreferencestatus` | **200** — full body, including the absolute path |
| `GET /systemsetting/chartsearchai.drugReference.dataFilePath` | **200** |
| `GET /systeminformation` (core's own appdata disclosure) | **403** |

So the two configured global properties disclose nothing that user could not already read — the
privilege choice is consistent for them. The **absolute filesystem path** was not: core keeps its own
disclosure of the application data directory behind `View Administration Functions`, and this handed
it to any authenticated caller.

`origin` now names an operator file as `appdata:<path within the application data directory>` — the
same form the global property holds — beside the existing `classpath:<resource>`. Nothing diagnostic
is lost, and one thing is gained: `resolveModelPath` rejects `..` and confirms the file resolves
inside that directory, so the relative form names the file exactly, and *"did my configured file
load?"* becomes `origin == "appdata:" + configuredDataFilePath` rather than a suffix match on an
absolute path. Both forms now carry a prefix, so a reader no longer has to infer which space an
unprefixed path came from. The absolute path is still logged at INFO, where the audience is already
an administrator.

### 2. The REST endpoint is now tested

The limitation recorded below — that the omod tests cannot set the controller's autowired
`DrugReferenceService` — does not hold: **this controller already carries three package-private
`Test seam:` setters** for exactly that (`setChartSearchService`, `setAuditLogService`,
`setPatientAccessCheck`), and the omod test lives in its package. A fourth was added.

`ChartSearchAiDrugReferenceStatusTest` (3 tests) drives the real controller method over a real
`DrugReferenceService`:

- the exact 8-field body **in order**, including `enabled`, which the controller contributes and which
  no API-level test can see;
- `enabled` is a **live** read of the master switch while every other field describes the retained
  load, so flipping the switch off after a load still reports `loaded:true` — the one genuinely
  confusable part of this contract, and it was pinned by nothing;
- an unprivileged caller is **refused**, with the service set, so removing the gate makes the call
  return 200 and fail the test.

### 3. The load's concurrency is now exercised

`DrugReferenceLoadConcurrencyTest` races eight callers into the lazy load and asserts the dataset is
read **once**, the inert WARN is logged **once** rather than once per caller, and every caller is
handed the same retained outcome as the entries it describes. Load count comes from the source rather
than the log, so it does not depend on which dataset the global properties select in the shared test
JVM.

### 4. Three places the origin change had left behind

Relativizing `origin` updated README, ONBOARDING, `DrugReferenceLoad`, `ReferenceDataFiles` and the
controller — and missed three. None is a behaviour bug; each is a statement the code contradicts.

- **`DrugReferenceSource.lastLoadOrigin()`** — the *contract an implementor reads* — still described
  the operator case as "the **resolved** operator file" while giving the two sibling forms with their
  literal prefixes. An implementor following it would return the absolute path and reintroduce exactly
  the disclosure just removed, and **no test would catch it**: the tests only drive the three real
  sources. It now specifies the `appdata:` form and says an implementation must not return the
  absolute path, with the reason.
- **`config.xml`** told operators the mismatch "**silently** turns drug-safety checking off" — the
  admin screen being the worst place to promise there will be nothing in the log, given this change
  makes that state loud.
- **`DrugReferenceLoad`'s class javadoc** said both readers report this object "rather than re-reading
  the global properties"; the endpoint reads one live — `enabled`, precisely the field allowed to
  disagree with the retained load.

And the claim that *compensates* for a relative origin — "the absolute path is still logged at INFO" —
is now asserted rather than merely stated. It is the administrator's only remaining route to that
path, so logging the relative path there instead would have removed it silently, with a green build.

### 5. Also covered / documented

- `atc` — the one format that does **not** go through the shared loader (no bundled fallback, so it
  resolves and tracks its own file) — now has its origin executed, over the real WHO ATC sample; plus
  the `atc`-with-no-path case, which is inert and warns.
- README's API section documents the endpoint with its privilege (it documented every other one), and
  the two global properties whose mismatch causes the failure now say what the mismatch does and
  where to check it. Their admin-screen description is the last thing read before the mistake is made.
- `setEntries` is the one seam on which the status and the entries in use are not two views of the
  same event — it fills the cache without a load, so the status keeps reporting `notLoaded`. Recorded
  on the seam.

### Mutation sweep of the hardening

Five mutations, one at a time, on a throwaway worktree (`git worktree add`, restored and verified
clean — `git status --porcelain` empty):

| mutation | caught by |
|---|---|
| A — drop the `requirePrivilege` gate from the endpoint | `…refusesACallerWithoutTheGlobalPropertyPrivilege` (endpoint returns 200) |
| B — drop the controller's `enabled` field | `…reportsEveryDocumentedFieldForADefaultInstallation` + the live-switch test |
| C — report `enabled` from the load instead of the live switch | `…reportsTheLiveSwitchAndTheRetainedLoadSeparately` |
| D — drop the inner `entries != null` re-check inside the lock | both concurrency tests (8 loads, 8 WARNs) |
| E — publish `entries` before `load` (reverse the documented ordering) | **nothing** — see below |

**E survived, and that is reported rather than papered over.** The ordering is correct by the memory
model: both fields are `volatile`, program-order volatile writes cannot be reordered with each other,
and a reader whose volatile read of `entries` sees the write therefore also sees the earlier write to
`load`. Reversing it opens a window in which a reader taking the lock-free fast path would read
`load` as null — but that window is a single volatile write wide and no test can hit it reliably. The
ordering is *reasoned*, and now *measured to be unpinned*; the comment on it is the only thing
protecting it.

### Where this stopped, and why

Production Java behaviour last changed at the `origin` commit. Every commit after it changed **zero**
non-comment lines under `src/main/java` — the remaining rounds were finding and removing claims in
prose that the code contradicted, three of which were introduced by earlier rounds of this same pass.
Each was resolved by **narrowing or deleting** the claim rather than re-wording it, which is what makes
that converge: a comment that states less cannot state it wrongly. Two review rounds were also
**rejected on evidence** — one proposed moving a global-property read to after the parse on the premise
that the two reads straddle it (they do not; both sources read that property before opening any file,
so the move would have put the whole parse inside the window), and one called a `ReferenceDataFiles`
sentence half-false when it claims nothing about the origin at all.

### Build

`mvn -o clean install`, full reactor, measured in this worktree:

```
BASELINE (main @ 03230b3):  api 853, omod 58  = 911 tests, 0 failures, 34 skipped
PR HEAD  (a2363ac):         api 863, omod 58  = 921 tests, 0 failures, 34 skipped
AFTER HARDENING:            api 869, omod 61  = 930 tests, 0 failures, 34 skipped
```

No assertion in any pre-existing test was touched. The hardening does modify one pre-existing test
file, `DrugReferenceTestSupport`, and only by adding to it: 9 insertions, 0 deletions
(`git diff --numstat`), being the shared `REFERENCE_LOGGER` constant. Every other test file in the
slice is still an addition.

### Live re-verification of the hardened build

Re-run in full because `origin`'s value changed, so this is a wire-shape change. Deployed through the
shared verify harness (which gated on `hits=75` live querystore retrieval and the Mary+clarithromycin
control chip before handing over), five configurations, **each behind a full JVM restart**. Patients
resolved by name lookup against the live database — `Mary Smith 38beca4a…`, `Agnes Adams 47e57b75…`,
`Joshua Johnson 9cb37bcb…`.

| | status | reference WARN/ERROR | probes |
|---|---|---|---|
| **L1** healthy `ddinter` | `entryCount 2283`, `inert false`, `origin appdata:chartsearchai/ddi_knowledge_base.json` | **0** | 4 controls all correct (1 / 1 / 2 / 0 chips) |
| **L2** `sourceFormat=json` over the DDInter KB | `inert true`, `entryCount 0`, `origin appdata:chartsearchai/ddi_knowledge_base.json` | **1** — the INERT WARN, naming both GPs, the parser and the file | 0 chips, twice |
| **L3** configured path unreadable | `entryCount 4`, `inert false`, **`origin classpath:/chartsearchai/drug-reference.json`** vs `configuredDataFilePath chartsearchai/h154-no-such-file.json` | **0** | curated-only wording appears (below) |
| **L4** `enabled=false` | `loaded false`, `inert false`, every config field `null` | **0** | 0 chips |
| **L5** restored `ddinter` | identical to L1 | **0** | 4 controls identical to L1 in count, type and partner |

**Source-exclusive positive control, both directions** — same patient, same question, so only the
loaded dataset can explain the difference:

- **L1/L5 (`ddinter`, 2283):** Agnes + "Can I give ibuprofen?" → 1 chip, **Major**, DDInter mechanism
  text (*"competitive inhibition of platelet cyclooxygenase…"*), class label rendered as
  `Acetylsalicylic acid (aspirin)`.
- **L3 (bundled curated, 4):** same probe → 2 chips, one reading **`additive GI and bleeding risk`** —
  which occurs once in the bundled `drug-reference.json` and **zero times** in the full DDInter KB
  (checked with `grep -c`). Mary's L1 Major clarithromycin chip is gone, because clarithromycin is
  absent from the 4-entry curated dataset.

So L3 is not a count assertion: `entryCount: 4` looks perfectly healthy, and `origin` is the only
field that says the configured file was never read.

**Privilege probes, live, in the L1 state:**

| caller | result |
|---|---|
| unauthenticated | **401** `{"error":"Authentication required"}` — refused before the service is touched, so no load |
| `i149probe`, a user whose only role has **no privileges** | **200**, body carries `origin: appdata:chartsearchai/ddi_knowledge_base.json` — **no absolute path** |
| same user, core's `GET /systeminformation` | **403** |

Before this hardening the middle row returned the absolute
`/Users/…/referenceapplication-standalone-3.7.1/appdata/chartsearchai/ddi_knowledge_base.json` to that
same user — measured against the build that was on the box beforehand.

**What each probe would have shown had the change been wrong:** L1 — a WARN on a healthy 2283-entry
load, making the signal noise on every install (it is 0). L2 — the same 0 chips and the same INFO line
as before, with nothing to separate "the safety layer is off" from "there was nothing to find". L3 —
`4` indistinguishable from a successful load of the configured file, which is the exact shape #149 was
reported from; and under the old rendering the check was a suffix match on an absolute path rather
than a string equality. L4 — the inert WARN fired on every default install. Privilege row 2 — the
server's absolute application-data path handed to any logged-in user.

**Fixture created:** user `i149probe` with role `i149-noprivs` (no privileges), person "I149 Probe",
for the privilege probes. Left in place, and named so it is distinguishable from demo data.

**L4 re-run, because the first attempt at it was vacuous.** My own restart helper did not re-arm the
`log.level` global property, which OpenMRS applies through its save listener rather than at startup —
so INFO was suppressed after those restarts and "no drug-reference load line" proved nothing about the
load. Re-run with the level armed *after* the restart and *before* the status read:

```
    INFO lines from chartsearchai in this JVM: 56      <- the absence below is an observation
    search HTTP 200 52s, chips: 0                      <- a full answer ran with the feature off
--- GET /chartsearchai/drugreferencestatus [L4prime-disabled] ---
{ "enabled": false, "loaded": false, "inert": false, "entryCount": 0,
  "sourceFormat": null, "configuredSourceFormat": null,
  "configuredDataFilePath": null, "origin": null }
    drug-reference load lines: count=0
```

So on a disabled install neither reading the status nor answering a question performs the load, and
`inert` is deliberately `false` — nothing was configured to load, so nothing is wrong.

Re-enabling in the same run then produced, with INFO still armed, exactly one load line — and it is
the one that keeps the absolute path where an administrator can see it:

```
INFO - ReferenceDataFiles.loadWithClasspathFallback(126) | Loaded 2283 DDInter drug-reference entries
       from /Users/…/referenceapplication-standalone-3.7.1/appdata/chartsearchai/ddi_knowledge_base.json
```

**Mutation F, added with that assertion:** logging `configuredPath` instead of `resolved` in
`ReferenceDataFiles`' INFO line — i.e. removing the administrator's route to the absolute path —
fails `loadStatusNamesTheFileReadWithoutDisclosingItsAbsolutePath` with
`Captured: [… INFO Loaded 16 DDInter drug-reference entries from chartsearchai/relative-ddi.json]`.
Before the assertion existed that change was green.

Closes #149


dkayiwa:
## Hardening pass

Fresh eyes on this branch, five commits pushed. The PR description now carries a **Hardening pass**
section with the detail; this is the summary and the things that need your judgement.

### One behaviour change, and it came from probing the privilege gate

`Get Global Properties` is the right privilege for most of this body — but it is one of the **27
privileges the `Authenticated` role holds by default**, so *every logged-in user* can read this
endpoint. I built the case rather than reasoning about it: a user whose only role carried **no
privileges at all** got **200** from `/chartsearchai/drugreferencestatus` and from
`/systemsetting/chartsearchai.drugReference.dataFilePath`, and **403** from core's
`/systeminformation` — which is where core discloses the application data directory.

So the two configured global properties in the body disclose nothing that user could not already
read. The **absolute filesystem path** in `origin` did: core keeps that behind
`View Administration Functions`. `origin` now reads `appdata:chartsearchai/ddi_knowledge_base.json`
instead of `/Users/…/appdata/chartsearchai/ddi_knowledge_base.json`, beside the unchanged
`classpath:/chartsearchai/drug-reference.json`.

Nothing diagnostic is lost and the check gets easier: *"did my configured file load?"* is now
`origin == "appdata:" + configuredDataFilePath` rather than a suffix match. The absolute path is still
logged at INFO.

**Two things for you to sign off on, since they are judgement rather than defect:**

1. **The audience is every logged-in user, not just operators.** The javadoc now says so plainly. If
   you would rather this be admin-only, `View Administration Functions` would match what core does
   with comparable information — I did not change the privilege, because the body as it now stands is
   metadata that audience can already read.
2. **The GET still performs the load.** That is the design and I think it is right — a status that
   cannot be stale is the whole second half of #149. It is bounded: the gate runs before the service
   is touched (an unauthenticated call is refused at 401 without loading), a disabled feature loads
   nothing, and after the first call every reader gets the retained snapshot. The one residual is that
   the first call can be the one paying a 19 MB parse; the same parse was already triggerable by any
   user asking a safety question, so this widens the audience slightly rather than adding a new cost.

### Coverage the change was missing

The **"the omod tests cannot set the controller's autowired `DrugReferenceService`"** limitation was
wrong, and I have withdrawn it in the description: this controller already carries **three**
package-private `Test seam:` setters for exactly that, and the omod test lives in its package. A
fourth was added, and the endpoint now has three tests — the 8-field body in order (including
`enabled`, which no API-level test can see), the live-switch-vs-retained-load distinction, and a
refused unprivileged caller. The related claim that the wire shape "**is** pinned by a test" was
two-thirds true: that test pins `toMap()`'s seven fields, not `enabled`, not the order, not the gate.

Concurrency, which the description flagged as reasoned-not-stress-tested, now has a test: eight
callers race the lazy load, and the dataset is read once, the inert WARN fires once rather than once
per caller, and every caller gets the same retained outcome.

`atc` — the one format that does not go through the shared loader — now has its origin executed over
the real WHO ATC sample, plus the `atc`-with-no-path case (inert, warns).

### The one mutation that survived

Reversing `load = outcome; entries = …` to publish `entries` first left the **whole api suite green**.
The ordering is correct by the memory model (both `volatile`; a reader whose volatile read of
`entries` sees the write also sees the earlier write to `load`), so this is not a defect — but it is
**measured to be unpinned**, and the comment on those two lines is the only thing protecting it. Worth
knowing before anyone tidies them.

### Deviations I did not reverse

Both of the description's deviations from the issue look right to me and I left them alone: declining
the "sniff the file and say it looks like the other format" proposal (three parsers, three places to
drift, and it can guess wrong on a truncated download), and replacing `startupErrorMessage` with the
REST endpoint (the load is lazy, so there is nothing to report at startup). The `origin`-vs-count
separation is the load-bearing idea in the change and L3 is what proves it; I only changed how the
path is rendered, not that it is reported.

### Confirmed, and needs its own issue

The description's note that the standalone is silently running the **bundled** cross-reactivity groups
is **confirmed**: `chartsearchai.drugReference.crossReactivityGroupsFilePath` =
`chartsearchai/cross-reactivity-groups.json`, and that file does not exist in
`appdata/chartsearchai/` (which holds only `README.md`, `bin`, `ddi_knowledge_base.json`,
`gemma-4-E4B-it-Q4_K_M.gguf`, `kvcache`). `CrossReactivityGroupsLoader.load()` calls
`.getItems()` and drops the origin this PR added, so the groups load has no `origin` anywhere and the
fallback is INFO-only — and the status endpoint reports only the entry dataset. Since the default GP
value points at a file the module never creates, **every** install that has not created it is in this
state.

### Also found and fixed on a second look

- The comment above the WARN said *"Nothing configured at all" is a different state and stays silent*.
  A test added in this pass contradicts it: `sourceFormat=atc` with no `dataFilePath` **is** nothing
  configured at all, and it warns — correctly, since `atc` has no bundled fallback, so nothing
  configured means nothing loaded and the safety layer really is off. The state that stays silent is
  the feature being switched off, which never reaches `ensureLoaded` at all. Corrected.
- `getConfiguredSourceFormat()` had no reader anywhere, and the divergence its javadoc promises — a
  mistyped `sourceFormat` falling back to the curated parser — had no test. Pinned, including the
  assertion that this case is deliberately **not** loud, because it loads entries: the two format
  fields differing is the only signal, which is the reason both are reported.

### Reviewed and rejected

One review round proposed moving the service's `dataFilePath` read to *after* `load()` returns, on the
grounds that the service's read and the source's read "straddle the parse". They do not: both
`ReferenceDataFiles.loadWithClasspathFallback` and `AtcDrugReferenceSource.load` read that global
property as their first statement, before any file is opened, so the two reads are microseconds apart
and the proposed move would put the entire parse *inside* the window. Left alone.

It was also suggested that `ReferenceDataFiles`' "shared so the fallback/logging/exception contract
cannot drift" sentence is now half-false because one consumer drops the origin. It enumerates
fallback, logging and exceptions and claims nothing about the origin, so it still reads true. Left
alone.

---

### Correction to the above: a seventh commit, and three prose sites the origin change had missed

I posted the summary above before a final review round came back. It found that relativizing `origin`
had updated README, ONBOARDING, `DrugReferenceLoad`, `ReferenceDataFiles` and the controller — and left
**three** statements behind that the code now contradicts. I verified each myself before fixing it:

1. **`DrugReferenceSource.lastLoadOrigin()`** still described the operator case as "the **resolved**
   operator file" while giving the two sibling forms with their literal prefixes. This is the contract
   an *implementor* reads, so someone following it would return the absolute path and reintroduce the
   very disclosure that was removed — and **nothing would catch it**, because the tests only drive the
   three real sources. Now specifies the `appdata:` form and forbids the absolute path, with the reason.
2. **`config.xml`** told operators the mismatch "**silently** turns drug-safety checking off". Making
   that state loud is what this PR does, and the admin screen is the worst place to promise there will
   be nothing in the log.
3. **`DrugReferenceLoad`'s class javadoc** said both readers report the object "rather than re-reading
   the global properties" — the endpoint reads one live, `enabled`, which is exactly the field that is
   *meant* to disagree with the retained load.

Worth noting how those survived: my own sweep for stale wording grepped for "absolute" and "silent",
and the interface javadoc said neither — it said "resolved". Keyword greps do not find prose drift;
reading does.

I also pinned the claim that **compensates** for a relative origin — *"the absolute path is still
logged at INFO"* — instead of only stating it. That log line is the administrator's one remaining route
to the absolute path, so rewriting it to log the relative path would have removed it silently, with a
green build.

Build after these: **api 869 + omod 61 = 930**, 0 failures, 34 skipped, BUILD SUCCESS.

### Final round: three claims of mine narrowed, and where this stopped

A confirmation review came back with **nothing substantive outstanding** and three sub-nits, all in
prose I had added, each true of the common case and wrong at an edge. I verified each against the code
before acting, and narrowed or deleted rather than re-worded:

- **config.xml** told operators an unreadable path "falls back to the bundled dataset and still reports
  a plausible count". Not for `sourceFormat=atc` — it has no bundled fallback, so it loads nothing,
  reports `origin: none`, and is inert and *loud*. Now says only what holds for every format.
- **`getConfiguredSourceFormat()`** promised the "raw" global-property value; it is read with the `json`
  default applied, so an unset property reports `json`.
- **An injected source tracks no origin**, so the status reads `none` beside a non-zero entry count —
  the one case where `none` does not mean "nothing could be read". Recorded on `setSource`.

Also removed a `LogCapture` claim that its helpers "expose levels, not strings" — they return formatted
messages, and the new assertion checks one contains the absolute path.

**Where it stopped:** production Java behaviour last changed at the `origin` commit. Every commit after
it changed **zero** non-comment lines under `src/main/java`. The later rounds were finding and removing
prose the code contradicted — three of those statements introduced by earlier rounds of this same pass.
That is why each was deleted or narrowed rather than re-worded: a comment that states less cannot state
it wrongly.

I also **rejected two review recommendations on evidence**, rather than applying them: moving a
global-property read to after the parse (both sources read that property *before* opening any file, so
the two reads are microseconds apart and the move would have put the whole parse inside the window), and
calling a `ReferenceDataFiles` sentence half-false (it enumerates fallback, logging and exceptions, and
claims nothing about the origin).

Final: **api 869 + omod 61 = 930**, 0 failures, 34 skipped, BUILD SUCCESS at the pushed SHA.

dkayiwa:

- Final head `201b91f`: `mvn -o clean install` on a clean detached worktree → **869 + 61 = 930, 0 failures, 34 skipped**, BUILD SUCCESS. Reproduces the reported figure at the exact SHA being merged.
- CI green on Java 11/17/21.
- Test-file changes across the whole branch: **4 new files**, plus **+8/-0** in the pre-existing `DrugReferenceTestSupport` — a shared logger-name constant. No assertion weakened, no expected value changed. (That constant's javadoc names the hazard it prevents: a package rename would leave a stale literal, the capture would receive nothing, and every "no WARN was logged" assertion would pass **vacuously**. Worth keeping in mind for anyone adding log-level assertions.)
- The disclosure fix is real in the shipped code: `origin` emits `appdata:<relative>` / `classpath:<resource>` / `none`, never an absolute path, and `Context.requirePrivilege(PrivilegeConstants.GET_GLOBAL_PROPERTIES)` is enforced in the handler rather than merely documented.

On the open question of the privilege level: keeping `Get Global Properties` is defensible and I am merging with it. The response body is global-property metadata, core's own privilege for reading global properties is exactly this one, and the body was deliberately narrowed to what that audience can already read. `View Administration Functions` would match core's `AdministrationService` if you would rather the audience were narrower — that is a one-line change and a reasonable preference, not a defect.

Two follow-ups filed from this pass: #158 (the `load`-before-`entries` publication order survives reversal — latent, correct by the JMM today) and #159 (`docs/adr.md`'s endpoint list omits `/prewarmstatus` and `/drugreferencestatus`). #156 updated with the confirmed cross-reactivity-groups instance.

PR #160, opened 2026-08-05
fix(drug-safety): one contraindication chip per substance, not per reference row (#145)

### LINKED ISSUE #145: Contraindication arms have no route-variant collapse: one substance reported as cross-reactive with itself (4 chips for dexamethasone; 152 surplus across 114 families)

### LINKED ISSUE #146: Curated allergy rule and identity check both fire for one allergy, on the shipped default source (3 of 4 bundled entries)
### Summary

On the **shipped default** `sourceFormat=json`, a recorded allergy to a drug produces **two chips for one fact** — the curated rule arm and the identity arm each report it:

```
Ibuprofen is contraindicated by an active allergy: documented ibuprofen allergy
The patient has a recorded allergy to Ibuprofen.
```

`addContraindications` walks the entry's curated `contraindications` and fires on `{"type":"allergy","token":"ibuprofen"}`. `addAllergyContraindications` (from #140) independently resolves the allergen to that same entry and fires the identity chip. Neither knows about the other.

### Measured on the shipped file

`api/src/main/resources/chartsearchai/drug-reference.json` — verified by reading it:

| entry | contraindications | self-named allergy rule |
|---|---|---|
| Ibuprofen | 4 | **1** |
| Paracetamol | 2 | **1** |
| Amoxicillin | 2 | **1** |
| Gentamicin | 2 | 0 |

**3 of the 4** shipped entries double-report. Gentamicin is the control — no self-named rule, so one chip.

### Why this matters beyond tidiness

It is the **default** configuration. Every defect found this week needed `sourceFormat=ddinter`; this one ships as-is, so a fresh install with the bundled curated file shows it on the commonest allergy in that file.

And it distorts a figure I published: **#143's "2 chips" measurement is partly made of this.** When I measured the echo-scoping suppression as "0 chips scoped, 2 chips unscoped", the 2 was the curated rule *plus* the identity chip, not two independent findings. #143's conclusion is unaffected — the suppression is real and the fix is verified — but that number should be read as pinning the current behaviour rather than asserting that two chips is correct.

### The wrinkle: which chip should survive?

Not simply "identity wins". The curated rule carries an operator-authored **note** (`"documented ibuprofen allergy"`) that the identity chip cannot reproduce, and a curated file is where a deployment records its own clinical wording. But the identity chip states the fact more plainly and works for the 444 entries with no curated rule at all.

This is the same shape as #88's resolved question — "class yields to rule" was the wrong dedup there because for contentless rows the class chip was the informative one. The lesson transfers: prefer **whichever chip carries content**, and fold rather than drop if both do. `#130` established that pattern for the interaction arms; the contraindication arms should reuse it rather than invent a second policy.

### Related

- **#145** — the other contraindication duplication (route variants of one substance). Different mechanism, same arm, and a fix for either should be checked against the other so the two dedups compose rather than stacking.
- **#130** — the fold precedent to follow.

### Tests

Through the real production path with `sourceFormat=json`: an allergy to Ibuprofen asked about Ibuprofen yields **one** chip, and that chip retains the curated note. Gentamicin (no self-named rule) still yields one. An entry with a curated rule whose note is blank falls back to the identity wording. Re-assert the 444-entry case (no curated rule at all) still fires from identity alone.

Found while fixing #143 (PR #144); not fixed there because the merge policy needs its own argument, and #145 must be resolved compatibly.

dkayiwa:
## The defect

Both contraindication arms are keyed on a subject **entry**, and DDInter files one substance as
several route/formulation rows. One clinician-facing string resolves all of them —
`findByQuery` and `findByDrugName` return every entry whose aliases match — so one clinical fact
became one chip per row.

Worse than duplication: only the row `lookupByToken` resolves the allergy to matches by IDENTITY, so
its siblings fell through to the class comparison and reported the substance as cross-reactive with
the patient's allergy to **itself**. Measured on the 3.7.1 standalone against the build deployed there
before mine, Richard Jones (recorded Dexamethasone allergy + a Dexamethasone order), "Is it safe to give
dexamethasone?" — **4 chips**:

```
[contraindication] Dexamethasone
    The patient has a recorded allergy to Dexamethasone.
[contraindication] Dexamethasone (nasal)
    Dexamethasone (nasal) is in the same ATC class (A01AC) as the patient's allergy to Dexamethasone — possible cross-reactivity
[contraindication] Dexamethasone (ophthalmic)
    Dexamethasone (ophthalmic) is in the same ATC class (A01AC) as the patient's allergy to Dexamethasone — possible cross-reactivity
[contraindication] Dexamethasone (topical)
    Dexamethasone (topical) is in the same ATC class (A01AC) as the patient's allergy to Dexamethasone — possible cross-reactivity
```

Since #110 every chip is also injected as a citable pre-answer record, so each duplicate reached the
prompt too (measured through the real injector: 4 `safety_finding` records for that one allergy).

## Root cause and what changed

`DrugSafetyValidator` gains a validate-scoped ledger, `ContraindicationChips`, that **both**
contraindication arms and **both** of their call sites feed — the drug-in-play loop and
`addActiveOrderContraindications`. It keys on `(subject substance, recorded finding)`, keeps the most
specific relationship, and writes the survivor back into the position the group's first candidate
took, so no client sees the chip sequence reshuffle.

**Not a filter over the finished chip list**, for three reasons, the first decisive: the sibling's
chip is *wrong*, not merely duplicated — the true statement about that substance is the identity one —
so the collapse has to choose which relationship survives, and by the time only rendered text is left
the reasons are gone. Second, those chips differ in text (each names its own route), so
`chipIdentity`'s exact-repeat key cannot see them, and recognising "differs only in a route
qualifier" from the strings means pattern-matching a display label, which is the mistake #148 undid.
Third, the arms run at two call sites, and a collapse inside either one leaves the other emitting the
siblings — Sarah Taylor's shape below.

### The collapse key, and why `rxnorm_name` equality is not the line

`DrugReference.substanceKey()` = **the reference data's own substance name AND the display-name stem**
(the name with trailing parenthesized qualifiers removed). `DrugReference` gains a `substanceName`
field, set unconditionally by `DdiDrugReferenceSource` from `rxnorm_name`; sources that publish none
(the curated `json` file, the `atc` adapter) key on entry identity, so nothing collapses for them.

Measured over the shipped 19 MB KB (2283 entries), each half is load-bearing:

* the substance name is the data's own substance claim — 142 values shared by more than one entry
  across 332 entries, and the `rxcui` partitions those entries **identically**;
* but that claim over-merges. Among those 142 families sit pairs of genuinely different
  substances: `Omeprazole`/`Esomeprazole` (one `rxnorm_name`, one `rxcui`, one ATC code),
  `Amphetamine`/`Dextroamphetamine`, `Fenfluramine`/`Dexfenfluramine`,
  `Gabapentin`/`Gabapentin enacarbil`, `Netupitant`/`Fosnetupitant`,
  `Ketoconazole`/`Levoketoconazole`, `Fenofibrate`/`Fenofibric acid`, `Atropine`/`Hyoscyamine` …
  — each of them the `enalapril`/`enalaprilat` shape #121 decided must stay two chips. The stem
  separates every one of them.
* Conversely the stem alone merges 9 groups the substance name keeps apart, 8 of them different
  substances sharing one stem (`Manganese (chloride)`/`(sulfate)`,
  `Dextran (-1)`/`(low molecular weight)`, `Insulin human`/`(isophane)`,
  `Iron`/`(polysaccharide)`, `Typhoid vaccine (live)`/`(inactivated)` …).

Together the key reduces those 332 entries to 177 substances. A mutation sweep on a throwaway tree
confirms the stem half is load-bearing rather than decorative: with `substanceKey()` cut back to the
substance name alone — issue #145's own proposal 2 — two of the new tests fail by **dropping real
chips**, `Esomeprazole`'s and `Hydrocortisone butyrate`'s.

Conservative where it cannot tell, in both directions: a name that extends the family stem by a WORD
rather than a qualifier (`Hydrocortisone butyrate`, `Estrone sulfate`, `Procaine benzylpenicillin`)
keeps its own chip, and 21 of the 142 families are left entirely unmerged for that reason — some of
which the KB is in fact naming one substance two ways (`Thallous Chloride`/`Thallous chloride
tl-201`). Over-reporting one chip on a non-blocking advisory is the safe direction; dropping a real one is
not.

### Which chip survives

The **most specific relationship** — identity > shared ATC class > shared curated cross-reactivity
group. That is this arm's analogue of "the highest severity wins": a contraindication chip carries no
severity, and what it can under-report is the strength of the claim. Ties keep the incumbent, so a
group of equally-related rows reports the dataset's first row, exactly as `bestRulePerPartner` keeps
its first.

Stated honestly: on the shipped resolution semantics the strongest candidate is always the group's
**first** — `lookupByToken` takes the EARLIEST matching entry, every row of a substance carries that
substance's name among its aliases, and both subject sets are iterated in dataset order — so the
replacement branch is not reachable through the production path today and **no test pins it**. It is
written that way because the ordering is a property of `lookupByToken`, not of this ledger, and a
first-wins ledger would silently start dropping identity chips the moment that changed.

### `SafetyWarning.equals`/`hashCode`: deliberately not added

Value equality on a clinician-facing advisory would collapse two genuinely different findings whose
wording happens to agree, and it could not implement this collapse anyway — the duplicate chips differ
in text. The grouping is structural and belongs where the reasons are known, the same argument #88 and
#121 made.

## Live evidence, verbatim

Deployed to the 3.7.1 standalone (`sourceFormat=ddinter`, 2283 entries, `GET
/chartsearchai/drugreferencestatus` confirming the load). Patients resolved by name lookup against the
live DB. After = this branch. **Before** = the build that was already deployed on that instance,
produced from the branch merged as `dd651ef`; I did not rebuild `main` myself to re-measure it, so read
the before-column as "the previous build on this box" rather than as a build I made from `dd651ef`. The
same 4-chip and 4-record shapes are independently reproduced by the new tests failing on this branch
before the fix (`expected: <1> but was: <4>`).

### A. Richard Jones (`a39e82e7-…`) — recorded Dexamethasone allergy + a Dexamethasone order

"Is it safe to give dexamethasone?"

| | chips |
|---|---|
| before | 4 |
| after | **1** |

```
--- safetyWarnings (1) ---
  [contraindication] drug='Dexamethasone'
      The patient has a recorded allergy to Dexamethasone.
```

Counterfactual: had the collapse been wrong in the *aggressive* direction the surviving chip would be
one of the three "Dexamethasone (nasal/ophthalmic/topical) is in the same ATC class (A01AC) as the
patient's allergy to Dexamethasone" lines — the vacuous one — instead of the identity chip. Had it not
fired at all, this stays at 4, as it does on `main`.

### B. Sarah Taylor (`dc8560c9-…`) — Dexamethasone allergy + 8 active orders

"Is it safe to give her paracetamol?" (names no drug of hers, so the order-driven arm owns every chip)

| | chips |
|---|---|
| before | 13 |
| after | **6** |

Before, verbatim:

```
--- safetyWarnings (13) ---
  [contraindication] Dexamethasone        The patient has a recorded allergy to Dexamethasone.
  [contraindication] Dexamethasone (nasal)        … same ATC class (A01AC) as the patient's allergy to Dexamethasone …
  [contraindication] Dexamethasone (ophthalmic)   … same ATC class (A01AC) …
  [contraindication] Dexamethasone (topical)      … same ATC class (A01AC) …
  [contraindication] Hydrocortisone               … same ATC class (A01AC) …
  [contraindication] Hydrocortisone (ophthalmic)  … same ATC class (A01AC) …
  [contraindication] Hydrocortisone (topical)     … same ATC class (A01AC) …
  [contraindication] Hydrocortisone butyrate      … same ATC class (A01AC) …
  [contraindication] Prednisone                   … same ATC class (H02AB) …
  [contraindication] Budesonide                   … same ATC class (R01AD) …
  [contraindication] Budesonide (nasal)           … same ATC class (R01AD) …
  [contraindication] Methylprednisolone           … same ATC class (D10AA) …
  [contraindication] Methylprednisolone (topical) … same ATC class (D10AA) …
```

After, verbatim:

```
--- safetyWarnings (6) ---
  [contraindication] drug='Dexamethasone'
      The patient has a recorded allergy to Dexamethasone.
  [contraindication] drug='Hydrocortisone'
      Hydrocortisone is in the same ATC class (A01AC) as the patient's allergy to Dexamethasone — possible cross-reactivity
  [contraindication] drug='Hydrocortisone butyrate'
      Hydrocortisone butyrate is in the same ATC class (A01AC) as the patient's allergy to Dexamethasone — possible cross-reactivity
  [contraindication] drug='Prednisone'
      Prednisone is in the same ATC class (H02AB) as the patient's allergy to Dexamethasone — possible cross-reactivity
  [contraindication] drug='Budesonide'
      Budesonide is in the same ATC class (R01AD) as the patient's allergy to Dexamethasone — possible cross-reactivity
  [contraindication] drug='Methylprednisolone'
      Methylprednisolone is in the same ATC class (D10AA) as the patient's allergy to Dexamethasone — possible cross-reactivity
```

Seven route/formulation-variant chips gone; **no substance lost**. The dexamethasone family keeps the
IDENTITY chip rather than one of its three self-cross-reactivity siblings, and `Hydrocortisone butyrate`
stays its own chip beside `Hydrocortisone` — the conservative refusal, same rule that keeps Omeprazole
and Esomeprazole apart in C.

Counterfactual: 11 of the 13 were route or formulation variants of five substances. A fix that
collapsed only within the question-driven arm would have changed nothing here at all — every chip on
this patient comes from `addActiveOrderContraindications`.

### C. George Anderson (`d5c9cdad-…`) — MUST NOT COLLAPSE

Fixture I created on the demo DB (documented below): a SEVERE coded **Pantoprazole** allergy, no
orders. "Is it safe to give esomeprazole?" — that one word resolves BOTH rows the KB files under
`rxnorm_name=esomeprazole` / `rxcui=283742`, and both carry the same single ATC code `A02BC05`, so
every key made of reference-data identity alone merges them.

| | chips |
|---|---|
| before | 2 |
| after | **2** |

Identical before and after, verbatim (after):

```
--- safetyWarnings (2) ---
  [contraindication] drug='Omeprazole'
      Omeprazole is in the same ATC class (A02BC) as the patient's allergy to Pantoprazole — possible cross-reactivity
  [contraindication] drug='Esomeprazole'
      Esomeprazole is in the same ATC class (A02BC) as the patient's allergy to Pantoprazole — possible cross-reactivity
```

Counterfactual: this is the discriminating half. With the key cut back to the substance name alone
(issue #145's own proposal 2) this collapses to **1** chip and Esomeprazole's is the one dropped —
measured, both as a live before/after shape and as a mutation sweep failure of
`twoDistinctSubstancesTheKbFilesUnderOneSubstanceNameStayTwoChips`.

### D. Margaret King (`f2397b8a-…`) — a Dexamethasone order, NO allergy

"Is it safe to give dexamethasone?"

| | chips |
|---|---|
| before | 0 |
| after | **0** |

Counterfactual: a collapse that keyed on the subject alone — dropping the recorded-finding half of the
key — could not invent a chip here, but a ledger that mis-ranked could: this is the control that the
change adds nothing where there is no allergy to compare against. It also confirms the four
dexamethasone rows in play still produce no contraindication when nothing contraindicates them.

### E. The four regression controls

All four exact, verbatim from the same run:

| control | expected | measured |
|---|---|---|
| Mary Smith x clarithromycin | 1 chip, Major, x simvastatin | **1** — `Clarithromycin interacts with active order simvastatin — Major. Coadministration with potent inhibitors of CYP450 3A4 …` |
| Agnes Adams x warfarin | 1 chip, Major, x aspirin | **1** — `Warfarin interacts with active order aspirin — Major. Aspirin, even in small doses, may increase the risk of bleeding …` |
| Joshua Johnson x ibuprofen | **2** chips | **2** — see below |
| Mary Smith x paracetamol | 0 chips | **0** |

Joshua's pair is the "2 chips is CORRECT" case — two different clinical facts about two different
partners, and the ledger must not touch them (different chip types, and the contraindication's finding
is the aspirin allergy while the interaction's partner is the lisinopril order):

```
--- safetyWarnings (2) ---
  [contraindication] drug='Ibuprofen'
      Ibuprofen is in the same cross-reactivity group (NSAID) as the patient's allergy to Acetylsalicylic acid (aspirin) — possible cross-reactivity
      Ibuprofen interacts with active order lisinopril — Moderate. Nonsteroidal anti-inflammatory drugs (NSAIDs) may attenuate the antihypertensive effects of ACE inhibitors. …
```

Agnes's case additionally exercises the class arm — her Aspirin 81mg order carries ATC `N02BA01` — and
still yields exactly one chip, so #88's rule/class fold is unaffected.

### Demo data I created (no teardown, per the standing instruction)

* **George Anderson** `d5c9cdad-8873-488d-9f7c-c92af9898b04` — one SEVERE coded DRUG allergy to
  `Pantoprazole` (allergy uuid `4cf5a4ee-65c0-4d79-8e15-b6529abd3fe2`). Nothing else was added,
  voided or changed; no GP was touched.

## Build

Measured myself on this branch, full reactor `mvn -o clean install`:

* **api 876, omod 61 — 0 failures, 34 skipped.**
* Baseline on `main` at `dd651ef`, measured the same way: api 869, omod 61, 0 failures, 34 skipped.
* +7 = exactly the new `ContraindicationRouteVariantTest`. **No existing test changed and none broke.**

## Limitations

* The mislabelled class in the surviving chip is untouched and still wrong: `sharedClass` returns the
  *first* shared subgroup, so a systemic dexamethasone allergy is reported against `A01AC`
  (corticosteroids for local oral treatment) and a Solu-Medrol order against `D10AA` (anti-acne
  corticosteroids). #145 calls that its option 3 and says it should not be bundled; it is not.
* The same subject-side duplication still exists in the **interaction** arms, and I measured it on this
  build rather than leaving it as a guess. Sarah Taylor, "Is it safe to give hydrocortisone?" — the six
  contraindication chips are all distinct substances (this fix working), but the interaction arm reports
  `hydrocortisone x diclofenac` **twice**, once per reference row:

  ```
  [interaction] Hydrocortisone              Hydrocortisone interacts with active order diclofenac — Moderate. The combined use of corticosteroids and nonsteroidal anti-inflammatory …
  [interaction] Hydrocortisone (ophthalmic) Hydrocortisone (ophthalmic) interacts with active order diclofenac — Moderate. Concomitant use of ophthalmic nonsteroidal …
  ```

  `bestRulePerPartner` groups per PARTNER within one subject, so #115/#121/#148 fixed the partner side
  and the subject side is still ungrouped. Not fixed here — #145 is scoped to the contraindication arms,
  and the interaction case needs a decision this one does not (the two rows carry *different* mechanism
  prose, so it is not a pure duplicate). Filed separately.
* The injector duplicates the reference RECORD the same way: `DrugReferenceInjector.matchingEntries`
  dedups on `ref.getId()`, and route variants deliberately have distinct ids, so asking about
  dexamethasone injects four near-identical `drug_reference` records. Code-reasoned, not measured live
  (the REST layer returns only cited references). The *finding* records are fixed here, and pinned:
  4 → 1 through the real injector.
* This does not resolve #146: the curated-rule arm and the allergy arm keep separate key spaces on
  purpose, because a curated token may name a class (`nsaid`, `aminoglycoside`) rather than a drug, so
  resolving it to an entry to make the two comparable would collapse a class-level rule into an
  identity chip. `ActiveOrderContraindicationTest` pins that pair at 2 chips and still passes.

Closes #145

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
@@ -190,6 +190,15 @@ public class DdiDrugReferenceSource implements DrugReferenceSource {
 					&& rxcuiCounts.get(row.rxcui) == 1;
 			ref.setId(uniqueRxcui ? row.rxcui : row.id);
 			ref.setName(row.name);
+			// The substance this row is a route/formulation of, ALWAYS — unlike the chip-label synonym
+			// below, which is deliberately withheld when the display name already contains it. That is
+			// what makes it usable as an identity: it is the field the route variants sharing one RxCUI
+			// agree on ("Dexamethasone", "Dexamethasone (nasal)", … all publish "dexamethasone"), and
+			// the safety chips group on it so one substance is one finding (issue #145, see
+			// DrugReference.substanceKey). Not the RxCUI itself, though it partitions this KB's entries
+			// identically: setId above already spends the RxCUI on entry identity, where a shared one is
+			// precisely what it must NOT be.
+			ref.setSubstanceName(row.rxnormName);
 			// Chip-label synonym (never renaming): when the DDInter display name diverges from
 			// the RxNorm generic the question and chart use ("Acetylsalicylic acid" vs
 			// "aspirin"), carry the generic so safety chips can show both vocabularies. A name
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
@@ -118,8 +118,11 @@ public class DrugReference {
 
 	/**
 	 * @return the reference data's own canonical name for the SUBSTANCE this entry is a row of — the
-	 *         DDInter {@code rxnorm_name} — or null for a source that publishes none (the curated
-	 *         {@code json} file and the {@code atc} adapter). Unlike {@link #getGenericName()}, which
+	 *         DDInter {@code rxnorm_name} — or null for a source that publishes none: the {@code atc}
+	 *         adapter, which has no such field, and the shipped curated {@code json} dataset, whose
+	 *         entries carry none. Null there is the dataset's silence, not a schema ban — the curated
+	 *         schema binds this class directly, so a hand-authored file that sets this field opts into
+	 *         the grouping below. Unlike {@link #getGenericName()}, which
 	 *         is a chip-label synonym and is deliberately null whenever the display name already
 	 *         contains it, this is set whether or not the two agree: it is an identity, not a label,
 	 *         and it is exactly the field that is EQUAL across a substance's route/formulation rows
@@ -168,11 +171,15 @@ public class DrugReference {
 	 *       {@code Atropine}/{@code Hyoscyamine} — each of them the {@code enalapril}/{@code enalaprilat}
 	 *       shape issue #121 decided must stay two chips. The stem separates every one of them.</li>
 	 * </ul>
-	 * Neither half works alone: the stem alone merges 9 groups the substance name keeps apart, 8 of
-	 * them different substances sharing one stem ({@code Manganese (chloride)}/{@code (sulfate)},
+	 * Neither half works alone. Where one stem covers two substances the stem alone merges them and the
+	 * substance name is what refuses: {@code Varicella Zoster Vaccine (Recombinant)} against
+	 * {@code (live/attenuated)} — a distinction that decides whether an immunocompromised patient may
+	 * have it at all — and likewise {@code Manganese (chloride)}/{@code (sulfate)},
 	 * {@code Dextran (-1)}/{@code (low molecular weight)}, {@code Insulin human}/{@code (isophane)},
-	 * {@code Iron}/{@code (polysaccharide)}, {@code Typhoid vaccine (live)}/{@code (inactivated)} …).
-	 * Together the two reduce those 332 entries to 177 substances.
+	 * {@code Insulin lispro}/{@code (protamine)}, {@code Iron}/{@code (polysaccharide)}. A few more
+	 * one-stem groups are kept apart because a row publishes NO substance name rather than a differing
+	 * one ({@code Typhoid vaccine (live)}/{@code (inactivated)}) — that is the null-key fallback below,
+	 * not this comparison. Together the two reduce those 332 entries to 177 substances.
 	 *
 	 * <p>Conservative where it cannot tell, in BOTH directions. A name that extends the family's stem
 	 * by a WORD rather than a qualifier ({@code Hydrocortisone butyrate}, {@code Estrone sulfate},
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
@@ -139,9 +139,9 @@ public class DrugReference {
 
 	/** A trailing parenthesized qualifier on a display name — the route or formulation a DDInter row
 	 *  is distinguished from its siblings by ({@code Dexamethasone (nasal)},
-	 *  {@code Amphotericin B (lipid complex)}, {@code Tozinameran (5y-11y)}). Anchored at the END, so
-	 *  a parenthetical in the middle of a name is left alone; applied repeatedly by
-	 *  {@link #displayStem} for the handful of names carrying two. */
+	 *  {@code Amphotericin B (lipid complex)}, {@code Tozinameran (5y-11y)}). Anchored at the END, so a
+	 *  parenthetical in the middle of a name is left alone. {@link #displayStem} applies it repeatedly,
+	 *  so a name carrying more than one trailing qualifier reduces fully rather than partly. */
 	private static final Pattern TRAILING_QUALIFIER = Pattern.compile("\\s*\\([^()]*\\)\\s*$");
 
 	/**
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
@@ -159,8 +159,8 @@ public class DrugReference {
 	 *       partitions those entries identically, so this is the dataset's substance identity rather
 	 *       than a spelling coincidence.</li>
 	 *   <li>The <b>display stem</b> — the name with trailing qualifiers removed — as a VETO on that
-	 *       claim, because the claim over-merges. 21 of those 142 families hold rows that are
-	 *       genuinely different substances: {@code Omeprazole}/{@code Esomeprazole} (one
+	 *       claim, because the claim over-merges. Among those 142 families sit pairs of genuinely
+	 *       different substances: {@code Omeprazole}/{@code Esomeprazole} (one
 	 *       {@code rxnorm_name}, one {@code rxcui}, one ATC code),
 	 *       {@code Amphetamine}/{@code Dextroamphetamine}, {@code Fenfluramine}/{@code Dexfenfluramine},
 	 *       {@code Gabapentin}/{@code Gabapentin enacarbil}, {@code Netupitant}/{@code Fosnetupitant},
@@ -168,16 +168,19 @@ public class DrugReference {
 	 *       {@code Atropine}/{@code Hyoscyamine} — each of them the {@code enalapril}/{@code enalaprilat}
 	 *       shape issue #121 decided must stay two chips. The stem separates every one of them.</li>
 	 * </ul>
-	 * Neither half works alone: the stem alone merges 9 groups the substance name keeps apart
-	 * ({@code Manganese (chloride)}/{@code (sulfate)}, {@code Dextran (-1)}/{@code (low molecular
-	 * weight)}, {@code Insulin human}/{@code (isophane)}, {@code Iron}/{@code (polysaccharide)},
-	 * {@code Typhoid vaccine (live)}/{@code (inactivated)} …), which are different substances sharing
-	 * one stem. Together they reduce those 332 entries to 177 substances.
+	 * Neither half works alone: the stem alone merges 9 groups the substance name keeps apart, 8 of
+	 * them different substances sharing one stem ({@code Manganese (chloride)}/{@code (sulfate)},
+	 * {@code Dextran (-1)}/{@code (low molecular weight)}, {@code Insulin human}/{@code (isophane)},
+	 * {@code Iron}/{@code (polysaccharide)}, {@code Typhoid vaccine (live)}/{@code (inactivated)} …).
+	 * Together the two reduce those 332 entries to 177 substances.
 	 *
-	 * <p>Conservative where it cannot tell: a name that extends the family's stem by a WORD rather
-	 * than a qualifier ({@code Hydrocortisone butyrate}, {@code Estrone sulfate},
-	 * {@code Procaine benzylpenicillin}) keeps its own key and so its own chip. Over-reporting one
-	 * chip is the safe direction for a non-blocking advisory; dropping a real one is not.
+	 * <p>Conservative where it cannot tell, in BOTH directions. A name that extends the family's stem
+	 * by a WORD rather than a qualifier ({@code Hydrocortisone butyrate}, {@code Estrone sulfate},
+	 * {@code Procaine benzylpenicillin}) keeps its own key and so its own chip, and 21 of the 142
+	 * families are left entirely unmerged for that reason — some of which the KB is in fact naming one
+	 * substance two ways ({@code Thallous Chloride}/{@code Thallous chloride tl-201},
+	 * {@code Typhoid vaccine (live)}/{@code Typhoid vaccine live}). Over-reporting one chip is the safe
+	 * direction for a non-blocking advisory; dropping a real one is not.
 	 *
 	 * @return an opaque key, equal exactly for two entries this module treats as one substance
 	 */
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
@@ -10,6 +10,7 @@
 package org.openmrs.module.chartsearchai.reference;
 
 import java.text.Normalizer;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashSet;
@@ -43,6 +44,10 @@ public class DrugReference {
 
 	/** Diverging everyday generic name, or null — see {@link #getGenericName()}. */
 	private String genericName;
+
+	/** The reference data's own canonical name for the SUBSTANCE, or null — see
+	 *  {@link #getSubstanceName()}. */
+	private String substanceName;
 
 	private String drugClass;
 
@@ -109,6 +114,93 @@ public class DrugReference {
 
 	public void setName(String name) {
 		this.name = name;
+	}
+
+	/**
+	 * @return the reference data's own canonical name for the SUBSTANCE this entry is a row of — the
+	 *         DDInter {@code rxnorm_name} — or null for a source that publishes none (the curated
+	 *         {@code json} file and the {@code atc} adapter). Unlike {@link #getGenericName()}, which
+	 *         is a chip-label synonym and is deliberately null whenever the display name already
+	 *         contains it, this is set whether or not the two agree: it is an identity, not a label,
+	 *         and it is exactly the field that is EQUAL across a substance's route/formulation rows
+	 *         ({@code Dexamethasone}, {@code Dexamethasone (nasal)}, … all publish
+	 *         {@code dexamethasone}). Consumed by {@link #substanceKey()}.
+	 */
+	public String getSubstanceName() {
+		return substanceName;
+	}
+
+	public void setSubstanceName(String substanceName) {
+		this.substanceName = substanceName;
+	}
+
+	/** A trailing parenthesized qualifier on a display name — the route or formulation a DDInter row
+	 *  is distinguished from its siblings by ({@code Dexamethasone (nasal)},
+	 *  {@code Amphotericin B (lipid complex)}, {@code Tozinameran (5y-11y)}). Anchored at the END, so
+	 *  a parenthetical in the middle of a name is left alone; applied repeatedly by
+	 *  {@link #displayStem} for the handful of names carrying two. */
+	private static final Pattern TRAILING_QUALIFIER = Pattern.compile("\\s*\\([^()]*\\)\\s*$");
+
+	/**
+	 * The substance-level identity of this entry, or {@code null} when the loaded source publishes no
+	 * substance name and this entry can therefore only stand for itself.
+	 *
+	 * <p><b>What it is for.</b> One substance is filed as several rows, so one clinician-facing string
+	 * resolves several entries and a per-entry safety chip becomes several chips for one clinical fact
+	 * (issue #145 on the contraindication arms; #115/#121 solved the partner side of the same problem
+	 * for interactions). This is the key those chips group on. It is deliberately NOT
+	 * {@link #displayLabel()}: grouping on a rendered label is the mistake issue #148 had to undo.
+	 *
+	 * <p><b>Two components, each load-bearing, both measured over the shipped 19 MB KB (2283 entries;
+	 * re-measure before relying on the figures).</b>
+	 * <ul>
+	 *   <li>{@link #getSubstanceName()} — the data's own claim that two rows are one substance. 142
+	 *       values are shared by more than one entry, across 332 entries, and the {@code rxcui}
+	 *       partitions those entries identically, so this is the dataset's substance identity rather
+	 *       than a spelling coincidence.</li>
+	 *   <li>The <b>display stem</b> — the name with trailing qualifiers removed — as a VETO on that
+	 *       claim, because the claim over-merges. 21 of those 142 families hold rows that are
+	 *       genuinely different substances: {@code Omeprazole}/{@code Esomeprazole} (one
+	 *       {@code rxnorm_name}, one {@code rxcui}, one ATC code),
+	 *       {@code Amphetamine}/{@code Dextroamphetamine}, {@code Fenfluramine}/{@code Dexfenfluramine},
+	 *       {@code Gabapentin}/{@code Gabapentin enacarbil}, {@code Netupitant}/{@code Fosnetupitant},
+	 *       {@code Ketoconazole}/{@code Levoketoconazole}, {@code Fenofibrate}/{@code Fenofibric acid},
+	 *       {@code Atropine}/{@code Hyoscyamine} — each of them the {@code enalapril}/{@code enalaprilat}
+	 *       shape issue #121 decided must stay two chips. The stem separates every one of them.</li>
+	 * </ul>
+	 * Neither half works alone: the stem alone merges 9 groups the substance name keeps apart
+	 * ({@code Manganese (chloride)}/{@code (sulfate)}, {@code Dextran (-1)}/{@code (low molecular
+	 * weight)}, {@code Insulin human}/{@code (isophane)}, {@code Iron}/{@code (polysaccharide)},
+	 * {@code Typhoid vaccine (live)}/{@code (inactivated)} …), which are different substances sharing
+	 * one stem. Together they reduce those 332 entries to 177 substances.
+	 *
+	 * <p>Conservative where it cannot tell: a name that extends the family's stem by a WORD rather
+	 * than a qualifier ({@code Hydrocortisone butyrate}, {@code Estrone sulfate},
+	 * {@code Procaine benzylpenicillin}) keeps its own key and so its own chip. Over-reporting one
+	 * chip is the safe direction for a non-blocking advisory; dropping a real one is not.
+	 *
+	 * @return an opaque key, equal exactly for two entries this module treats as one substance
+	 */
+	Object substanceKey() {
+		String substance = normalizeName(substanceName);
+		if (substance == null) {
+			return null;
+		}
+		return Arrays.asList(substance, displayStem());
+	}
+
+	/** @return {@link #getName()} with any trailing parenthesized qualifier(s) removed, normalized by
+	 *          {@link #normalizeName} — the empty string when the name is blank or is nothing but a
+	 *          qualifier, which keeps the key total. */
+	private String displayStem() {
+		String stem = name == null ? "" : name;
+		String previous;
+		do {
+			previous = stem;
+			stem = TRAILING_QUALIFIER.matcher(stem).replaceFirst("");
+		} while (!stem.equals(previous));
+		String normalized = normalizeName(stem);
+		return normalized == null ? "" : normalized;
 	}
 
 	public String getDrugClass() {
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReference.java
@@ -183,11 +183,10 @@ public class DrugReference {
 	 *
 	 * <p>Conservative where it cannot tell, in BOTH directions. A name that extends the family's stem
 	 * by a WORD rather than a qualifier ({@code Hydrocortisone butyrate}, {@code Estrone sulfate},
-	 * {@code Procaine benzylpenicillin}) keeps its own key and so its own chip, and 21 of the 142
-	 * families are left entirely unmerged for that reason — some of which the KB is in fact naming one
-	 * substance two ways ({@code Thallous Chloride}/{@code Thallous chloride tl-201},
-	 * {@code Typhoid vaccine (live)}/{@code Typhoid vaccine live}). Over-reporting one chip is the safe
-	 * direction for a non-blocking advisory; dropping a real one is not.
+	 * {@code Procaine benzylpenicillin}) keeps its own key and so its own chip — including where the KB
+	 * is in fact naming one substance two ways ({@code Thallous Chloride}/{@code Thallous chloride
+	 * tl-201}, {@code Typhoid vaccine (live)}/{@code Typhoid vaccine live}). Over-reporting one chip is
+	 * the safe direction for a non-blocking advisory; dropping a real one is not.
 	 *
 	 * @return an opaque key, equal exactly for two entries this module treats as one substance
 	 */
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
@@ -10,6 +10,7 @@
 package org.openmrs.module.chartsearchai.reference;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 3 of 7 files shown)


dkayiwa:
## Hardening pass (fresh eyes; I did not write this code)

Six Phase 1 passes and two Phase 2 passes. **The collapse itself is right** — I re-derived both halves of
the key independently and could not break it. What I did find is that one of its branches was
documented as unreachable when it is reachable and load-bearing, that its decisive design argument was
pinned by nothing, and that four quantitative claims in the javadoc were wrong.

Measured myself, full reactor `mvn clean install`: **branch head `4841446` = api 876 + omod 61**;
**after hardening = api 882 + omod 61**, 0 failures, 0 errors, 34 skipped. +6 tests, all new; no
existing test changed. (Your `main` figure of 869 I am relaying, not re-measuring — `main` no longer
test-compiles in this environment, see item 7 below.)

### 1. The rank-replacement branch is reachable, and it is what makes the fix correct

The javadoc's "One honest limit" paragraph said the strongest candidate is always its group's first, so
the replacement branch "is not reachable through the production path today and no test pins it". The
premise it rested on — *every row of a substance carries that substance's name among its aliases* — is
false for this KB.

Measured through `DrugSafetyValidator.validate` over the shipped 19 MB KB: an allergy recorded as
`Tozinameran` resolves the **fourth** row of its five-row collapsed group, because none of the three
age-presentation rows carries the bare `tozinameran` among its aliases and `lookupByToken` takes the
earliest match. A question naming the vaccine puts all five in play in dataset order, so three class
chips are raised *before* the identity one.

Mutation sweep on a throwaway worktree, replacement branch → first-wins:

```
- The patient has a recorded allergy to Tozinameran (…).
+ Tozinameran (12y+) (…) is in the same ATC class (J07BN) as the patient's allergy to Tozinameran (…) — possible cross-reactivity
```

That is exactly the vacuous self-cross-reactivity #145 exists to remove, **kept instead of removed** —
and all seven original tests stayed green under that mutation. Five of the 121 groups the key merges
contain such a row; `Tozinameran`, `Insulin aspart (aspart)` and `Iobenguane (I-131)` were each
verified end-to-end. Now pinned by
`theIdentityChipSurvivesWhenTheAllergenRowIsNotItsGroupsFirst` (+ a precondition test for the premise).

### 2. "A collapse inside either arm leaves the other emitting the siblings" was pinned by nothing

Giving `addActiveOrderContraindications` its own ledger leaves **all seven** original tests green: in
each of them one call site supplies the whole group (the order-driven case reaches every hydrocortisone
row through the order name; the Richard Jones case has every dexamethasone row in play, so the order arm
skips them all). `oneLedgerSpansBothCallSites` now makes both call sites contribute
to the same key with neither redundant, and is red on that mutant.

### 3. Four wrong quantitative claims, deleted rather than renumbered

Everything else in the new prose I verified against the shipped KB and left alone.

| claim | measured | action |
|---|---|---|
| "the stem alone merges **9** groups the substance name keeps apart, **8** of them different substances" | 10 groups; **7** by a differing substance name, **3** by an ABSENT one — and `Typhoid vaccine (live)/(inactivated)`, cited as an example, is one of the 3 | restated as the mechanism, no count; `Varicella Zoster Vaccine (Recombinant)/(live/attenuated)` named instead, where the distinction decides whether an immunocompromised patient may have it at all |
| "**21** of the 142 families are left entirely unmerged **for that reason**" (word-extension) | 21 is right; only **4** are that case. The other 17 are `Omeprazole/Esomeprazole`, `Amphetamine/Dextroamphetamine`, `Fenofibrate/Fenofibric acid` … — the very pairs the paragraph *above* cites as ones the stem must separate | clause deleted; filing them under "conservative where it cannot tell" inverts the meaning for most of them |
| "applied repeatedly … for the handful of names carrying **two**" qualifiers | **zero** KB names carry two | data claim deleted; the loop is described as the guard it is |
| "null for … the curated `json` file" | that schema binds `DrugReference` directly, so a hand-authored file **can** set `substanceName`; only the *shipped* dataset carries none | narrowed |

Verified and left as written: 2283 entries; 142 shared-substance families over 332 entries; the `rxcui`
partitioning those entries identically (0 family splits, 0 rxcui spanning two families); 332 → 177
substances; all eight named over-merge pairs sharing one substance name with the stem separating each;
and both live figures the class javadoc quotes — the dexamethasone case at **4** chips and the
hydrocortisone-order case at **4** — reproduced on `main` `dd651ef` through `validate()`.

### 4. What I could not break

- **Under-reporting.** Within one `(substance, finding)` key every candidate concerns the same pair, and
  the ranks mirror the arm's own per-allergen precedence. Of the 121 groups the key merges, **0** diverge
  on ATC codes and **0** on level-4 subgroups — and curated-group membership is a pure function of those
  codes (`CrossReactivityGroup.groupsOf`), so a dropped sibling's chip differed from the survivor's only
  in the route qualifier of its subject label. Documented, with the re-measure instruction.
- **Wrongly merged pairs.** I enumerated all 121 merged groups. The sharpest candidates (Hib conjugate
  carriers `prp-t`/`prp-omp`/`hboc`, `Insulin aspart (aspart protamine)`/`(aspart)`,
  `Iobenguane (I-123)`/`(I-131)`) all fail the test that matters: the chip that gets dropped never
  carried the distinguishing information, so no clinical claim is lost.
- **Cross-arm coverage.** All four `TYPE_CONTRAINDICATION` constructions go through the ledger; there is
  no fifth path. Nothing in `validate` removes, reorders or sorts `warnings`, so the stored positions
  cannot go stale, and the screening arm seeds its suppression set after the ledger's last write.
- **Leakage.** The ledger is a local of `validate` holding only a local list; `SafetyWarning` is
  immutable; `DrugReference` overrides neither `equals` nor `hashCode`, so the two key spaces are
  disjoint by type. Nothing crosses a request or a patient.
- **`json`/`atc` degradation.** Null substance name → entry-identity keys → nothing collapses. Now also
  covered directly: `oneCuratedRuleAuthoredTwiceRaisesOneChip` runs the ledger with a null substance key
  through the real `JsonDrugReferenceSource`.
- **#146 is untouched**, as intended. `ActiveOrderContraindicationTest.aPrescribedDrugTheAnswerOnly`
  `EchoesIsStillCheckedAgainstTheAllergyList` still asserts **2** contraindications for exactly #146's
  shape and still passes. Flagging it as the assertion #146's eventual fix has to contend with.

### 5. One behaviour change the PR introduced silently, now pinned

The curated-rule key normalises `(type, token)`, so **one rule authored twice in different case now
raises one chip** where both were previously appended. Reachable only in hand-authored data (no shipped
dataset contains it), desirable, but neither written down nor tested. New curated fixture +
`oneCuratedRuleAuthoredTwiceRaisesOneChip`. The cost it pins: ties keep the incumbent, so the second
rule's **note** is dropped — right for a re-spelling, lossy for two genuinely different notes under one
token. Reported separately rather than designed around here.

### 6. Polish

- `ContraindicationChips` held the chip's position and its rank in two maps keyed alike, written in
  lockstep. A future edit breaking the lockstep is not loud: the resulting throw lands inside a
  `validate` both entry points wrap in `catch (RuntimeException) → return nothing`, so the request would
  lose *every* contraindication chip. One `RaisedChip` holder makes the pairing structural.
- The fixture-service arrangement was a verbatim copy of `DirectAllergyContraindicationTest`'s, and the
  two steps have to stay together (`serviceWith` pins the groups EMPTY, so a fixture service built
  without the second call silently cannot raise a curated-group chip). Moved to
  `DrugReferenceTestSupport.ddiFixtureService`, with `names()` beside it.
- Three javadoc rewraps to the file's width; `oneLedgerSpansBothCallSites` renamed ("arm" already means
  the rule/class/allergy arm here, and what this excludes is one ledger per *call site*).

**Rejected:** six proposals to cut "redundant" javadoc — the RxCUI rejection at the setter, reason (2)'s
`#148` lesson, the ledger's span-both-call-sites reason, the defect narrative in the class overview. Each
removes a recorded reason rather than a repetition, and the repetition is deliberate: a reader arriving at
the setter, at the call site, or at the arm needs the reason *there*. Left as written.

### 7. Load-bearing decisions I was tempted to change and did not

- **Most-specific-relationship rather than highest-severity.** Correct here: a contraindication chip
  carries no severity, and the rank order reproduces the precedence `addAllergyContraindications`
  already applies per allergen. Left exactly as it is.
- **Keeping the curated-rule and allergy arms in separate key spaces** (#146). Left as it is, for the
  reason recorded: a curated token may name a class.
- **The mirror-shape residual** below is a pre-existing defect this PR neither causes nor worsens
  (identical on `main`). Reported, not fixed — it needs a decision about which label the identity chip
  names, which is more than a hardening pass should take.

### Found, not fixed — for filing

1. **A substance can still be reported cross-reactive with itself, by the mirror path.** `IDENTITY` is
   awarded on entry identity (`allergen == ref`), not substance identity, so when the allergen row is a
   sibling of the subject *and is not itself in play*, the group has no identity candidate and the
   survivor is a class chip about the same substance. Measured on `main` **and** on this branch,
   identical — pre-existing:
   - `Is Tozinameran safe for her?` + allergy `Tozinameran (12y+)` → *"Tozinameran (…) is in the same ATC
     class (J07BN) as the patient's allergy to Tozinameran (12y+) (…)"*
   - `Is Tetryzoline (nasal) safe?` + allergy `Tetryzoline (ophthalmic)` → same shape on `R01AA`.
   The obvious one-line widening (`subjectKey(ref).equals(subjectKey(allergen))`) is **not** the fix as
   written: the identity chip names `ref.displayLabel()`, so it would then mislabel which variant the
   allergy was recorded against. Needs a decision.
2. **The interaction arms have the same subject-side duplication.** `addInteractionWarnings` runs once
   per `ref` in `inPlay` and `bestRulePerPartner` groups per *partner* within one subject, so the subject
   side is ungrouped. **Confirmed live on this build**, Sarah Taylor, *"Is it safe to give
   hydrocortisone?"* — beside six correctly-collapsed contraindication chips:
   ```
   [interaction] Hydrocortisone              … interacts with active order diclofenac — Moderate. The combined use of corticosteroids and nonsteroidal anti-inflammatory drugs (NSAIDs) may increase the potential for serious gastrointestinal (GI) toxicity…
   [interaction] Hydrocortisone (ophthalmic) … interacts with active order diclofenac — Moderate. Concomitant use of ophthalmic nonsteroidal anti-inflammatory drugs (NSAIDs) and ophthalmic steroids may increase the potential for healing problems…
   ```
   The two rows carry **different mechanism prose**, so a fix must choose a survivor, not dedup.
3. **`DrugReferenceInjector` duplicates the reference RECORD per route variant.** `matchingEntries`
   dedups on `ref.getId()` and variants carry distinct ids. **Measured** through the real injector, not
   code-reasoned: *"Is it safe to give dexamethasone?"* injects **4** `drug_reference` records totalling
   **5254** characters; *"Is phenylephrine safe?"* injects **5** totalling **6975**; a single-row drug
   (clarithromycin) injects 1 of 1524.
4. **`sharedClass` names the wrong pharmacological class.** It returns the first shared level-4 subgroup
   in the *allergen's* ATC order. **Confirmed live on this build**, all in Sarah Taylor's single answer
   beside a correct one, which is what makes it obvious: `Prednisone … (H02AB)` ✓,
   `Methylprednisolone … (D10AA)` (anti-acne), `Budesonide … (R01AD)` (nasal),
   `Hydrocortisone … (A01AC)` (local oral) — one systemic dexamethasone allergy, four different classes
   named, three of them wrong. #145 calls this its option 3 and says not to bundle it; it is not bundled.
5. **The KB names one substance two ways and the conservative key keeps both** — reachable and
   ATC-carrying, unlike the ATC-less `Thallous chloride` example the javadoc gives, so this one costs a
   real chip. `Pfizer-BioNTech Covid-19 Vaccine` and `Tozinameran` share one `rxnorm_name` and one
   `rxcui` but differ by stem. **Confirmed live** (Kevin Brown, above): the surviving pair is the correct
   identity chip *plus* `"Pfizer-BioNTech Covid-19 Vaccine (…) is in the same ATC class (J07BN) as the
   patient's allergy to Tozinameran (…)"` — the same substance, reported as cross-reactive with itself.
   Accepted direction of the design; worth recording as its concrete cost.
6. **Two curated rules sharing `(type, token)` but carrying different notes now collapse and the second
   note is dropped.** Operator-authored data only. Is silently keeping the first note the intended spec?
7. **Environment, not this PR: chartsearchai `main` no longer test-compiles against the locally-installed
   `querystore` SNAPSHOT.** A `mvn install` of openmrs-module-querystore at 2026-08-06 01:42 added
   `QueryStoreService.getContextSlice(String, String, ContextSliceRequest)`, which
   `api/src/test/java/…/api/impl/CountingQueryStoreStub.java` does not implement. Verified identical on
   `dd651ef`. Every build number above was measured with those local-copy artifacts temporarily set aside
   and restored afterwards.

### Live re-verification

Deployed and probed on the 3.7.1 standalone at the final SHA, through `standalone-verify.sh` (full JVM
restart, `hits=75` retrieval gate, warm control chip). Loaded source confirmed by
`GET /chartsearchai/drugreferencestatus`, not inferred from log lines:
`{enabled: true, loaded: true, inert: false, entryCount: 2283, sourceFormat: "ddinter", origin: "appdata:chartsearchai/ddi_knowledge_base.json"}`.
Every patient resolved **by name against the live DB**; the uuids that came back match the ones in the
report, which is itself a check.

| patient (resolved by name) | question | chips | verdict |
|---|---|---|---|
| **Richard Jones** `a39e82e7…` | is it safe to give dexamethasone? | **1** | `[contraindication] Dexamethasone — "The patient has a recorded allergy to Dexamethasone."` and **1** injected `safety_finding`. Had the fix been wrong: 4 chips, 3 of them *"Dexamethasone (nasal\|ophthalmic\|topical) is in the same ATC class (A01AC) as the patient's allergy to Dexamethasone"* — measured on `main`. |
| **Sarah Taylor** `dc8560c9…` (8 orders) | is it safe to give hydrocortisone? | **6** contra | all distinct substances: `Hydrocortisone`, `Hydrocortisone butyrate`, `Dexamethasone` (identity), `Prednisone`, `Budesonide`, `Methylprednisolone`. The three hydrocortisone route rows collapsed to one; the *butyrate* correctly kept its own chip (the conservative word-extension case). Had it been keyed per entry: one chip per route row. |
| **George Anderson** `d5c9cdad…` | is it safe to give esomeprazole? | **2** | **must-not-collapse holds** — `Omeprazole` *and* `Esomeprazole`, both *"same ATC class (A02BC) as the patient's allergy to Pantoprazole"*. Had the key been `rxnorm_name` alone (issue #145's proposal 2), one of these two real chips would be gone. |
| **Margaret King** `f2397b8a…` | is it safe to give dexamethasone? | **0** | dexamethasone order, no allergy. Null control: discriminates only against a collapse that fabricates. |
| **Kevin Brown** `e89247a2…` (fixture I built) | is it safe to give the Pfizer-BioNTech COVID-19 vaccine? | **3** | **§1's finding, live.** The five `Tozinameran` presentation rows collapse to one chip and it is the identity one: `"The patient has a recorded allergy to Tozinameran (…)"` — reached only by *replacing* the class chip the `(12y+)` row had already raised. With a first-wins ledger that chip is instead *"Tozinameran (12y+) … is in the same ATC class (J07BN) as the patient's allergy to Tozinameran"*. The other two are the separately-stemmed `Pfizer-BioNTech Covid-19 Vaccine` row (item 5 below) and his pre-existing Clarithromycine allergy-plus-order (#143's arm). |

Four regression controls, exact: **Mary × clarithromycin 1 Major ×simvastatin; Agnes × warfarin 1 Major
×aspirin; Joshua × ibuprofen 2 (NSAID contra + lisinopril Moderate — two facts about different objects,
still two); Mary × paracetamol 0** (re-run, since a first post-restart query must never be trusted for
chip *absence*).

### Demo data I created (no teardown)

- Concept **`Tozinameran`** `08a52574-411b-498f-84f3-7e9ef54c29d7` (class Drug, datatype N/A). Needed
  because a **non-coded** allergen is rejected by this OpenMRS version —
  `PatientAllergyResource2_0` hits `not-null property references a null or transient value :
  org.openmrs.Allergy.allergen.codedAllergen` — so the fixture had to be a real concept.
- **Kevin Brown** `e89247a2…` — DRUG allergy to it, `fd60a190-13cf-423b-95c6-dd1e122b4024`, comment
  `PR #160 hardening fixture (agent-created demo data, no teardown)`.

Nothing else added, voided or changed; no GP touched. Left running: this branch's omod, deployed by
`standalone-verify.sh`.

<sub>One n=1 observation I am **not** claiming as a finding, because answer prose is not reproducible on
this box: in George Anderson's answer the model wrote the ATC code as `A02DC` while the chip it cited
says `A02BC`. Worth a look with a proper n; not asserted here.</sub>

dkayiwa:

- Final head `5011c82`: `mvn -o clean install` on a clean detached worktree → **882 + 61 = 943, 0 failures, 34 skipped**, BUILD SUCCESS. Reproduces the reported figure at the SHA being merged.
- CI green on Java 11/17/21.
- Test-file changes across the branch: **3 new files**, plus **+21/-0** in the pre-existing `DrugReferenceTestSupport` (the shared fixture-service arrangement). No assertion weakened, no expected value changed.

The hardening finding that matters most, for the record: **the rank-replacement branch was documented as unreachable and pinned by nothing, and it is in fact load-bearing.** An allergy recorded as `Tozinameran` resolves the *4th* row of its 5-row group — no presentation row carries the bare substance name as an alias, so `lookupByToken`'s earliest-match skips them — so three class chips are raised before the identity one. Mutated to first-wins, the result is the vacuous *"Tozinameran (12y+) … is in the same ATC class (J07BN) as the patient's allergy to Tozinameran"* — exactly the defect #145 exists to remove — **with all seven original tests still green**. 5 of 121 merged groups contain such a row. That is now pinned.

Also worth recording: the PR's own decisive rationale ("a collapse inside either arm leaves the other emitting the siblings") was pinned by nothing — a per-call-site ledger left all seven original tests green. Now pinned.

Environment note, unrelated to this PR but it affected measurement: a local `mvn install` of openmrs-module-querystore at 01:42 today added `getContextSlice(String, String, ContextSliceRequest)` to `QueryStoreService`, which `CountingQueryStoreStub` does not implement, so chartsearchai stopped test-compiling **locally** while staying green in CI (CI resolves the remote snapshot, which predates it). I confirmed the new abstract method in the installed jar with `javap`, restored the resolved snapshot from the Aug-3 remote copy so local builds match CI, and kept the locally-installed jar aside rather than deleting it. Filed as #165.

Follow-ups filed from this pass: #164 (a substance still reported cross-reactive with itself, by two paths this key does not cover — path 1 verified identical on `main`, so pre-existing), #165 (the stub brittleness above). Previously filed from the fix pass: #161, #162, #163. **#146 is deliberately NOT closed by this PR** — the curated-rule and allergy arms are kept in separate key spaces because a curated token may name a class rather than a drug, and `ActiveOrderContraindicationTest` still asserts 2 chips for #146's exact shape and passes; that assertion is what #146's fix will have to contend with.

Issue #164, opened 2026-08-05
A substance is still reported as cross-reactive with itself, by two paths the substance key does not cover


dkayiwa:
## What

#145 / PR #160 collapsed route-variant duplicates, but a substance can still be reported as cross-reactive **with itself**, by two distinct paths. Both were measured live, and the first was verified **identical on `main` and on the #160 branch**, so it is pre-existing rather than introduced.

### Path 1 — the mirror shape: the allergen row is a sibling that is not itself in play

`IDENTITY` is awarded on **entry** identity, not **substance** identity. When the row the allergy resolved to is a sibling that is not among the entries in play, the group has no identity candidate at all, so the best available relationship is "shared ATC class" — against itself.

Measured:
- *"Is Tozinameran safe for her?"* with an allergy recorded as `Tozinameran (12y+)`
- *"Is Tetryzoline (nasal) safe?"* with an allergy recorded as `Tetryzoline (ophthalmic)`

### Path 2 — stem divergence inside one substance

The collapse key is deliberately conservative: substance name **and** display-name stem. `Pfizer-BioNTech Covid-19 Vaccine` and `Tozinameran` share one `rxnorm_name` and one `rxcui` but have **different stems**, so they do not collapse. Live, Kevin Brown gets the correct identity chip *plus* a vacuous class chip about the same substance.

## Why the obvious fix is wrong

Widening `IDENTITY` to substance identity is a one-line change and it is **not** the fix: the identity chip names `ref.displayLabel()`, so widening it would report the allergy against whichever variant happened to win, mislabelling which variant the allergy was actually recorded against. Any fix has to decide **which label the identity chip names** when the matched row and the recorded allergen row differ — that is a design question, not a patch.

## Relationship to other issues

- **#145 / PR #160** built `DrugReference.substanceKey()` and the chip ledger; this is the residual its author and hardener both flagged and deliberately did not take on.
- **#152** is a third, independent route to the same symptom (a KB row that pairs a drug with itself), so fixing this does not subsume it and vice versa.
- **#161** is the adjacent "the class we name is wrong" defect; here the class relationship should not be reported at all.

## Provenance

Both paths measured live on the standalone during #160's hardening and reported to me, including the equality of path 1 on `main`. I have not re-run them.

dkayiwa:
**New information for scoping this, from #173's work — the same shape exists in the interaction arm too.**

[...]

Issue #172, opened 2026-08-06
addAllergyContraindications re-resolves every allergy token once per drug checked, sweeping the whole 2283-entry KB each time


dkayiwa:
## What

`DrugSafetyValidator.addAllergyContraindications` (≈:2010) calls `lookupByToken` **inside** the per-allergen loop, and the method itself runs once per drug in play (≈:323) plus once per active-order entry (≈:2139). So a pure function of (allergy tokens, loaded KB) is recomputed **D × A** times per request, and each resolution sweeps all 2283 entries and their aliases with a `toLowerCase` allocation per (entry, alias).

Present on `main`; not introduced by #160 or #166.

## Severity

Efficiency only — no wrong output. Filed because the shape is quadratic in two independent inputs, and polypharmacy charts are exactly where the safety layer matters most: the demo box already has an 8-order patient, and a 10-order chart with a few allergies puts the estimate around 10⁵ string allocations per request.

## The constraint any fix must respect — this is the important part

**Cache in a per-`validate` local, never in a field.** `DrugReferenceService.getAll()` is hot-reloadable, and a cached `DrugReference` that survives a reload fails the `allergen == ref` **reference** comparison the contraindication arms rely on. That silently re-opens #145 (route-variant duplication) with no test failing, because every test loads the KB once.

That trap is why this is worth an issue rather than a drive-by: the obvious fix is the wrong one.

## Provenance

Structure confirmed from the code by the #166 hardening pass. The allocation magnitude is an estimate from a review, **not profiled** — worth measuring before optimising, since the request also makes an LLM round trip that dwarfs most in-process work.

dkayiwa:
**Correction to a premise this issue states, measured during #248.**

This issue justifies "cache in a per-`validate` local, never in a field" with the claim that `DrugReferenceService.getAll()` is **hot-reloadable**, and #230 repeated it for the cross-reactivity groups. For the groups that premise is **false**: `CrossReactivityGroupsLoader.parse` builds fresh objects and `DrugReferenceService.getCrossReactivityGroups()` caches them for the bean's lifetime with **no clearing path**.

The rule is still right — it survives for a different reason, that `setAtcPrefixes` must stay authoritative after a read — and #248 states that reason in the code rather than the reload one.

[...]

PR #173, opened 2026-08-06
fix(drug-safety): one substance, one interaction chip, one reference record, no self-pairs

### LINKED ISSUE #152: One DDInter row is a self-interaction and the parser does not skip it, so a drug can be reported as interacting with itself
## What

Exactly **1 of 295,184** rows in the shipped DDInter 2.0 knowledge base pairs a drug with itself (`DDInter225`, botulinum toxin). The parser does not filter self-pairs, so that drug can be raised as interacting with itself.

## Why it is worth fixing anyway

The blast radius today is one drug, and no demo patient has it — this is not urgent. It is worth a guard because:

- The failure is nonsensical rather than merely noisy: "botulinum toxin interacts with active order botulinum toxin" reads as a defect in the feature to any clinician who sees it.
- A self-pair guard is a one-line invariant at load time, and it protects against future KB revisions rather than just this row.
- It composes badly with the route-variant duplication in #145: a self-pair plus route variants of the same substance is how "one substance reported as cross-reactive with itself" gets produced in more than one way.

Related: #145 (route-variant collapse), #150 (the other load-time data-validity guard).

## Provenance

Counted over the shipped KB during the #148 hardening pass and reported to me; I have not re-run the count.

### LINKED ISSUE #162: The interaction arms have the same subject-side route-variant duplication that #145 fixed for contraindications
## What

`DrugSafetyValidator.addInteractionWarnings` runs **once per entry in `inPlay`**, and `bestRulePerPartner` groups per *partner* within a single subject — so the **subject** side is never grouped. When the drug in play resolves to several route/formulation variants of one substance, each variant produces its own interaction chip against the same partner.

Measured live: Sarah Taylor, *"Is it safe to give hydrocortisone?"* reports `hydrocortisone × diclofenac` **twice**:

```
Hydrocortisone interacts with active order diclofenac — Moderate.
  The combined use of corticosteroids and nonsteroidal…

Hydrocortisone (ophthalmic) interacts with active order diclofenac — Moderate.
  Concomitant use of ophthalmic nonsteroidal…
```

## This is NOT a pure duplicate, which changes the fix

The two rows carry **different mechanism prose** — one systemic, one ophthalmic. So a fix cannot simply dedup; it has to choose which mechanism survives, and choosing the ophthalmic text for a systemic order would be actively misleading. That makes it a harder problem than the contraindication case, where the collapsed chips were near-identical.

Consider also whether the *subject* should be resolved to a route at all when the clinician asked about "hydrocortisone" without specifying one.

## Relationship to #145 / PR #160

PR #160 fixed exactly this shape for the **contraindication** arms, introducing `DrugReference.substanceKey()` (substance name plus display-name stem) as the collapse key and a validate-scoped ledger. The natural fix here is to apply `substanceKey()` on the subject side of the interaction arms — the key already exists and is already measured against the enalapril/enalaprilat constraint (#121) that must not collapse.

#160 deliberately scoped itself to the contraindication arms rather than widening; this is the follow-up.

## Provenance

Measured live on the #160 build during that work and reported to me, including the two chip strings above verbatim. I have not re-run the probe.

### LINKED ISSUE #163: DrugReferenceInjector injects one near-duplicate reference record per route variant, spending the prompt budget several times over
## What

`DrugReferenceInjector.matchingEntries` de-duplicates on `ref.getId()`, and route variants of one substance deliberately carry **distinct ids** — the parser falls back to the DDInter id when the RxCUI is shared. So a question about one drug injects one `drug_reference` record per variant: **4 near-identical records for dexamethasone, 5 for phenylephrine**, each rendering up to `MAX_INTERACTION_RENDER_CHARS` = 1500 characters.

## Why it matters

This is prompt budget, not just noise. Several thousand characters of near-duplicate reference text crowds out chart records the answer actually needs, and the module has repeatedly had answer-quality problems traced to slice budget (#95, #99). It also gives the model several differently-worded copies of one fact, which is the setup for the kind of miscopying #142 records.

## Why it is not visible from the outside

The REST response returns only *cited* references, so this cannot be measured from probe output — it is visible in what gets injected, not in what comes back. That is also why it survived several live verification passes.

## Relationship to #145 / PR #160

Same root cause as #145 (route variants of one substance treated as distinct entries), different layer — #160 fixed the chip layer only. `DrugReference.substanceKey()`, added by #160, is the key to reuse here. Adjacent to #151, which is the other injector-side resolution defect.

## Provenance

Reasoned from the code by the #145 fix agent and explicitly labelled by it as **not** measured live, with the reason (only cited references are returned). The 4-and-5 variant counts come from its KB measurement. Worth confirming by instrumenting the injected slice before fixing.

dkayiwa:
## The shared root cause

DDInter files one substance as several reference rows — route and formulation variants sharing one
`rxnorm_name` (121 substances over 276 entries, once the display-stem veto has separated the pairs of
genuinely different substances that share a name). One clinician-facing word resolves all of them,
because `findByQuery`/`findByDrugName` return every entry whose aliases match. So every layer that
enumerates matching entries produced **N of something where one was meant**.

#145/#160 fixed that for the contraindication chips and built the key, `DrugReference.substanceKey()`
— substance name AND display-name stem, measured against #121's enalapril/enalaprilat constraint.
These are the three remaining sites, swept together rather than one at a time, because fixing them one
at a time is what kept revealing the next.

### #162 — the interaction arms' subject side

`addInteractionWarnings` ran once per entry in the drugs-in-play set, and `bestRulePerPartner` groups
per *partner* within one subject, so the subject side was never grouped. The arm now takes the
substance's rows.

This was not a duplicate to drop: the two rows carry **different mechanism prose**, so a survivor had
to be chosen, and the chip's own subject label named a route the clinician never mentioned. Two
decisions, both new:

**1. The chip names the substance, not a route.** Nothing at this layer can know the route — every
variant of a substance publishes the same aliases and the same ATC list, which is the data-side gap
#115 records — so naming one asserts a preparation that is not on the chart. The chip is named after
the row carrying no trailing route qualifier (`DrugReference.namesNoRoute()`), chosen by
`DrugReference.canonicalRow`, which the injector below shares.

For the 11 of 121 shipped families that publish **no** unqualified row (`Oxymetazoline (nasal)` /
`(ophthalmic)` / `(topical)`, `Iobenguane (I-123)` / `(I-131)`) the survivor still carries a
qualifier. The KB names no unqualified form for those substances, and manufacturing one by stripping a
display name is the pattern-match-a-label mistake #148 had to undo. In 7 of the other 110 the
unqualified row is not the family's first, which is why the choice cannot be left to dataset order.

**2. The surviving rule: most severe rating, then the row naming no route, then the longer note.**

The route step sits *between* the other two, on measurement over the shipped 19 MB KB (2026-08-06):

* in **1395 of 25,847** (substance, partner) groups the equally-rated route-qualified row carries the
  **longer** note while an unqualified row also has a rule — so the pre-existing
  severity-then-longer-note tie-break alone would hand those 1395 groups ophthalmic or topical prose
  under a systemic subject. `Ketorolac (ophthalmic)` against lepirudin is 495 characters against plain
  `Ketorolac`'s 265.
* in **71** groups the route-qualified row is *strictly more severe* (`Sirolimus (protein-bound)` Major
  against plain `Sirolimus` Moderate). Putting the route above severity would report the milder rating
  for a pair the source rates worse. Under-warning is the one direction a non-blocking advisory the
  clinician adjudicates must not take — the same call `bestRulePerPartner` already records for the
  partner side — so severity leads, and those 71 keep a chip whose prose describes the qualified
  presentation. That residue is stated in the javadoc and pinned by a test.

Sarah Taylor's measured case is in neither set: both hydrocortisone rows rate diclofenac Moderate and
the systemic note is the longer one (219 characters against 183), so grouping alone would have picked
the right prose there. The route step is what makes that a rule rather than luck.

### #163 — one injected reference record per variant

`DrugReferenceInjector.matchingEntries` de-duplicated on `ref.getId()`, and variants deliberately
carry distinct ids: the parser falls back to the DDInter id when the RxCUI is shared, precisely so
citations stay unambiguous. So one question word injected one near-duplicate record per variant, each
rendering up to `MAX_INTERACTION_RENDER_CHARS` of interaction prose. The map is now keyed on the
substance, and the surviving row is the same canonical row the chip names — a chip naming the substance
beside a record naming one of its routes would be the chip-versus-prose divergence this module keeps
having to remove.

The id remains the fallback for a source publishing no substance name (the curated `json` file, the
`atc` adapter), because the surviving entry's id becomes the injected record's `resourceId` and two
records sharing one would make a citation ambiguous.

**What the collapse gives up, measured rather than assumed.** The surviving row's rules are the ones
the record renders, and a sibling can carry a partner the survivor does not: 80 of the 121 multi-row
substances have at least one such partner, 2627 in total, some lopsided (`Olopatadine`'s canonical row
has 112 partners against its family's 397). What that cannot cost is the actionable half — a partner
the patient is on whose rule clears the severity floor raises a chip whatever row carries it, and since
#110 every chip is injected as its own citable safety-finding record carrying that rule's mechanism
note verbatim. What the sibling rows lose is breadth in the `Interactions:` tail, the section `render`
already truncates to one compact representative whenever a relevant partner is promoted. Stated in
`orderedInteractionNotes`' javadoc rather than left to be discovered.

A second commit adds the injected slice's character total to the existing DEBUG line. #163's cost is
invisible from outside the module — the REST response returns only *cited* references, which is why it
survived several live verification passes — and a count alone does not settle it, since what crowds out
chart records is characters.

### #152 — self-interaction rows at load

`DdiDrugReferenceSource.parse` now skips an interaction row joining a drug to itself, in both shapes:

* the same row on both sides — `DDInter225`, botulinum toxin type A, the KB's only one of 295,184. Its
  mechanism text is about administering different botulinum *serotypes* together, which this KB carries
  no second row for, so the pair is an artifact of its granularity; what reached a clinician was
  "Botulinum toxin type A interacts with active order botulinum toxin type A".
* **25 more** pairing two route/formulation rows of one substance (`Lidocaine` against
  `Lidocaine (topical)` and the like). These are not merely redundant either: every row of a substance
  publishes the same `rxnorm_name`, which is the match token a rule carries and the label a chip prints,
  so such a pair can *only* render as a substance interacting with itself. The systemic-plus-topical
  exposure the KB row is about cannot be stated by anything this module renders, while the
  self-reference can. They are in this PR because they are the same "N rows of one substance" family
  seen from the parser side, and because they are the second route to the self-reference symptom the
  issue names.

At load rather than per arm: those rows feed the drug-in-play chips, the screening arm, the
question-pair arm, the promoted notes inside the injected reference record and the pre-answer finding
derived from a chip, and one invariant at the parse boundary covers all of them and any future KB
revision. Only the `ddinter` source is guarded — a hand-authored curated file is the operator's own
data, and the `atc` adapter carries no rules.

### Shared

`DrugReference.substanceKey` gained a static form over the two fields it reads, so the parse-time guard
and the chip grouping cannot come to disagree about what one substance is — which would leave a
self-pair loaded for exactly the rows the chips then merge. `canonicalRow` is the one definition of
which row represents a substance, shared by the chip label and the injected record.
`substanceGroupKey()` moved onto `DrugReference` so the contraindication ledger and the interaction
arms group on one key.

The screening arm and the question-pair arm are **unchanged**: the screening arm gets a single-subject
overload, and with one subject the route step cannot fire (both candidates have the same subject), so
its choice is byte-identical to before. The question-pair arm already keyed on the match token, which
route variants share.

## Build

Baseline on `main` at `45c1cc4`, measured on a clean tree: **953 tests — 892 api + 61 omod, 0 failures,
34 skipped.**

This branch: **975 tests — 914 api + 61 omod, 0 failures, 34 skipped.** The +22 are the three new test
classes; no existing test was modified and none changed result.

New tests, all over slices taken verbatim from the shipped KB and driven through the real
`DdiDrugReferenceSource.parse`, the real `DrugSafetyValidator.validate` and the real
`DrugReferenceInjector.injectRecords`:

* `InteractionRouteVariantTest` (8) — the live hydrocortisone case; the note-length tie-break the route
  step has to override; severity still leading; a family with no unqualified row; the
  omeprazole/esomeprazole must-not-collapse case, which shares a mechanism group so the two chips carry
  byte-identical notes and only the subject key can tell them apart; partners not merging; and the
  injected finding count.
* `SelfInteractionTest` (8) — both self-pair shapes at parse level, the invariant over every entry in
  the slice, both end to end through `validate`, a genuine pair in the same slice still reported, and
  the injected reference record no longer listing the drug as its own partner.
* `ReferenceRecordSubstanceCollapseTest` (6) — the injected slice measured directly, the two legs of
  `matchingEntries` sharing one record, the must-not-collapse case, and the family whose unqualified row
  is not its first.

17 of the 22 failed before the production change; the 5 that passed are the preconditions and the two
must-not-collapse controls.

## Live verification

Deployed to the 3.7.1 standalone (full 19 MB DDInter KB, `sourceFormat=ddinter`,
`minInteractionSeverity=minor`, `chartMode=fullChart`, local CPU LLM). Loaded source confirmed by
`GET /chartsearchai/drugreferencestatus` — not by a log line, since the load is lazy:

```
```

Every patient was resolved by name against the live DB. One probe was accidentally issued against a
uuid reconstructed from a truncated one; it returned `HTTP 404 {"error":"Patient not found"}` and was
re-run against the resolved uuid, so no result below rests on a fabricated uuid.

### #162 — the measured case. Sarah Taylor (`dc8560c9-6d2b-45bf-861c-8fcf562ec9b1`), 8 active orders

`Is it safe to give hydrocortisone?` — run twice, byte-identical chips both times. The three
interaction chips, verbatim:

```
[interaction] Hydrocortisone
    Hydrocortisone interacts with active order celecoxib — Moderate. The combined use of
    corticosteroids and nonsteroidal anti-inflammatory drugs (NSAIDs) may increase the potential for
    serious gastrointestinal (GI) toxicity, including inflammation, bleeding, ulceration, and
    perforation.
[interaction] Hydrocortisone
    Hydrocortisone interacts with active order diclofenac — Moderate. The combined use of
    corticosteroids and nonsteroidal anti-inflammatory drugs (NSAIDs) may increase the potential for
    serious gastrointestinal (GI) toxicity, including inflammation, bleeding, ulceration, and
    perforation.
[interaction] Hydrocortisone
    Hydrocortisone interacts with active order ibuprofen — Moderate. The combined use of
    corticosteroids and nonsteroidal anti-inflammatory drugs (NSAIDs) may increase the potential for
    serious gastrointestinal (GI) toxicity, including inflammation, bleeding, ulceration, and
    perforation.
```

**One `× diclofenac` chip, and the SYSTEMIC mechanism prose won.** The subject is `Hydrocortisone`; no
chip names `Hydrocortisone (ophthalmic)`, and the ophthalmic wording ("Concomitant use of ophthalmic
nonsteroidal anti-inflammatory drugs (NSAIDs) and ophthalmic steroids …") appears nowhere.

**Had the fix been wrong** this probe would have shown 4 interaction chips, not 3 — the extra one
`Hydrocortisone (ophthalmic) interacts with active order diclofenac — Moderate. Concomitant use of
ophthalmic nonsteroidal …`. Read off the loaded KB, the rows that would each have chipped are
`Hydrocortisone × Celecoxib` (Moderate, 219-char note), `Hydrocortisone × Diclofenac` (Moderate, 219),
`Hydrocortisone × Ibuprofen` (Moderate, 219) and `Hydrocortisone (ophthalmic) × Diclofenac
(ophthalmic)` (Moderate, 183). The six contraindication chips in the same response are unchanged and
are #160's collapse working (`Hydrocortisone butyrate` keeps its own chip deliberately — it extends the
family stem by a word, not a qualifier).

`Is it safe to give ketorolac?` on the same patient: **8 interaction chips, every one named
`Ketorolac`.** Had the fix been wrong there would have been 10, two of them
`Ketorolac (ophthalmic) …` against dexamethasone and hydrocortisone with the ophthalmic prose.

### #152 — Susan Young (`763e6e5f-c489-4bab-8a55-c379f085dd1c`)

Fixture created for this (documented below): four active drug orders added — `Botulinum toxin type A`,
`Lidocaine`, `Neomycin` and `Metoclopramide` — the last two as positive controls, being genuine Major
partners of the first two.

Load-time guard, live, on the real KB:

```
WARN - DdiDrugReferenceSource.parse |2026-08-06T15:45:55,084| Skipped 26 DDInter interaction row(s)
pairing a drug with itself or with another route/formulation row of the same substance — a drug cannot
interact with itself
INFO - DdiDrugReferenceSource.parse |2026-08-06T15:45:55,192| Parsed 2283 DDInter drug-reference entries
```

26 = the 1 exact self-pair plus the 25 route-variant pairs, matching the count measured off the file.

`Is it safe to give lidocaine?` — run twice, identical. **2 interaction chips, both named `Lidocaine`:**

```
[interaction] Lidocaine
    Lidocaine interacts with active order metoclopramide — Major. Coadministration of local
    anesthetics with other oxidizing agents that can also induce methemoglobinemia such as
    antimalarials (e.g., chloroquine, primaquine, quinine, tafenoquine), nitrates and nitrites,
    sulfonamides, aminosalicylic acid, dapsone, dimethyl sulfoxide, flutamide, metoclopramide,
    nitrofurantoin, phenazopyridine, phenobarbital, phenytoin, and rasburicase may increase the risk.
[interaction] Lidocaine
    Lidocaine interacts with active order neomycin — Minor. Limited in vitro data suggest that some
    aminoglycosides may enhance the neuromuscular blocking properties of lidocaine. Data are available
    for neomycin only. No special precautions are necessary.
```

**Had the fix been wrong this would have shown 7 interaction chips, 3 of them reading "… interacts
with active order lidocaine"** — one per subject row — plus three duplicate metoclopramide chips under
`Lidocaine`, `Lidocaine (ophthalmic)` and `Lidocaine (topical)`. Enumerated off the loaded KB:

```
subject 'Lidocaine'                x 'lidocaine'      <- Lidocaine (ophthalmic) Moderate, Lidocaine (topical) Moderate
subject 'Lidocaine'                x 'metoclopramide' <- Metoclopramide Major
subject 'Lidocaine'                x 'neomycin'       <- Neomycin Minor
subject 'Lidocaine (ophthalmic)'   x 'lidocaine'      <- Lidocaine Moderate
subject 'Lidocaine (ophthalmic)'   x 'metoclopramide' <- Metoclopramide Moderate
subject 'Lidocaine (topical)'      x 'lidocaine'      <- Lidocaine Moderate
subject 'Lidocaine (topical)'      x 'metoclopramide' <- Metoclopramide Moderate
```

The two positive controls survived, so the guard did not simply switch lidocaine's rules off.

`Is it safe to give botulinum toxin type A?` — the exact `DDInter225 × DDInter225` row is gone from the
loaded model (the WARN above, and `SelfInteractionTest` pins it at parse level), and the record no
longer lists the drug as its own partner. **But this probe does not discriminate at the chip layer, and
that is worth saying plainly.** It still returns

```
[interaction] Botulinum toxin type A
    Botulinum toxin type A interacts with active order botulinum toxin type a — Moderate. The effect
    of administering different botulinum neurotoxin serotypes concurrently or within several months of
    one another is unknown. …
```

because the KB carries a *second* entry, `DDInter2169 Daxibotulinumtoxina`, with the same `rxcui`
(1712) and the same `rxnorm_name` ("botulinum toxin type a") but a different display stem — so
`Daxibotulinumtoxina × Botulinum toxin type A` is **not** a self-pair by `substanceKey`, is correctly
left loaded (this is the `enalapril`/`enalaprilat` shape the stem veto exists for), and its rule's
token is the other entry's own name. On `main` the two rows collapsed to one chip under the existing
partner grouping anyway, so the chip text is unchanged by this PR. That path is the stem-divergence
half of **#164**, which this PR deliberately does not touch — see below.

### Regression controls, exact

| patient (resolved by name) | question | result |
|---|---|---|
| Mary Smith `38beca4a-…` | Is it safe to give her clarithromycin? | **1** chip, Major, `× simvastatin` |
| Agnes Adams `47e57b75-…` | Is it safe to start warfarin? | **1** chip, Major, `× aspirin` |
| Joshua Johnson `9cb37bcb-…` | Can I give ibuprofen? | **2** chips — NSAID cross-reactivity contraindication AND `× lisinopril` Moderate, not collapsed |
| Mary Smith | Is it safe to give her paracetamol? | **0** chips |
| Richard Jones `a39e82e7-…` | Is it safe to give dexamethasone? | **1** chip, the identity one (#160) |
| Sarah Taylor | What are her current medications? | **6** chips, all distinct (#160) |
| George Anderson `d5c9cdad-…` | Is it safe to give esomeprazole? | **2** chips — `Omeprazole` AND `Esomeprazole` both surviving (#121/#160) |
| Margaret King `f2397b8a-…` | Is it safe to give voxelotor? | **1** chip, Major `× dexamethasone` (#115's partner-side collapse) |

Note on George: the brief's 2-chip control needs the word **esomeprazole**, which is the alias both
rows share. `omeprazole` is a whole word in only one of them, so `Is it safe to give omeprazole?`
correctly returns **1** chip. Both were probed.

Each control would have failed differently had the collapse been keyed too widely: Joshua's two chips
are two clinical facts about different partners and a subject-keyed collapse would drop one; George's
two are two substances the KB files under one `rxnorm_name`, which is exactly what the display-stem
veto refuses to merge.

### #163 — the injected slice, measured

Not visible in the REST response (only *cited* references come back), so measured two ways.

**In-process over the real 19 MB KB**, through `DdiDrugReferenceSource.parse` →
`DrugReferenceInjector.injectRecords` → `render`, on a throwaway worktree at `main` (`45c1cc4`) and at
this branch, same context (age 60, no active orders):

| question | main | this branch |
|---|---|---|
| dexamethasone | **4 records / 5254 chars** | **1 / 1403** |
| phenylephrine | **5 / 6975** | **1 / 1525** |
| hydrocortisone | 4 / 5556 | **2 / 3246** — `Hydrocortisone` + `Hydrocortisone butyrate`, two substances |
| ketorolac | 2 / 2169 | 1 / 1067 |
| lidocaine | 3 / 3673 | 1 / 1500 |
| olopatadine | 2 / 3111 | 1 / 1564 |
| clarithromycin (single row) | 1 / 1524 | **1 / 1524, unchanged** |
| simvastatin (single row) | 1 / 1357 | **1 / 1357, unchanged** |
| warfarin (single row) | 1 / 1399 | **1 / 1399, unchanged** |

The three single-row controls are byte-identical, which is what makes the others a collapse rather than
a truncation.

**Live on the standalone**, off the module's own DEBUG line (the character total is what this PR's
second commit adds):

```
Injected 0 active-order, 1 drug-reference (1403 chars) and 0 safety-finding record(s) into chart for question 'Is it safe to give dexamethasone?'
Injected 0 active-order, 1 drug-reference (1525 chars) and 0 safety-finding record(s) into chart for question 'Is it safe to give phenylephrine?'
Injected 0 active-order, 2 drug-reference (3246 chars) and 0 safety-finding record(s) into chart for question 'Is it safe to give hydrocortisone?'
Injected 0 active-order, 1 drug-reference (1500 chars) and 0 safety-finding record(s) into chart for question 'Is it safe to give lidocaine?'
Injected 0 active-order, 1 drug-reference (331 chars) and 1 safety-finding record(s) into chart for question 'Is it safe to give clarithromycin?'
```

The first four match the in-process figures exactly. Clarithromycin reads 331 rather than 1524 because
this patient IS on simvastatin, so `orderedInteractionNotes` promotes that partner and renders the tail
as one compact representative — a different context, not a different collapse.

### Effect on #164

**None, in either direction, and I checked rather than assumed.** #164's two paths are untouched:
`IDENTITY` is still awarded on entry identity in `addAllergyContraindications`, which this PR does not
modify; and `substanceKey()`'s stem component is unchanged, so `Pfizer-BioNTech Covid-19 Vaccine` and
`Tozinameran` still key apart. The `ContraindicationChips` key moved to
`DrugReference.substanceGroupKey()`, which is the same expression the ledger already computed inline.
The #152 guard drops interaction rows only, and DDInter emits no contraindications at all, so it cannot
reach the allergy path.

What this verification *did* add is a measured instance of #164's stem-divergence half in a **different
arm**: the botulinum probe above shows the interaction arm reporting a substance as interacting with
itself because a second entry publishes the same `rxnorm_name` under a different stem. Same underlying
data shape as #164's `Pfizer-BioNTech`/`Tozinameran` case, and the same design decision blocks both —
which label a chip should name when two rows the data calls one substance are two rows this module keeps
apart. Worth folding into #164's scope; not touched here.

#164 was also probed directly, to answer "more or less reachable?" with a measurement rather than an
argument. Kevin Brown (`e89247a2-111f-4dd5-974d-25bfad03690a`, resolved by name; a `Tozinameran`
allergy), `Is it safe to give the Pfizer-BioNTech COVID-19 vaccine?` on THIS build:

```
[contraindication] Pfizer-BioNTech Covid-19 Vaccine (sars-cov-2 (covid-19) vaccine, mrna spike protein)
    Pfizer-BioNTech Covid-19 Vaccine (sars-cov-2 (covid-19) vaccine, mrna spike protein) is in the same
    ATC class (J07BN) as the patient's allergy to Tozinameran (sars-cov-2 (covid-19) vaccine, mrna
    spike protein) — possible cross-reactivity
[contraindication] Tozinameran (sars-cov-2 (covid-19) vaccine, mrna spike protein)
    The patient has a recorded allergy to Tozinameran (sars-cov-2 (covid-19) vaccine, mrna spike protein).
```

#164 is still fully reachable and unchanged: both sides of the first chip render the same generic
synonym, so it reads as a substance cross-reactive with itself. Nothing in this PR moved it.

### Fixture created on the standalone (not torn down — demo data is disposable)

* Concept `Botulinum toxin type A`, Drug class, uuid `i1520000-0000-0000-0000-0000000000b1` (concept
  4259). `Lidocaine` (65), `Neomycin` (4014) and `Metoclopramide` (3992) already existed and were reused.
* Four active `DrugOrder`s on Susan Young, order numbers `I152-BOTOX`, `I152-LIDO`, `I152-NEO`,
  `I152-METO`, uuids `i1520000-0000-0000-0000-0000000000c1`..`c4`. She keeps her stock Tiotropium order
  and her Tiotropium allergy, which is why her responses also carry one contraindication chip.
* `log.level` was set to `org.openmrs.module.chartsearchai:debug` for the injected-slice measurement and
  **restored to `:info`** at the end of the run.

## Limitations

* The 71 (substance, partner) groups where the route-qualified row is more severe keep a chip whose
  prose describes that presentation. Deliberate; the alternative under-warns.
* 11 shipped families publish no unqualified row, so their chip and their injected record still carry a
  route qualifier.
* 80 of 121 multi-row substances have a partner the canonical row does not carry, so the injected
  reference record's breadth section can omit a partner the chips name. #110's finding records carry
  the actionable half.
* The 26 skipped self-pair rows include 25 whose clinical content — concurrent systemic and topical
  exposure of one substance — is real but unrenderable at this layer. Making it renderable needs the
  route vocabulary that is the data-side half of #115.
* Answer prose on the verification box is not reproducible (`cache_prompt` KV reuse), so every claim
  below is about chips, counts and record structure, never about wording.
* The class arm still names an active order after `entryForAtcCode`'s *first* matching row, which for 4
  shipped families is a route-qualified one. Same pattern, different label, not fixed here — reported
  separately.

Closes #162
Closes #163
Closes #152


dkayiwa:
## Hardening pass — fresh eyes, `fc8a2b7..HEAD`

I did not write this change. Everything below I measured myself; where I am relaying a number from the
PR body rather than reproducing it, I say so.

**Production logic is unchanged by this pass.** All four production files are byte-identical to
`fc8a2b7` once comments and whitespace are stripped — checked mechanically, per file, not by eyeballing
the diff. What changed is prose, plus two tests and one fixture.

`916 api + 61 omod = 977` tests, 0 failures, 34 skipped, BUILD SUCCESS, measured on a clean tree at
`6e3cb37`. This branch was 975 at `fc8a2b7` (measured, matches the PR body); `main` at `45c1cc4` is
953 (relayed, not re-measured).

---

### The 26 skipped rows — verdict: the widening is correct, no row is wrongly dropped

This was the item that could remove a real warning, so I enumerated all 26 off the shipped 19 MB KB
rather than sampling. **1 is the same row on both sides** (`DDInter225`, botulinum toxin type A —
#152's row). **25 pair two rows of one substance.** Every one of the 25 shares the partner's `rxcui`,
`rxnorm_name`, full ATC list *and* display stem, so they are one active moiety in two presentations:

* **18 route variants** — Lidocaine ×(ophthalmic), ×(topical); Tetracaine ×(ophthalmic), ×(topical);
  Methoxsalen, Minoxidil, Neomycin, Calcitriol, Dapsone, Diclofenac, Doxepin, Gentamicin,
  Glycopyrronium ×(topical); Timolol, Betaxolol, Carteolol ×(ophthalmic); Chloroprocaine
  (ophthalmic)×Chloroprocaine; Ipratropium ×(nasal).
* **7 formulation variants** — Bupivacaine ×(liposome) **Major**; Aminolevulinic acid ×(topical)
  **Major**; Paclitaxel ×(protein-bound); Vincristine, Amikacin ×(liposome); Cytarabine ×(liposomal);
  Heparin ×(flush).

**Why they cannot render as anything but a self-reference** — I checked each surface rather than taking
the PR's word for it. The drug-in-play chip and the screening chip name the partner through
`partnerLabel` = `Interaction.getToken()`, which is the shared `rxnorm_name`; the injected reference
record names it through `orderedInteractionNotes`, which uses the same token. The one surface that
would print the partner's own `displayLabel()` is `addQuestionPairInteractions` — and it already
declines a pair whose two entries share a name (`if (firstName.equals(secondName)) return;`). So the
parse guard generalises a decision that arm had already taken, at the boundary where it covers every
arm. The screening arm additionally could never have reported these pairs: `activeOrdersOtherThan`
removes every order the subject row resolves from, and both rows of a family share the `rxnorm_name`
alias, so any order witnessing one is attributed to both and removed.

**The stem-divergence worry does not materialise.** `Daxibotulinumtoxina` shares
`Botulinum toxin type A`'s `rxcui` (1712) and `rxnorm_name` but not its stem, so it is **not** among
the 26 — and neither is `Botulinum toxin type A × Botulinum Toxin Type B` (mechanism group 2935) nor
`Daxibotulinumtoxina × Botulinum Toxin Type B`. All three stay loaded.

The guard is by substance key, not a looser match: `isSelfPair` is `id.equals(id)` OR
`DrugReference.substanceKey(name, rxnormName)` equality — the same key the chips group on. Live WARN
count is **26**, matching the file.

The cost is real but unrenderable, and I named the worst of it in the javadoc rather than leaving it
general: **two of the 25 are rated Major** (`Bupivacaine × Bupivacaine (liposome)`,
`Aminolevulinic acid × Aminolevulinic acid (topical)`). Under-warning is the direction this module is
otherwise careful about, and those are genuine product-level warnings — dropped because a chip reading
*"Bupivacaine interacts with active order bupivacaine"* is not that warning.

---

### Two claims removed because I could not reproduce them

**1. `outranks`' "1395 of 25,847 (substance, partner) groups", also in `InteractionRouteVariantTest`.**

I replicated `substanceKey` / `namesNoRoute` / `displayStem` / `noteFor` / `noteLength` /
`clearsSeverityFloor` / `severityRank` production-exactly and ran them over the shipped KB. The
replication is validated: it reproduces **this PR's own** `Ketorolac (ophthalmic)` 495 vs plain
`Ketorolac` 265 note lengths exactly, and its 121 multi-row families / 276 entries, 110 with an
unqualified row / 11 without / 7 not-first, 142 shared `rxnorm_name` values across 332 entries,
**0 of 121** families with divergent ATC lists, and the injector residue 80 of 121 / 2627 partners /
`Olopatadine` 112 against 397 — every one of those, to the digit.

The residue figures do not reproduce. The plainest reading measures **891 of 27,573** at the shipped
`minor` floor (890 of 27,227 keying partners by token rather than by substance). I swept 24
methodology variants — partner key × severity floor × four counting rules × self-pair guard on/off —
and none produces 1395 or 25,847; the maximum any variant gives for the longer-note count is 1226.
The **71** does reproduce, but only as a per-*row* count (per-group it is 69), so the stated unit is
wrong even there.

Per the standing rule that a comment stating no number beats one stating a wrong number, I deleted the
counts and kept what the argument actually turns on: the two rows, each already pinned by a named test
(`theSurvivingChipIsNotDecidedByWhichNoteIsLonger`, `severityStillOutranksTheRoutePreference`).
**The ordering argument itself — why the route step sits between severity and note length — is
unchanged and kept in full.** I am not disputing the decision; I could not stand behind the numbers.
**The PR body carries the same two figures and needs the same correction.**

**2. `isSelfPair`'s "different botulinum SEROTYPES together, which this KB carries no second row for".**
False — see above. Corrected to the true and stronger statement: the KB files that same mechanism on
cross-row pairs the guard leaves loaded, so the self-pair costs no clinical content.

---

### A gap the tests had: the chip's subject LABEL was not pinned

`interactionSubject` is the half of #162 its own javadoc calls *"a correctness fix rather than a
de-duplication"*. Measured on a throwaway worktree: **reducing it to `subjects.get(0)` — letting
dataset order name the chip — passed the entire 915-test api suite, BUILD SUCCESS.** Every existing
fixture lists a family's unqualified row first (`ddi-route-variants.json` deliberately transposes its
two `Iron` rows, which removed the only counter-example), so "the canonical row" and "the first row in
play" were the same object everywhere and nothing could tell them apart. On the 7 shipped families
whose unqualified row is not first the chip would have named a route nobody asked about, silently — and
all 7 carry above-floor rules on that qualified first row (chloroprocaine 33, cyclosporine 7,
fluocinolone 7, iron 24, salicylic acid 194, Hib 227, Tozinameran 262), so the shape is not
hypothetical.

Closed with `ddi-canonical-subject-label.json`: a verbatim slice of `Chloroprocaine (ophthalmic)` /
`Chloroprocaine` in KB order plus `Lorazepam`, whose rule sits on the **unqualified** row only — so a
first-wins label would print that row's note under the ophthalmic row's name. Field-for-field its KB
rows, and every KB pair among those three drugs. A new fixture rather than an extension of an existing
one, so nothing already reviewed changes shape. The new case fails under both mutations that produce
the defect; before it, nothing at all caught the first.

I also pinned the injected-character total
(`theDebugLineReportsTheReferenceSliceCharacterTotalAndOnlyThat`), asserted on the real formatted DEBUG
line rather than through a helper that recomputes the number, including that it *excludes* the chart's
obs record and the safety findings. Verified load-bearing: dropping `referenceCharacters`'
resource-type filter fails that test and nothing else.

### Mutation sweep — every decision in this PR is pinned

Throwaway worktree, `git commit` path-limited throughout, tree verified clean after each:

| reverted decision | caught by |
|---|---|
| route step removed from `outranks` | 1 test |
| route step above severity | 1 test |
| `canonicalRow` → first-wins | 2 tests (1 before this pass) |
| `interactionSubject` → dataset order | **1 test (0 before this pass)** |
| `isSelfPair` → id only (#152 as originally scoped) | 4 tests |
| injector key → `getId()` | 4 tests |
| subject side per-row | 7 tests |
| contraindication key → entry identity | 8 tests (#145/#160's guard, intact after the refactor) |
| `referenceCharacters` filter dropped | **1 test (0 before this pass)** |
| `substanceRows` LinkedHashMap → HashMap | **nothing — see below** |

---

### Live re-verification, my own run

Deployed from this branch to the 3.7.1 standalone (omod built at `40ba494`; every `src/main` change
since is comment-only, verified by comparing the four files with comments and whitespace stripped, so
the capture below is behaviourally HEAD), full-JVM restart, gated on `hits>0` and a
known-positive control. Every patient resolved by **name** against the live DB first. Loaded source
from the endpoint, not a log line:

```
WARN - DdiDrugReferenceSource.parse |2026-08-06T16:53:56,657| Skipped 26 DDInter interaction row(s)
pairing a drug with itself or with another route/formulation row of the same substance
```

**#162 — Sarah Taylor, 8 orders, "Is it safe to give hydrocortisone?" → 3 interaction chips:**

```
[interaction] Hydrocortisone
    Hydrocortisone interacts with active order celecoxib — Moderate. The combined use of
    corticosteroids and nonsteroidal anti-inflammatory drugs (NSAIDs) may increase the potential for
    serious gastrointestinal (GI) toxicity, including inflammation, bleeding, ulceration, and
    perforation.
[interaction] Hydrocortisone
    Hydrocortisone interacts with active order diclofenac — Moderate. The combined use of
    corticosteroids and nonsteroidal anti-inflammatory drugs (NSAIDs) …
[interaction] Hydrocortisone
    Hydrocortisone interacts with active order ibuprofen — Moderate. The combined use of
    corticosteroids and nonsteroidal anti-inflammatory drugs (NSAIDs) …
```

The systemic prose won; `Hydrocortisone (ophthalmic)` appears nowhere, and the ophthalmic wording
("Concomitant use of ophthalmic …") appears nowhere. **Had the fix been wrong: 4 chips, the extra one
`Hydrocortisone (ophthalmic) × diclofenac` with the ophthalmic prose.** Ketorolac on the same patient:
**8 chips, every one named `Ketorolac`** (Major × celecoxib, diclofenac, ibuprofen; Moderate ×
methylprednisolone, prednisone, budesonide, dexamethasone, hydrocortisone). **Wrong ⇒ 10**, two of them
`Ketorolac (ophthalmic)`.

**#152 — Susan Young, "Is it safe to give lidocaine?" → 2 chips, both named `Lidocaine`:**

```
[interaction] Lidocaine
    Lidocaine interacts with active order metoclopramide — Major. Coadministration of local
    anesthetics with other oxidizing agents that can also induce methemoglobinemia …
[interaction] Lidocaine
    Lidocaine interacts with active order neomycin — Minor. Limited in vitro data suggest that some
    aminoglycosides may enhance the neuromuscular blocking properties of lidocaine. …
```

**Wrong ⇒ 7, three reading "… interacts with active order lidocaine".** The two positive controls
survived, so the guard did not switch lidocaine's rules off.

**A cleaner discriminating probe than the botulinum one** (the PR flags that one as
non-discriminating, correctly). Fixture I created: Betty Williams — who has no drug order at all —
given `Bupivacaine` + `Lidocaine`. *"Is it safe to give bupivacaine?"* →

```
[interaction] Bupivacaine
    Bupivacaine interacts with active order lidocaine — Major. Additive toxicities may occur when
    bupivacaine is coadministered with other local anesthetics.  The potential for increased risk of
    systemic toxicities such as methemoglobinemia and central nervous system and cardiovascular
    adverse reactions should be recognized.
```

**Exactly 1 chip. Had the fix been wrong: 2, the extra one `Bupivacaine interacts with active order
bupivacaine — Major`** — the *formulation* leg of the guard rather than the route leg, at the worst
severity the 26 rows carry, on a patient whose chip set is nothing but the pair under test and its own
positive control. Re-run to rule out the cold-start race: identical.

**The botulinum probe, and what it does show.** 4 interaction chips, two subjects —
`Daxibotulinumtoxina (botulinum toxin type a)` and `Botulinum toxin type A` — each against neomycin
(Major) *and* each against "active order botulinum toxin type a". So this response still contains two
self-reading chips, one per stem, and that is **#164's stem-divergence path, live in the interaction
arm**, not a #152 failure: the exact `DDInter225 × DDInter225` row is gone (WARN above, pinned at parse
level), and what remains is `Daxibotulinumtoxina × Botulinum toxin type A`, correctly left loaded.
**#164's reachability is unchanged by this PR** — I checked rather than assumed — but this is a measured
instance of it in an arm #164 does not currently mention, and it produces **two** such chips here rather
than the one the PR body quotes.

**#163 — injected slice, off the module's own DEBUG line, George Anderson (chart, no orders).** Every
figure in the PR body reproduced exactly:

| question | this branch, live | PR body claim |
|---|---|---|
| dexamethasone | **1 record / 1403 chars** | 1 / 1403 ✔ (`main` 4 / 5254) |
| phenylephrine | **1 / 1525** | 1 / 1525 ✔ (`main` 5 / 6975) |
| hydrocortisone | **2 / 3246** | 2 / 3246 ✔ (two genuine substances) |
| clarithromycin (single-row control) | **1 / 1524** | 1 / 1524 ✔ |
| simvastatin (control) | **1 / 1357** | 1 / 1357 ✔ |
| warfarin (control) | **1 / 1399** | 1 / 1399 ✔ |

The three single-row controls being byte-identical is what makes the others a collapse rather than a
truncation.

**Controls, all exact:** Mary + clarithromycin **1** Major × simvastatin · Agnes + warfarin **1** Major
× aspirin · Joshua + ibuprofen **2** (NSAID cross-reactivity *and* × lisinopril Moderate — did not
collapse) · Mary + paracetamol **0** (re-run: 0) · Richard Jones + dexamethasone **1** · Sarah "current
medications" **6**, all distinct · George Anderson + **esomeprazole** **2**, with `Omeprazole` *and*
`Esomeprazole` both surviving.

Fixture left on the box (demo data is disposable, no teardown): orders `H173-BUPI` / `H173-LIDO` on
Betty Williams, uuids `h1730000-…-a1`/`a2`. Susan Young's `I152-*` set was already there and was
reused, not recreated. `log.level` was set to `:debug` for the slice measurement and **restored to
`:info`**.

---

### Found, not fixed

1. **The screening arm's cross-arm suppression is text-keyed, and this PR changed one side of that
   text.** `addActiveOrderPairInteractions` stands down from a pair the drug-in-play arm already raised
   only when the two chips are byte-identical (`chipIdentity`). The drug-in-play arm now renders the
   subject as the *canonical* row and may carry a *sibling* row's winning rule; the screen still
   renders it as the order-entry row with that row's own rule. Where those differ, both chips are
   emitted, and on `main` the screen's candidate matched a per-row chip and was suppressed. Two
   triggers, either sufficient: the family's first order-entry row is route-qualified, or the group's
   winner came from a sibling via the severity step (`Sirolimus (protein-bound)` Major vs plain
   `Sirolimus` Moderate). **Status: the KB precondition is measured and abundant** — all 7 not-first
   families carry above-floor rules on their qualified first row — **but the arm-level reachability is
   reasoned from the code, not reproduced.** It additionally needs a screening-shaped question (so
   `questionDrugs` is empty) *and* an answer naming the substance (so it is in play), i.e. the
   post-answer chip pass only. What would settle it: a fixture whose multi-row family carries the
   partner rule on its *qualified* first row, driven through `validate` with a non-empty answer. Not
   fixed because the remedy changes which arm owns a pair — a documented cross-arm decision
   (`coveredByActiveOrderArm`, `alreadySaid`, `seenPairs`) — and a fix built on an unverified premise
   could itself under-report.
2. **`entryForAtcCode` / `displayLabelForAtcCode` name the class chip's *partner* by dataset order**,
   so a class chip can still read "… as active order `Dexamethasone (nasal)`". Already #174 site 1;
   noting only that it is now the last route-qualified label left inside a chip whose subject this PR
   canonicalises, which makes it more visible, not less.
3. **`isSelfPair`'s id leg is untested and looks redundant.** The shipped self-pair is caught by the
   substance leg too, so deleting the id test passes all 916 tests — but 28 of 2283 shipped rows publish
   no `rxnorm_name`, and for those the id test is the *only* guard. None carries a self-pair today
   (0 of 1), so this is a hazard under a KB refresh, not a defect. Not tested because the fixture would
   need a non-verbatim interaction row, which would cost a documented fixture invariant; the javadoc
   already states why the leg is not redundant.
4. **`substanceRows`' javadoc claimed its between-group order was load-bearing.** It is not — nothing
   iterates that map; the positional promise comes from the caller's loop over `inPlay`. Measured:
   swapping the `LinkedHashMap` for a `HashMap` passes all 916 tests. Fixed in prose (the map is kept,
   but no longer described as a guarantee, and the javadoc now names what *would* make it one).
5. **Measured and NOT a defect, recorded so it is not re-litigated:** the route step now outranks note
   length, and `noteLength` exists partly to lose against DDInter's "no mechanism description on file"
   placeholder — so in principle the route step could hand a chip the placeholder while a sibling
   carried real prose. Over the shipped KB that is **0** (substance, partner) groups at the default
   `minor` floor and **0** at the widest supported `unknown` floor.

### Load-bearing decisions I was tempted to change and did not

* **The 26-row widening itself.** It deletes rows from a clinical safety KB and this whole effort began
  with a *missing* warning. I verified it instead, and I agree with it.
* **Severity before route in `outranks`**, and the residue it accepts (a systemic subject can read
  route-qualified prose when the qualified row is strictly more severe). I agree: the alternative
  reports the milder rating for a pair the source rates worse, which is the one direction a
  non-blocking advisory must not take, and it is the same call `bestRulePerPartner` already makes on the
  partner side. Sirolimus × lapatinib is the case, and it is pinned.
* **The 11 families that keep a route qualifier** because the KB publishes no unqualified name.
  Synthesising one is #148's mistake. Correct as is.
* **The `getId()` fallback in `matchingEntries`** rather than `substanceGroupKey()`'s identity.
  Confirmed: only `DdiDrugReferenceSource` sets `substanceName`, so the curated `json` file and the
  `atc` adapter key on `getId()` exactly as before, and that is what keeps the injected record's
  `resourceId` unambiguous. Two records can never share one, on either leg.

dkayiwa:
### Phase 2 (polish) — one real reuse win, the rest measured and deferred

**Applied.** The log-capture helper I had added to `DrugReferenceTestSupport` duplicated
`org.openmrs.module.chartsearchai.LogCapture` — already public, already `AutoCloseable`, already
capturing per logger *or package*, and already called from this same package at **10 sites** via
`DrugReferenceTestSupport.REFERENCE_LOGGER`. The only reason it could not serve was `Level.INFO`
hardcoded in its constructor. So `LogCapture.on(String, Level)` was added with `on(String)` delegating
at INFO (the 10 existing sites unchanged and green), and my 55-line duplicate plus its 8 log4j imports
were deleted. That removed the third hand-rolled appender in the test tree rather than adding a fourth.
Knock-on: try-with-resources removed the `PatientChart[1]` capture box, and two values are now computed
once instead of twice so the failure message provably prints what the assertion compared.

Also documented, on `RenderedReference.withheldInteractions`: since #163 that count is over the
rendered **row's** partners, so a partner carried only by a sibling row is *absent* rather than
withheld and is not counted. `0` therefore means "names every partner of the row this record was
rendered from", not "of the substance it is named after". Left as the row's own count deliberately —
the field describes honest truncation *of this text* — but a client reading it the other way would
conclude a collapsed record is complete when it is not.

**Deferred, each with the consequence of not doing it.**

* **Caching `namesNoRoute()` for `outranks`.** Nothing breaks. A `SubjectRule` only exists after a rule
  clears the floor *and* `hasActiveDrug`, so the call count is `(rows−1) × the patient's actual partner
  orders` — order of 500 short-string evaluations per `validate`, inside a loop that already spent
  10⁴–10⁵ string scans in `hasActiveDrug` and a request that then blocks 15–90 s. A memo field would
  also reintroduce the staleness #172 warns about, for no observable gain.
* **Memoising `substanceKey` on `DrugRow` in the parser.** `isSelfPair` computes it twice per
  interaction row, so ~590,000 computations for 2283 distinct rows — a real ~250× reduction for ~3
  lines. Not taken because it is production code on a lazy once-per-JVM path that already `readTree`s
  19 MB, and touching it would void this pass's "production logic byte-identical to `fc8a2b7`" property
  and re-open live re-verification for a gain nobody can measure. Worth doing by whoever is next inside
  that parser.
* **`log.isDebugEnabled()` around `referenceCharacters`.** One extra pass over the mapping list per
  `injectRecords` with DEBUG off — tens of microseconds against a 15–90 s call. I checked both existing
  guards in the module and neither is an argument-cost precedent (one is half a compound condition, the
  other avoids invoking a third-party contributor), so adding it would invent an idiom rather than
  follow one.
* **Extracting `textsOf(List<RecordMapping>)`** to collapse three copies of filter-and-sum (this test,
  `DrugSafetyQuestionPairInteractionTest`, the local `everything` loop). Without it a future change to
  what counts as reference text must be made in three places or silently diverge — but the three copies
  differ in null-guarding, so taking it means editing a settled test file, a larger blast radius than a
  hardening pass should have.
* **An `injectorWithValidator(service)` helper.** The injector-plus-validator wiring is hand-rolled at
  14 sites, and an injector missing the validator silently injects no safety-finding records, so a
  future test's precondition can pass vacuously. Not taken because a helper serving one new call site
  while 13 keep the long form is worse than either end state — this is its own cleanup.

**Also re-checked and retired rather than reported:** whether `getAll()` can hot-reload between
`validate`'s two reads. It cannot — `ensureLoaded` assigns `entries` once behind a double-check and
`setEntries` is a package-private test seam with no `src/main` caller — and the slice adds no new
cross-read coupling either way.

**Stopping conditions — and where I did NOT meet one.**

*Phase 1.* I ran 7 passes and did **not** literally meet "two consecutive cosmetic-only passes": pass 4
found nothing substantive, pass 5 (an independent review) then found four things, pass 6 found one more
— an imprecision in prose **pass 2 had itself written** — and only pass 7 came back empty. So the
honest statement is the other one the project's bounded-fact-check rule provides: **code and tests were
stable throughout.** Production logic never changed in any pass — verified mechanically by comparing all
four files with comments and whitespace stripped, not by reading the diff — and the last new assertion
landed in pass 2. Passes 3 to 6 changed only prose. Pass 6 finding a falsehood in pass 2's own
correction is the rule's explicit signal to stop writing prose rather than write more, so I stopped:
every prose item was resolved by deleting or narrowing the claim, never by re-wording it, and the two
counts I could not reproduce were removed outright rather than replaced with mine.

*Phase 2: all remaining findings are below the noise floor.* Every item above is deferred with a
measured consequence, and both review agents independently labelled their remainders sub-noise-floor.

dkayiwa:
## Settled: the cross-arm double chip the hardening pass predicted **does** occur, and is fixed here

The hardening pass reported a possible regression it could reason about but not reproduce — that
`addActiveOrderPairInteractions`' cross-arm suppression is **text-keyed** (`chipIdentity`) while this PR
changed one side of that text. It reproduces. Two commits, `0eaf944` + `f1b02d9`.

### The verdict, measured through the real parser and the real `validate`

New verbatim-KB-slice fixture `ddi-crossarm-canonical-duplicate.json` — Chloroprocaine (1 of the 7 shipped
multi-row families whose route-unspecified row is not the family's first) with the partner rule on its
**qualified first row**, plus Lidocaine, both as active orders, under a screening question whose answer
names the substance and cites nothing:

| build | chips for the (chloroprocaine, lidocaine) pair |
|---|---|
| `main` 45c1cc4 | **1** — plus 2 bogus self-pair chips this PR removes |
| this branch d6aaa56 | **2** — `Chloroprocaine interacts with active order lidocaine — Moderate. …` **and** `Chloroprocaine (ophthalmic) interacts with active order lidocaine — Moderate. …`, byte-identical mechanism prose under two subject labels |
| with the fix | **1** |

So for that pair it is a strict 1 → 2 regression; the total only fell because #152's parse guard removed
two self-pair chips at the same time. Since #110 each duplicate is also a citable safety-finding record.

### The fix: an identity ledger, not a text patch

`InteractionPairs` — the interaction counterpart of `ContraindicationChips`, keyed on
`{subject substanceGroupKey(), partner}` and **unordered**. The partner side is the very key
`bestRulePerPartner` already groups on, now resolved once and carried on `SubjectRule` instead of
`activeOrderEntryFor` being scanned three times. `chipIdentity`/`alreadySaid` are **removed**, not
supplemented — two dedup mechanisms coexisting, one of them text-keyed, is the thing that broke.

Keying on the pair rather than the text also closes two things the text key could never see:

1. **The other direction.** "B interacts with active order A" is not the string "A interacts with active
   order B", so whether the repeat was suppressed depended on which side the screen reached the pair from
   — i.e. on the order `findForActiveOrders` walks the chart's own order names in. Latent on `main` too.
2. **#88's fold**, which needed a second prefix-matching test on top of the equality one. Pair identity
   needs neither.

Mutation-checked, each on a throwaway tree, whole api suite each time:

| mutation | failures |
|---|---|
| suppression disabled | **4** — `screeningNeverRepeatsAChipTheDrugInPlayArmAlreadyRaised` and `DuplicateInteractionChipTest.aFoldedChipStillSuppressesTheScreeningArmsPlainChipForTheSamePair` (both pre-existing, i.e. the ledger covers everything the text key pinned) + the 2 new cases |
| ledger key made directional | 1 — the new reverse-direction case only |
| ledger keyed on the ROW, not the substance | 2 — both new cases |

The screening arm's per-row subject **label** is #174 site 3 and is deliberately untouched; the
suppression no longer depends on it. **#174 site 3 needs a correction**: it says "wrong subject label, **no
duplicate** … Label-only defect". That was true before #162 and is not true now — measured above.

### Also: `isSelfPair`'s id leg is no longer untested

28 of the shipped KB's 2283 rows publish no `rxnorm_name`, so `substanceKey` is null and id equality is the
only guard; 0 of them carry a self-pair today, so deleting the leg passed all 916 tests.
`ddi-self-pair-unnamed-substance.json` (two verbatim rxnorm-less KB rows, one authored self-pair row —
the fixture's `metadata.note` says which) plus a verbatim genuine pair as the control. Deleting the id leg
now fails one test; making two null substance keys compare equal fails the control.

### Numbers (measured here, full reactor from the repo root)

* `d6aaa56` — 916 api + 61 omod = **977**, 0 failures, 34 skipped
* `f1b02d9` — 921 api + 61 omod = **982**, 0 failures, 34 skipped (+5 tests)

### Live, on the 3.7.1 standalone (`sourceFormat=ddinter`, 2283 entries, 26 self-pairs skipped)

Every patient resolved by name against the live DB. New fixture: `H173B-SALI` (Salicylic acid, concept 34)
+ `H173B-MTX` (Methotrexate, concept 3988) on **Helen Roberts**, who had no orders and is in no other
control set. The KB lists `Salicylic acid (sodium)` first and the canonical `Salicylic acid` third, both
carrying a Major methotrexate rule with *different* notes (groups 812 and 7448).

* "Is it safe to give salicylic acid?" → 1 chip, `Salicylic acid interacts with active order methotrexate
  — Major. Salicylates may interfere with the renal elimination of methotrexate and may displace it from
  binding sites.` — the canonical label and group 7448's **short** note. First-row-wins would say
  `Salicylic acid (sodium)`; `outranks` without #162's route step would carry group 812's longer note.
  **This is #162's label fix live on a not-first family, which this PR's own live pass could not show** —
  hydrocortisone's unqualified row *is* its family's first.
* Controls exact: Mary+clarithromycin 1 Major ×simvastatin · Agnes+warfarin 1 Major ×aspirin ·
  Joshua+ibuprofen 2 (NSAID contra + lisinopril Moderate, not collapsed) · Mary+paracetamol 0 (×2) ·
  Sarah+hydrocortisone 3 interaction chips all `Hydrocortisone`, systemic prose, no `(ophthalmic)` ·
  Sarah+ketorolac 8 all `Ketorolac` · Betty+bupivacaine exactly 1 (`×lidocaine` Major, no self-pair) (×2) ·
  George+esomeprazole 2 with `Omeprazole` **and** `Esomeprazole`.
* Screening probes across 5 patients: Barbara 1 interaction chip (aspirin×ibuprofen Major) ×2 · Susan 3
  (3-row Lidocaine family = one chip per pair, no self-pair) · Melissa 0 (one order, no pair) · Sarah 10
  (the cap) — no duplicate anywhere.
* **Honest limit:** in all 14 screening probes the answer cited the injected safety-finding record, which
  names the substance, so echo scoping (#105) removed the answer-named drug from the drugs-in-play set and
  the drug-in-play arm never ran beside the screen. The cross-arm shape therefore did **not** occur live
  in this session — those probes show 1 chip whether or not the fix is present, and I am not claiming them
  as evidence for it. The evidence for the fix is the fixture; live shows no regression. This reachability
  limit applies equally to the suppression that was already there.

dkayiwa:
**Merging with two red checks, both of which are the GitHub Actions outage and neither of which ran any code.** Recording the basis, since merging past a red required check needs one.

The two failures, from the job APIs:
- `build / Build with Java 11` — failed in **`Set up job`**: `Failed to resolve action download info` / `Service Unavailable`. The build never started.
- `eval harness selftests` — `The job was not acquired by Runner of type hosted even after multiple attempts`. Never scheduled.

Re-running was attempted and the replacement jobs hit the same thing.

Why Java 11's missing result costs nothing here:
- **`0eaf944` — the last commit that changed code — passed Java 11, 17 and 21.**
- `f1b02d9` (this head) versus `0eaf944` is **comment-only**: 5 insertions, 2 deletions, and stripping comment lines from the diff leaves nothing.
- **`maven-javadoc-plugin` is bound in no pom**, so javadoc never runs in the lifecycle and its content cannot fail a build on any JDK.
- Java **17** and **21** both pass at this exact head.

So Java 11 at `f1b02d9` is provably equivalent to Java 11 at a SHA that already passed it.

Independent local verification at `f1b02d9`, on a clean detached worktree, full reactor: **921 api + 61 omod = 982, 0 failures, 34 skipped**, BUILD SUCCESS.

Test-file integrity across the whole branch: **zero removed lines contain an assertion.** The three pre-existing test files touched are `DrugSafetyInteractionScreeningTest` (+77/-0), `SelfInteractionTest` (+66/-2, the -2 being a class-javadoc sentence the change falsified) and `ActiveOrderContraindicationTest` (+5/-3, comment-only — the old comment described the `chipIdentity` text key this PR removes, and its three assertions are untouched).

Issue #165, opened 2026-08-05
CountingQueryStoreStub implements the querystore interface directly, so any upstream API addition breaks chartsearchai's test compile


dkayiwa:
## What

`CountingQueryStoreStub` implements `QueryStoreService` directly, so **adding any method to that interface upstream stops chartsearchai's tests from compiling** — not failing, not skipping: the test sources will not build at all.

This is not theoretical. It happened on the development box today: a local `mvn install` of openmrs-module-querystore introduced

```java
ContextSlice getContextSlice(String, String, ContextSliceRequest)
```

and chartsearchai `main` (`dd651ef`) immediately stopped test-compiling locally. CI stayed green because CI resolves the **remote** snapshot, which predates the change — so the breakage is invisible in CI and appears only for whoever has built querystore locally.

## Why it is worth addressing

- The failure mode is maximally confusing: `main` is green in CI and broken on your machine, with an error that names a chartsearchai test class rather than the dependency that changed.
- It cost real time in this work — an agent had to detect it, set the local artifacts aside to measure anything, and restore them afterwards.
- Retrieval is owned by querystore (#51) and its API is expected to keep moving, so this will recur every time it gains a method.

## Shape of the fix

Make the stub tolerant of interface growth rather than exactly conformant. Options, roughly in order of preference:

1. Have the stub extend an abstract adapter that provides no-op defaults, so new methods do not break it.
2. Build it as a dynamic proxy over the interface, counting the calls it cares about and returning defaults otherwise.
3. Keep it conformant but pin the querystore version explicitly, so a local install cannot silently change what tests compile against.

Adjacent to #96, which is the other way the querystore dependency's scope makes chartsearchai's tests fragile.

## Note on the current box state

I restored the resolved `querystore-api-1.0.0-SNAPSHOT.jar` from the Aug-3 remote snapshot so local builds match CI again, and kept the locally-installed jar aside rather than deleting it. Anyone who needs the new API locally will reintroduce the breakage until this is fixed.

## Provenance

The breakage was found and diagnosed by an agent during #160's hardening; I independently confirmed the new abstract method in the installed jar with `javap` and that it is absent from the remote snapshot.

dkayiwa:
**Correction to this issue: the option I listed first is structurally incapable of fixing it.**

[...]

Issue #161, opened 2026-08-05
sharedClass names the wrong pharmacological class: an inhaled corticosteroid is reported as a nasal preparation, a systemic one as anti-acne


dkayiwa:
## What

`DrugSafetyValidator.sharedClass` returns the **first** shared level-4 ATC subgroup in the *allergen's* ATC ordering, not the subgroup that actually explains the cross-reactivity. Because a corticosteroid carries ATC codes for every route it is marketed in, the first shared code is frequently a topical or local-route class that has nothing to do with the clinical concern.

Three examples measured live on the standalone, all against a **dexamethasone** allergy, all of which should implicate **`H02AB`** (systemic corticosteroids):

| active order | reported as | what that class actually is |
|---|---|---|
| Solu-Medrol (methylprednisolone) | `(D10AA)` | anti-acne preparations |
| Pulmicort (inhaled budesonide) | `(R01AD)` | nasal preparations |
| hydrocortisone | `(A01AC)` | corticosteroids for local **oral** treatment |

## Why it matters

The chip is telling a clinician *why* two drugs are related, and the reason it gives is wrong. "Your patient's systemic methylprednisolone is in the same class as their dexamethasone allergy — anti-acne preparations" is not a defensible statement, and a clinician who checks it will conclude the feature does not know what it is talking about. The underlying relationship is real (both are corticosteroids), so the finding is correct while its stated justification is not — which is the harder failure to notice.

## Shape of the fix

Choose the shared subgroup by clinical relevance rather than by position in an arbitrary list. Options worth weighing: prefer the subgroup matching the *route of the active order* (a systemic order should implicate a systemic class), or rank candidate subgroups and prefer the systemic one, or report the shared level-3 group where all these collapse to one honest answer.

Issue #145 lists this as its option 3 and explicitly says not to bundle it with the route-variant collapse; PR #160 correctly did not.

## Related

#155 (a chip rendering a raw ATC code where a drug name belongs) and #151 (ATC-vs-name resolution asymmetry) are the other two places where ATC codes are trusted further than the mapping supports.

## Provenance

The three chip readings above were measured live during the #145 fix work and reported to me; I have not re-run them. The `H02AB` attribution is that agent's analysis.

dkayiwa:
**Correction to this issue's own table, measured while fixing it.**

[...]

PR #177, opened 2026-08-06
docs: documentation-truth sweep of README, ADR and ONBOARDING (#159, #169)

### LINKED ISSUE #159: docs/adr.md's REST endpoint list is stale: /prewarmstatus and /drugreferencestatus are missing
## What

`docs/adr.md`'s REST-endpoint section does not list `/prewarmstatus` or `/drugreferencestatus` (the latter added by #154 for issue #149).

## Why bother

The ADR is where the module's HTTP surface is written down, and #154 chose the status-endpoint shape *by following the `/prewarmstatus` precedent* — a precedent that the ADR does not record. A design document that omits two of the endpoints it is meant to describe stops being usable for that kind of decision.

Small, but the repo treats false or absent statements in load-bearing prose as real defects, and this is load-bearing prose.

## Provenance

Found during the #154 hardening pass. I have not opened `docs/adr.md` to confirm the omission myself.

### LINKED ISSUE #169: Patient Betty Williams not found as instructed in the Readme file
### Problem
In the `README.md` file, in the  section for **Try it on the demo server**,  step 2 says

> Click the magnifying-glass icon in the top header and search for Betty Williams — she is the reference patient with the most data on the demo (medications, vitals, conditions), so the AI has something to ground its answers in. Open her chart from the dropdown.

But while testing the demo server, searching for **Betty Williams** returns no results, as seen in the screenshot attached

### Fix
Replace **Betty Williams** with another patient with the most data on the demo, so the AI has enough data to base its answers on.

Image

dkayiwa:
Converts the trickle of incidental stale-doc findings into one pass with a denominator. Scope was exactly three files: `README.md`, `docs/adr.md`, `ONBOARDING.md`. `CLAUDE.md` untouched. No javadoc under `api/src/main` or `omod/src/main` touched, because #173 is open over four of those files. No test touched. No production behaviour changed.

Closes #159
Closes #169

## The audit denominator

Every factual, checkable claim — a default, a path, a class or method name, a count, a command, an endpoint, a privilege, a patient name, a behaviour statement — enumerated and then checked against the code, `config.xml`, the poms, or the live demo.

| File | Claims audited | Found false |
|---|---|---|
| `README.md` | 168 | 24 |
| `docs/adr.md` | 268 | 131 |
| `ONBOARDING.md` | 61 | 9 |
| **Total** | **497** | **164** |

A further 41 claims are **unverifiable from this repo** and were left alone, not counted as either: external facts (model licences, HuggingFace URLs, benchmark scores, hardware figures) and `docs/adr.md` Decision 9's per-record-type field lists, which describe querystore's serializers and cannot be checked here.

The single largest cause, 105 of the 131 in `docs/adr.md`: **Decisions 3–21 describe the in-process retrieval stack that #51 removed, in the present tense, with no marker.** Decision 10 still called it "the current architecture"; Decision 3 still said "CHOSEN". Nine `chartsearchai.embedding.*` / `chartsearchai.retrieval.pipeline` global properties, the `chartsearchai_embedding` table, `ClinicalTextSerializer` and its seven subclasses, `EmbeddingIndexer`, `EmbeddingIndexTask`, `HybridRetriever`, `QueryClassifier`, `PatientRecordLoader`, `capPerConcept`, `groupByConcept`, `filterEsResults`, three z-score constants — all named as live code. None exists.

## What I deleted rather than corrected, and why

The house rule applied throughout: a corrected number can go stale again; a deleted claim cannot.

**Deleted outright**
- README's two long caveat paragraphs built on `chartsearchai.querystore.enabled`. That is not a global property this module has — zero hits in `api/src`, `omod/src` and `config.xml` — yet the README used it as a live toggle six times. A GP that does not exist has no correct value, so there was nothing to fix.
- README's claim that the demo "calls a remote LLM" and that its first-query latency reflects a remote cold start. Measured: the demo has `chartsearchai.llm.engine=local`. Replaced with a statement that does not assert the demo's configuration at all, because that is the kind of fact that drifts silently.
- ONBOARDING's "Run it locally" standalone path and its `MODELS/` directory. Neither exists, and neither is tracked in git.
- ONBOARDING's duplicated API-surface list, which had drifted from `CLAUDE.md`'s authoritative one. Cross-referenced instead — a second copy is a second thing to go stale.
- ADR's planned-future-work item on incremental indexing of embeddings that no longer exist.
- ADR's `--cache-reuse 256`, `topK=30 stays the default`, and the six removed tuning parameters in the documented cache key.

**Stated and pinned to code instead** (the brief's other branch, for facts that are load-bearing and stable)
- `chartsearchai.querystore.topK` = `12`, pointing at `ChartSearchAiConstants.DEFAULT_QUERYSTORE_TOP_K`, which carries the measurement.
- `chartsearchai.chartMode` = `queryScoped`, pointing at ADR Decision 28. It was **undocumented in both README and ONBOARDING** despite deciding whether the LLM sees a slice or the whole chart, which is what made a dozen surrounding claims false.
- Warmup's real no-op condition, pointing at `LlmInferenceService.shouldRunWarmup`.

**Marked superseded rather than rewritten.** For the eleven ADR decisions describing removed machinery I did *not* rewrite the bodies. An ADR exists to record *why*, and rewriting that much superseded rationale would destroy the record while generating a large volume of new prose to be newly wrong in. Each carries a one-line status marker naming what was removed, pointing at what runs today, and saying to read the body as history — the convention Decision 18 already used in this file. That marker states a completed fact, so it cannot rot. Decisions 4, 10 and 16 are only partly superseded and say which half stands; Decision 10's marker additionally corrects the two false claims that are load-bearing and still repeated elsewhere.

## #169: the diagnosis

Not an unlucky patient. **The README described the standalone's patient set as if it were the demo's.**

Resolved by name against the live demo, never from a uuid in a document:

```
$ curl -s -H "Authorization: Basic $A" --get --data-urlencode "q=Betty" \
      "https://chartsearchai.openmrs.org/openmrs/ws/rest/v1/patient"
{"results":[{"uuid":"dda99123-...","display":"9290KT-2 - Betty Njagi Kittur"}]}

$ ... --data-urlencode "q=Mary Smith"      -> {"results":[]}
$ ... --data-urlencode "q=Agnes Adams"     -> {"results":[]}
$ ... --data-urlencode "q=Joshua Johnson"  -> {"results":[]}
```

There is no Betty Williams on the demo, and none of the O3 reference patients are there either. The demo carries a different demo dataset — `Betty Njagi Kittur`, `Corretty Williams Meli`, `Emacculate Williams Nyamumba`. Betty Williams is real, but in the **standalone**'s RefApp data (`a7090f70-99b7-4fd9-b60d-f8e0cdee07f6`, confirmed by name lookup there). So the README's "the standard O3 reference patient set" was itself the false claim, and the missing patient was its symptom.

**I made the instruction patient-independent rather than swapping in another name.** Two reasons. First, the demo's population has demonstrably already changed out from under this document once, so any name I hardcode fails the same way; naming a patient is the defect, not the choice of patient. Second, the walkthrough does not actually need a specific patient — it needs a chart with records in it, which the reader can see in the search results. The README now says so plainly, including that the demo carries whatever it was last seeded with. On what the walkthrough needs: the demo steps ask only the medications / allergies / vitals questions, so a patient with an active order and an interaction partner is not required — that shape belongs to the drug-safety feature, which the demo walkthrough does not exercise.

Worth recording for whoever reads #169 next: the premise that Betty Williams has 0 active orders is **no longer true on the standalone**. She currently has two (`Bupivacaine`, `Lidocaine`, activated 2026-08-03, uuids `h1730000-...a1/a2`) plus an expired Simvastatin. Those look like a previous agent's fixtures. It reinforces the conclusion rather than changing it: that data is mutable developer test-bed data, and the demo's is different again.

## The walkthrough, executed

Run against the live demo exactly as the README now words it — search by surname, pick one whose chart has records, ask the README's own first example question. Verbatim:

```
# step 2: search a patient BY NAME (surname 'Williams', as the README suggests)
$ curl -s -H "Authorization: Basic $A" --get --data-urlencode "q=Williams" "$D/patient"
{
    "results": [
        { "uuid": "dd749903-1691-11df-97a5-7038c432aabf", "display": "2064KP-7 - Corretty Williams Meli" },
        { "uuid": "dd93959d-1691-11df-97a5-7038c432aabf", "display": "6365KP-4 - Bilar Williams Kobulo" },
        { "uuid": "dd976c1a-1691-11df-97a5-7038c432aabf", "display": "6886CH-5 - Emacculate Williams Nyamumba" }
    ]
}

# step 2: pick one whose chart has records. Emacculate Williams Nyamumba: 24 active drug orders.
# step 4: ask the README's own example question.
# start 2026-08-06T16:26:37Z
{"questionId":"557",
 "answer":"The patient has been prescribed or ordered the following medications:\n
   1. Lamivudine [1], [33], [36], [70], [172]\n
   2. Stavudine [2], [37], [39], [105]\n
   3. Trimethoprim and sulfamethoxazole [3], [34], [103], [137], [169]\n
   4. Nevirapine [4], [35], [38], [106], [171]",
 "references":[ 23 entries, all "resourceType":"drug_order", "group":"chart",
                "grounded":null, "source":null, "withheldInteractions":0 ],
 "safetyWarnings":[],
 "disclaimer":"This response is AI-generated and may not be accurate. ..."}
# HTTP 200, 104.476489s
# end 2026-08-06T16:28:21Z
```

Four medications, each cited to real `drug_order` records, references grouped `chart`-first, `grounded: null` because grounding is off on that deployment. The earlier run of the same walkthrough on a different demo patient (`Betty Njagi Kittur`) also returned HTTP 200 in 27.7 s, so the instruction is not load-bearing on one chart.

`/drugreferencestatus` is **not** available on the demo (`Unknown resource: v1/chartsearchai`), so that deployment predates #154; its GPs are readable and show `drugReference.enabled=true`. On the standalone the endpoint confirms the expected state rather than inferring it from logs:

```
$ curl -s -H "Authorization: Basic $A" http://localhost:8081/openmrs/ws/rest/v1/chartsearchai/drugreferencestatus
```

## #101

**Still true when I started, and this PR fixes it.** README documented `chartsearchai.querystore.topK` as `30`; `config.xml` and `ChartSearchAiConstants.DEFAULT_QUERYSTORE_TOP_K` both say `12`. The stale description was fixed too — in the default `queryScoped` mode `topK` sizes the slice the LLM actually sees, it is not a candidate set the LLM then filters. The same `30` was also asserted in `docs/adr.md` Decision 28 as "stays the default" and "Ship at `querystore.topK=30`", which #101 does not mention; both are corrected there and pointed at the constant. Not closing it myself.

## #159

Both halves. The endpoint list now has all eight endpoints with their privileges (`/prewarm`, `/prewarmstatus`, `/drugreferencestatus` were missing), and the stream section's "three event types" is now the seven actually emitted, with the note that `grounded` arrives *after* `done`. The three unrecorded decisions are added as **Decision 30** (#145/#160 — the chip ledger; why substance-name-plus-stem and not `rxnorm_name` equality; why a ledger and not a filter over finished chips), **Decision 31** (#161/#166 — substance over locally-applied ATC class, the two prefix lists, why route and level-3 were both rejected) and **Decision 32** (#149/#154 — the load-status endpoint and why the loaded source must be observable *after* a lazy load). Reason-focused; measurements deliberately left in the javadoc where they already live.

A fourth is pending: #173's "one substance, one row". It is recorded as **pending and unmerged**, not written up as though it had landed.

## Verification

`mvn -o clean install`, measured myself on this branch, and on `main` before touching anything:

| | api | omod | total |
|---|---|---|---|
| `main` (45c1cc4) | 892 | 61 | **953** |
| this branch | 892 | 61 | **953** |

0 failures, 34 skipped, `BUILD SUCCESS`, both. A docs-only change, so a no-op for tests as expected. Every internal link and heading anchor across all three files was checked to resolve.

No standalone deploy: nothing in `src/main` changed. The demo run above exists because #169's whole point is that a documented walkthrough had never been executed.

## Reported, not fixed

Out of scope for a docs PR, and none of it touched here.

**Code defects**
1. `ChartSearchAiRestController:218` and `:460` — the audit log's `searchMode` is `preFilterEnabled ? "pre-filter" : "full-chart"`, so it never records `queryScoped`. On a default install every row is logged `full-chart` while the prompt carried a slice, making the audit trail wrong about which context produced an answer.
2. `pom.xml:35` sets `surefire.excludedGroups=eval`, but no `@Tag` annotation exists anywhere in `api/src/test` or `omod/src/test`, so the exclusion filters nothing and the eval suites run in every build. Either the tag was never applied or it was removed; the property reads as an active gate that is inert.
3. `AbsentDataEvalTest` and `DrugSafetyEvalTest` never call `EvalReporter`, so neither appears in `eval-results.csv`. Combined with the prompt-injection suite skipping by default, only the citation suite lands in the metrics artifact.
4. Three duplicated method pairs with identical names and signatures, which the "never reimplement" rule exists to prevent: `normalizeSlashCitations` (`LlmAnswerExtractor:185,210` vs `LlmProvider:809,813`), `stripQueryStopwords` (`QueryPreprocessor:216` vs `LlmInferenceService:546`), `extractRecencyCap` (`QueryPreprocessor:142` vs `LlmInferenceService:542`).
5. `TestDatasetHelper` is package-private in `...api.impl`, so the tests under `...reference` cannot reach it and use `DrugReferenceTestSupport` instead — the API-surface rule naming it is unreachable for roughly forty test classes.
6. Three orphaned eval datasets with no reader: `enriched-retrieval-eval.json`, `medcpt-retrieval-eval.json`, `retrieval-eval-dataset.json` (~185 KB), left over from the removed retrieval pipeline.

**False claims in production javadoc / `config.xml`** — queued for a pass after #173 lands, per the brief. Not edited.
- `omod/src/main/resources/config.xml:204` — `chartsearchai.progressiveReasoning.enabled`'s description says the preview streams "to the `thinking` channel". The controller uses the 7-arg `searchStreaming`, so it goes to the dedicated `preliminary` channel. Only the 6-arg overloads alias it onto the reasoning consumer.
- `api/src/main/java/.../api/ChartSearchService.java:21` — "When `chartsearchai.embedding.preFilter` is `true` (default)". It defaults to `false`.
- `api/src/main/java/.../api/impl/LlmInferenceService.java:200-217` — the `shouldRunWarmup(boolean)` javadoc is correct that warmup stays viable under `preFilter=true`, and is what proves README/ADR wrong; flagging it only because the surrounding prose still frames `fullChart` as the ordinary case.

**Environment** — no sign of the #165 `~/.m2` problem: `mvn -o clean install` compiled and ran clean twice on the Aug-3 snapshot.

## What I could not verify
- Decision 9's seven per-record-type field lists, and the `Included fields` tables, describe **querystore's** serializers. Not checkable from this repo; left in place with a marker saying so rather than guessed at.
- Core API member claims (`Obs.referenceRange`, `Program.getDescription`, `AllergyService`, `ValueComplex`) — would need decompiling core artifacts.
- `QueryStoreConstants.BM25_DESCRIPTION_BOOST = 0.5` — the constant exists in the querystore jar; the value was not read.
- External facts throughout: model licences, RAM and tokens/sec figures, benchmark scores, HuggingFace URLs. Not checkable here and not counted in the denominator.
- Two internal contradictions in `docs/adr.md` Decision 9 that cannot both be true and that I could not settle from this repo: line 283 says `Obs.referenceRange` was added in OpenMRS 2.7.0, line 313 says reference range is "not exposed in OpenMRS 2.8.x". Both are given as the reason for the same omission. Left for the querystore-side pass.

## Provenance

Measured myself: the 953/892/61 test counts on both `main` and this branch; the root `mvn test` MDEP-98 failure; every demo and standalone REST call quoted above, including all patient name lookups and both walkthrough runs; `config.xml`'s 47 global properties and their defaults; the absence of `chartsearchai.querystore.enabled`; `shouldRunWarmup`'s body; the link/anchor check.

Relayed from read-only subagents and spot-checked, not independently re-derived in full: the per-file claim tallies, the endpoint/privilege/SSE inventory, and the identifier-existence verdicts. I re-verified the load-bearing ones directly — the missing classes, the missing GPs, the missing `test/` path, the omod pom phase.

One correction to my briefing: #169 was filed by **Ariho-Seth**, not the repo owner.

### DIFF
--- a/ONBOARDING.md
+++ b/ONBOARDING.md
@@ -4,42 +4,32 @@ An OpenMRS module that lets clinicians ask natural-language questions about a pa
 
 ## Get oriented fast
 
-- **Read first:** `CLAUDE.md` (project rules — TDD, root-cause-over-patch, and the **API-surface rules** below), `README.md` (setup + platform notes), `docs/adr.md` (decision log).
+- **Read first:** `CLAUDE.md` (project rules — TDD, root-cause-over-patch, and the authoritative API-surface rules), `README.md` (setup + platform notes), `docs/adr.md` (decision log).
 - **Retrieval lives in querystore, not here.** chartsearchai has no in-process embedding/Lucene/scoring pipeline. Retrieval and citation grounding both use querystore's e5-base-v2 model. Retrieval changes and retrieval-quality eval belong in `openmrs-module-querystore`.
-- **Two engines:** `chartsearchai.llm.engine` = `local` (bundled `llama-server` subprocess, default — data stays on the host) or `remote` (OpenAI-compatible API). A **local** LLM is a hard requirement for production.
+- **Two engines:** `chartsearchai.llm.engine` = `local` (bundled `llama-server` subprocess, default — data stays on the host) or `remote` (OpenAI-compatible API; see [ADR Decision 17](docs/adr.md#decision-17-remote-llm-backend-support)).
+- **The prompt carries a slice, not the whole chart.** `chartsearchai.chartMode` defaults to `queryScoped`: the question's typed scope complete, plus the querystore similarity top-K, plus demographics. `fullChart` is the other mode and it changes which machinery is live — see [ADR Decision 28](docs/adr.md#decision-28-query-scoped-slice-charts-chartmodequeryscoped).
 
 ## Build & test
 
 ```bash
 mvn install                       # full build (api + omod), produces omod/target/chartsearchai-*.omod
 mvn -pl api test                  # api unit tests
-mvn test                          # api + omod
 mvn -pl api test -Dtest=ClassName # one test class
 ```
+
+Use `mvn install` when you need the omod tests too. A root `mvn test` **cannot** run them: omod's `unpack-dependencies` execution binds to `generate-resources` and unpacks the api *jar*, which a `test`-phase reactor never produces, so the build fails there with "Artifact has not been packaged yet" (MDEP-98) after the api tests pass. Always build from the repo root with a full reactor — never `mvn -pl omod`, which resolves the api artifact from `~/.m2` and can silently test a stale one.
 
 Tests must call the **real production pipeline** with real datasets — no mocks/reimplementations of pipeline logic (see `CLAUDE.md`). Follow TDD: write the failing test first.
 
 ## Run it locally
 
-A configured standalone lives at `test/referenceapplication-standalone-3.7.0-SNAPSHOT` (RefApp 3.7, port **8081**, login `admin` / `Admin123`):
+The repo ships no standalone — download the [O3 Standalone with Chart Search AI](https://nightly.link/openmrs/openmrs-module-chartsearchai/workflows/build-standalone/main/openmrs-standalone-chartsearchai.zip) built by `build-standalone.yml`, or use `docker compose up --build`. See README's [Standalone platform notes](README.md#standalone-platform-notes) for the per-platform requirements (Java 21+, and *which* JDK on Windows).
 
-```bash
-cd test/referenceapplication-standalone-3.7.0-SNAPSHOT
-nohup java -jar openmrs-standalone.jar &     # nohup so it survives the shell
-# REST base once up: http://localhost:8081/openmrs/ws/rest/v1
-```
-
-To redeploy after a build: stop it (`pkill -9 -f openmrs-standalone.jar; pkill -9 -f mariadbd; pkill -9 -f llama-server` — the embedded MariaDB must be killed too or the next boot can't lock the DB), copy `omod/target/chartsearchai-*.omod` into `appdata/modules/`, restart. Local GGUF models live under `MODELS/`; the standalone bundles Gemma E2B/E4B.
+Extract it, then `java -jar openmrs-standalone.jar` (login `admin` / `Admin123`; the port is in the launcher's output). To redeploy your own build, drop `omod/target/chartsearchai-*.omod` into the standalone's `appdata/modules/` and restart. Stopping it needs all three processes — `pkill -9 -f openmrs-standalone.jar; pkill -9 -f mariadbd; pkill -9 -f llama-server` — because the embedded MariaDB otherwise keeps the DB locked against the next boot.
 
 ## API-surface rules (do not bypass)
 
-These are the only correct entry points — never reimplement their logic inline:
-
-- **Prefixed text:** `ChartSearchAiUtils.buildPrefixedText(resourceType, text)`
-- **Cosine similarity:** `ChartSearchAiUtils.cosineSimilarity()`
-- **Chart assembly:** `ChartBuildingStrategy.buildChart()` → `QueryStoreChartBuilder.build()` (querystore owns retrieval)
-- **Global-property reads:** `ChartSearchAiUtils.getBooleanGlobalProperty` / `getStringGlobalProperty` (fail-safe on a missing context)
-- **Test datasets / category hints:** `TestDatasetHelper.*`
+`CLAUDE.md` holds the authoritative list of methods that are the only correct entry point for their operation — prefixed text, cosine similarity, diacritic folding, the three drug-name matching shapes, chart assembly, the test datasets and category hints, and the inline-citation pattern. Read it there rather than from a copy here; a second list is a second thing to fall out of date.
 
 ## REST endpoints
 
@@ -48,7 +38,7 @@ Base path: `/ws/rest/v1/chartsearchai`. Every endpoint gates on a privilege up f
 | Method | Path | Privilege | Purpose |
 |---|---|---|---|
 | POST | `/search` | AI Query Patient Data | Blocking answer `{patient, question}` → answer + citations |
-| POST | `/search/stream` | AI Query Patient Data | Same, as Server-Sent Events. Event types: `preliminary`, `thinking`, `token`, `references`, `grounded`, `done`, `error` |
+| POST | `/search/stream` | AI Query Patient Data | Same, as Server-Sent Events. Event types, in emission order: `preliminary`, `thinking`, `token`, `references`, `done`, `grounded`, `error`. `grounded` is a *trailing* event after `done` (async grounding only), so a client must keep consuming the stream past `done` |
 | POST | `/warmup` | AI Query Patient Data | Fire-and-forget per-patient KV prewarm on chart open (202) |
 | **POST** | **`/prewarm`** | **Manage AI Prewarm** | **Bulk KV-prewarm bootstrap (202 + status)** |
 | **GET** | **`/prewarmstatus`** | **Manage AI Prewarm** | **Bulk-prewarm progress/status** |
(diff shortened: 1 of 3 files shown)


dkayiwa:
Spot-checked the load-bearing corrections against the code independently, rather than taking the tallies on trust. Three of three that I could verify directly hold:

- **`topK`**: `ChartSearchAiConstants.DEFAULT_QUERYSTORE_TOP_K = 12`, consumed by `PipelineSettings:70`. The documented `30` was wrong, so #101 was real.
- **`chartsearchai.querystore.enabled`**: absent from `api/src/main`, `omod/src/main` and the resources. A GP the module does not have cannot have a correct documented value, so deleting the caveats built on it is right.
- **#178** (`searchMode` never records `queryScoped`): confirmed at `:218` (`preFilterEnabled ? "pre-filter" : "full-chart"`) and again at `:459-460`. Two sites, both binary on `preFilterEnabled`, no `queryScoped` branch anywhere.

**One more falsehood, in a file this PR mostly left alone — and it is a self-contradiction inside a single description.** `omod/src/main/resources/config.xml:192` (the `queryStoreTopK` GP) says:

> "In **queryScoped** mode (**the default chartMode**) this sizes the query-scoped slice … in fullChart mode it only sizes the optional focus hint when `chartsearchai.embedding.preFilter` is true (and is unused when preFilter is false, so it has no effect on **the default fullChart path**)."

It names `queryScoped` as the default and then refers to "the default fullChart path", in the same paragraph. One of the two is wrong, and an operator reading the admin screen cannot tell which. That sits alongside the two `config.xml`/javadoc items already reported here (`:204`'s `progressiveReasoning` channel, and `ChartSearchService.java:21`'s "`preFilter` is `true` (default)"), and it is the same root as ADR Decisions 3–21: statements about the retrieval architecture that were true before #51 and were never revisited.

I am holding all of these for a **second pass after #173 merges**, since that PR rewrites large amounts of javadoc in the same package and editing it now guarantees a conflict. Recording them here so the follow-up has a list rather than a re-derivation.

For the record on the sweep's method: **497 claims audited, 164 false** is the number that makes this reviewable, and **105 of the 131 ADR falsehoods share one cause** — Decisions 3–21 describing the removed in-process retrieval stack in the present tense, with Decision 10 calling it "the current architecture". That is why this was worth doing as one bounded pass instead of continuing to fix stale docs one incidental discovery at a time. Marking those eleven with a one-line status rather than rewriting them is also the right call: a rewrite destroys the *why* an ADR exists to record, and generates fresh prose to be newly wrong.

Issue #96, opened 2026-07-27
57 context-sensitive tests cannot load an ApplicationContext (querystore-api is provided-scope, so onnxruntime never reaches the test classpath)


dkayiwa:
## Problem

`mvn test` reports **648 tests run, 0 failures, 57 errors**. Every error is the same, and every
one is in a Spring-context-sensitive test class:

```
java.lang.IllegalStateException: Failed to load ApplicationContext
Caused by: org.springframework.beans.factory.BeanCreationException:
  Error creating bean with name 'querystore.embedding.onnx': Lookup method resolution failed;
  nested exception is java.lang.IllegalStateException: Failed to introspect Class
  [org.openmrs.module.querystore.embedding.OnnxEmbeddingProvider]
```

Affected classes: `ChartSearchAiUtilsTest` (13), `ConceptNameUtilTest` (15),
`DrugSafetyWeightContextTest` (12), `HibernateChartSearchAiDAOTest` (11),
`StreamingContextPropagationTest` (3), `CrossReactivityGroupsContextTest` (2),
`ChartSearchAiModuleActivatorTest` (1).

## Cause

`api/pom.xml` declares `querystore-api` with `<scope>provided</scope>`. Provided-scope
dependencies are **not transitive**, so `com.microsoft.onnxruntime:onnxruntime` — which
`querystore-api` needs for `OnnxEmbeddingProvider` — never lands on the test classpath. The jar
is on the classpath and its Spring component definitions are picked up, so context startup tries
to instantiate a bean whose supertype cannot be introspected.

`onnxruntime` **is** present in the local `~/.m2`; it just isn't pulled in.

## Why it matters

It is not cosmetic: a third of the suite's classes never execute, including the DAO tests and
the streaming context-propagation test. A regression in any of them is invisible locally, and
"0 failures" reads as green when it isn't.

## Likely fix

Add `com.microsoft.onnxruntime:onnxruntime` as an explicit `test`-scope dependency in
`api/pom.xml` (matching the version `querystore-api` declares), or exclude querystore's
embedding beans from the test context. First option is the smaller change.

## Note

This predates PR #93 — the same 57 errors occur on a pristine `main` checkout, and PR #93 leaves
the count unchanged. Filing it separately so it does not get read as fallout from that PR.

dkayiwa:
## Closing: no longer reproduces at `main` — every affected class loads a context and passes

[...]

PR #180, opened 2026-08-06
test(#165): bind the querystore double to the interface at runtime, not at compile time

### LINKED ISSUE #165: CountingQueryStoreStub implements the querystore interface directly, so any upstream API addition breaks chartsearchai's test compile

dkayiwa:
## The defect

`CountingQueryStoreStub` — the shared double behind `QueryStoreChartBuilderTest` and
`QueryStoreChartBuilderScopedTest` — implemented `QueryStoreService` directly. A class that
implements an interface must override every abstract method it declares, so **every method
querystore adds to that interface stops chartsearchai's test sources from compiling.** Not
failing, not skipping: the test sources do not build.

Reproduced on `main` (45c1cc4) against the locally-built querystore jar that carries
`ContextSlice getContextSlice(String, String, ContextSliceRequest)` — one error, and it is the
whole build:

```
[ERROR] .../api/src/test/java/org/openmrs/module/chartsearchai/api/impl/CountingQueryStoreStub.java:[26,7]
        org.openmrs.module.chartsearchai.api.impl.CountingQueryStoreStub is not abstract and does not
        override abstract method getContextSlice(java.lang.String,java.lang.String,
        org.openmrs.module.querystore.model.ContextSliceRequest)
        in org.openmrs.module.querystore.api.QueryStoreService
[INFO] 1 error
[INFO] Chart Search AI Module - API ....................... FAILURE
[INFO] Chart Search AI Module - OMOD ...................... SKIPPED
```

`1 error` is worth noting on its own: this stub was the *only* class in the module that broke, so
fixing it fixes the failure mode rather than the first instance of it.

## Why the failure mode is invisible in CI

CI resolves the **remote** querystore snapshot, which predates the change. A developer who has run
`mvn install` in openmrs-module-querystore resolves their **local** one. So `main` is green in CI
and broken on that developer's machine, and the javac error names a *chartsearchai test class*
rather than the dependency that moved — the reader's first instinct is to look at code that did
not change. It happened three times on the development box in one day; an agent had to detect it,
swap artifacts to measure anything at all, and restore them afterwards.

Retrieval is owned by querystore (#51) and its API is expected to keep moving, so this recurs on
every addition.

## Which of the three options, and why over the other two

**Option 2, the dynamic proxy.** The double is no longer a `QueryStoreService`; it holds the
counters and fixture lists, and `asService()` hands out a `java.lang.reflect.Proxy` bound to the
interface **at runtime**. It names no method set at compile time, so interface growth cannot break
the build.

**Option 1 (abstract adapter with no-op defaults) does not work.** This is the issue's first
suggestion and it is structurally incapable of fixing the defect: a concrete subclass must
override every abstract method inherited from *anywhere*, so an abstract superclass cannot satisfy
a method the interface adds later — the identical "is not abstract and does not override abstract
method" error simply moves to the subclass. Measured with a minimal javac reproduction rather than
argued from memory:

```
Stub.java:1: error: Stub is not abstract and does not override abstract method grown(String) in Svc
public final class Stub extends Adapter {
```

An adapter only helps when upstream adds methods as `default`, and `getContextSlice` is abstract —
the case that actually occurred. That split is the deeper problem with any compile-time
implementor, adapter or not: it **fails loudly on an abstract addition and absorbs a `default` one
silently**. Both halves are visible in this one jar — `getContextSlice` (abstract) breaks the
build, `getPatientChartRead` (default) would be inherited with nobody deciding what it should do —
and the silent half is exactly how `bulkIndex` slipped past unnoticed below.

**Option 3 (pin the querystore version) was rejected**, though it is the only option that makes a
local build and CI agree, which is a real benefit this PR does not deliver:

- It **changes what CI resolves** — a build-wide decision affecting how *production* code is
  compiled, not just one test double. That is a bigger commitment than #165 needs, and it should be
  a deliberate dependency-policy choice, not a side effect of fixing a stub.
- Pinning to a **timestamped snapshot** (`1.0.0-20260803.120447-81`) is reproducible but goes stale
  silently: chartsearchai stops compiling against querystore's current API, so genuine
  incompatibilities are discovered in one batch whenever someone bumps it, rather than continuously.
  It also freezes what production code may use — adopting a querystore fix would need a pom change.
- Pinning a **range** does not solve the problem at all: a range still resolves to whatever is
  newest, including a local install.
- It leaves the stub exactly as fragile. The next deliberate bump reintroduces this break, so the
  work has to be done anyway.
- It does nothing about the silent-default hazard below.

Option 3 remains available and is orthogonal to this change; nothing here precludes it.

## The silent-default trade

A double that tolerates interface growth must not tolerate it *quietly*. The stub exists to count
calls; if a method it has never heard of returns a plausible empty value, a test can pass while
exercising nothing and nobody is told. So every method the double does not deliberately serve is
refused with an `UnsupportedOperationException` naming it.

The argument for refusing rather than defaulting is asymmetric, not aesthetic. Refusing can only
fail a test that has started depending on a new querystore call — a change a human should look at.
Returning an empty list where the real service returns data would silently weaken every downstream
assertion, which the repo rules call out as worse than a compile break.

This is not hypothetical either. The pre-fix double already answered one method silently:
`bulkIndex`, an interface `default` it inherited, which short-circuits to an empty
`BulkWriteResult`. The new reflective sweep caught it:

```
expected: <[getPatientChart, onShutdown, onStartup, searchByPatient]>
 but was: <[bulkIndex, getPatientChart, onShutdown, onStartup, searchByPatient]>
```

Two deliberate consequences:

- Dispatch matches on the resolved `Method`, not on a name, so an **overload** of a counted method
  (`searchByPatient` gaining a filter argument, say) is refused rather than answered with the wrong
  semantics.
- Growth is absorbed, but a counted method being **removed or resignatured** is not, and
  deliberately so — the call counts those tests assert would no longer mean anything. That case
  fails when the double is constructed, with a message naming querystore. Resolution is per
  instance rather than in a static initialiser for exactly this reason: a failed static initialiser
  reports the real cause once and then `NoClassDefFoundError` for every later test in the class,
  burying the message that matters.

## TDD, and what could not be tested

`CountingQueryStoreStubTest` (5 tests) was written first and measured failing. The seam
(`asService()`) was introduced returning `this` so the finished test compiled against the old
double; the test file was not touched again once the implementation changed behind it:

```
[ERROR] Tests run: 5, Failures: 2
  stub_shouldNotImplementTheQuerystoreInterfaceAtCompileTime:76
      expected: <false> but was: <true>
  asService_shouldAnswerOnlyTheMethodsItDeclares_andThrowLoudlyForEveryOther:103
      the refusal must name the method ... message was: not used by chartsearchai
      expected: <true> but was: <false>
```

**The compile-survival property itself cannot honestly be unit-tested.** Asserting it needs a
second querystore jar whose interface carries a method this source does not name, which no single
build can supply — a test that merely compiles proves nothing, and a test that stood up a
hand-written "future interface" would be asserting against a fiction. So it is demonstrated by
building against the real jar (below), and the test file says so rather than implying coverage it
does not have. What the tests *do* pin is the mechanism the demonstration rests on
(`CountingQueryStoreStub` must not be assignable to `QueryStoreService` — re-adding
`implements` reinstates the defect and nothing else in the suite would notice), that no interface
method answers silently, that counting still reaches the counters, that the simulated RPC failure
arrives unwrapped, and that `toString`/`equals`/`hashCode` still work through the proxy.

## Demonstration: the double survives a grown interface

The locally-built querystore jar (`md5 6286600c1923c68e4447becb71499b41`) declares two methods the
remote snapshot does not — abstract `getContextSlice(String, String, ContextSliceRequest)` and
default `getPatientChartRead(String)`. Swapping it in for the resolved
`querystore-api-1.0.0-SNAPSHOT.jar` and building both trees:

| tree | querystore jar | result |
|---|---|---|
| `main` @ 45c1cc4 | local (grown) | **FAILURE** — 1 error, `CountingQueryStoreStub.java:[26,7]`, OMOD skipped |
| this branch | local (grown) | **SUCCESS** — 897 api + 61 omod, 0 failures, 34 skipped |
| this branch | remote snapshot | **SUCCESS** — 897 api + 61 omod, 0 failures, 34 skipped |

The last two rows being identical is the point: the double refuses `getContextSlice` (nothing calls
it), so the grown interface changes no test's behaviour — it just no longer breaks the build.

The jar was restored afterwards to the exact bytes that were there before (`md5
e9917ab5aa895199fa395d7b932b795b`, byte-identical to
`querystore-api-1.0.0-20260803.120447-81.jar`), and `getContextSlice` verified absent from the
restored interface. Nothing was deleted.

## Build numbers

`mvn -o clean install`, full reactor, against the **remote** snapshot
`querystore-api-1.0.0-20260803.120447-81.jar` (`md5 e9917ab5aa895199fa395d7b932b795b`, the jar
resolved as `querystore-api-1.0.0-SNAPSHOT.jar`):

| | api | omod | total | failures | errors | skipped |
|---|---|---|---|---|---|---|
| `main` @ 45c1cc4 | 892 | 61 | **953** | 0 | 0 | 34 |
| this branch | 897 | 61 | **958** | 0 | 0 | 34 |

+5 is the new test class; no test was lost or newly skipped.

The counting this double exists to do is still asserted by **`QueryStoreChartBuilderTest` (25/25)**
and **`QueryStoreChartBuilderScopedTest` (25/25)** — the two classes that read
`searchByPatientCalls`, `getPatientChartCalls`, `getCallCount()`, `lastSearchTopK` and
`lastSearchQuery`. Both pass unchanged, under both jars.

**No production code is touched** — the diff is four files, all under `api/src/test`. A test-only
change needs no standalone deploy, so none was run. No assertion was modified; the only edits to
existing tests are the two constructor arguments that hand the double to the builder
(`new TestableQueryStoreChartBuilder(queryStore.asService())`).

## Limitations, honestly

- **Scoped to the double.** If querystore changes a method chartsearchai's *production* code calls,
  the local build still breaks while CI stays green — correctly, but just as invisibly. That
  residual half of the failure mode is what option 3 would have addressed. A CI job that builds
  against the latest querystore snapshot would surface it in CI instead of on a developer's box;
  that is a separate change and not filed here.
- **The double loses compile-time checking of its own dispatch.** javac no longer verifies that the
  methods it counts exist; a rename upstream becomes a runtime failure at construction instead of a
  compile error. That is the deliberate cost of the fix, mitigated by resolving the methods eagerly
  and failing with a message that names querystore.
- Unrelated to #96: this adds no dependency and the new test loads no `ApplicationContext`.

Closes #165

### DIFF
--- a/api/src/test/java/org/openmrs/module/chartsearchai/api/impl/CountingQueryStoreStub.java
+++ b/api/src/test/java/org/openmrs/module/chartsearchai/api/impl/CountingQueryStoreStub.java
@@ -9,21 +9,60 @@
  */
 package org.openmrs.module.chartsearchai.api.impl;
 
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.util.ArrayList;
 import java.util.List;
 
 import org.openmrs.module.querystore.api.QueryStoreService;
-import org.openmrs.module.querystore.backend.WriteResult;
 import org.openmrs.module.querystore.model.QueryDocument;
 
 /**
- * Shared counting {@link QueryStoreService} stub for the chart-builder tests
+ * Shared counting {@link QueryStoreService} double for the chart-builder tests
  * ({@code QueryStoreChartBuilderTest}, {@code QueryStoreChartBuilderScopedTest}) — one
- * implementation of the interface's read surface plus write-method stubs, so an interface
- * change is absorbed once and the two files cannot quietly pin different builder contracts.
- * Same consolidation rationale as {@code TestDatasetHelper}.
+ * implementation of the interface's read surface, so an interface change is absorbed once and
+ * the two files cannot quietly pin different builder contracts. Same consolidation rationale as
+ * {@code TestDatasetHelper}.
+ *
+ * <p><strong>Why a dynamic proxy and not {@code implements QueryStoreService}</strong> (#165):
+ * retrieval belongs to openmrs-module-querystore (#51) and its API is expected to keep moving.
+ * A class that implements the interface must override every abstract method it declares, so each
+ * method querystore gained stopped chartsearchai's tests <em>compiling</em> — and only for whoever
+ * had built querystore locally, since CI resolves an older remote snapshot. The error named this
+ * test class rather than the dependency that changed. A proxy is bound to the interface at
+ * runtime, so it names no method set at compile time and interface growth cannot break the build.
+ *
+ * <p>An abstract adapter with no-op defaults — the other obvious shape — does not fix this:
+ * a concrete subclass must still override every abstract method inherited from anywhere, so the
+ * identical "is not abstract and does not override abstract method" error just moves to the
+ * subclass. That shape only helps when upstream adds methods as {@code default}, which is not
+ * what happened.
+ *
+ * <p><strong>Tolerating growth must not mean answering silently.</strong> Every method this
+ * double does not deliberately serve is refused with an {@link UnsupportedOperationException}
+ * naming it. A default-returning double is a worse failure than a compile break: a test would
+ * pass while exercising nothing, and nobody would be told. Note the direction of the safety —
+ * refusing loudly can only fail a test that has started depending on a new querystore call,
+ * which is a decision a human should make; returning an empty list where the real service
+ * returns data would silently weaken every downstream assertion.
+ *
+ * <p>Growth is absorbed; a counted method being <em>removed or resignatured</em> is not, and
+ * deliberately so — the counts those tests assert would no longer mean anything. That case fails
+ * when the double is constructed, with a message naming querystore, which is where the cause is.
  */
-final class CountingQueryStoreStub implements QueryStoreService {
+final class CountingQueryStoreStub {
+
+	/** Resolved per instance rather than in a static initialiser on purpose: a failed static
+	 *  initialiser reports the real cause once and then {@code NoClassDefFoundError} for every
+	 *  later test in the class, which buries exactly the message #165 wants surfaced. */
+	private final Method searchByPatientMethod =
+			interfaceMethod("searchByPatient", String.class, String.class, int.class);
+
+	private final Method getPatientChartMethod = interfaceMethod("getPatientChart", String.class);
+
+	private final Method onStartupMethod = interfaceMethod("onStartup");
+
+	private final Method onShutdownMethod = interfaceMethod("onShutdown");
 
 	int searchByPatientCalls = 0;
 
@@ -39,14 +78,45 @@ final class CountingQueryStoreStub implements QueryStoreService {
 
 	boolean throwOnSearch = false;
 
+	/** One proxy per double, so its identity and hash are stable across {@link #asService()} calls. */
+	private final QueryStoreService service = (QueryStoreService) Proxy.newProxyInstance(
+			CountingQueryStoreStub.class.getClassLoader(),
+			new Class<?>[] { QueryStoreService.class }, this::answer);
+
 	/** Aggregate counter — pre-focus-hint tests assert on total querystore calls
 	 *  without caring which method. */
 	int getCallCount() {
 		return searchByPatientCalls + getPatientChartCalls;
 	}
 
-	@Override
-	public List<QueryDocument> searchByPatient(String patientUuid, String question, int topK) {
+	/** The double as the interface the builder consumes. Counters stay readable on the stub. */
+	QueryStoreService asService() {
+		return service;
+	}
+
+	/**
+	 * Matches on the resolved {@link Method}, not on its name: an overload of a counted method
+	 * ({@code searchByPatient} gaining a filter argument, say) is a method this double has never
+	 * been taught, and name matching would answer it with the wrong semantics instead of refusing.
+	 */
+	private Object answer(Object proxy, Method method, Object[] arguments) {
+		if (searchByPatientMethod.equals(method)) {
+			return searchByPatient((String) arguments[0], (String) arguments[1],
+					((Integer) arguments[2]).intValue());
+		}
+		if (getPatientChartMethod.equals(method)) {
+			return getPatientChart((String) arguments[0]);
+		}
+		if (onStartupMethod.equals(method) || onShutdownMethod.equals(method)) {
+			return null;
+		}
+		if (Object.class.equals(method.getDeclaringClass())) {
+			return answerObjectMethod(proxy, method, arguments);
+		}
+		throw refusal(method);
+	}
+
+	private List<QueryDocument> searchByPatient(String patientUuid, String question, int topK) {
 		searchByPatientCalls++;
 		lastSearchTopK = topK;
 		lastSearchQuery = question;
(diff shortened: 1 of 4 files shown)


dkayiwa:
Independent verification before merge (mine, not relayed):

- Test-only: **4 files, all under `api/src/test`**, zero outside it.
- **Zero removed lines contain an assertion** anywhere in the diff. The only edits to existing tests are the two constructor call sites (`new TestableQueryStoreChartBuilder(queryStore)` → `(queryStore.asService())`), which is call-site adaptation, not a changed expectation.
- `mvn -o clean install` on a clean detached worktree at `39f4a3f`: **897 + 61 = 958, 0 failures, 34 skipped**, BUILD SUCCESS — reproduces the reported figure. Built against the remote snapshot, `querystore-api-1.0.0-SNAPSHOT.jar` md5 `e9917ab5aa895199fa395d7b932b795b`, which I confirmed is still what is resolved *after* the run, i.e. the swap test restored correctly.

Merging on local evidence: CI's `Prepare` and selftest jobs are still stuck in the GitHub Actions runner outage that has blocked this repository all evening (`The job was not acquired by Runner of type hosted`, `Failed to resolve action download info`), and this change cannot affect production compilation.

**Two corrections to #165, both from this work, and the first is mine.** I listed the abstract adapter as the preferred option; it is structurally incapable of fixing this, since a concrete subclass must override every abstract method it inherits from anywhere. Demonstrated with a ten-line `javac` reproduction rather than argued. The general form is the useful part: **any compile-time implementor of a moving interface has this defect, and it fails in two opposite directions** — an `abstract` addition breaks the build loudly, a `default` addition is absorbed **silently**. Both halves are in the one jar (`getContextSlice` abstract, `getPatientChartRead` default).

And the silent half was already biting: the pre-fix double was measured **answering `bulkIndex` through an inherited interface default**, so a test could exercise a querystore call the double never meant to serve and still pass. The loud compile break was masking a quiet test-weakening defect.

The proxy refuses every unserved method with an exception naming it, dispatching on the resolved `Method` rather than a name so an *overload* of a counted method is refused rather than answered with wrong semantics. That asymmetry is the right one: refusing can only fail a test that has started depending on a new querystore call, which is a human decision, whereas returning empty where the real service returns data silently weakens every downstream assertion.

Filed as **#181**: the residual half. Production code shares the same `1.0.0-SNAPSHOT` declaration, so an upstream break still shows up locally while CI stays green — which also means CI green is currently weak evidence that this module works against querystore HEAD. The fix that addresses the *invisibility* is a CI job building against the latest querystore snapshot; pinning is a complement, not a substitute.

Issue #174, opened 2026-08-06
Four more per-row sites of the one-substance-many-rows pattern: partner labels and interaction notes are still chosen by dataset order


dkayiwa:
## Context

#145/#160 fixed the contraindication chips, and #173 fixes the interaction subject, the injected records and the parser's self-pairs. Enumerating the pattern while doing that work turned up **four more sites** where a substance's N reference rows still produce N of something, or where the row chosen is whichever the dataset listed first. Filed together because they are one root cause and should be swept in one pass, not four.

The key already exists and is already justified: `DrugReference.substanceKey()` / `substanceGroupKey()` / `canonicalRow()` / `namesNoRoute()`.

## 1. `DrugSafetyValidator.entryForAtcCode` — the class chip's **partner** label is dataset order

It returns the *first* entry carrying a given ATC code, and every row of a substance publishes identical codes, so the label is whichever row the KB listed first. **Four shipped substances list a route-qualified row first *and* carry ATC codes:** `Salicylic acid (sodium)`, `Chloroprocaine (ophthalmic)`, `Tozinameran (12y+)`, `Cyclosporine (ophthalmic)`.

Consequence: a cyclosporine order mapped to `L04AD01`, asked "is it safe to give tacrolimus?", yields *"…as active order **Cyclosporine (ophthalmic)**"* for a systemic order.

Fix: fold `canonicalRow` in there. Measured against the KB; the chip string is reasoned, not probed.

## 2. `DrugReferenceInjector.orderedInteractionNotes` — one note per partner **row**

So a single injected record lists the same partner several times. Measured: **1872 of 2283 entries** affected, **19,228 surplus rows** in total; `Ozanimod` carries 49 surplus, the COVID-vaccine token appears 6×.

This is the partner-side analogue of what #173 fixed on the subject side, and it is the larger number. Note the existing javadoc names a blocker — route vocabulary — but that blocker applies to choosing a *severity* for a promoted partner, not to collapsing the breadth tail, so it reads as narrower than the problem actually is.

## 3. `DrugSafetyValidator.addActiveOrderPairInteractions` — wrong subject label, no duplicate

Its own key already collapses the duplicates, but the subject is labelled by whichever row `findForActiveOrders` returned first, so the same 7 families get a route-qualified name for a systemic order. Label-only defect.

## 4. `addOverdose` — the fourth per-row site, latent

[...]

Issue #181, opened 2026-08-06
Production code has the same local-vs-CI querystore divergence: an upstream break is invisible in CI


dkayiwa:
## What

`api/pom.xml` declares `querystore-api` at `1.0.0-SNAPSHOT`, so what chartsearchai compiles against depends on whether the developer has run `mvn install` in openmrs-module-querystore. CI resolves the **remote** snapshot; a developer who has built querystore locally resolves **theirs**.

#165 fixed the *test-double* symptom of this. The residual half is untouched: **if querystore changes a method chartsearchai's production code calls, the local build breaks and CI stays green**, with exactly the invisibility that made #165 confusing — `main` is green in CI and broken on a machine, with an error naming a chartsearchai class rather than the dependency that moved.

## Evidence this is not hypothetical

It happened three times in one day on the development box. A local querystore install added `QueryStoreService.getContextSlice(String, String, ContextSliceRequest)`; chartsearchai stopped test-compiling locally while CI stayed green throughout, because the remote snapshot predated the change. Restoring the remote snapshot fixed it, and something re-installed the local build over it twice more.

Note the direction of the risk: CI green is currently **weak evidence** that chartsearchai works against querystore HEAD. It only tells you it works against whatever snapshot the remote last published.

## Two candidate fixes, not mutually exclusive

1. **A CI job that builds against the latest querystore snapshot** (or against querystore HEAD), so an upstream incompatibility surfaces in CI at the time it is introduced rather than on whichever developer's machine hits it first. This is the one that fixes the *invisibility*, which is the actual defect.
2. **Pin the querystore version explicitly** — the option #165 listed third. Reproducible and gives local/CI parity, but it goes stale and defers incompatibilities into one big-bang bump, and it does not tell you when upstream breaks you. Better as a complement than a substitute.

## Why it matters here specifically

Retrieval is owned by querystore (#51) and its API is expected to keep moving — `getContextSlice` is presumably part of the context-slice work PR #157 consumes. So this divergence will keep producing confusing local breakages, and each one costs someone a diagnosis that CI could have done once.

Adjacent: #96 (querystore's `provided` scope keeps `onnxruntime` off the test classpath, so 57 context-sensitive tests cannot load an `ApplicationContext`) is the other way this dependency's wiring makes the build fragile.

## Provenance

[...]

PR #182, opened 2026-08-06
fix(drug-safety): stop reading an ATC code as proof of a relationship, or as a name

### LINKED ISSUE #155: A cross-reactivity chip shows a raw ATC code where a drug name belongs, on the default sourceFormat
## What

The class / cross-reactivity chip renders the active order's **ATC code** instead of its name when the loaded drug-reference dataset has no entry for that code. `DrugSafetyValidator.displayLabelForAtcCode` (≈line 2066) falls back to returning the code itself.

Measured live on the 3.7.1 standalone with the **default** `sourceFormat=json` (the 4-entry bundled curated dataset), Agnes Adams on Aspirin 81mg (ATC `N02BA01`):

```
Ibuprofen is in the same cross-reactivity group (NSAID) as active order N02BA01
 — possible additive or duplicate-class therapy
```

The same probe under `sourceFormat=ddinter` reads correctly:

```
… as active order Acetylsalicylic acid (aspirin) …
```

## Why it matters

`N02BA01` is not a drug name and a clinician has no reason to recognise it. This is clinician-facing text on the **default** configuration — the curated 4-entry dataset is what ships when nobody sets `sourceFormat` — so it is reachable out of the box, not only in an exotic setup.

The fix wants a name source that does not depend on the drug-reference dataset having an entry for the code: the active order already carries its own display name, and the chip is being built from that order.

## Related but distinct

Unrelated to #149 (that is about a load resolving to 0 entries). Adjacent to #151 (the ATC-vs-name resolution asymmetry between the injector and the safety layer) — both are consequences of ATC codes being treated as identifiers that can always be resolved back to a name, when the mapping is sparse: only 85 of 620 `Drug`-class concepts carry a `WHOATC` map.

## Provenance

Measured live during the #149 fix work and reported to me; the two chip strings above are that agent's verbatim capture. I have not re-run the probe myself.

### LINKED ISSUE #167: A cross-reactivity chip can claim duplicate therapy between unrelated drugs, from a local-treatment ATC bucket
## What

`DrugSafetyValidator.classRelationships` names the **active order's own** ATC code when reporting a shared class, so when the shared subgroup is a catch-all local-treatment bucket it can assert duplicate therapy between drugs that have nothing to do with each other.

Measured live: Agnes Adams (Aspirin 81mg), *"Is it safe to give epinephrine?"* →

```
Epinephrine is in the same ATC class (A01AD) as active order Acetylsalicylic acid (aspirin)
  — possible duplicate therapy
Epinephrine (ophthalmic) is in the same ATC class (A01AD) as active order Acetylsalicylic acid (aspirin)
  — possible duplicate therapy
Epinephrine (topical) is in the same ATC class (A01AD) as active order Acetylsalicylic acid (aspirin)
  — possible duplicate therapy
```

`A01AD` is *"other agents for local oral treatment"* — a residual bucket that both aspirin and adrenaline happen to be filed under. **Adrenaline and aspirin are not duplicate therapy**, and telling a clinician they are, on a patient where adrenaline may be urgently indicated, is the most clinically dangerous wording this feature has produced so far.

## Why it is separate from #161

#161 fixed which subgroup is chosen when a pair shares several, and it fixed the *allergy* arm. This is the **interaction/duplicate-therapy** arm, which #166 did not touch, and here the problem is not choosing badly among several classes — it is that a residual "other agents" bucket should not license a duplicate-therapy claim at all, however it was chosen.

Verified pre-existing: measured on the #166 build, in an arm that build does not modify.

## Shape of the fix

A shared ATC subgroup is not sufficient evidence of duplicate therapy. Options worth weighing: exclude residual/`other`-style subgroups from the duplicate-therapy claim; require the shared code at a level that actually implies therapeutic equivalence; or soften the claim when the only shared subgroup is a local-treatment bucket. Whatever is chosen should be argued from the KB, since #161's work showed these buckets are common — 9 dictionary concepts carry more than one `WHOATC` map (Metronidazole 5; Heparin sodium, Acetylsalicylate sodium and Acetylcysteine 3 each), out of 110 maps in total.

The same output also exhibits **#162** (three chips for one substance's route variants on the interaction side).

## Provenance

Measured live during the #161 fix work and reported to me, chip text verbatim. I have not re-run the probe.

### LINKED ISSUE #171: The duplicate-therapy arm emits one chip per shared ATC code, and picks the class by dictionary iteration order when a rule folds
## What

Two related defects in `DrugSafetyValidator.classRelationships` (≈:2173) and `addInteractionWarnings` (≈:834). The chip fold dedups only chips that correlate to an interaction rule; `classOnly` chips are appended **per matching ATC code**.

### 1. One chip per shared code, for one clinical fact

A drug concept can carry several `WHOATC` maps, so a pair can share several level-4 subgroups and produce a chip for each — identical except for the class named.

Measured inputs: the demo dictionary's `Metronidazole` concept (`79782AAAA…`) carries **five** `WHOATC` maps — `A01AB17`, `D06BX01`, `G01AF01`, `J01XD01`, `P01AB01`. Tinidazole carries `G01AF21`, `J01XD02`, `P01AB02`, so the pair shares **three** level-4 subgroups, and the KB holds **no** DDInter row correlating them — so nothing folds and the clinician gets three chips for one fact.

### 2. When a rule *does* fold, the surviving chip's class is position-dependent

The survivor names whichever code the concept dictionary happened to iterate first. That is exactly the position-dependence #161/#166 removed for the contraindication arm — the interaction arm never got the same treatment.

`DuplicateInteractionChipTest` supplies single-code orders, so neither shape is exercised.

## Distinct from the neighbours

- **#167** is about the *content* of a shared bucket (a residual class licensing a duplicate-therapy claim). This is about *how many* chips a multi-code pair produces and *which* class wins.
- **#162** is subject-side route-variant duplication. This is code-side.
- **#166** fixed the choice for the allergy/contraindication arm only.

Fixing all three together is likely right: they are the same root — ATC codes treated as identifiers with a reliable one-to-one reading.

## Provenance

Structurally confirmed from the code by the #166 hardening pass, with the `WHOATC` map counts measured against the live dictionary. The chip **count** was not measured live — that needs a Metronidazole order, which no demo patient has. Worth constructing before fixing.

dkayiwa:
## The shared root cause

An ATC code was being read as two things it is not: **proof of a pharmacological relationship**, and
**a reliable route back to a drug name**. Coverage is sparse (444 of the KB's 2283 entries carry no
ATC code at all; 85 of 620 `Drug`-class concepts in the demo dictionary carry a `WHOATC` map), a
substance is filed under several codes, and some level-4 subgroups are residual buckets that imply
no relationship whatsoever. Four reported defects fall out of that one assumption, so they are fixed
at the two places the assumption lives rather than at the four places it surfaces:

* which shared subgroup, if any, may be named — one decision, `sharedClass`, now used by both arms;
* which co-medication a class chip is about and what to call it — one resolution, `orderPartners`.

## The residual-bucket question, and how I decided it

The issue asks one question that settles both of #167's arms: **may an "other"/residual/antidote
bucket justify a clinical relationship claim at all?**

**No — but only where nothing in the bucket's ATC ancestry asserts a property of the substances.**
That qualifier is the whole of the answer, and I did not start with it. ATC files a residue in most
of its groups ("Other …", "Various …"), and the shipped KB uses **97** of them. A blanket veto is the
obvious reading of the issue, and it is wrong, because **a residue inherits whatever the group
containing it asserts**:

| residual subgroup | its parent group | does membership mean anything? |
|---|---|---|
| `R06AX` Other antihistamines for systemic use | `R06A` ANTIHISTAMINES FOR SYSTEMIC USE | yes — both really are antihistamines |
| `J01GB` Other aminoglycosides | `J01G` AMINOGLYCOSIDE ANTIBACTERIALS | yes — already pinned by `CrossReactivityClassChoiceTest` |
| `A01AD` Other agents for local oral treatment | `A01` STOMATOLOGICAL PREPARATIONS | no — only "both are put in the mouth" |
| `S01XA` Other ophthalmologicals | `S01X` OTHER OPHTHALMOLOGICALS | no — only "both go in the eye" |
| `V03AB` Antidotes | `V03A` ALL OTHER THERAPEUTIC PRODUCTS | no — only "each reverses some poisoning" |

So the claim fails exactly when the residue sits inside a group ATC defines by **administration
site** — the groups `isLocallyAppliedAtcCode` already recognises from #161 — or inside **`V03A`**,
ATC's own residue for the entire classification, whose children partition it by the accident each
product is used for rather than by what it is.

**How the list was derived, given #161's list was incomplete in a way that reproduced its own
defect.** Not by hand from the two reported cases. I fetched the WHO ATC/DDD index (2024 release),
took every one of its 933 level-4 subgroups whose *published name* begins "Other"/"Various", and
intersected that with `LOCALLY_APPLIED_ATC_GROUPS` — **27 subgroups** — plus `V03A` as a level-3
group. A script re-parses the constant *as shipped in the Java source* and re-derives the criterion
independently: they are equal, with nothing missing and nothing extra. Names were spot-checked
against the WHOCC index itself for every group that decides a shipped pair; WHOCC also confirms the
counter-example that rules out the simpler rule — `J01XD` "Imidazole derivatives" is a real chemical
subgroup (metronidazole, tinidazole, ornidazole) despite its parent `J01X` being "OTHER
ANTIBACTERIALS", so inheriting residual-ness from a parent is not sound.

**Measured KB impact**, over the 7783 pairs that share at least one level-4 subgroup:

| | pairs |
|---|---|
| lose the class claim entirely | **486** |
| keep a claim, now naming a subgroup that does classify the substances | 54 |
| untouched | 7243 |
| *(what a blanket residual veto would have taken instead)* | *1974* |

The 1488-pair difference between the two rules is exactly the legitimate relationships the ancestry
qualifier preserves: every antihistamine pair (`R06AX`, 88), antidepressant (`N06AX`, 136),
antiepileptic (`N03AX`, 91), aminoglycoside (`J01GB`, 45), opioid (`N02AX`) and "other antineoplastic"
(`L01XX`, 276) pair in the KB.

**Is any legitimate relationship lost?** Yes, and it is worth stating rather than rounding to zero:
two of `V03A`'s children do share a mechanism — `V03AC` iron chelators (3 pairs) and `V03AE`
potassium/phosphate binders (6 pairs) — and they lose their claim along with the antidotes. That is
9 pairs, against the 135 antidote pairs the same rule removes. I kept the group-level rule rather
than carving those two out, because the criterion is ATC's own words about a group and a per-child
exception list is precisely the hand-tuning that made #161's list incomplete. The trade is recorded
in the javadoc.

## What changed, per site

**#167 — both arms, one exclusion.** New `DrugReference.isUnclassifyingAtcCode(code)` over a
28-prefix list with the criterion, the derivation and the measured impact in its javadoc.
`DrugSafetyValidator.sharedClass` skips such a subgroup outright in *both* of its tiers rather than
demoting it — a demotion cannot work here, because potassium iodide and acetylcysteine share one
locally-applied bucket (`S01XA`) *and* one systemic bucket (`V03AB`), so every tier a demotion could
fall through to is occupied by another bucket that means nothing either. When nothing is left, the
arms fall through to the curated cross-reactivity groups, which is the only class statement this
module makes from clinician-curated data rather than from a code.

**#171 — one claim per co-medication, and a class chosen by rule not by position.**
`classRelationships` now groups the active-order codes by the co-medication they identify and words
one sentence per partner, and the class named is `sharedClass`'s choice over the whole shared set.
That is the same function the allergy arm has used since #161/#166, so the two arms can no longer
describe one pair through different subgroups, and neither depends on the order a dictionary
published its mappings in. `ruleAbout` takes the partner's whole code set (sorted) so a rule filed
under one of a substance's five codes still correlates.

**#155 — a name that does not depend on the reference dataset.** The partner label follows one
ladder: the dataset's name for the substance → the contributing **order's own display name** → the
bare code. Only the last rung can still print a code, and only where the context carries no
identified order at all (issue #118's flattened-set fallback). *Side effect worth flagging:* because
naming the order and correlating it are the same resolution, grouping an order's codes also closed
part of the residue `addInteractionWarnings` documented — on the curated default the rule chip
(cites `B01AC06`) and the class chip (matches under `N02BA01`) are now recognised as one pair and
fold into one chip. Both halves are pinned.

**#174 site 1 — the partner label is no longer dataset order.** `entryForAtcCode` folds in
`DrugReference.canonicalRow`, so it returns the row a chip should name rather than the row the KB
listed first. #174 stays open for sites 2–4.

## Tests

Eleven new cases, each failing first against `main` with the exact defect string, in three classes
driven through the real `DrugSafetyValidator.validate` over verbatim KB slices parsed by the real
`DdiDrugReferenceSource` (and, for #155, the real bundled curated dataset — the production default):

* `ResidualAtcClassClaimTest` — both #167 arms, plus **the boundary case**: loratadine ×
  desloratadine must still chip `R06AX`. That one passes before this change as well as after,
  deliberately: the risk in fixing #167 is over-vetoing.
* `MultiCodeClassChipTest` — one chip for metronidazole's five codes; the same chip when the code set
  is iterated in reverse; and a folded chip naming `M01AE` rather than the topical `M02AA`.
* `ClassChipPartnerLabelTest` — the order's display name instead of `N02BA01`; the code still printed
  when nothing names the order; `Cyclosporine` instead of `Cyclosporine (ophthalmic)`.

No existing test was modified, and none needed to be: nothing in the suite asserted any of these
defects as expected output. `CrossReactivityClassChoiceTest`'s `J01GB` case is the near miss — a
blanket residual veto would have broken it, which is a large part of why the rule is not one.

## Live evidence

Built from this branch, deployed to the 3.7.1 standalone, probed against `main` (`fe2e38f`) first and
then this build, on the same box with the same fixtures. Patients were resolved by name against the
live DB. Chip text is verbatim.

Three fixtures were created for shapes the demo data could not express (agent-created demo data,
tagged `I167`/`I171`/`I174`, left in place): an **Acetylcysteine allergy** on Susan Hall, a
**Metronidazole order** (`I171-METRO`) on James Thompson, and a **Salicylic acid order**
(`I174-SALI`) on Richard Brown together with the two `WHOATC` maps the shipped KB publishes for that
substance (`S01BC08`, `D01AE12`), which the stock dictionary's concept lacked.

### #167, interaction arm — Agnes Adams (aspirin 81mg), "Is it safe to give epinephrine?"

before (`main`) — 1 chip, and the false claim reached the **answer**, not just the chip:
```
[interaction] Epinephrine is in the same ATC class (A01AD) as active order Acetylsalicylic acid (aspirin) — possible duplicate therapy
answer: No — Epinephrine should not be given: Epinephrine is in the same ATC class (A01AD) (A01AD)
        as active order Acetylsalicylic acid [112], possible duplicate therapy [115].
```
after — **0 chips**, and the answer stops refusing adrenaline:
```
answer: The records do not address the safety of giving Epinephrine.
```
Had the exclusion missed `A01AD`, the chip above would simply still be there.

### #167, contraindication arm — Susan Hall (acetylcysteine allergy), "Is it safe to give potassium iodide?"

before — 1 chip:
```
[contraindication] Potassium Iodide is in the same ATC class (V03AB) as the patient's allergy to Acetylcysteine — possible cross-reactivity
answer: No — Potassium Iodide should not be given: … suggesting possible cross-reactivity [462].
```
after — **0 chips**; `The records do not address the safety of giving potassium iodide.` Had only the
site-group half of the rule shipped, `S01XA` would have been vetoed and this chip would have read
`(V03AB)` unchanged — which is exactly why `V03A` is in the list.

### #171 — James Thompson (metronidazole, five WHOATC maps), "Is it safe to give tinidazole?"

This shape had never been measured live. before — **3 chips for one fact, and 3 injected citations**:
```
[interaction] Tinidazole is in the same ATC class (J01XD) as active order Metronidazole — possible duplicate therapy
[interaction] Tinidazole is in the same ATC class (G01AF) as active order Metronidazole — possible duplicate therapy
[interaction] Tinidazole is in the same ATC class (P01AB) as active order Metronidazole — possible duplicate therapy
answer: … (J01XD, G01AF, P01AB) … [337], [338], [339].
```
after — **1 chip, 1 citation**, naming the systemic subgroup:
```
[interaction] Tinidazole is in the same ATC class (J01XD) as active order Metronidazole — possible duplicate therapy
```
The finding itself is real (both are nitroimidazoles) and survives; only the triplication goes. Note
the before-order — `J01XD, G01AF, P01AB` — is the dictionary's mapping order, not sorted, which is
the position-dependence the second half of #171 is about.

### #155 — Agnes Adams on the DEFAULT `sourceFormat=json`

The load is lazy, so the source was proved switched rather than assumed, via
`GET /chartsearchai/drugreferencestatus`:
```
{"enabled":true,"loaded":true,"inert":false,"entryCount":4,"sourceFormat":"json",
 "configuredSourceFormat":"json","configuredDataFilePath":"","origin":"classpath:/chartsearchai/drug-reference.json"}
```
(The first attempt flipped only the format and left `dataFilePath` on the 19 MB DDInter file; the
same endpoint reported `inert:true entryCount:0` and caught it. Both GPs have to move.)

after — the partner is named by the order, and the source-exclusive control note proves which
dataset produced it (`additive GI and bleeding risk` exists only in the curated seed):
```
[interaction] Ibuprofen interacts with active order aspirin — additive GI and bleeding risk. Ibuprofen is in
              the same cross-reactivity group (NSAID) as active order Aspirin 81mg — possible additive or
              duplicate-class therapy
```
Same on a second run. Issue #155 recorded `… as active order N02BA01` for this configuration; that
is also what the new test reproduces in-process against the unfixed code. **Limitation:** the live
before-capture for this arm is the issue's, not mine — my before-pass ran on `ddinter`. This capture
also had to use a module (not JVM) restart, which is known to leave querystore retrieval dead, so
only the chip is evidence here — not the prose or the references.

### #174 site 1 — Richard Brown (salicylic acid order), "Is it safe to give diclofenac?"

before / after:
```
- [interaction] Diclofenac is in the same ATC class (S01BC) as active order Salicylic acid (sodium) — possible duplicate therapy
+ [interaction] Diclofenac is in the same ATC class (S01BC) as active order Salicylic acid — possible duplicate therapy
```
The class is unchanged and correct: `S01BC` is a real subgroup, so this doubles as evidence the
exclusion is not over-broad — the same order's *other* code, `D01AE12`, IS on the veto list.

### The must-keep case, and the controls

Joshua Johnson + ibuprofen — **2 chips before, the same 2 after**, byte-identical, including the
class relationship that is genuine:
```
[contraindication] Ibuprofen is in the same cross-reactivity group (NSAID) as the patient's allergy to
                   Acetylsalicylic acid (aspirin) — possible cross-reactivity
[interaction]      Ibuprofen interacts with active order lisinopril — Moderate. …
```

| control | expected | measured |
|---|---|---|
| Mary Smith + clarithromycin | 1 Major × simvastatin | 1 ✓ |
| Agnes Adams + warfarin | 1 Major × aspirin | 1 ✓ |
| Mary Smith + paracetamol | 0 | 0 ✓ |
| Joshua Johnson + ibuprofen | 2, incl. NSAID contra | 2 ✓ |
| Betty Williams + bupivacaine | exactly 1 | 1 ✓ |
| George Anderson + esomeprazole | 2, both PPIs | 2 ✓ (`A02BC` × pantoprazole allergy) |
| Sarah Taylor + hydrocortisone | 3 named `Hydrocortisone` | 3 such chips ✓, in a set of 9 — see below |

Sarah Taylor returns 9 chips rather than 3, and the extra 6 are not this change: she now carries
**three** NSAID orders (celecoxib, diclofenac, ibuprofen) rather than the one #173 measured, so the
three interaction chips are one per order — all named `Hydrocortisone` with systemic prose, which is
the property that control exists to protect — and the remaining chips are answer-driven steroid names
checked against her dexamethasone allergy, via `H02AB` and `R01AD`. Neither is a residual bucket.

## Build

Measured with `mvn -o clean install` (full reactor) in this environment, both numbers mine:

| tree | api | omod | total |
|---|---|---|---|
| `main` @ `fe2e38f` | 926 | 61 | **987** |
| this branch | 937 | 61 | **998** |

0 failures, 34 skipped on both; the +11 is exactly the eleven cases added here. Resolved
`querystore-api-1.0.0-SNAPSHOT.jar` md5 `e9917ab5aa895199fa395d7b932b795b` (the Aug-3 remote
snapshot) before and after every build. CI has been in a runner outage, so there may be no CI result
to read.

## Scope

`Closes #167`
`Closes #171`
`Closes #155`

**#174 stays open.** Only its site 1 (`entryForAtcCode`) is addressed here. Sites 2–4 —
`orderedInteractionNotes` emitting one note per partner row (1872 of 2283 entries, 19,228 surplus
rows), `addActiveOrderPairInteractions`' subject label, and `addOverdose` — belong to the
one-substance-one-row pattern #173 swept, and are deliberately left for that sweep.


dkayiwa:
**Standalone state left behind** (3.7.1 box), so a later reader can tell fixture from demo data.

Left running this branch's build, GPs back as found: `sourceFormat=ddinter`, `dataFilePath=chartsearchai/ddi_knowledge_base.json`, verified after a clean full-JVM start — `{"loaded":true,"inert":false,"entryCount":2283,"sourceFormat":"ddinter"}`, querystore `hits=75`, Mary+clarithromycin control chip present.

Created and **left in place** (agent-created demo data, all tagged):
- `I167` — Acetylcysteine allergy (concept 130) on **Susan Hall**, allergy uuid `i1670000-…a1`. Gives the `V03AB` antidote-bucket pair a live shape.
- `I171-METRO` — Metronidazole order (concept 56, five `WHOATC` maps) on **James Thompson**, order uuid `i1710000-…a1`.
- `I174-SALI` — Salicylic acid order (concept 34) on **Richard Brown**, order uuid `i1740000-…a1`, plus two `WHOATC` concept_reference_terms/maps added to concept 34 (`S01BC08`, `D01AE12`) — the stock dictionary's Salicylic acid concept carried no ATC map, and site 1 needs an ATC-mapped order.

Three patients chosen because each had a real chart but neither an allergy nor a drug order, so each probe's chip set is exactly the case under test.

One incident worth recording: the JVM hit `OutOfMemoryError` during the **third** `moduleaction` module restart within one JVM lifetime, while restoring the GPs after the #155 capture. All evidence was already captured; the box was recovered with a full-JVM restart. Treat ~2 module restarts per JVM as the practical ceiling on this box.

dkayiwa:
## Hardening pass

Fresh eyes: three Phase 1 passes, a three-agent Phase 2 fan-out, and one targeted return to Phase 1 when polish found a test that could pass vacuously. **No decision in this PR was reversed.** The exclusion list gained two members that change nothing on shipped data, and nine statements that this change had made false were repaired. Everything below that carries a number I measured myself unless it says otherwise.

### The list was re-derived independently, and it reproduces

I did not read the constant and check its arithmetic. I re-derived the criterion from the WHOCC ATC/DDD index (which today serves the **2026** release — I could not retrieve a 2024 snapshot, so the "933 level-4 subgroups in the 2024 index" figure is not something I can confirm; every *name* checks out), enumerated every level-4 subgroup under each of `LOCALLY_APPLIED_ATC_GROUPS`, and re-measured the KB impact by driving the **production predicates** (`DrugReference.isLocallyAppliedAtcCode` / `isUnclassifyingAtcCode`, compiled against the built api jar) over the shipped 19 MB file.

| | PR says | I measured |
|---|---|---|
| level-4 residues under the locally applied groups | 27 | **27**, same set |
| pairs sharing ≥1 level-4 subgroup | 7783 | **7783** |
| lose the class claim | 486 | **486** |
| keep a claim, naming a classifying subgroup | 54 | **54** |
| untouched | 7243 | **7243** |
| of the 27, used by the KB | 20 | **20** |
| `V03A` children used by the KB | 8 | **8** |

**And it does not repeat #161's failure mode.** That was the specific worry: #161's list was incomplete in a way that *reproduced* the defect, because the fix itself moved 21 pairs onto a group it had missed. I checked the same thing here — of the 54 pairs this change moves onto a different subgroup, **zero** land on a residual bucket. They land on `S01AA`, `S01GA`, `S01BC`, `S01EA`, `R01AA`, `R02AB`, `S01EB`, all of which classify. The `J01GB` test does real work: reverting the veto to a tier-1-only demotion fails both of `ResidualAtcClassClaimTest`'s arms, and reverting `sharedClass` to the pre-#171 first-shared-code walk fails four tests including the code-order one.

### Two members were missing

Both added; **both match no shipped-KB entry**, so no number above moves and no live capture changes. They are on the list for the same reason four of `SYSTEMIC_USE_EXCEPTIONS` are — the criterion decides membership, not the test suite.

* **`S02DC` "Indifferent preparations"** (ear-wax oils) under `S02D` "OTHER OTOLOGICALS", inside `S`. A bucket ATC fills by exclusion whose published name begins with neither "Other" nor "Various" — so the derivation's name test *structurally cannot find it*. WHOCC's own name search returns exactly this one row for "indifferent" and nothing for "miscellaneous", which is why naming one such bucket is enough. (`D09AX` "Soft paraffin dressings" and `D07XA`–`D07XD` are also inside `D` and also not named "Other …", but they name what their members are, so they classify and stay out.)
* **`V07A` "ALL OTHER NON-THERAPEUTIC PRODUCTS"** — `V03A`'s structural twin. They are the only two groups in the index whose own name begins "ALL OTHER", and both sit directly under `V` "VARIOUS", so nothing above either names a body system or a property. Writing the rule for one and not the other was an accident of which one produced a bug report.

### On the 9-pair loss you asked me to judge

**I agree with keeping the group-level rule. I do not agree that the cost is 9 pairs — it is 116, and the javadoc now says so.**

The "9" counts only `V03AC` and `V03AE`. But the same argument applies wherever the vetoed subgroup's own published name states a therapy, and ten of them do. Measured over the 486:

`D06AX` 33 · `D05AX` 26 · `D01AE` 25 · `S01GX` 12 · `V03AE` 6 · `D10AX` 5 · `V03AC` 3 · `G01AX` 3 · `S01AX` 2 · `M02AX` 1 — **116**.

Concretely, and neither was acknowledged: **calcipotriol × calcitriol** are both topical vitamin-D analogues and share only `D05AX`; **azelastine × cetirizine** are both H1 antihistamines and share only `S01GX`. And **451 of the 486 carry no DDInter rating at all**, so for those the class chip was the only chip — the loss is not softened by a rule arm picking it up.

Why I still agree with the decision:

1. The alternative is a per-group pharmacological judgement ("does `D06A` 'Antibiotics for topical use' mean it, or is it just a bucket?"), which is exactly the hand-tuning that left #161's list incomplete, and which this module has no curated data to make. It is also #183, which is deliberately deferred.
2. The 486 buys the removal of claims that are not merely weak but false — "adrenaline duplicates aspirin", "potassium iodide cross-reacts with acetylcysteine".
3. Every vetoed pair still falls through to the curated cross-reactivity groups, which is the one class statement this module makes from clinician-curated data.

What I have changed is the accounting. The javadoc said *"That is 9 pairs, against the 135 antidote pairs the same rule removes"*, which reads as the whole cost; it now carries the 116 with its breakdown and the two examples.

### Verified, not assumed

* **A demotion cannot work** — confirmed, not taken on trust. Reverting the both-tier skip to a tier-1-only demotion fails `aLocalOralTreatmentBucketRaisesNoDuplicateTherapyChip` **and** `anAntidoteBucketRaisesNoCrossReactivityChip`, the second reporting `S01XA` — the pair falling through to the other meaningless bucket, exactly as the javadoc predicts.
* **Partner grouping is by identity, not text**, and it is not a third dedup mechanism — it keys on the same `DrugReference.substanceGroupKey()` the contraindication ledger and the interaction subject side already use. No key space can collide: a `List`, a `DrugReference`, an `ActiveDrugOrder` and a `String` cannot be equal to one another.
* **The #155 ladder's last rung is justified, not unreachable.** `PatientClinicalContextBuilder` only adds an `ActiveDrugOrder` when `orderNames` is non-empty, so a listed order always has a display name; the bare code survives for a context built from the flattened set alone and for the builder's own KNOWN GAP, and `withNoIdentifiedOrderToNameTheCodeIsStillTheLastResort` pins it. Removing the order rung fails `anOrderTheDatasetDoesNotCoverIsNamedByItsOwnDisplayName`.
* **`sharedGroupForCode` → `sharedGroupForCodes` needs no deprecated delegate.** The module is `1.0.0-SNAPSHOT` and its only tag/release is a demo-data pre-release, so no consumer can be compiled against the old signature; it had one in-package caller. A one-code delegate would also be actively harmful — it would re-offer the per-code reading #171 exists to remove.
* **The perf hazard is a hazard.** Measured through the real `validate` over the real 19 MB KB, branch vs `main`: a 20-active-order chart **23.7 ms → 45.8 ms**, a 100-order chart **127 ms → 155 ms**. Against a 15–90 s answer. No memoisation added, which is right — #172 means any memo must be a per-`validate` local and never a field, and the javadoc now records that constraint where a future reader will hit it.

### Nine statements this change had made false

Repaired in this pass; details in the commits. The sharpest was `DrugSafetyValidator:880`, which still described the class arm as *"walked per active-order CODE … once per code"* twenty-four lines above the slice's own comment saying #171 removed exactly that. Also: `atcSubgroups()`'s *"the one shared definition used by both the injector and the validator"* (the two have deliberately diverged — the injector keeps the raw intersection because it is choosing what to put in front of the model, where an extra record is noise); `InteractionPairs`' stated reason for excluding class chips; the class javadoc's *"'Same class' means a shared ATC level-4 chemical subgroup"*; ADR Decision 27's *"only when no ATC subgroup is shared"*; Decision 31's *"two prefix lists"*; both `warnOnInteractions` and `warnOnContraindications` GP descriptions in `config.xml`; and `docs/drug-kb-demo.md`'s field table. **ADR Decision 33** is added — the log had no entry for #167/#171/#155/#174-site-1 at all.

Two of the slice's own new claims were also false and are now corrected:

* `classRelationships`' `@return` promised *"in the context's own code order (so chip order is unchanged)"*. It is partner **first-appearance** order, and the difference is observable: with a union of `[D06BX01, G01AF02, J01XD01]` and a tinidazole question, the old walk emitted clotrimazole then metronidazole and the new one emits them the other way round. Chip content is unchanged; two class-only chips can swap places.
* `orderPartners`' *"Resolved once per pass"* — it runs once per **subject**, and `ruleAbout` re-runs `entryForAtcCode` rather than reading the carried resolution.

### Two of the new tests could pass vacuously

Polish surfaced one structural item, so it went back through Phase 1. Both fixes are additions; no existing assertion is relaxed, changed or removed.

* **`ClassChipPartnerLabelTest` had no fixture precondition** — and it is the file that most needs one. Its #174 site-1 case distinguishes `canonicalRow` from first-wins *only* because `ddi-class-partner-canonical-row.json` happens to list `Cyclosporine (ophthalmic)` at index 1 and `Cyclosporine` at index 2, both publishing `L04AD01`. I confirmed that ordering; nothing asserted it. Regenerate that slice in another row order and both resolutions agree, the case stays green, and the regression walks back in — the same trap `DDI_ROUTE_VARIANTS`' javadoc already warns about for its own slice. `theFixtureReallyListsTheRouteQualifiedRowFirst` now pins it.
* **`theFixtureReallySharesThreeSubgroupsWithNothingToFoldInto` checked only the first half of its name.** A rated row correlating tinidazole with metronidazole would fold the three sentences into one chip whatever `classRelationships` did, so "three before, one after" would stop measuring anything. The fixture's three interaction rows name only Ketoprofen, Metronidazole and Ibuprofen — verified, and now asserted.

Also from that pass: **`UNCLASSIFYING_ATC_SUBGROUPS` → `UNCLASSIFYING_ATC_GROUPS`.** Two of its thirty entries (`V03A`, `V07A`) are level-3 groups, not the `ATC_SUBGROUP_PREFIX_LENGTH` subgroups the name claimed — the same mixed-level shape its sibling `LOCALLY_APPLIED_ATC_GROUPS` is correctly named for, consumed by the same `fallsUnderAnyGroup` helper. Private, no call site outside the file. On a PR whose subject is not over-reading an ATC level, the constant should not misstate its own.

### Polish deliberately not applied

Each with the consequence spelled out rather than a severity label:

* **An order-side early return in `classRelationships`** (`Collections.disjoint(atcSubgroups(union), refClasses)` plus a group check, before the partner scan) would skip the whole measured 22–30 ms on every validate that cannot raise a class chip. Ship without it and nothing breaks: the cost is 22–30 ms against a 15–90 s answer, i.e. 0.03%, and it is production code changing after the live pass. The strongest of the deferred items; worth its own PR with a before/after measurement rather than a late edit here.
* **`if (rules.isEmpty()) return null;` at the top of `ruleAbout`** saves ~3 ms of scans in the class-only case. Ship without it and nothing breaks — the loop it guards already does nothing when `rules` is empty.
* **`DrugReference.publishesAtcCode(String)`** to replace `normalizedAtcCodes().contains(code)` inside `entryForAtcCode`'s scan would remove ~400 KB of garbage per scan. Ship without it and nothing breaks; it also changes the null-argument contract from `false` to NPE, which wants its own reasoning.
* **Hoisting `normalizedAtcPrefixes()` out of `containsAnyCode`'s loop.** Ship without it and nothing breaks: the shipped curated file has one group with two prefixes, so it is at most five two-element sets per partner.
* **Consolidating the `shared()` fixture-characterisation helper**, now written three times (two of them new here) against `DrugReferenceTestSupport`'s own stated rule. Ship without it and nothing breaks today, but the three copies already differ in strictness — one asserts `size()==1`, one only `!isEmpty()` — so the next class-choice test will copy whichever it lands next to, and the fixture preconditions will disagree about what "resolves" means. The consolidation touches `CrossReactivityClassChoiceTest`, which predates this PR; better as its own change.
* **Narrowing `CrossReactivityGroup.sharedGroupForCodes` and `containsCode` to package-private.** Ship without it and nothing breaks — no consumer exists. Doing only these two would make the class's visibility *less* uniform than it is now (`containsAnyCode`, `groupsOf`, `normalizedAtcPrefixes`, `sharedGroup` are all public with only in-package callers); it is all five or none, which is not this PR's subject.

### Live re-verification

Full-JVM restart, `ddinter` / 2283 entries confirmed by `GET /chartsearchai/drugreferencestatus` (not a log line). Run against `56d5275`; the only non-comment production change since is the private constant rename above (same contents, same behaviour — `git diff 56d5275 -- api/src/main/java` filtered to non-comment lines is exactly those two lines), so the evidence stands. Every patient resolved by **name lookup against the live DB**. Two fixtures created for this pass and left in place: a **Metronidazole** order on James Thompson (concept `79782AAAA…`, five `WHOATC` maps — neither he nor Richard Brown had any active drug order left, so the author's cases could not be re-run as-is) and a **Salicylic acid** order on Richard Brown (concept `83645AAAA…` → `D01AE12` + `S01BC08`).

| probe | result | what it would have shown if the change were wrong |
|---|---|---|
| **Joshua Johnson + ibuprofen (must-keep)** | **2 chips**, incl. `Ibuprofen is in the same cross-reactivity group (NSAID) as the patient's allergy to Acetylsalicylic acid (aspirin) — possible cross-reactivity` | the NSAID relationship is genuine and cross-branch; an over-broad veto silences it |
| Agnes Adams + epinephrine | **0 chips**, 0 refs (re-run to clear the cold-start race) | `…same ATC class (A01AD) as active order Acetylsalicylic acid (aspirin)` |
| Susan Hall + potassium iodide | **0 chips** (re-run) | `…same ATC class (V03AB) as the patient's allergy to Acetylcysteine` |
| James Thompson + tinidazole | **1 chip** `(J01XD)`, **1 citation** | 3 chips + 3 citations — `G01AF`, `J01XD`, `P01AB`, one per shared subgroup |
| Richard Brown + diclofenac | `…same ATC class (S01BC) as active order **Salicylic acid**` | `Salicylic acid (sodium)`, the first row listed. Also proves the veto is not over-broad: the same order's other code `D01AE12` **is** vetoed and `S01BC` survives |

Controls: Mary+clarithromycin 1 Major ×simvastatin ✓ · Agnes+warfarin 1 Major ×aspirin ✓ · Mary+paracetamol 0 ✓ · Betty+bupivacaine exactly 1 ✓ · George+esomeprazole 2, both PPIs via `A02BC` ✓.

**Sarah Taylor's expectation needs restating, not the code.** She returns **9** chips, not the 3 #173 recorded, because she has acquired three NSAID orders (Advil, Celebrex, Diclofenac). Corrected expectation: **9 chips, of which 4 are named `Hydrocortisone`** — one contraindication via `H02AB` against her Dexamethasone allergy, plus **3 interaction chips, one per NSAID order**, all with systemic corticosteroid/NSAID prose, which is the property that control exists to protect — and 5 answer-driven steroid names (Hydrocortisone butyrate, Dexamethasone direct-allergy, Prednisone, Budesonide via `R01AD`, Methylprednisolone). Neither `H02AB` nor `R01AD` is a residual bucket, so this PR moves none of them.

**One caveat left open.** I did not re-run the `sourceFormat=json` capture. Flipping that GP needs another restart, and a third `moduleaction` restart in one JVM OOMs this box. That arm is pinned in-process by `ClassChipPartnerLabelTest.anOrderTheDatasetDoesNotCoverIsNamedByItsOwnDisplayName`, which drives the **real** bundled seed through the **real** validator; removing the order rung makes it fail with `as active order A01AD05`. `docs/drug-kb-demo.md` now points at that test rather than asserting a live reading.

### Build

`mvn -o clean install`, full reactor, all four numbers mine:

| tree | api | omod | total |
|---|---|---|---|
| `main` @ `fe2e38f` (clean worktree, checked out and built there) | 926 | 61 | **987** |
| this branch as I received it @ `9c6c320` | 937 | 61 | **998** |
| this branch after hardening | 938 | 61 | **999** |

0 failures, 34 skipped throughout. The +1 is the one test added below. Resolved `querystore-api-1.0.0-SNAPSHOT.jar` md5 `38205608891f10b68929c35fb3eace7e` before and after every build — that is the **Aug-6 remote snapshot `-82`**, not the Aug-3 `-81` (`e9917ab…`); a newer published snapshot, not a rogue local reinstall, and it compiles and passes. GitHub Actions is in a runner outage — `gh pr checks 182` reports *"no checks reported on the 'fix/atc-class-claims' branch"*, so this PR has no CI result at all.

### Found, not fixed — worth filing

1. **`A16AX` and `N07XX` license the very claim #167 removes, at 91 and 55 shipped-KB pairs.** Both are residues of a level-2 group that is itself ATC's residue for a whole main group (`A16`/`A16A` and `N07`/`N07X`, both "OTHER …") — structurally what `D11AX` is, and `D11AX` is vetoed only because dermatologicals happen to be locally applied. Eliglustat × givosiran; pitolisant × inotersen. Smaller siblings `V04CX` 21, `G02CX` 3, `R07AX` 1, `M09AX` 1. Not folded in here because it would move every number in this PR and the live evidence measured against it — and because taking the whole family costs real relationships too (`G02CX` bremelanotide × flibanserin, `M09AX` onasemnogene × risdiplam, and `A16AX` itself holds miglustat × eliglustat). Measured by me.
2. **A drug can be reported as duplicating itself.** The "restating existing therapy" skip tests ATC codes, not substance identity, so when a KB row's ATC list omits the substance's real WHO code the order is not recognised as the same drug. Seven shipped rows have the shape — `Omeprazole` carries only `A02BC05` and its `rxnorm_name` is literally "esomeprazole", while `A02BC01` appears nowhere in the KB — giving *"Omeprazole is in the same ATC class (A02BC) as active order Omeprazole 20mg"*. Pre-existing; #155's label ladder is what turns it from an obscure code into visible nonsense. **Not reachable on the 3.7.1 demo dictionary** — I checked live, its Omeprazole concept carries no `WHOATC` mapping at all.
3. **`orderPartners`' ladder splits one order into two partners when the dataset covers it only partly**, giving two chips for one co-medication. Documented in the javadoc by this pass; the fix is to group by ORDER first and resolve the entry from the group, which is a behaviour change needing its own measurement. Same root makes the builder's nameless-order KNOWN GAP produce one partner per code.
4. `ruleAbout`'s `TreeSet` determinism guard is **unpinned** — reverting it to plain iteration leaves 937/937 green. Its observable effect needs an entry-less multi-code partner with two co-equal rules, which the shipped KB cannot produce.
5. The fold's `else` branch is **never executed** by the suite — verified by making it throw.
6. `CrossReactivityGroup.containsCode(String)` is `public` and, now that this PR removed `sharedGroupForCode`, has no caller outside `containsAnyCode` in its own class.
7. Stale docs beyond what I repaired: ADR Decision 32 still says #173 is *"pending, not yet merged … add it when it lands"* (it merged as `f502ec4`); `README.md` and `docs/drug-knowledge-base-comparison.md` still describe `atcSubgroups()` as "the shared definition behind both" the injector and the validator; the 22 captures under `eval/drift-metric/fixtures/probe-safety/**` embed pre-#173 `detail` strings (their assertions still pass, since the scorer reads only `type`/`drug`/`resourceUuid`).

dkayiwa:

- `mvn -o clean install` on a clean detached worktree at `e8c0e84`: **938 + 61 = 999, 0 failures, 34 skipped**, BUILD SUCCESS. Reproduces the hardening figure exactly. `main` at `fe2e38f` = **926 + 61 = 987**, also measured by me.
- **No pre-existing test file is touched at all** — six new files, and zero removed lines contain an assertion anywhere in the branch.
- Querystore jar `38205608891f10b68929c35fb3eace7e`, confirmed to be `querystore-api-1.0.0-20260806.153951-82.jar`, the newest **remote** snapshot. So the resolution moved for a legitimate reason and this is not #181 recurring.

Merging with **no CI result at all** — GitHub Actions has been in a hosted-runner outage and no workflow has run for this branch (`gh pr checks` reports "no checks reported"). Basis: the local full-reactor build above, three JDK-independent facts (no production API change outside this module, no new dependency, no lifecycle change), and the live pass below.

**The hardening found the gap I asked it to look for, and it was real.** The exclusion list was built by taking every WHO ATC level-4 subgroup whose published name begins "Other"/"Various" and intersecting with the locally-applied groups — so a residue whose name begins with *neither* word is structurally invisible to that test. **`S02DC` "Indifferent preparations"** (under `S02D` OTHER OTOLOGICALS) is exactly that case, and **`V07A` "ALL OTHER NON-THERAPEUTIC PRODUCTS"** is `V03A`'s twin — the only two groups in the index named "ALL OTHER". Both added; both match no shipped-KB entry, so they are on the list by criterion rather than by measured impact, which the constant's javadoc now states as the deliberate rule.

**It also corrected the stated cost by an order of magnitude, in the direction that matters.** The PR claimed 9 pairs lose a real relationship; the measurement is **116** — with `D06AX` 33, `D05AX` 26, `D01AE` 25, `S01GX` 12 leading — and it named two the original missed: **calcipotriol × calcitriol** and **azelastine × cetirizine**. Sharper still: **451 of the 486 affected pairs carry no DDInter rating**, so for those the class chip was the only chip and the veto means silence rather than a downgrade. The javadoc's justification was false for ten of the 27 members and now says the over-reach is deliberate and carries the 116.

I agree with keeping the group-level rule. The alternative is per-child exceptions, which is precisely the hand-tuning that left #161's list incomplete in a way that reproduced the defect it fixed. But the corrected figure is now on **#183**, which is where that trade gets decided properly, alongside **#184** (residues of a level-2 group that is itself ATC's residue — `A16AX` 91 pairs, `N07XX` 55).

On the stopping rule it was honest rather than convenient: passes 1 and 2 both found substantive items, pass 3 found falsehoods in its *own* pass-2 corrections, and it invoked §5 to stop writing prose rather than claim a convergence it had not observed. Code and tests were stable from pass 1 onward.

Also filed from this pass: **#185** (a drug reported as duplicating itself — the existing-therapy skip compares ATC codes, not substance identity; seven shipped rows, unreachable on the demo dictionary but not on a mapped one) and **#186** (`orderPartners` splitting one order into two partners under partial dataset coverage).

PR #187, opened 2026-08-07
fix(drug-safety): decide identity by substance, and name the patient's own record

dkayiwa:
## The defect

A substance was reported as cross-reactive, or interacting, **with itself**, by two routes the
substance key of #145/#160/#173 does not cover. Neither is a key that was too wide.

**The mirror path.** `IDENTITY` was awarded on ENTRY identity — `allergen == ref`. When the row an
allergy resolves to is a sibling the question does not itself put in play, the group holds no
identity candidate at all, so the best available relationship is the class comparison, against the
same substance.

**Stem divergence.** `Pfizer-BioNTech Covid-19 Vaccine` and `Tozinameran` publish one `rxcui`
(2468231) and one `rxnorm_name` but no display stem in common, so the stem VETO kept them apart —
correctly, as far as a display name can tell — and the chip read as a self-reference anyway. The
same shape sits in the interaction arm: `Daxibotulinumtoxina` and `Botulinum toxin type A` share
`rxcui` 1712 and one `rxnorm_name` and differ in stem.

## What distinguishes two rows of one substance from two substances under one name

The reference data's substance name over-merges. #160 measured that `rxnorm_name` and `rxcui`
partition the 2283 entries identically — 142 families, 332 entries, 0 families disagreeing in either
direction — and rejected that partition as the collapse key, because among those 142 families sit
`Omeprazole`/`Esomeprazole`: one `rxnorm_name` (`esomeprazole`), one `rxcui` (283742), one ATC code
(`A02BC05`), and two substances. That refusal was right and stands.

What was wrong was the veto. The display stem is a **display-name heuristic**: it cannot tell a
second SUBSTANCE from a second NAME, so it separates the two PPIs correctly and separates
`Tozinameran` from its own brand row incorrectly.

The reference data can tell them apart, and does, in a field the module was not reading: a
**`drugbank_id`**, an identifier of a *substance* rather than of a presentation. 2077 of the 2283
rows carry one; a row that carries none makes no claim to be a substance of its own, which is what
the rule below turns on.

```
Tozinameran                       rxcui 2468231  rxnorm_name "SARS-CoV-2 …"  drugbank DB15696
Pfizer-BioNTech Covid-19 Vaccine  rxcui 2468231  rxnorm_name "SARS-CoV-2 …"  drugbank null
Botulinum toxin type A            rxcui 1712     rxnorm_name "botulinum …"   drugbank DB00083
[...]

PR #192, opened 2026-08-08
fix(drug-safety): resolve an allergen to the entry that IS it, not the earliest that claims its name

### LINKED ISSUE #176: 206 of 2283 drug-reference entries do not resolve to themselves, so a chip can be subjected on the wrong variant of the drug ordered
## What

`lookupByToken` resolves a drug name to the **earliest** matching entry, and **206 of 2283** entries do not resolve to themselves — the token in their own name matches an earlier entry first. So the chip's subject can be a different variant of the substance than the one actually ordered.

Live-observed during #173's hardening: an order for **Botulinum toxin type A** is subjected on `Daxibotulinumtoxina (botulinum toxin type a)` — a distinct product whose parenthesised qualifier happens to contain the ordered name.

## Why it is worth its own issue

- **206 entries is 9% of the KB**, so this is not an edge case; it is a systematic property of earliest-match resolution over a corpus where names nest.
- It is **upstream of the whole cluster we have been fixing.** #145/#160/#162/#173 have made chip *counts* and *labels* correct given a correctly resolved entry. This defect means the entry itself can be the wrong one, so those fixes cannot help — and the substance key deliberately keeps `Daxibotulinumtoxina` and `Botulinum toxin type A` apart, which is right, and does not stop the misresolution.
- It is already partially documented in `addAllergyContraindications`, i.e. known and never quantified until now.

## Relationship to the rest

- **#164** is the labelling question for two rows that *share* a name. This is the converse: two rows that do **not** share a name, one resolving to the other.
- **#86** established that unanchored substring matching produces clinically wrong chips (`opium` ⊂ `tiotropium`); the matchers were anchored in #128/#148. This is the residue that anchoring cannot fix, because `botulinum toxin type a` really is a whole-word match inside `Daxibotulinumtoxina (botulinum toxin type a)` — the problem is *earliest-match*, not boundaries.
- A plausible fix direction: prefer an exact name match over any alias/parenthetical match before falling back to earliest, and measure how many of the 206 that resolves.

## Provenance

The 206 figure and the live observation are #173's hardening measurements, relayed; I have not re-run them. The connection to #86's residue is my reading.

dkayiwa:
## The defect

`DrugReferenceService.lookupByToken` resolved a clinician-entered drug name — an allergen as the chart
records it — to the **earliest** entry any of whose names matched. Reference names nest, so the row it
landed on could be another presentation of the substance, or a different substance.

Since #187 that row is what **all three** chips the allergy arm can raise NAME
(`allergen.displayLabel()` in the identity chip's subject and detail, and in both class chips' "the
patient's allergy to …"), so the misresolution is printed to the clinician. It is upstream of
#145/#160/#173/#182/#187/#188: those made counts, labels, classes and identity correct *given a
correctly resolved entry*.

It is not the boundary problem #128/#148 fixed. `botulinum toxin type a` is a whole-**string** match
inside the `Daxibotulinumtoxina` row's alias list, because that row's `rxnorm_name` IS that name — no
boundary rule can see it. Nor is it `canonicalRow`'s question: both botulinum rows name no route, so
that fold keeps the earliest, which is the wrong one, and applying it here would rename a charted
`Ketorolac (ophthalmic)` allergy to `Ketorolac` — the mislabelling #187 exists to avoid.

## The re-measured population, and the method

The issue's headline figure predates #187/#188, so it was re-derived on `main` (88fdb5f) first. **Both
figures below come from production code** — `DdiDrugReferenceSource.parse` over the standalone's
19 MB `ddi_knowledge_base.json`, then `DrugReferenceService.lookupByToken`,
`DrugReference.substanceGroupKey`, `DrugReference.isNamed`, `DrugReference.canonicalRow` — driven from a
throwaway JUnit case, never from a reimplementation (§10). Re-measure before relying on them.

| measured on `main` (88fdb5f), asking for each of the 2283 entries by its own display name | |
|---|---|
| entries that do **not** resolve to themselves | **206** — the issue's figure, reproduced exactly |
| … of which resolve to a **different substance** (`substanceGroupKey`) | **54** |
| …… where the earlier row does **not** claim the name (a fragment matched) | 44 |
| …… where the earlier row claims the name **in full** (alias collision) | 10 |
| … of which resolve to **another row of the same substance** | **152** |
| …… where that row is not even the family's `canonicalRow` | 7 |

So the count is unchanged at 206, but what it means is not: after #187, 152 of the 206 are a
**labelling** defect (the verdict is right, the name printed is a different product), and 54 are a
**clinical** defect (a different substance — the identity chip becomes a cross-reactivity hedge, or
disappears entirely).

**The issue's headline example still reproduces, and changed category.** `Botulinum toxin type A` and
`Daxibotulinumtoxina` publish one `rxnorm_name` and one `drugbank_id`, so #187 made them one substance
— they are in the 152, not the 54. The chip's verdict is right and its subject names a different
product. The issue describes the trap as a "parenthesised qualifier"; in the data it is not one. The
Daxibotulinumtoxina row is *named* `Daxibotulinumtoxina` and carries `botulinum toxin type a` as its
`rxnorm_name`; the parentheses are `displayLabel()` rendering that synonym. A parenthesis-ranking rule
would not have fired on it at all.

After the change, measured the same way: **0 of 2283**, and 0 to a different substance.

Over every distinct name string the KB publishes (2283 display names + all aliases = **7452**):
**0 stopped resolving**, and 471 resolve differently — 412 decided by the display-name rank (108 of
those landing on a different substance from earliest-match) and **59 by the alias rank**, all 59 a
different substance.

## What changed

**`DrugReference.nameMatchStrength(String)`** — one definition of how strongly an entry claims a
clinician-entered name: its own display name **is** that name (2) > one of its other names is (1) >
the name occurs inside one of its names as a bounded token (0) > no match (-1).

**`DrugReferenceService.lookupByToken`** takes the strongest claim, earliest on a tie. It is gated on
the same `matchesDrugName` the old scan used, so **the set of entries a name can resolve to is
unchanged and only the choice within it moves** — it can never resolve a name that resolved to nothing,
nor fail to resolve one that did (0 of 7452, above). A recorded name no entry is *named* — a localized
display name with a strength appended, the ordinary chart shape — still resolves to the earliest match,
which is what keeps #147/#129's localized-allergen coverage intact.

**The alternatives, and why not.** *Longest matching alias* (the figure this javadoc used to record)
resolves the 44 fragment cases and none of the 10 alias collisions; the claim rank resolves both, and
it is a criterion rather than a tie-break by length. *`canonicalRow`* answers a different question — it
picks a substance's representative row for display — and using it here renames the charted allergen,
which is the regression #187's own commit message argues against. *Ranking a parenthesised qualifier
below the base name* does not fire on the headline case (see above), and where a stored qualifier is
involved (`Iron (polysaccharide)` → `Iron (bisglycinate)`) the display-name rank already settles it,
while demoting the qualifier would push a charted `Iron (polysaccharide)` allergy onto a row the chart
does not name. *Earliest-match as the final fallback is kept*, because with the two exact ranks in
place nothing is left for a length heuristic to fix on this corpus (0 of 2283) and the fallback is what
resolves inflected names at all.

**`DrugSafetyValidator.addAllergyContraindications`: the ledger's allergy finding key moves from the
resolved ROW to its SUBSTANCE.** This is required, not incidental. Two allergy records for two
presentations of one substance are one clinical fact, and they used to collapse *only because both
misresolved onto the same earlier row*: with `Hydrocortisone (ophthalmic)` and `Hydrocortisone
(topical)` both recorded, a correct resolution keyed on the row reports the patient twice. Measured on
the verbatim PPI/hydrocortisone slice: `main` 2 chips, resolution fix alone 4, resolution fix plus this
key 2.

## Call sites audited

`lookupByToken` has exactly **one** production call site: `addAllergyContraindications`, reached from
two places (the drugs-in-play loop in `validate`, and `addActiveOrderContraindications`, #143's arm).
Everything else naming it is javadoc or a test.

- **the identity chip** (subject + detail) — now names the row the chart's own string is;
- **the two class chips** ("… as the patient's allergy to X") — same;
- **`sharedClass(refClasses, allergen)` / `CrossReactivityGroup.sharedGroup(refGroups, allergen)`** —
  these read the *resolved* row's ATC codes, so a wrong resolution compared the wrong drug's
  classification. This is where the 54 hurt: `Enalaprilat` resolving to `Enalapril` left the arm
  comparing the wrong substance, and since the `Enalaprilat` row carries no ATC code at all the finding
  disappeared instead of being mislabelled (0 chips → 1, pinned by a test);
- **`seenAllergens`** — an identity `Set` of resolved allergens, so a more precise resolution can put
  two entries where one was. Handled by the ledger key above, and pinned;
- **#187's `allergen == ref` reference comparison** (#172's trap) — no longer exists: #187 replaced it
  with `substanceGroupKey().equals(...)`, which is value equality over the substance, so a changed
  resolution cannot silently flip an identity comparison. Both operands were re-read to confirm it.
- **`DrugReferenceInjector`** — resolves nothing by allergen name; its findings are rendered from the
  chips, so it inherits the fix.

## Live evidence

Built from this branch, deployed to the 3.7.1 standalone, full-JVM restart, `sourceFormat=ddinter`,
2283 entries. Every patient resolved **by name against the live DB**. The same probe script ran first
against `main` (88fdb5f) on the same instance, so the before column is measured, not reasoned.

**Fixture created** (agent-created demo data, kept — `$SP/h176-fixture.sh`, idempotent): a coded DRUG
allergy on three patients that are not regression controls — `Botulinum toxin type A` on **Margaret
King**, `Esomeprazole` on **Melissa Wright**, `Enalaprilat` on **Michael Turner** — plus a new
Drug-class concept `Enalaprilat` (the dictionary had none; `Botulinum toxin type A` and `Esomeprazole`
it already had). Non-coded allergens are rejected by this OpenMRS version, so an allergy has to be
coded.

### Confirming the source

```
GET /chartsearchai/drugreferencestatus
```

Patient state was read from the **database**, not from the REST order list: that list returns no results
for a DrugOrder it cannot serialize, and it does so at `v=full` too, not only at `v=custom` — Melissa
Wright reads 0 orders there and has an active Spironolactone order (order_id 337). Reported separately.

### The predictive check, before any probe

Of the 16 allergy tokens the probe patients actually carry, driven through the real `lookupByToken` over
the real 19 MB KB on both builds, exactly **three** move — and they are the three discriminators. Every
control token is identical:

```
                                    main (88fdb5f)                     this branch
'botulinum toxin type a'   -> [1300] Daxibotulinumtoxina      [1390] Botulinum toxin type A
'esomeprazole'             -> [ 380] Omeprazole               [1924] Esomeprazole
'ciprofloxacin lactate'    -> [  19] Lactic acid              [1600] Ciprofloxacin
'enalaprilat'              -> [1142] Enalaprilat              [1142] Enalaprilat        (unchanged)
'acetylsalicylate sodium'  -> null                            null                     (unchanged)
'aspirin'                  -> [1112] Acetylsalicylic acid     [1112] Acetylsalicylic acid
'ketoconazole'             -> [  11] Ketoconazole             [  11] Ketoconazole
'ibuprofen'                -> [2174] Ibuprofen                [2174] Ibuprofen
'pantoprazole'             -> [ 432] Pantoprazole             [ 432] Pantoprazole
'dexamethasone'            -> [1744] Dexamethasone            [1744] Dexamethasone
'tiotropium'               -> [ 907] Tiotropium               [ 907] Tiotropium
'tetryzoline (ophthalmic)' -> [ 876] Tetryzoline (ophthalmic) [ 876] Tetryzoline (ophthalmic)
'levothyroxine'            -> [  73] Levothyroxine            [  73] Levothyroxine
'tozinameran'              -> [1295] Tozinameran              [1295] Tozinameran
'ciprofloxacin'            -> [1600] Ciprofloxacin            [1600] Ciprofloxacin
'hydrocortisone'           -> [2156] Hydrocortisone           [2156] Hydrocortisone
```

### 1. Margaret King — the issue's headline, verbatim

`Is it safe to give botulinum toxin type A?` (chart records an allergy to **Botulinum toxin type A**;
one active Dexamethasone order)

**main (88fdb5f):**
```
--- safetyWarnings (1) ---
  [contraindication] drug='Daxibotulinumtoxina (botulinum toxin type a)'
      The patient has a recorded allergy to Daxibotulinumtoxina (botulinum toxin type a).
--- ANSWER ---
No — botulinum toxin type A should not be given: the patient has a recorded allergy to
Daxibotulinumtoxina [119], [122].
```
The wrong product name reaches the **answer prose**, not only the chip.

**this branch:**
```
--- safetyWarnings (1) ---
  [contraindication] drug='Botulinum toxin type A'
      The patient has a recorded allergy to Botulinum toxin type A.
--- ANSWER ---
No — botulinum toxin type A should not be given: The patient has a recorded allergy to
Botulinum toxin type A [122].
```

*Had the fix been wrong* (resolution unchanged) the chip would still have said `Daxibotulinumtoxina`.
*Had it over-reached* (e.g. by using `canonicalRow`) it would have said `Daxibotulinumtoxina` too,
because both botulinum rows name no route and that fold keeps the KB's earlier one.

### 2. Melissa Wright — a direct allergy attributed to a different substance

`Is it safe to give esomeprazole?` (chart records an allergy to **Esomeprazole**; active Spironolactone)

**main (88fdb5f):**
```
--- safetyWarnings (2) ---
  [contraindication] drug='Omeprazole'
      The patient has a recorded allergy to Omeprazole.
  [contraindication] drug='Esomeprazole'
      Esomeprazole is in the same ATC class (A02BC) as the patient's allergy to Omeprazole — possible cross-reactivity
```
Both chips are wrong in the same pass: the direct allergy is booked to omeprazole, and esomeprazole —
the drug the chart actually names — is downgraded to a cross-reactivity hedge.

**this branch:**
```
--- safetyWarnings (2) ---
  [contraindication] drug='Omeprazole'
      Omeprazole is in the same ATC class (A02BC) as the patient's allergy to Esomeprazole — possible cross-reactivity
  [contraindication] drug='Esomeprazole'
      The patient has a recorded allergy to Esomeprazole.
```
The two sentences have swapped subjects: the direct allergy is now booked to the drug the chart names,
and the hedge to the other substance. Still 2 chips — the must-not-collapse decision of #121 holds.

*Had the fix been wrong*, the chip count would be the same 2 and the two sentences would still have
named Omeprazole as the allergen — which is why this probe asserts the sentences, not the count.

### 3. Dorothy Odinga — the finding that vanished

`Is it safe to give ciprofloxacin?` (chart records an allergy to **Ciprofloxacin lactate**, one of
ciprofloxacin's own CIEL names; also an Ibuprofen allergy and an active `Aspirine Co 81mg`)

**main (88fdb5f):** 2 chips, **neither about the recorded ciprofloxacin allergy**
```
--- safetyWarnings (2) ---
  [interaction] drug='Ciprofloxacin'
      Ciprofloxacin interacts with active order aspirin — Moderate. …
  [contraindication] drug='Acetylsalicylic acid (aspirin)'
      Acetylsalicylic acid (aspirin) is in the same cross-reactivity group (NSAID) as the patient's allergy to Ibuprofen — possible cross-reactivity
```
The allergy resolved to `Lactic acid`, which carries no ATC code, so the class comparisons had nothing
to compare and the direct allergy reached the clinician neither as a chip nor as prose.

**this branch:** 3 chips — the missing finding is back, and the answer leads with it
```
--- safetyWarnings (3) ---
  [contraindication] drug='Ciprofloxacin'
      The patient has a recorded allergy to Ciprofloxacin.
  [interaction] drug='Ciprofloxacin'
      Ciprofloxacin interacts with active order aspirin — Moderate. …
  [contraindication] drug='Acetylsalicylic acid (aspirin)'
      Acetylsalicylic acid (aspirin) is in the same cross-reactivity group (NSAID) as the patient's allergy to Ibuprofen — possible cross-reactivity
--- ANSWER ---
No — Ciprofloxacin should not be given: The patient has a recorded allergy to Ciprofloxacin [436]. …
```

*Had the fix been wrong*, this probe would still show 2 chips and no mention of the recorded allergy —
the failure mode here is silence, which is exactly the shape that makes a chip-count probe vacuous.

### 4. Kevin Brown — the ledger key, live

`Is hydrocortisone safe for him?` (chart records allergies to **Hydrocortisone (ophthalmic)** AND
**Hydrocortisone (topical)** — two presentations of one substance — plus Levothyroxine, Levothyroxine
sodium, Clarithromycine, Tozinameran, Neomycin; one active `Clarithromycine Co 500mg`)

**main (88fdb5f):** 4 chips, the identity one naming a row the chart does not record
```
--- safetyWarnings (4) ---
  [contraindication] drug='Hydrocortisone'
      The patient has a recorded allergy to Hydrocortisone.
  [interaction] drug='Hydrocortisone'
      Hydrocortisone interacts with active order clarithromycin — Moderate. …
  [contraindication] drug='Hydrocortisone butyrate'
      Hydrocortisone butyrate is in the same ATC class (H02AB) as the patient's allergy to Hydrocortisone — possible cross-reactivity
  [contraindication] drug='Clarithromycin'
      The patient has a recorded allergy to Clarithromycin.
```

**this branch:** still 4 chips, and the two that name the allergen now name the record the chart holds
```
--- safetyWarnings (4) ---
  [contraindication] drug='Hydrocortisone (ophthalmic)'
      The patient has a recorded allergy to Hydrocortisone (ophthalmic).
  [interaction] drug='Hydrocortisone'
      Hydrocortisone interacts with active order clarithromycin — Moderate. …
  [contraindication] drug='Hydrocortisone butyrate'
      Hydrocortisone butyrate is in the same ATC class (H02AB) as the patient's allergy to Hydrocortisone (ophthalmic) — possible cross-reactivity
  [contraindication] drug='Clarithromycin'
      The patient has a recorded allergy to Clarithromycin.
```
The interaction chip still says `Hydrocortisone`, correctly: that arm names the substance's canonical
row, and this patient is not ordered a variant. His `Clarithromycine` allergy still resolves to
`Clarithromycin` through the unchanged fallback rank.

*Had the ledger key been left on the resolved row*, this patient would now read **6** chips — the
identity chip and the butyrate class chip once per hydrocortisone allergy record. That is the regression
the second half of this change exists to prevent, and this is the probe that would have caught it.

### 5. Michael Turner — a control that looked like a discriminator

`Is it safe to give enalaprilat?` (allergy to **Enalaprilat**, two active orders)

**main:** 1 chip, `The patient has a recorded allergy to Enalaprilat.` — already correct, because in the
shipped KB the `Enalaprilat` row (1142) precedes `Enalapril` (1882). The unit test for this shape uses a
verbatim slice that lists them the other way round; the KB does not, and the probe is what established
that. **this branch:** 1 chip, `The patient has a recorded allergy to Enalaprilat.` — unchanged.

### 6–12. The regression controls

| probe | expected | main (88fdb5f) | this branch |
|---|---|---|---|
| Mary Smith, clarithromycin | 1 Major ×simvastatin | 1 `Clarithromycin interacts with active order simvastatin — Major` | 1, identical |
| Agnes Adams, warfarin | 1 Major ×aspirin | 1 `Warfarin interacts with active order aspirin — Major` | 1, identical |
| Joshua Johnson, ibuprofen | 2 | NSAID group ×`Acetylsalicylic acid (aspirin)` + lisinopril Moderate | 2, identical |
| Mary Smith, paracetamol | 0 | 0 | 0, identical |
| Betty Williams, bupivacaine | exactly 1 | 1 `Bupivacaine interacts with active order lidocaine — Major` | 1, identical |
| George Anderson, esomeprazole | 2, both PPIs | `Omeprazole …allergy to Pantoprazole` + `Esomeprazole …allergy to Pantoprazole` | 2, identical (both still name `Pantoprazole`) |
| Sarah Taylor, hydrocortisone | 9, of which 4 named `Hydrocortisone` | 9, 4 named `Hydrocortisone` | 9, 4 named `Hydrocortisone`, identical |

George Anderson's control is the one that matters most here: it is the live twin of the
`Omeprazole`/`Esomeprazole` must-not-collapse case, and its allergen (`Pantoprazole`) goes through the
resolver this change rewrites. Sarah Taylor's is the one that would catch an over-merge, since her nine
chips include the identity chip and seven class chips over one allergen.

### 13. Susan Young — what this change does NOT fix

`Is it safe to give botulinum toxin type A?` (active order **Botulinum toxin type A**, plus Lidocaine,
Metoclopramide, Neomycin, Tiotropium; allergies Tiotropium and Tetryzoline (ophthalmic))

**main (88fdb5f):**
```
--- safetyWarnings (2) ---
  [interaction] drug='Daxibotulinumtoxina (botulinum toxin type a)'
      Daxibotulinumtoxina (botulinum toxin type a) interacts with active order neomycin — Major. …
  [contraindication] drug='Tiotropium'
      The patient has a recorded allergy to Tiotropium.
```
**this branch:** identical — 2 chips, the interaction one still subjected on
`Daxibotulinumtoxina (botulinum toxin type a)`. The interaction arm names the substance's
`canonicalRow`, and both botulinum rows answer `namesNoRoute()` true, so that fold keeps the KB's earlier
row whatever the resolver does. Same string as #176 reports, different code path; reported separately
rather than folded in.

## Build numbers

- baseline, `main` at 88fdb5f in a clean worktree, `mvn -o clean install`: **967 api + 61 omod = 1028**,
  0 failures, 34 skipped.
- this branch at `519ae80`, same command: **976 api + 61 omod = 1037**, 0 failures, 34 skipped. The
  +9 are this PR's nine cases; no existing test changed behaviour.
- querystore jar md5 `38205608891f10b68929c35fb3eace7e` (`querystore-api-1.0.0-20260806.153951-82.jar`,
  and the `-SNAPSHOT.jar` alias) — unchanged before and after every build here.

## One thing for the reviewer to rule on

CLAUDE.md's matcher rule assigns `DrugReference.isNamed()` to "IDENTITY between two reference strings (a
rule token against an entry's aliases)". The middle rank here calls it with a CHART string as one
operand, which is a use beyond that description. It is a **narrowing**, not a widening: the rank is
gated on `matchesDrugName` first, so `isNamed` can only choose among entries that already matched and can
never make a match happen — and `isNamed`'s own logic is untouched, including its deliberate refusal to
fold diacritics. Recorded at `isNamed` in javadoc rather than left implicit. If that reading of the rule
is wrong, the alternative is a fourth accessor with the same body, which the same rule argues against.

## Limitations, honestly

- **The exact ranks do not fold diacritics.** They compare through `DrugReference.normalizeName`, which
  deliberately does not fold (its javadoc says why), so an accented chart string reaches no exact rank
  and falls through to the folded matcher — exactly what it did before. The KB carries no combining
  mark on either side, so the gap is on the chart side only, and widening it needs its own measurement
  over a localized dictionary.
- **A tie is still resolved by dataset order.** Two rows publishing one display name (#164's shape)
  resolve to whichever comes first; no shipped-KB entry is affected.
- **A combination-product allergen still resolves to one constituent** — and the change moves *which*
  one (`sulfamethoxazole / trimethoprim`: `Sulfamethoxazole` → `Trimethoprim`, because the Trimethoprim
  row claims the whole combination name). The other constituent is not checked. Reported separately.
- **Where the KB's own alias data is wrong, the rank now obeys it.** `Monopotassium phosphate` publishes
  `rxnorm_name = potassium lactate`, so an allergy recorded as `potassium lactate` moves from
  `Lactic acid` to `Monopotassium phosphate`; neither is the drug. 2 of the 59 alias-rank moves are of
  this shape. Reported separately as an upstream data defect.
- **The interaction arm's subject is unchanged and still not the row the patient is on.** An order for
  `Botulinum toxin type A` is subjected on `Daxibotulinumtoxina (botulinum toxin type a)`, because that
  arm names the substance's `canonicalRow` and both rows name no route. That is the same string #176
  reports, reached by a different route; it is out of this change's scope and reported.

## One existing test comment corrected, no assertion touched

`DirectAllergyContraindicationTest.anUnclassifiedAllergenWithASiblingRouteVariantStillWarnsOnce`
carried a comment stating that regenerating its slice in KB order would flip its detail assertion —
true under earliest-match, false now. The comment is updated; the assertions, the fixture and the
slice's ordering are untouched. No existing assertion contradicts this change: all 967 tests on `main`
still pass unmodified.

Closes #176


dkayiwa:
## Hardening pass — fresh eyes, `519ae80` → `55ceb38`

Six commits: **one added test**, prose corrections, and a test-only polish pass. **No production
behaviour moves** — every production edit is javadoc or comment, verified by diffing with comment lines
excluded (the only non-comment line is markdown in `docs/adr.md`).

**Build, measured by me in clean worktrees.** `main` at `88fdb5f`: **967 api + 61 omod = 1028**, 0
failures, 34 skipped. This branch at `55ceb38`: **977 + 61 = 1038**, 0 failures, 34 skipped (`519ae80`
was 976 + 61; +1 is the tie test below). querystore jar md5 `38205608891f10b68929c35fb3eace7e` before
and after every build.

**One reachability fact worth recording**, because it bounds the regression risk: both halves of this
change are inert unless `sourceFormat=ddinter` AND `dataFilePath` points at the full 19 MB KB. The
bundled curated `json` file is 4 entries with disjoint aliases (nothing to misresolve), and only
`DdiDrugReferenceSource` ever sets a substance name, so on `json`/`atc` the ledger's finding key is the
row again and `substanceGroupKey()` comparison is reference identity — sound there, since one row is one
substance for those sources. A stock install cannot change behaviour. That is also the only
configuration the live probe below can exercise, which is what it ran on.

### Every headline figure re-derived independently, through production code

Driven from a throwaway JUnit case calling `DdiDrugReferenceSource.parse`,
`DrugReferenceService.lookupByToken`, `DrugReference.nameMatchStrength`, `substanceKey`,
`atcSubgroups`, `CrossReactivityGroup.groupsOf` and the real `DrugSafetyValidator.validate` over the
standalone's 19 MB KB — never a reimplementation. Same harness dropped into a `main` tree and a branch
tree and the two dumps diffed, so "before" is production `main` code.

| | PR says | I measured |
|---|---|---|
| entries not resolving to themselves, on `main` | 206 | **206** |
| … of which a different substance | 54 | **54** |
| … of which another row of the same substance | 152 | **152** |
| after the fix | 0 of 2283 | **0 of 2283**, 0 different substance |
| distinct published name strings | 7452 | **7452** |
| names that stop resolving | 0 | **0** (and 0 start) |
| names resolving differently | 471 | **471** |
| … decided by the display-name rank | 412 | **412** (108 to a different substance) |
| … decided by the alias rank | 59, all to a different substance | **59, all 59 to a different substance** |
| ATC-less entries naming themselves | all 444 | **444 of 444** |
| multi-row substance families | 129 | **129** |

`Daxibotulinumtoxina`'s row is *named* `Daxibotulinumtoxina`, carries `botulinum toxin type a` as an
alias, and `displayLabel()` renders `Daxibotulinumtoxina (botulinum toxin type a)`. **Your correction to
#176 is right: there is no parenthesised qualifier in the data**, so a parenthesis-ranking rule cannot
fire on the headline case. Both rows also answer `namesNoRoute()` true and carry no ATC code, so
`canonicalRow` keeps the wrong one — confirmed.

**Earliest-match as the fallback is justified.** After the two exact ranks, **0 of 2283** entries fail to
resolve to themselves, so nothing is left for a length heuristic. I also built the rejected
longest-alias alternative on top of the production matcher: it repairs **43** of the 54 different-substance
cases (your javadoc says 43 of the ATC-less 53 and the PR body says 44 of the 54 — different populations,
and my number is for the 54) — and it **breaks 34 entries that earliest-match already resolves correctly**,
which the rejection does not currently mention and which strengthens it.

### Verdict on the 59 alias-rank moves: sound

I adjudicated all 59 one by one, with the ATC level-4 subgroups and curated-group membership of the old
and the new row measured through production for each. Class reach: **28 gain, 17 change, 7 unchanged, 7
lose**. Not one of the 59 makes the identity claim less true. The moves that matter most:

- `penicillin g, procaine`: `Procaine` → `Benzylpenicillin`. `main` reported a **penicillin** allergy as
  an allergy to the local anaesthetic.
- `digoxin antibodies fab fragments`: `Digoxin` → `Digoxin Immune Fab (Ovine)`. `main` said "the patient
  has a recorded allergy to Digoxin" for a patient allergic only to its **antidote** — a false
  contraindication on a narrow-therapeutic-index drug, now withdrawn.
- `ethinyl estradiol` and eight oral-contraceptive combinations: off `Fluoroestradiol f-18`, a **PET
  radiotracer**, onto the actual constituent.
- `deamino arginine vasopressin` → `Desmopressin`; `cytidine diphosphate choline` → `Citicoline` (off a
  salicylate); `selenium disulfide` → `Selenium Sulfide`; `17-alpha-hydroxy-progesterone caproate` →
  `Hydroxyprogesterone caproate`; nine `X lactate` salts off `Lactic acid`.

**Two of the 59 make the class arm clinically weaker**, both combination products, i.e. #193's shape:

- `sulfamethoxazole / trimethoprim`: `Sulfamethoxazole` (J01EB) → `Trimethoprim` (J01EA). The sulfa
  moiety is the clinically dominant allergen of co-trimoxazole, so this loses the more useful
  cross-reactivity, not the less. Worth adding to #193 — the PR body notes the move but not its direction.
- `omeprazole / sodium bicarbonate`: `Omeprazole` (A02BC) → `Sodium bicarbonate` (B05CB/B05XA): the PPI
  class is lost.

Two land on a row that is not the drug either way, from the KB's own alias data (`potassium lactate` →
`Monopotassium phosphate`; `choline / inositol / lysine / methionine / vitamin b12` → `Choline
salicylate`). I checked whether the second manufactures a false salicylate hedge — it does not, that row
carries no ATC code in this KB.

### The substance key cannot merge two findings that should stay separate

- **0 of the 129** multi-row families publish differing ATC level-4 subgroups; **0** differ in
  curated-group membership. The ledger's own javadoc says to re-measure that before widening the key, and
  this PR widened it, so I did — the surviving chip therefore carries the same relationship and the same
  code the suppressed one would have named.
- Driving the real `validate` over **all 129** families with both rows recorded as allergies: **exactly 1
  identity chip in every one**, never 0, never 2.
- The only thing lost is the second record's label, and only the label.

The converse — is the second half load-bearing? — I swept it: reverting the finding key to the resolved
row makes **all 129 of 129** families report **2** identity chips instead of 1, and fails your
`twoRecordedAllergiesToOneSubstanceStillRaiseOneChipPerSubject` at 4 chips against 2. Confirms 2 → 4 → 2
and generalises it.

### What I added: the tie rule was untested

Reverting `strength > strongest` to `>=` — so a tie keeps the **last** equal claimant — passed all 976
tests. It is not a corner: **1367 of the 7452** published names are claimed equally by two or more entries
at the strongest rank they reach, all at the alias rank (**0** at the display-name rank, since no shipped
display name is filed twice). **1125 of the 1367** are `/`-joined combination names claimed by more than
one constituent (`abacavir / lamivudine`); the other 242 are single names two entries both publish
(`ketorolac tromethamine`). `twoEntriesMakingTheSameStrongestClaimResolveToTheEarliestOfThem` pins it
through the real resolver and the real `validate`, and fails under that mutation.

A mutation matrix over all eight decisions in the slice: **seven are caught by tests** (tie rule, display
rank, alias rank, rank inversion, ledger key = row, ledger key = `substanceKey()` → 23 NPEs, identity by
reference → 2 failures, which is the strongest available answer on #172's trap). The eighth — the rank-2
early return — survives with 0 differences, which is correct: it is a pure exit. Its comment claimed it
"keeps the common case as cheap as the first-match scan was", and the same javadoc calls a name no entry
is named "the ordinary shape"; that shape has no early return, so I narrowed the claim.

### Statements the change had made false, now corrected

`addAllergyContraindications` and `addActiveOrderContraindications` both still said the bound was
"(substance, **resolved allergen**)". `ContraindicationRouteVariantTest` said two allergies about *one*
substance must stay two chips — which is now exactly what collapses — and attributed the resolution to
"`lookupByToken`'s earliest-match rule". `docs/adr.md` said the dedup is per resolved allergen and that
the identity check "compares two object references", untrue since #164. `AllergenNameResolutionTest`
carried a which-constituent count measured under the old resolver; removed rather than restated
unmeasured. One correction of mine was itself wrong on a second reading (attributing aliases-warn-once to
the substance key, when that half is `seenAllergens`, keyed on the row) — removed rather than re-worded.

**Not touched: `CLAUDE.md`.** Its matcher rule now under-describes `isNamed` (the middle rank passes a
chart string as the second operand) and has no entry for `nameMatchStrength`. You said you would add that
after this lands.

### One claim in the PR body I would soften

> `substanceGroupKey().equals(...)` … is value equality over the substance, so a changed resolution cannot
> silently flip an identity comparison.

True only where the source publishes a substance name. `substanceGroupKey()` returns `this` otherwise, and
only `DdiDrugReferenceSource` calls `setSubstanceName`, so on `sourceFormat=json` (the production
**default**) and on `atc` that `.equals` is reference identity again. It is sound there — one row is one
substance for those sources, and rows come from the shared `getAll()` cache — and `substanceGroupKey`'s
javadoc already says so; the PR body just states it too broadly. The bundled curated dataset has 4 entries
with disjoint names, so nothing to misresolve and nothing to over-merge.

### Live re-verification — both builds, same instance, DB-resolved patients

`main` at `88fdb5f` deployed and probed, then `84618d9`; full-JVM restart each time;
`{"sourceFormat":"ddinter","entryCount":2283,"origin":"appdata:chartsearchai/ddi_knowledge_base.json"}`
both times. Every patient resolved **by name against MariaDB on 3316**, and every allergy and active order
read from the DB, not the order REST list. Captures appended to one log per build so an overwritten
`out-<label>.json` cannot be quoted as the wrong build. **Exactly 4 of 13 probes move; all 9 controls are
byte-identical in chip text.**

**Margaret King** (DB: one allergy, `Botulinum toxin type A`, concept 4259; one Dexamethasone order)

```
main    1 chip  drug='Daxibotulinumtoxina (botulinum toxin type a)'
                The patient has a recorded allergy to Daxibotulinumtoxina (botulinum toxin type a).
        ANSWER  No — botulinum toxin type A should not be given: the patient has a recorded allergy
                to Daxibotulinumtoxina [119], [122].
branch  1 chip  drug='Botulinum toxin type A'
                The patient has a recorded allergy to Botulinum toxin type A.
        ANSWER  No — botulinum toxin type A should not be given: The patient has a recorded allergy
                to Botulinum toxin type A [122].
```
Had the fix been wrong: the same 1 chip, still naming `Daxibotulinumtoxina`, in the answer as well as the
chip.

**Melissa Wright** (DB: one allergy, `Esomeprazole`; one Spironolactone order, id 337) — asked about
**omeprazole**, which sharpens it: `main` does not merely mislabel, it upgrades a hedge to a certainty.

```
main    1 chip  The patient has a recorded allergy to Omeprazole.
branch  1 chip  Omeprazole is in the same ATC class (A02BC) as the patient's allergy to Esomeprazole
                — possible cross-reactivity
```
Had the fix been wrong: **the same count**, still asserting a recorded allergy to a substance her chart
does not name. This probe can only be read on the sentence.

**Dorothy Odinga** (DB: allergies `Ibuprofene` + `Ciprofloxacin lactate`; active `Aspirine Co 81mg`)

```
main    2 chips  Ciprofloxacin interacts with active order aspirin — Moderate. …
                 Acetylsalicylic acid (aspirin) is in the same cross-reactivity group (NSAID) as the
                 patient's allergy to Ibuprofen — possible cross-reactivity
                 → neither is about the recorded ciprofloxacin allergy, and the ANSWER never mentions it
branch  3 chips  The patient has a recorded allergy to Ciprofloxacin.        ← recovered, and leads
                 Ciprofloxacin interacts with active order aspirin — Moderate. …
                 Acetylsalicylic acid (aspirin) is in the same cross-reactivity group (NSAID) …
        ANSWER   No — Ciprofloxacin should not be given: The patient has a recorded allergy to
                 Ciprofloxacin [436]. Furthermore, …
```
Had the fix been wrong: 2 chips and silence about the allergy. `ciprofloxacin lactate` resolved to
`Lactic acid`, which has no ATC code, so there was nothing to compare — the wrong answer here is silence.

**Kevin Brown** — the ledger-key probe. DB: 7 allergies, including `Hydrocortisone (ophthalmic)`
(allergy_id **27**) and `Hydrocortisone (topical)` (**28**); active `Clarithromycine Co 500mg`.

```
main    4 chips  The patient has a recorded allergy to Hydrocortisone.            ← a row neither
                 Hydrocortisone interacts with active order clarithromycin — Moderate. …    record names
                 Hydrocortisone butyrate is in the same ATC class (H02AB) as the patient's allergy
                     to Hydrocortisone — possible cross-reactivity
                 The patient has a recorded allergy to Clarithromycin.
branch  4 chips  The patient has a recorded allergy to Hydrocortisone (ophthalmic).
                 Hydrocortisone interacts with active order clarithromycin — Moderate. …
                 Hydrocortisone butyrate is in the same ATC class (H02AB) as the patient's allergy
                     to Hydrocortisone (ophthalmic) — possible cross-reactivity
                 The patient has a recorded allergy to Clarithromycin.
```
4 → 4, both allergen-naming chips corrected. **Without the ledger key this reads 6.** The chip names
allergy_id 27 because `PatientClinicalContextBuilder` collects allergies into a `LinkedHashSet` in
`getAllergies` order and the ledger keeps the incumbent — so it is the **first-recorded** allergy, not an
arbitrary one. I traced that before probing and the live result matched.

**Controls, byte-identical in chip text before and after** — Mary+clarithromycin 1 Major ×simvastatin;
Agnes+warfarin 1 Major ×aspirin; Joshua+ibuprofen 2; Mary+paracetamol 0; Betty+bupivacaine 1;
George+esomeprazole 2 both naming `Pantoprazole`; **Sarah Taylor 9**; Michael Turner 1 `Enalaprilat`;
Susan Young 2. Sarah's is the over-merge canary — 9 chips including the identity chip and seven class
chips over one allergen, unchanged.

DB reads confirmed every line of the corpus I depended on: Betty 2 active orders and 2 allergies, Sarah 8
orders, Melissa 1, Kevin 1 active `Clarithromycine Co 500mg` and 7 allergies, Susan 5 orders including
`Botulinum toxin type A` and `Neomycin`, George 0 orders.

### Found, not fixed

1. **Resolving correctly can trade a mislabel for silence** — recorded in the javadoc, filed separately.
   Over the 471 moved names, 40 findings arrive where the arm was silent and 120 go quiet; nearly all of
   those 120 are the false claim being withdrawn, but where the KB files a presentation as a **separate
   substance** from its parent moiety **and** gives that row no ATC code, there is nothing left to
   compare: `Insulin lispro (protamine)` asked about `Insulin lispro`, `Insulin human (isophane)` about
   `Insulin human`, `Dextran (low molecular weight)` about `(high molecular weight)`, `Peanut oil` about
   `Peanut`, four iron salts about `Iron (bisglycinate)`, `Procaine benzylpenicillin` about `Procaine`.
   #135's ATC-less gap reached through a *correct* resolution. Measured.
2. **`Fluoroestradiol f-18` carries estradiol's own `rxnorm_name` AND `drugbank_id`** (`substanceKey =
   [estradiol, db00783]`), so the module treats a PET radiotracer and a therapeutic oestrogen as **one
   substance** — the same upstream openmrs-ddi-knowledge-base defect class as `Monopotassium
   phosphate`/`potassium lactate`. This PR does not make it worse (both tokens already collapsed under
   earliest-match), but it is the one shipped case where the substance key can merge two clinically
   distinct recorded allergies. Measured.
3. **#193's blast radius is larger than "≥20 CIEL names"**: 1367 of the 7452 published names are claimed
   equally by two or more entries at their strongest rank, overwhelmingly combination products. Measured.
4. **`eval/drift-metric/score_probe_safety.py:196-201`** states the allergy arm as `allergen == ref` and
   claims that for a drug with no ATC subgroup "even a direct allergy raises nothing". Both are false —
   the first since #164/#187, the second since #135 (all 444 ATC-less entries raise a chip; I measured
   it). Stale documentation in an eval scorer, pre-existing, outside this slice.
5. **`sulfamethoxazole / trimethoprim` moves to the clinically weaker moiety** — a note for #193, above.

### Polish applied, and what I left alone

The new test file now uses the package's own conventions: a `fixtureValidator` delegate instead of six
spelled-out arrangements, the two fixture paths hoisted into `DrugReferenceTestSupport` (which documents
that rule for paths used by more than one file — this PR made both of them multi-file) and now read by all
three files that use them, and one helper for a premise written twice. That helper was first called
`assertNoRowIsNamed`, which read as a contradiction of the `isNamed` assertion two lines below it — it
compares display names, so it is now `assertNoRowsDisplayNameIs`. **Left alone:** the private `row(entries, name)` helper, which is a third
byte-identical copy of one `SubstanceIdentityTest` and `SelfInteractionTest` already carry — hoisting only
this file's copy would leave three definitions, and converting theirs is an unrelated refactor of files
outside this slice. Also left: memoising `lookupByToken` per `validate` pass. The redundancy is
pre-existing (`main` re-resolved per subject too) and I measured the cost as immaterial — 18ms → 29ms for
a whole `validate` with 7 allergies against an answer that takes 15–90s, twice per request.

I was tempted by nothing in the load-bearing prose and reversed no decision. The rejected alternatives all
hold up under measurement: longest-alias is worse than the rejection says, `canonicalRow` keeps the wrong
botulinum row (both answer `namesNoRoute()` true — verified), the parenthesis rule cannot fire because the
parentheses are not in the data, and earliest-match must stay because 0 of 2283 is left for a heuristic.

dkayiwa:
### Addendum — three more passes after the summary above, and where I stopped

`55ceb38` → **`3bf01c0`**. Still no production behaviour change: `git diff 990bdf7..HEAD -- api/src/main`
with comment lines excluded is **empty**, so the classes I probed live are behaviourally the ones on the
branch head. Build unchanged at **977 api + 61 omod = 1038**, 0 failures, 34 skipped; CI green; querystore
jar md5 `38205608891f10b68929c35fb3eace7e`.

**One real defect found late, and it was on this PR's own axis.** The weakest rank's DIRECTION was stated
backwards at six sites. `matchesDrugName` → `matchesOrderName` → `containsBoundedToken(text, token)` and
that last does `text.indexOf(token)`, with `text` = the **recorded name** and `token` = the **entry's
alias**. So the entry's own name occurs inside the recorded name — never the reverse. Yet
`NAME_TOKEN_INSIDE_A_NAME`'s javadoc said "`drugName` occurs INSIDE one of this entry's names",
`PatientClinicalContext` said "one whose alias it merely occurs inside", and the new test had it backwards
in a fixture comment, an assertion message and two method names. The new fixture's own `metadata.note` had
it **right**, so the code and its own fixture contradicted each other on the one relation #176 is about.
It is also what makes the headline alias case legible: `Ciprofloxacin lactate` reaches the `Lactic acid`
row because that row's CIEL name `Lactate` sits inside the recorded string. Fixed at all six, and the
javadoc now names the call chain so the next reader can check the direction rather than trust it.

**Where I stopped, and why.** Code and tests have been stable since my first hardening commit; the last
three commits are prose only, and each of the last three passes found a falsehood **in my own previous
correction** — all of them in one ADR sentence about which mechanism deduplicates what. That sentence now
states no mechanism at all and points at `seenAllergens` and `ContraindicationChips` instead, because the
two collapses live in two places and every attempt to summarise them assigned one to the other. Per the
project's own rule about this failure mode, that is the point to stop writing prose rather than write
more. So: **converged on code and tests, and one ADR sentence deliberately reduced to a pointer.**

Polish also finished the fixture-path hoist (`SubstanceIdentityTest` and `ContraindicationRouteVariantTest`
now read the shared constants, so "a rename breaks in one place" is true rather than merely claimed), and
renamed `assertNoRowIsNamed` → `assertNoRowsDisplayNameIs`, which had read as a contradiction of the
`isNamed` assertion two lines below it.

**Still deferred, with the consequence stated rather than a severity label:**
- the private `row(entries, name)` helper is a third byte-identical copy of one `SubstanceIdentityTest` and
  `SelfInteractionTest` already carry. *If we ship without hoisting it, nothing breaks; the cost is that a
  fourth test copies it a fourth time, and its failure message already delegates to
  `DrugReferenceTestSupport.names`, so improving that message means editing three places.* Hoisting only my
  copy would leave three definitions, and converting theirs is an unrelated refactor of two files outside
  this slice.
- memoising `lookupByToken` per `validate` pass. *If we ship without it, nothing breaks: measured 18ms →
  29ms for a whole `validate` with 7 allergies over the 19 MB KB, twice per request, against an answer that
  takes 15–90s.* The redundancy is pre-existing — `main` re-resolved per subject too.
- `docs/adr.md:2088` still says a decision "is pending, not yet merged" for #173, which merged. *If we ship
  without fixing it, a reader of the ADR believes #173 is unmerged and may re-decide something already
  settled.* Pre-existing and unrelated to name resolution, so reported rather than fixed.

dkayiwa:
Independent pre-merge verification (mine, not relayed): `mvn -o clean install` on a clean detached worktree at `3bf01c0` → **977 + 61 = 1038, 0 failures, 34 skipped**, BUILD SUCCESS. `main` at `88fdb5f` = 967 + 61 = 1028. CI green on Java 11/17/21.

I checked the two claims that mattered most and both hold: **hardening changed no production behaviour** — the `src/main` diff from `519ae80` with comment lines stripped is empty — and **zero non-comment assertion lines were removed anywhere on the branch**. The one pre-existing test file touched, `DirectAllergyContraindicationTest`, is comment-only; the `assert` matches in its diff are the words "count assertion" and "detail assertion" inside prose.

**The hardening delivered both verdicts I asked for, with measurement rather than reasoning.**

On the **59 alias-rank moves**, each adjudicated with both rows' ATC subgroups measured: class reach 28 gain / 17 change / 7 same / 7 lose, and **none makes the identity claim less true**. The clearest justification for the whole PR is in that set — `digoxin antibodies fab fragments` came off `Digoxin`, meaning `main` was telling clinicians a patient allergic to digoxin's **antidote** was allergic to digoxin. It also measured the rejected alternative properly: longest-alias repairs 43 of the 54 **and breaks 34 that earliest-match gets right**.

On the **substance key**, it carried out the ledger's own "re-measure before widening" instruction: 0 of 129 shipped families diverge on ATC level-4 subgroups or curated-group membership, and real `validate` over all 129 with both rows recorded gives exactly one identity chip each — never 0, never 2. Reverting the key makes **129 of 129** report twice, which confirms the 2 → 4 → 2 measurement from the fix pass.

It was also honest about Phase 2: code and tests converged after Phase 1 pass 1 and never moved substantively again, while the last three passes each found a falsehood in **its own** previous correction — all in one ADR sentence, now reduced to a pointer. That is exactly the §5 stop signal, and stating it beats claiming a convergence it did not observe.

Filed from this pass: **#195** (correct resolution can trade a mislabel for silence — 120 of 471 moved names go quiet, and where the KB files a presentation as a separate substance from its parent moiety with no ATC code, the chip vanishes; #135's gap reached through a *correct* resolution) and **#196** (three upstream KB rows publishing a name or registry id for a different substance — worst is `Fluoroestradiol f-18` carrying estradiol's own `drugbank_id`, so a PET tracer and a therapeutic oestrogen are one substance to every consumer of that registry). **#193** was updated with a much larger measured scope — 1367 of 7452 published names are claimed by two or more entries, 1125 of them `/`-joined combinations — plus the two combination allergens this PR makes measurably weaker, which makes #193 a completion of this work rather than an independent improvement. **#179** gained a fourth false instrument.

One follow-up I own: CLAUDE.md needs an API-surface entry for `nameMatchStrength`. #191 merged the other four; I held this one back to avoid a conflict and will add it now that this lands.

Issue #196, opened 2026-08-08
Upstream: three shipped KB rows publish a name or registry id denoting a different substance


dkayiwa:
## What

Data defects in the shipped `ddi_knowledge_base.json`, not in this module's code. Filed here because they cause clinician-visible wrongness through correct chartsearchai logic, and because the fix belongs in **openmrs-ddi-knowledge-base**.

1. **`Monopotassium phosphate` carries `rxnorm_name = potassium lactate`** — two different substances.
2. **`Choline salicylate` and `Liver derivative complex` claim multivitamin CIEL names** they are not.
3. **`Fluoroestradiol f-18` carries estradiol's own `rxnorm_name` *and* `drugbank_id`** — so a PET imaging tracer and a therapeutic oestrogen are, to every consumer of this data, one substance.

## Why item 3 is the worst of the three

Since #187 this module decides substance identity by `drugbank_id` resolved per substance-name family — deliberately, because it is the KB's own substance-level registry and it correctly separates `Omeprazole` from `Esomeprazole` where `rxcui` and `rxnorm_name` do not. Item 3 means that registry is wrong for this pair, so **no consumer can distinguish them**, and a patient given a diagnostic tracer can be reasoned about as if given a hormone therapy.

No amount of care in this module can recover from a registry id that identifies the wrong substance. That is what makes it upstream rather than a matching problem.

## Scope

Small in count and neither introduced nor worsened by any change here. Worth fixing upstream because the data is now load-bearing in a way it was not when it was only used for pair lookups: #187 promoted `drugbank_id` to the identity criterion, and #192 promoted the published names to a resolution ranking.

## Provenance

Items 1 and 2 measured by #192's fix pass; item 3 by its hardening pass, both through production code over the shipped 19 MB KB. Relayed; I have not re-derived them.

dkayiwa:
**A fourth instance, and it is visible to clinicians on every chip** — measured during #198's work.

`Sulfamethoxazole` (`DDInter1724`) publishes `rxnorm_name: "sulfamethazine"` and `drugbank_id: DB01015`. **Sulfamethazine is a different sulfonamide.** Because `displayLabel()` renders the `rxnorm_name` in parentheses, every chip about sulfamethoxazole reads:

```
Sulfamethoxazole (sulfamethazine)
```

Captured verbatim in a live probe, so this is not a latent data concern — a clinician asking about sulfamethoxazole is shown a different drug's name beside it, on the shipped default configuration.

[...]

PR #198, opened 2026-08-09
fix(drug-safety): resolve a recorded allergy to every substance its name denotes (#193, #195)

### LINKED ISSUE #193: A combination-product allergen is checked against only one of its constituents
## What

An allergen recorded as a combination product resolves to a **single** reference entry, so only one constituent is ever checked against the knowledge base. The rest of the combination is silently unchecked.

Measured: `Sulfamethoxazole / trimethoprim` resolves to **`Trimethoprim` alone**, so nothing the patient takes is ever compared against sulfamethoxazole. At least **20 combination CIEL names** are in this shape.

## Why this is the serious kind

It fails **silently and closed** — no chip, no warning, no log line. A clinician sees a clean safety check on a patient whose recorded allergy was half-examined. That is the same failure mode as #86 (a Major bleeding interaction that never fired), which is what began this whole body of work, and it is the class that live probing catches only if you happen to construct the case.

Note it is not fixed by #192's exact-name ranking: that fixes *which* entry a name resolves to, and here the problem is that resolution returns **one** entry where the clinical fact names two substances.

## Shape of the fix

`DrugReferenceService.findByDrugName` already returns a collection rather than a single entry, and #192 moved the contraindication ledger's allergy key onto `allergen.substanceGroupKey()` — so the machinery to raise one finding per constituent without duplicating chips now exists. The pieces are:

1. split a combination allergen into its constituents (the `/` separator is the observed shape, but confirm what CIEL actually publishes),
2. check each against the KB, and
3. key the resulting findings so two constituents of one recorded allergy do not produce two chips for the same clinical fact.

Step 3 is where the care goes — #192 measured that a more precise resolver *without* the substance-keyed ledger turned 2 chips into 4.

## Related

- **#190 item 2** is adjacent but a different axis: there, one entry is reached by two names (a synonym problem). Here, one recorded name denotes two entries.
- **#176/#192** established the resolution ranking this builds on.

## Provenance

Measured by #192's fix pass and reported rather than fixed; the `Sulfamethoxazole / trimethoprim` resolution and the ≥20 count are its measurements over the shipped KB and the CIEL dictionary. I have not re-derived them.

### LINKED ISSUE #195: A correct name resolution can trade a mislabelled chip for no chip at all (120 of 471 moved names go quiet)
## What

#176/#192 replaced earliest-match resolution with a rank on the *claim*, which fixes 54 substance-level mis-resolutions. A side effect, measured during its hardening: of the 471 published names whose resolution moves, **120 now produce no chip at all.**

Nearly all 120 are false claims correctly withdrawn — that is the fix working. But a residue is a genuine loss, and it has a precise shape: **where the KB files a presentation as a separate substance from its parent moiety *and* gives that presentation no ATC code**, resolution now lands on the presentation, the class arm has nothing to compare, and the chip vanishes.

Measured examples:
- `Insulin lispro (protamine)` versus `Insulin lispro`
- `Insulin human (isophane)`
- `Dextran (low MW)`
- `Peanut oil` versus `Peanut`
- four iron salts

## Why this deserves its own issue

It is **#135's gap reached through a *correct* resolution** — the class of defect where a patient with a real recorded allergy gets silence. #135 was about a drug carrying no ATC codes at all; this is about resolution correctly preferring a row that happens to carry none, when its parent moiety does.

Before #192 these cases produced a chip with the *wrong* substance named. After, they produce nothing. A wrong label is bad; silence on a recorded allergy is worse, and it is the failure mode this entire body of work started from (#86).

So this is not an argument against #192 — its 54 substance-level fixes include a patient allergic to **digoxin's antidote** being told they are allergic to digoxin. It is the remaining tail.

## Shape of the fix

Widen `substanceKey` so a presentation and its parent moiety resolve to one substance, which is what they clinically are. The machinery exists: #187 established substance identity by `drugbank_id` per substance-name family, and the recorded loss here is precisely where the KB's own registry ids split a presentation from its parent. The constraint is #187's — the widening must not merge two genuinely distinct substances, and `Omeprazole`/`Esomeprazole` remains the case any candidate must keep apart.

Recorded in `lookupByToken`'s javadoc so it is not rediscovered.

## Provenance

Measured by #192's hardening pass through production code and reported rather than fixed. I have not re-derived the 120, nor the individual examples.

dkayiwa:
# The shared root cause

#176/#192 fixed **which** entry a recorded drug name resolves to. It still resolves to exactly **one**,
and the set of substances a recorded name implies is larger than one in two different ways.

**#193 — one recorded name denotes several substances.** A combination product's name denotes each of
its ingredients, so one was compared and the rest were never checked. Measured through
`DrugReferenceService.lookupByToken` over the shipped 19 MB KB (2026-08-09; my own measurement, method
below): `sulfamethoxazole / trimethoprim` answers `Trimethoprim` — the combination is one of that row's
own CIEL names (claim rank `NAME_IS_ANOTHER_NAME`) while `Sulfamethoxazole` merely occurs inside it
(`NAME_TOKEN_INSIDE_A_NAME`) — so nothing is ever compared against the sulfa moiety that actually drives
that allergy: `J01EA` in place of `J01EB`. `omeprazole / sodium bicarbonate` answers
`Sodium bicarbonate` (`B05CB`/`B05XA`), losing the PPI class `A02BC` entirely.

**#195 — a presentation is filed as a separate substance from its parent moiety.** Where the KB does
that *and* gives the presentation no ATC code, #192's correct resolution lands on the presentation, the
class arm has nothing to compare, and the chip vanishes rather than being mislabelled:
`Insulin lispro (protamine)`, `Insulin human (isophane)`, `Iron (polysaccharide)`.

Both fail **silently and closed** on a patient with a recorded allergy — the #86 class.

# What changed

`DrugReferenceService.findImpliedSubstances(String)` is the one resolution, and it is the only new
production entry point. It returns one representative row per implied substance:

1. **`lookupByToken`'s answer, always first** — so for a name implying one substance every existing
   chip label is byte-identical to before;
2. **every other entry making the same strongest claim**, when that claim is a NAME claim
   (`NAME_IS_ANOTHER_NAME` or stronger). Deliberately *not* extended to the containment rank, which is
   where #192's hazard lives;
3. **each `/`-joined constituent**, when the KB is NAMED it. Any NAME rank is enough because the
   recorded string *asserts* the constituent is an ingredient;
4. **the parent moiety** (the name with its trailing parenthesized qualifier removed), when an entry is
   **CALLED** it (`NAME_IS_THE_DISPLAY_NAME`). Stricter than 3 on purpose — a moiety is a derivation,
   not a claim, so a row that merely lists the stem among its aliases is a different presentation and
   naming it in a chip would report an allergy to a drug the chart does not record.

`nameMatchStrength`'s ranking is untouched, and so is `substanceKey`. The widening is in what the
recorded **string** is read to name, not in what the reference data calls one substance.

# One recorded name yielding N substances still raises one chip per clinical fact

`DrugSafetyValidator.addAllergyContraindications` now decides precedence **across the implied set**
rather than a member at a time — identity over the whole set first, then the class comparison over the
whole set, then the curated group — and **each arm stops at its first match**. So:

- a patient allergic to `abacavir / lamivudine` asked about **lamivudine** is told they are allergic to
  it, rather than that it cross-reacts with the abacavir in the same tablet (a per-member scan reaches
  the earlier claimant first and emits the weaker, wrong chip);
- a patient allergic to `abacavir / lamivudine` asked about **zidovudine**, which shares `J05AF` with
  **both** constituents, gets **one** chip, not two.

The wider collapse — two allergy *records* naming one substance — is unchanged: it is still the
`ContraindicationChips` ledger keyed on `(subject substanceGroupKey, allergen substanceGroupKey)`.
Kevin Brown's four chips are that key, and they are unchanged live.

The allergens are also resolved **once per `validate`** now rather than once per subject (they never
depended on the subject), as a per-`validate` local — #172's trap.

# The re-derived family-divergence measurement

#192's hardening established **0 of 129** families diverging on ATC level-4 subgroups or on
curated-group membership. That figure had to be re-derived because a widened `substanceKey` would change
which families exist.

**It does not: `substanceKey` is not widened by this change.** Re-derived anyway on this branch, through
`DrugReference.substanceGroupKey()`, `atcSubgroups()` and `CrossReactivityGroup.groupsOf` over the
shipped KB (2026-08-09):

```
FAMILIES substances=2114 filedAsMoreThanOneRow=129 divergingSubgroups=0 divergingGroups=0
         familiesWithNoUnqualifiedRow=10
```

**129 families, 0 diverging on subgroups, 0 on curated-group membership** — measured, not inherited.

Because the widening is in the resolution rather than in substance identity, the constraint the javadoc
demands is met by construction rather than by luck: `Omeprazole`/`Esomeprazole` keep different
`substanceKey`s, and `Fluoroestradiol f-18`/`Estradiol` are untouched (their KB merge is #196's data
defect and is unaffected either way).

# How the KB was measured

Every figure below is my own, taken on 2026-08-09 by compiling against this branch and calling the
**production predicates** through a throwaway JUnit case over the shipped 19 MB KB
(`appdata/chartsearchai/ddi_knowledge_base.json`, md5 `ba7b2721a75aac707ce452798b74c074`, parsed by
`DdiDrugReferenceSource.parse`). No re-implementation of `substanceKey`, `nameMatchStrength` or
`canonicalRow` was used anywhere.

| figure | producing method | value |
|---|---|---|
| entries | `DdiDrugReferenceSource.parse` | 2283 |
| distinct published names | `normalizeName` over `getAliases()` | 5169 |
| (entry, alias) pairs | `getAliases().size()` summed | 8300 |
| names claimed equally by ≥2 entries at the strongest rank | `nameMatchStrength` | **1367** |
| … of which `/`-joined | | **1125** |
| … of which covering ≥2 substances | `substanceGroupKey` | 1110 |
| names denoting ≥2 substances *after* this change | `findImpliedSubstances` | 1159 (1097 `/`-joined), widest set 6 |
| substances | `substanceGroupKey` | 2114 |
| substances filed as more than one row | | **129** |
| … diverging on ATC level-4 subgroups | `atcSubgroups()` | **0** |
| … diverging on curated-group membership | `CrossReactivityGroup.groupsOf` | **0** |
| … with no unqualified row | `namesNoRoute()` | 10 |
| names whose resolution the #192 ranking moves | `findByDrugName().get(0)` vs `lookupByToken` | 265 |
| … raising no contraindication chip on `main` | real `DrugSafetyValidator.validate` | 87 |
| … raising no contraindication chip on this branch | real `validate` | **64** (23 recovered, **0** newly silent) |
| published names carrying a trailing qualifier | `parentMoietyName` | 313 |
| … gaining a substance from the moiety leg | `findImpliedSubstances` | **10** |
| `/`-joined published names | `combinationConstituents` | 1461 |
| … gaining a substance from the constituent leg | `findImpliedSubstances` | 1073 |

**The 1367 and the 1125 in #193 reproduce exactly.** Its denominator does not: I cannot reach 7452 by
any definition I tried — 5169 distinct normalized published names, 8300 (entry, alias) pairs. #192 also
recorded 471 moved names / 120 silent; over the distinct-name universe I measure 265 / 87. Both
differences are consistent with a larger denominator having been used there, and I have not been able to
reconstruct it. Reported rather than adopted.

**The moiety leg audited exhaustively rather than argued.** All 10 links it creates KB-wide:

```
[Moderna covid-19 vaccine (6y+ bivalent booster)]      -> [Moderna covid-19 vaccine]
[Moderna COVID-19 Vaccine (6m-5y bivalent booster)]    -> [Moderna covid-19 vaccine]
[Moderna COVID-19 Vaccine (bivalent booster)]          -> [Moderna covid-19 vaccine]
[Moderna COVID-19 Vaccine (6m-5y)]                     -> [Moderna covid-19 vaccine]
[Moderna COVID-19 Vaccine (6y-11y)]                    -> [Moderna covid-19 vaccine]
[Moderna COVID-19 Vaccine (booster only)]              -> [Moderna covid-19 vaccine]
[Iron (polysaccharide)]                                -> [Iron]
[Multivitamin (prenatal)]                              -> [Multivitamin]
[Insulin human (isophane)]                             -> [Insulin human]      (moiety carries A10AB..A10AF)
[Insulin lispro (protamine)]                           -> [Insulin lispro]     (moiety carries A10AB/AC/AD)
```

Every one is a presentation reaching its own parent moiety. The sibling pairs that share a stem and
must not merge — `Varicella Zoster Vaccine (Recombinant)`/`(live/attenuated)`,
`Manganese (chloride)`/`(sulfate)`, `Typhoid vaccine (live)`/`(inactivated)`,
`Dextran (low)`/`(high molecular weight)` — are all absent, because the KB publishes no row *named*
their stem (dextran's bare name is an alias of the high-MW row, not any row's display name).

# Mutation sweeps

Run on a throwaway worktree; nothing below was ever in the committed tree.

| reverted decision | what failed |
|---|---|
| class arm no longer stops at the first implied substance | `aCombinationNamingSeveralSubstancesStillRaisesOneChipPerClinicalFact` — 1 chip became 2, both about the same recorded allergy |
| moiety gate relaxed from the display-name rank to the alias rank | `aMoietyTheKbOnlyLISTSRatherThanNAMESIsNotReached` — the dextran pair merged, 0 chips became 1 |
| constituent leg removed | all three #193 outcome cases returned to 0 chips |

# Limitations, stated rather than implied

- **The bare-word moiety is not reached.** `Peanut oil` against `Peanut`, `Dextran 40`,
  `penicillin g, procaine`, `loteprednol etabonate`, `multivitamin with iron`, `factor ix complex` —
  all still silent. That shape is spelled exactly like `Digoxin Immune Fab (Ovine)` against `Digoxin`,
  a patient allergic to digoxin's **antidote**, which #192 measured and separated; separating them needs
  a judgement about substances, not about names. `PresentationMoietyAllergenTest` pins the peanut and
  dextran cases as bounds so a later widening cannot claim them silently.
- **A constituent the KB publishes no name for is dropped.** 768 distinct such parts across the 1461
  `/`-joined names — `apple pectin`, `dextran 70`, `polysaccharide iron complex`, `iron,peptonized`,
  `acellular pertussis vaccine`. The `isNamed` gate is also what makes the split safe: influenza strain
  designations (`A/California/7/2009`) split into fragments and every one is rejected.
- **The moiety leg only strips a trailing parenthesized qualifier**, not a nested one and not a
  constituent's own qualifier. Measured: 313 of the 5169 published names carry such a qualifier and 10
  gain anything from it, so extending the derivation is not where the remaining coverage is.
- Chip **wording** on this instance is not reproducible (`cache_prompt` KV reuse); every claim below is
  about chips and structure, never about the answer's phrasing.
- `Fluoroestradiol f-18` and `Estradiol` already share a `substanceKey` on the shipped KB because the
  tracer row carries estradiol's own `drugbank_id` (#196). Nothing here changes that either way, and it
  is not used as evidence in either direction.

# Follow-ups found while doing this, not fixed here

1. **KB data defect, #196's family.** `Sulfamethoxazole` (`DDInter1724`) publishes
   `rxnorm_name: "sulfamethazine"` and `drugbank_id: DB01015`. Sulfamethazine (sulfadimidine) is a
   different sulfonamide, so `displayLabel()` renders every chip about it as
   `Sulfamethoxazole (sulfamethazine)` — visible in the live capture above. Measured; upstream.
2. **A stale denominator in an existing test's prose.**
   `AllergenExactNameResolutionTest.twoEntriesMakingTheSameStrongestClaimResolveToTheEarliestOfThem`
   states "of the 7452 name strings the shipped KB publishes". Its 1367 and 1125 both reproduce exactly;
   the 7452 does not (5169 distinct, 8300 (entry, alias) pairs). Reported and deliberately **not edited**
   — it is an existing test.
3. **Corpus facts in the shared brief contradicted by the database** (port 3316, `SELECT` on
   `orders`/`drug_order`, active = not voided, `date_stopped IS NULL`, unexpired): Betty Williams has
   **2** active drug orders (the brief says 0), Sarah Taylor **8**, Kevin Brown **1**, George Anderson
   **0**, James Thompson **2** (both Metronidazole).
4. **Hazard, not a defect.** `combinationConstituents` splits on `/` unconditionally; a future KB that
   published a row *named* an influenza strain fragment (`california`, `victoria`) would let that
   fragment through the gate. Reasoned, not observed — 0 such cases in the shipped KB.
5. **CLAUDE.md's API-surface list should gain an entry** for this: *"Allergen substance resolution: use
   `DrugReferenceService.findImpliedSubstances(name)`. `lookupByToken` answers ONE row and is the
   primitive underneath it; calling it directly for a recorded allergy re-opens #193/#195."* Not added
   here — I do not edit `CLAUDE.md`.

# Live verification

Both builds were deployed to the 3.7.1 standalone by `standalone-verify.sh` (full JVM restart, module
start confirmed, querystore retrieval gated on `hits>0`, known-positive control chip gated) and probed
with the **same 15-probe script**. `main` was captured first at `18eb0a0`, then this branch. Patients
were resolved **by name against the database** (MariaDB on 3316), not from reconstructed uuids.

**Fixture created and left in place** (`$SP/i193-fixture.sh`, idempotent): three NON-CODED drug
allergies on **James Thompson** `49485340-1c54-4d60-b328-52368114f08d` —
`Sulfamethoxazole / trimethoprim`, `Omeprazole / sodium bicarbonate`, `Insulin lispro (protamine)`.
He was chosen from the DB: 201 obs (so retrieval is non-empty — measured `hits=337`), **2 active
Metronidazole orders**, and no pre-existing allergies. Metronidazole shares no ATC level-4 subgroup with
any of the three, so the three probes below do not cross-contaminate. Non-coded rather than coded
because the standalone's concept-name index is stale (`POST /concept` fails with
`DuplicateConceptNameException`); `PatientClinicalContextBuilder` reads `nonCodedAllergen` as an allergy
token either way.

Both runs passed the script's gates identically — `querystoreBuild … hits=75`, warm control chip = 1 —
so neither capture is vacuous. The omods differ (`36229052` vs `36232361` bytes), so the deploy really
changed. The querystore artifact was untouched: `querystore-api-1.0.0-SNAPSHOT.jar` md5
`38205608891f10b68929c35fb3eace7e` before and after.

## #193 — the sulfa moiety is checked

`Is it safe to give sulfamethoxazole?` — James Thompson, chart records an allergy to
**`Sulfamethoxazole / trimethoprim`**, active Metronidazole.

**main (`18eb0a0`) — 1 chip, and it is not about the allergy:**
```
[interaction] Sulfamethoxazole (sulfamethazine) interacts with active order metronidazole — Minor. …
ANSWER: No — Sulfamethoxazole should not be given: Sulfamethoxazole (sulfamethazine) interacts with
        active order Metronidazole — Minor. …
```
The recorded co-trimoxazole allergy produces **nothing** — no chip, no injected finding, no mention in
the answer.

**this branch — 2 chips, the allergy is back and leads the answer:**
```
[contraindication] The patient has a recorded allergy to Sulfamethoxazole (sulfamethazine).
[interaction]      Sulfamethoxazole (sulfamethazine) interacts with active order metronidazole — Minor. …
ANSWER: No — sulfamethoxazole should not be given: the patient has a recorded allergy to
        Sulfamethoxazole [339], and Sulfamethoxazole interacts with active order metronidazole [340].
```

*Had the fix been wrong*, this probe would still show exactly 1 chip, still about metronidazole, and the
answer would still not mention the allergy — a plausible-looking "No" for the wrong reason. The
`(sulfamethazine)` synonym is `displayLabel()`'s and is an upstream KB defect (follow-up 1), not this
change.

## #193 — `omeprazole / sodium bicarbonate` regains `A02BC`

`Is it safe to give pantoprazole?` — same patient, chart records an allergy to
**`Omeprazole / sodium bicarbonate`**.

**main — 0 chips:**
```
ANSWER: The records do not address the safety of giving Pantoprazole.
```
**this branch — 1 chip:**
```
[contraindication] Pantoprazole is in the same ATC class (A02BC) as the patient's allergy to
                   Omeprazole — possible cross-reactivity
ANSWER: No — Pantoprazole should not be given: Pantoprazole is in the same ATC class (A02BC), which is
        the same class as the patient's allergy to Omeprazole, suggesting possible cross-reactivity [339].
```
On `main` the allergy resolves to `Sodium bicarbonate` (`B05CB`/`B05XA`), which shares nothing with
`A02BC`. *Had the fix been wrong*, this would still be a flat "the records do not address" — the answer
`main` actually gave.

## #195 — the presentation reaches its parent moiety

`Is it safe to give insulin lispro?` — same patient, chart records an allergy to
**`Insulin lispro (protamine)`**.

**main — 0 chips, and no deterministic finding at all:**
```
--- safetyWarnings (0) ---
--- references (1) ---
  #336 type=allergy group=chart
ANSWER: No — insulin lispro should not be given: the patient has a documented allergy to
        Insulin lispro (protamine) [336].
```
The model happened to read the raw allergy record — that is the LLM, not the safety layer, and it is
exactly what the deterministic layer exists to stop depending on. **No `safety_finding` record was
injected**, so nothing pinned it.

**this branch — the finding exists and is cited:**
```
[contraindication] The patient has a recorded allergy to Insulin lispro.
[contraindication] The patient has a recorded allergy to Insulin lispro (protamine) (insulin lispro
                   protamine, human).
--- references (2) ---
  #334 type=allergy group=chart
  #339 type=safety_finding group=reference
ANSWER: No — Insulin lispro should not be given: The patient has a recorded allergy to Insulin lispro
        [339], which is supported by the recorded allergy to Insulin lispro (protamine) [334].
```
The load-bearing difference is the **first** chip and the `safety_finding` record: those come from the
pre-answer findings pass, which runs with an EMPTY answer and is therefore not answer-dependent. The
second chip arrives by the answer-echo path (#105 — the answer names the presentation and it is not
scoped out), which depends on the model's wording and citations and is **not** reproducible on this box;
it is a true statement about the chart either way, and `main` would raise it too if that path fired
there. Do not read a count into it.

*Had the fix been wrong*, this probe would show 0 chips and one `type=allergy` reference — `main`'s
result exactly.

## Must-not-regress (from #192) — all identical, chip for chip

| probe | main (`18eb0a0`) | this branch |
|---|---|---|
| Margaret King, `Is it safe to give botulinum toxin type A?` | 1 — `The patient has a recorded allergy to Botulinum toxin type A.` | identical |
| Melissa Wright, `Is it safe to give esomeprazole?` | 2 — `Omeprazole …allergy to Esomeprazole` + `The patient has a recorded allergy to Esomeprazole.` | identical |
| Dorothy Odinga, `Is it safe to give ciprofloxacin?` | 3, `The patient has a recorded allergy to Ciprofloxacin.` first | identical |
| **Kevin Brown**, `Is hydrocortisone safe for him?` | **4** | **4, byte-identical** |

Kevin Brown is the ledger-key canary: he carries `Hydrocortisone (ophthalmic)` **and**
`Hydrocortisone (topical)`, and `Levothyroxine` **and** `Levothyroxine sodium` — two pairs of records
naming one substance each. His allergen resolution now also runs the moiety leg (`hydrocortisone` **is**
a KB display name), and it adds no substance because that row is already the same `substanceKey`. Four
chips, not six. *Had the ledger key or the one-chip-per-recorded-allergy rule been broken*, this is the
probe that shows it.

## The exact controls — all identical

| probe | expected | main | this branch |
|---|---|---|---|
| Mary Smith, clarithromycin | 1 Major ×simvastatin | 1 | 1, identical |
| Agnes Adams, warfarin | 1 Major ×aspirin | 1 | 1, identical |
| Joshua Johnson, ibuprofen | 2 | 2 (NSAID group ×`Acetylsalicylic acid (aspirin)` + lisinopril Moderate) | 2, identical |
| Mary Smith, paracetamol | 0 | 0 | 0, identical |
| Betty Williams, bupivacaine | 1 | 1 (`×lidocaine — Major`) | 1, identical |
| George Anderson, esomeprazole | 2, both naming `Pantoprazole` | 2 | 2, identical |
| **Sarah Taylor**, hydrocortisone | 9, of which 4 named `Hydrocortisone` | 9, 4 | **9, 4, identical** |

Sarah Taylor is the over-merge canary — nine chips over one `Dexamethasone` allergen across eight
orders. A `substanceKey` widened too far collapses them; this change does not widen `substanceKey` at
all, and her count and text are unchanged.

**11 of the 14 probes are byte-identical between the two builds. The three that changed are exactly the
three cases #193 and #195 describe.**

# Build numbers

Measured by me, on clean worktrees, with the full reactor (`mvn -o clean install`):

| tree | api | omod | total |
|---|---|---|---|
| `main` at `18eb0a0` (the SHA this branches from) | 977, 0 failures, 34 skipped | 61, 0 failures | **1038** |
| this branch at `239af0d` | 992, 0 failures, 34 skipped | 61, 0 failures | **1053** |

+15 = the two new test classes (8 + 7). **No existing test was modified and none changed behaviour.**

Environment: querystore `querystore-api-1.0.0-SNAPSHOT.jar` md5 `38205608891f10b68929c35fb3eace7e`,
unchanged before and after.

Closes #193
Closes #195


dkayiwa:
**Re-probe stability check**, run against the same live build after the sweep, because the brief warns
never to trust a single query after a restart for chip presence or absence:

| probe | run 1 (in the sweep) | run 2 | run 3 |
|---|---|---|---|
| James Thompson, `Is it safe to give pantoprazole?` | 1 chip, `Pantoprazole is in the same ATC class (A02BC) as the patient's allergy to Omeprazole — possible cross-reactivity` | 1, identical | 1, identical |
| James Thompson, `Is it safe to give insulin lispro?` | 2 chips | 2, identical | 2, identical |

So the `A02BC` chip and both insulin chips are stable, not a first-query artefact. Answer *wording* still
is not reproducible on this box (`cache_prompt` KV reuse) — every claim in the PR body is about chips and
injected records, never about phrasing.

dkayiwa:
# Hardening pass

Fresh eyes; I did not write this change. Five commits on top of `239af0d`, **all of them tests and
prose — no production line changed**. `git diff 239af0d HEAD -- api/src/main` with comment lines
stripped is empty, so the live behaviour below is the behaviour this PR already had.

| tree | api | omod | total |
|---|---|---|---|
| `main` @ `18eb0a0`, measured in a clean worktree | 977, 0 failures, 34 skipped | 61 | **1038** |
| this PR @ `239af0d` | 992 | 61 | **1053** |
| after hardening @ `f870864` | 997 | 61 | **1058** |

`mvn -o clean install`, full reactor. querystore `querystore-api-1.0.0-SNAPSHOT.jar` md5
`38205608891f10b68929c35fb3eace7e` before and after. KB `ddi_knowledge_base.json` md5
`ba7b2721a75aac707ce452798b74c074`.

Every KB figure below is mine, taken by compiling against the branch and calling the **production**
predicates (`lookupByToken`, `findImpliedSubstances`, `nameMatchStrength`, `substanceGroupKey`,
`atcSubgroups`, `combinationConstituents`, `parentMoietyName`, `CrossReactivityGroup.groupsOf`) over
the shipped 19 MB KB. Nothing was re-expressed in Python.

---

## What hardening found: three of this PR's decisions had no test at all

A mutation sweep at `239af0d` (`mvn -o test -pl api` on a throwaway worktree, one decision reverted at
a time). Reverting each of these left **all 992 tests green**:

| reverted decision | result at `239af0d` |
|---|---|
| the equal-claimant leg removed entirely | **0 failures** |
| the `/` split narrowed to a spaced separator | **0 failures** |
| the constituent gate raised to the display-name rank | **0 failures** |

For contrast, the decisions that *were* held: reversing the implied set's order, removing the class
arm's `break`, deciding identity against `lookupByToken`'s answer only, relaxing the moiety gate, and
keying the identity chip on the row instead of the substance each failed 1–4 cases.

The equal-claimant leg is the one that matters: it is the widest leg, and **nothing in the repository
would have noticed its removal**.

### What was added

- **`aCombinationTheKbSpellsWithoutASeparatorIsStillCheckedAgainstEveryIngredient`** —
  `amoxicillin and clavulanic acid`, asked about amoxicillin. No separator and no qualifier, so
  neither the constituent leg nor the moiety leg can reach either ingredient; the name resolves to
  `Clavulanic acid`, which carries no ATC code, so both class comparisons have nothing to compare.
  **On `main` this is 0 chips — silence on a recorded penicillin allergy.** Kills the leg-2-removed
  mutation. The KB spells many combinations this way (`rifampicin isoniazid pyrazinamide and
  ethambutol`, `sultamicillin tosylate`, `potassium chloride-potassium gluconate`); 52 published names
  reach a second substance through this leg alone.
- **`aConstituentTheKbOnlyCONTAINSRatherThanNAMESIsNotReached`** — `benzocaine / clove oil`, asked
  about clove. The gate's lower bound, and it is doing real work rather than being satisfied by
  accident (below).
- **`anEntryClaimingTheRecordedNameOnlyByContainmentIsNotAnImpliedSubstance`** — `ketorolac
  tromethamine`, asked about tromethamine. Relaxing the equal-claim test to "any match" fails this and
  eight other cases across four files.
- **`aCombinationJoiningItsIngredientsWithoutSpacesStillReachesBothOfThem`** — `potassium
  citrate/potassium gluconate`. This one does **not** isolate the split; see below. Its comment and
  the javadoc say so.

Nine verbatim shipped-KB rows were added to `ddi-combination-allergen.json` by the rule its own
metadata states. I then checked every row of **both** new fixtures against the full KB through
production code: 16+13 rows (now 19+13), **0 divergences** in `substanceKey`, `atcSubgroups` or alias
list, and every one of the file's existing outcomes reproduces byte-for-byte on the full KB. That check
was not stated for the moiety fixture either; it passes.

---

## Verdict on leg 2 (the equal-claimant leg)

**Keep it, unchanged.** It is the widest leg and it is the only one that reaches a combination the KB
spells without a separator — 52 published names, including co-amoxiclav, the fixed-dose TB
combinations and `sultamicillin tosylate`. Removing or narrowing it re-opens #193 for all of them.

**I did find two entries making the same strongest claim that are not one clinical fact**, which is
what you asked for. The KB gives all three trastuzumab rows an **identical `ciel` alias list**:

```
Trastuzumab            rxnorm=trastuzumab             drugbank=DB00072  atc=L01FD01
Trastuzumab emtansine  rxnorm=ado-trastuzumab emtansine drugbank=DB05773 atc=L01FD03
Trastuzumab deruxtecan rxnorm=trastuzumab deruxtecan  drugbank=DB14962  atc=L01FD04
```

Three different antibody–drug conjugates, three DrugBank substances, one shared name list. Driven
through the real `DrugSafetyValidator.validate` over the shipped KB, allergy `Ado-trastuzumab
emtansine`:

```
main      "Is it safe to give trastuzumab deruxtecan?"  -> 3 chips
            The patient has a recorded allergy to Trastuzumab.
            Trastuzumab deruxtecan is in the same ATC class (L01FD) as ... allergy to Trastuzumab
            Trastuzumab emtansine  is in the same ATC class (L01FD) as ... allergy to Trastuzumab

branch    "Is it safe to give trastuzumab deruxtecan?"  -> 3 chips
            The patient has a recorded allergy to Trastuzumab.
            The patient has a recorded allergy to Trastuzumab deruxtecan.     <- FALSE
            The patient has a recorded allergy to Trastuzumab emtansine.      <- true, and new
```

The branch turns a true cross-reactivity chip about deruxtecan into a false identity claim, and at the
same time turns a *wrongly weak* chip about emtansine — the drug the chart actually records — into a
correct identity chip. So this is **not a reason to narrow leg 2**: `main` is not right here either
(it reports the Kadcyla allergy as "Trastuzumab"), and the wrongness is upstream. Filed below as a KB
data defect in #194/#196's family. Narrowing leg 2 to same-substance-name would cost the 52 genuine
combinations; that trade is clearly wrong.

**The `AllergenExactNameResolutionTest` compatibility claim checks out.** `Daxibotulinumtoxina` and
`Botulinum toxin type A` both carry `substanceKey = [botulinum toxin type a, db00083]` on the shipped
KB, so `findImpliedSubstances("botulinum type a toxin-haemagglutinin complex")` returns **1** row and
that test cannot move. Verified through `substanceGroupKey()`, not assumed. (In the fixture slice the
`Daxibotulinumtoxina` row publishes no `drugbank_id` of its own; the parser assigns the family's, which
is why the slice agrees with the full KB. I checked that rather than inferring it.)

---

## Verdict on the precedence-with-break ordering

**Correct, and pinned.**

What orders the set: `lookupByToken`'s answer first, then the other equal claimants in dataset order,
then the constituents in the order the recorded name lists them, then the parent moiety;
`addSubstance` keeps the first row per `substanceGroupKey()`.

Why the break cannot drop a more severe finding: **each arm scans the whole set before the next arm
runs.** Identity is decided over every implied substance, then the class arm over every one, then the
group arm. A member matching only by class therefore cannot pre-empt a member matching by identity,
and the break only ever chooses between findings of **equal** rank. What "first" actually decides is
which implied substance the chip *names* and which one the ledger keys on — not the chip's kind.

That it is pinned is measured, not argued: reversing the implied set's order
(`Collections.reverse` on the return) fails
`aCombinationNamingSeveralSubstancesStillRaisesOneChipPerClinicalFact`, which asserts the chip names
`Abacavir`. Removing the class arm's `break` fails the same case 1 → 2. Deciding identity against the
first member only fails four cases across both new files.

One ordering fact worth recording: across the whole KB, **no two published names resolving to the same
row produce the same implied set in a different order** (0 of 2283 rows), so a patient spelling one
allergy two ways cannot get two differently-keyed chips through order alone.

---

## The two canaries do NOT discriminate this fix

You asked me to confirm they would move if the fix were wrong. **They would not** — I measured it
rather than assuming it.

Both patients' allergens were read from the database, then driven through the real
`DrugSafetyValidator.validate` over the shipped KB against builds with one decision reverted at a time:

| build | Kevin Brown (7 allergy records) | Sarah Taylor (1 allergy record) |
|---|---|---|
| hardened, unmutated | 2 | 5 |
| implied set order reversed | 2, identical | 5, identical |
| equal-claimant leg removed | 2, identical | 5, identical |
| both gates relaxed | 2, identical | 5, identical |
| **identity chip keyed on the row, not the substance** | **3** | 5, identical |

(Offline counts; live they are 4 and 9 because the order-driven arm and the answer-echo path add more.
The deltas are the point.)

The reason is simple and checkable: **every one of Kevin Brown's seven allergy tokens, and Sarah
Taylor's one, has an implied set of size 1.** `Hydrocortisone (ophthalmic)` does reach the moiety leg
and the moiety row is already the same `substanceKey`, so nothing is added — exactly as the PR body
says, now verified. `Dexamethasone` implies only itself.

So:

- **Kevin Brown's 4 is a real canary for the ledger key** (it moves under that mutation, and
  `AllergenExactNameResolutionTest.twoRecordedAllergiesToOneSubstanceStillRaiseOneChipPerSubject`
  catches the same thing in CI). It is **not** a canary for anything the implied set does.
- **Sarah Taylor's 9-with-4-`Hydrocortisone` moves under none of the five mutations, including the
  ledger key.** She is a `substanceKey`-widening control — a real thing to hold, since #195 proposed
  exactly that widening and this PR declined it — but the PR body's "drops if the implied set merges
  too much" overstates her: her implied set cannot merge, because it has one member.

Both are worth keeping as regression controls. Neither is evidence that the fix works. The probes that
are evidence are the four below.

---

## The `/` split: judged, and the judgement recorded rather than pinned

**It should stay unconditional.** Measured:

- 1461 published names carry the separator: **1444 spaced, 17 bare**.
- Of the 1073 that gain a substance, **1072 are spaced and 1 is bare** —
  `potassium citrate/potassium gluconate`.
- **That one is also claimed at the alias rank by both its ingredients**, so the equal-claimant leg
  reaches the second ingredient without the split. Reverting `COMBINATION_SPLIT` to `" / "` therefore
  leaves the whole suite green, and the dataset offers **no** name that separates the two rules.

So a spaced-separator gate would cost nothing today. It is still the wrong rule, because it can only
ever **drop** an ingredient, and dropping is the one direction this change exists to prevent. I wrote
that judgement, and the fact that nothing pins it, into `combinationConstituents`' javadoc rather than
letting a test imply a guarantee it does not give.

**The safety property is not the accident the PR body suggests.** Of the constituent strings the gate
rejects, **341 match no entry at all** — that is where the influenza strain fragments (`california`,
`7`) sit, and they are refused by `matchesDrugName`, not by the rank. But **32 do match, at the
containment rank, and the gate is the only thing holding them back**:

```
lactated ringer's solution  -> Lactic acid                (from  glucose / lactated ringer's solution)
estradiol enantate          -> Fluoroestradiol f-18       (from  algestone acetophenide / estradiol enantate)
trolamine salicylate        -> Salicylic acid (sodium)    (from  capsaicin / trolamine salicylate)
dextran 70                  -> Dextran (high molecular weight)
l in normal saline          -> Sodium chloride            (from  potassium chloride 20 mmol/l in normal saline)
apple pectin                -> Pectin
clove oil                   -> Clove
```

Every one is #192's hazard shape. `benzocaine / clove oil` now pins it.

---

## Corrections to figures in the PR body

Everything else in the body reproduces exactly, including the whole moiety audit — I re-derived all
**10** links independently and got the same list, same rows, and `313` qualifier-carrying names.
`2283 / 5169 / 8300 / 1367 / 1125 / 1110 / 1159 / 1461 / 1073 / 313 / 10` and
`substances=2114 filedAsMoreThanOneRow=129 divergingSubgroups=0 divergingGroups=0` all reproduce.

Two things to correct:

1. **"768 distinct such parts" is not distinct parts — it is occurrences.** Measured through
   `combinationConstituents` + the production gate: **768 rejected occurrences** (707 inside
   spaced-separator names, 61 inside bare ones) over **373 distinct** constituent strings. The 768 is
   right as a count of occurrences; the word "distinct" is not.
2. **`divergingGroups=0` is true but its evidence base is thin, and the way it is measured matters.**
   `DrugReferenceService.setEntries` pins the curated groups **empty**, so a measurement that used the
   `setEntries` seam without `setCrossReactivityGroups` would report 0 vacuously. I measured it both
   ways and got 0 both times — but with the groups genuinely loaded, the shipped set contains **one**
   group (NSAID) and exactly **one** of the 129 multi-row families touches it. The number is correct;
   it is close to unfalsifiable on this data, and worth saying so.

---

## Live re-verification

Required — I changed only tests and comments, but the whole point of the PR is chip output, and a
`main` capture is the only thing that makes a silent-failure fix evidence rather than assertion.

Both builds deployed to the 3.7.1 standalone by `standalone-verify.sh` (full JVM restart, module start
confirmed, retrieval gated on `hits>0`, warm control chip gated). Both runs passed the gates
identically — `querystoreBuild hits=75`, warm control chip = 1 — so neither capture is vacuous. Omods
differ (`36229053` vs `36232365` bytes). Patients resolved **from the database** (MariaDB 3316), not
from REST.

**15 probes on each build. 11 are chip-identical. The 4 that changed are exactly the target cases.**

### #193 — the sulfa moiety (James Thompson, `49485340-1c54-4d60-b328-52368114f08d`)

**main `18eb0a0` — 1 chip, and it is not the allergy:**
```
[interaction] Sulfamethoxazole (sulfamethazine) interacts with active order metronidazole — Minor. …
--- references (3) ---  #337 drug_order  #338 drug_reference  #339 safety_finding
ANSWER: No — Sulfamethoxazole should not be given: Sulfamethoxazole (sulfamethazine) interacts with
        active order Metronidazole — Minor. …
```
The recorded co-trimoxazole allergy produces nothing.

**hardened — 2 chips, the allergy leads:**
```
[contraindication] The patient has a recorded allergy to Sulfamethoxazole (sulfamethazine).
[interaction]      Sulfamethoxazole (sulfamethazine) interacts with active order metronidazole — Minor. …
ANSWER: No — sulfamethoxazole should not be given: the patient has a recorded allergy to
        Sulfamethoxazole [339], and Sulfamethoxazole interacts with active order metronidazole [340].
```
*Had the fix been wrong*: one chip, about metronidazole, and a confident "No" for the wrong reason.

### #193 — `omeprazole / sodium bicarbonate` regains `A02BC` (same patient)

**main — 0 chips and 0 references:**
```
--- safetyWarnings (0) ---
--- references (0) ---
ANSWER: The records do not address the safety of giving Pantoprazole.
```
**hardened — 1 chip:**
```
[contraindication] Pantoprazole is in the same ATC class (A02BC) as the patient's allergy to
                   Omeprazole — possible cross-reactivity
--- references (1) ---  #339 type=safety_finding
```
*Had the fix been wrong*: the flat "the records do not address" `main` actually gave.

### #195 — the presentation reaches its moiety (same patient)

**main — 0 chips, and no deterministic finding at all:**
```
--- safetyWarnings (0) ---
--- references (1) ---  #336 type=allergy group=chart
ANSWER: No — insulin lispro should not be given: the patient has a documented allergy to
        Insulin lispro (protamine) [336].
```
The model read the raw allergy record. No `safety_finding` was injected, so nothing pinned it.

**hardened:**
```
[contraindication] The patient has a recorded allergy to Insulin lispro.
[contraindication] The patient has a recorded allergy to Insulin lispro (protamine) (insulin lispro
                   protamine, human).
--- references (2) ---  #334 type=allergy group=chart   #339 type=safety_finding group=reference
```
The load-bearing part is the **first** chip and `#339`, both from the pre-answer findings pass, which
`DrugReferenceInjector.preAnswerFindings` runs with an EMPTY answer — I read that call site to confirm
it. The second chip is the #105 echo path and depends on the model's wording; do not read a count into
it. *Had the fix been wrong*: 0 chips and one `type=allergy` reference — `main`'s result exactly.

### NEW — the equal-claimant leg, live for the first time

The leg that had no test also had no live probe. Fixture created and left in place
(`$SP/H198-fixture.sh`, idempotent): **one non-coded drug allergy `Amoxicillin and clavulanic acid` on
Susan Lopez `5e5c8ba4-4df3-4735-9e61-19cfc4ae298c`**, chosen from the DB because she has 89 obs (so
retrieval is non-empty) and **no other allergies and no active orders**, so nothing can contaminate the
probe.

**main — 0 chips, no deterministic finding:**
```
--- safetyWarnings (0) ---
--- references (1) ---  #157 type=allergy group=chart
ANSWER: No — Amoxicillin should not be given: the patient has an allergy to Amoxicillin and
        clavulanic acid [157].
```
**hardened — 1 chip and an injected finding:**
```
[contraindication] The patient has a recorded allergy to Amoxicillin.
--- references (2) ---  #157 type=allergy group=chart   #159 type=safety_finding group=reference
ANSWER: No — Amoxicillin should not be given: The patient has a recorded allergy to Amoxicillin [159],
        which is supported by the recorded allergy to Amoxicillin and clavulanic acid [157].
```
*Had the fix been wrong*: `main`'s result — a plausible "No" produced entirely by the model reading a
chart record, with the safety layer silent. This name carries no separator and no qualifier, so it is
the **only** one of the four probes that exercises the equal-claimant leg.

### Must-not-regress — all identical, chip for chip

| probe | main | hardened |
|---|---|---|
| Margaret King, botulinum toxin type A | 1 — `Botulinum toxin type A` | identical |
| Melissa Wright, esomeprazole | 2 — `Omeprazole` + `Esomeprazole` | identical |
| Dorothy Odinga, ciprofloxacin | 3, `Ciprofloxacin` first | identical |
| **Kevin Brown**, hydrocortisone | **4** | **4, identical** |
| **Sarah Taylor**, hydrocortisone | **9, of which 4 named `Hydrocortisone`** | **9, 4, identical** |
| Mary Smith, clarithromycin | 1 Major ×simvastatin | identical |
| Agnes Adams, warfarin | 1 Major ×aspirin | identical |
| Joshua Johnson, ibuprofen | 2 | identical |
| Mary Smith, paracetamol | 0 | identical |
| Betty Williams, bupivacaine | 1 (×lidocaine, Major) | identical |
| George Anderson, esomeprazole | 2, both naming `Pantoprazole` | identical |

---

## Found, not fixed

1. **KB data defect, #194/#196's family — three antibody–drug conjugates share one alias list.**
   `Trastuzumab` (DB00072), `Trastuzumab emtansine` (DB05773) and `Trastuzumab deruxtecan` (DB14962)
   carry byte-identical `ciel` arrays. Six published names are affected (`ado-trastuzumab emtansine`
   and the five `trastuzumab-anns/-dkst/-dttb/-pkrb/-qyyp` biosimilar codes). Effect on this branch,
   **measured** through the real `validate`: a recorded `Ado-trastuzumab emtansine` allergy asked about
   deruxtecan produces `The patient has a recorded allergy to Trastuzumab deruxtecan.` — a false
   statement about the chart, where `main` produced a true same-class chip. Upstream; the module cannot
   distinguish them by name.
2. **The four legs read the recorded string and never each other's output.** A trailing qualifier on a
   combination name switches the wider legs off instead of composing with them: `findImpliedSubstances`
   returns 2 rows for `abacavir / lamivudine` and **1** for `abacavir / lamivudine (oral)`, because the
   qualified string is no entry's name (claim drops to the containment rank, so the equal-claimant leg
   is skipped) and the constituent becomes `lamivudine (oral)`, which is no entry's name either.
   Measured; **not a regression** (`main` answers the same single row) and **0** of the shipped KB's 11
   slash-plus-qualifier names lose a substance, so it reaches production only through a free-text
   allergen. Recorded in the javadoc. Closing it means feeding each leg's output back through the
   others, which needs its own measurement.
3. **The group arm's `break` is unpinned, and unpinnable from shipped data.** Measured: **0** published
   names imply two members of a curated cross-reactivity group, and the shipped curated set has exactly
   **one** group (NSAID). *If we ship without it:* a later refactor that drops that `break` lets one
   recorded name implying two members of one curated group raise two chips for one clinical fact —
   #145's collapse — and no test notices. A test needs a second curated group **and** a fabricated
   combination name, which the fixtures' verbatim-slice convention rules out.
4. **The constituent gate's upper bound is unpinned.** Raising it from the alias rank to the
   display-name rank (the symmetric-with-the-moiety-gate move) costs exactly **2** published names
   KB-wide: `menthol / selenium disulfide` → `Selenium Sulfide`, `yohimbe preparation / zinc sulfate` →
   `Yohimbine`. *If we ship without it:* those two go silent and nothing notices. Left unpinned
   deliberately — two names did not justify a third fixture row set.
5. **29 published names keep a weaker-claiming representative.** `addSubstance` keeps the first row per
   substance, and the first is not always the strongest claimant: `4-aminobenzoic acid / salicylic acid`
   labels its chip `Salicylic acid (sodium)` although the constituent `salicylic acid` names
   `Salicylic acid` outright. **Not a regression** — that is the row `lookupByToken` already answered —
   and preferring the stronger claimant is `canonicalRow`'s relabelling question, which
   `nameMatchStrength`'s javadoc deliberately keeps out of this path. Now stated in `addSubstance`'s
   javadoc, which previously asserted the opposite.
6. **`recordedAllergens` de-dupes on the whole implied list, where the old `seenAllergens` de-duped on
   the resolved row.** Two allergy tokens resolving to the **same** row but implying **different** sets
   are now two records, and can raise two chips where `main` raised one (they key the ledger on
   different substances). **341** of the 2283 rows are the resolution answer for two or more published
   names with divergent implied sets. Reasoned and KB-measured; **not observed on a patient**. The
   javadoc documents the choice deliberately and over-reporting is this module's stated safe direction,
   so I left it — flagging it because the count change is real and nothing pins it.
6b. **…and it de-dupes on an ORDERED list, which the javadoc did not say.** `!out.contains(implied)` is
   `List.equals`, so two tokens naming the *same* substances in a *different order* survive as two
   records where the old row-keyed guard collapsed them. The identity arm still collapses (it matches
   whichever member IS the subject), but the class arm stops at its first match and keys the ledger on
   that member, so a subject related to two of them gets **two `SAME_CLASS` chips for one clinical
   fact**. *If we ship without this:* a patient whose chart records one combination product under two
   spellings in opposite ingredient order is told twice, in two differently-worded chips, about one
   allergy. Three independent reviewers found it. **Measured unreachable from the reference data** — no
   two published names resolving to one row produce the same substances in a different order (0 of
   2283) — so I could not write a failing test from real fixture rows, and changing production code at
   this point would invalidate the live verification for a case with no measured instance. I corrected
   the javadoc to describe the actual behaviour and left the code. The fix is to de-dupe on the SET of
   `substanceGroupKey()`s, which is also cheaper (`O(A)` hashing instead of an `O(A²·M)` list walk);
   the failing case needs a subject sharing an ATC subgroup with **both** constituents, which the
   current fixture cannot express.
7. **Two corpus facts the database contradicts** (port 3316, `orders` join `drug_order`, not voided,
   not stopped, unexpired, action ≠ DISCONTINUE), measured 2026-08-09: **Melissa Wright has 1** active
   drug order (Spironolactone), not the Spironolactone + Ondansetron the shape table lists; **Susan
   Young has 5** (Tiotropium, Botulinum toxin type A, Lidocaine, Neomycin, Metoclopramide), not the one
   Tiotropium the table lists — later agents added the other four. The DB-measured line added on
   2026-08-09 reproduces exactly: Betty Williams 2, Sarah Taylor 8, Kevin Brown 1, George Anderson 0,
   James Thompson 2.
8. **`AllergenExactNameResolutionTest`'s javadoc still carries the 7452 denominator.** Confirmed still
   present; deliberately not edited, per the note that it is being corrected separately.

## Polish, and what I deliberately did not take

Applied: `row(entries, name)` hoisted into `DrugReferenceTestSupport` (this diff had taken it from two
copies in the package to four, and CLAUDE.md puts shared drug-reference helpers there); four premise
assertions given messages, so a failure no longer prints `expected: <1> but was: <0>` with no drug
named.

Not taken, each with what breaks if we ship without it:

- **Fusing the equal-claimant scan into `lookupByToken`'s.** When a name is claimed at the alias rank
  `lookupByToken` cannot early-exit, so it visits all 2283 entries and `findImpliedSubstances` then
  visits them again computing the same predicate — a straight doubling. *If we ship without it:* one
  extra full scan per allergy token, tens of microseconds against an LLM round trip, twice per
  question. It is a real duplication and the right cleanup, but it changes `lookupByToken`'s shape, and
  `lookupByToken` is the method three issues' worth of measurement is pinned to.
- **Renaming `List<DrugReference> allergen` to `recordedAllergy`.** The identifier is singular, the type
  is a set of substances, and it meant one row before this change. *If we ship without it:* a reader
  who knows the old code reads `allergen` as one row and misreads the three arms. Deferred because the
  rename touches `{@link}`s in three javadocs that have already been through four correction rounds.
- **Guarding `recordedAllergens` on `!inPlay.isEmpty() || !orderEntries.isEmpty()`.** *If we ship
  without it:* a drugless question on a patient with allergies and no resolvable orders resolves every
  allergy token for a result nothing reads — a few full scans, discarded, twice per question.
- **Splitting the varicella/manganese bound case in two.** *If we ship without it:* when varicella
  fails, the manganese assertions never run, so one fix hides the other.
- **`Pattern.quote` on the combination separator.** *If we ship without it:* nothing today — `/` is not
  a metacharacter — but changing that constant to `|`, `.` or `+` would silently change the split
  rather than fail to compile.

## Fixture left on the standalone

`Amoxicillin and clavulanic acid`, non-coded DRUG allergy, on **Susan Lopez**
`5e5c8ba4-4df3-4735-9e61-19cfc4ae298c`. Idempotent script at `$SP/H198-fixture.sh`. James Thompson's
three allergies from the original pass are untouched. Nothing else on the instance was changed; no GP
was altered.

dkayiwa:
### Addendum — one more polish pass, and where I stopped

A second polish pass found one real thing, and it was mine: my fixture generator was run twice against
a file it had already appended to, so `ddi-combination-allergen.json`'s `metadata.note` carried the
same 1221-character block twice. Removed in `2b4b013`, along with the enumerator that still said "Two
shapes" before listing three. Data rows, interaction rows and mechanisms are untouched (19 / 38 / 9)
and the slice still reproduces the shipped KB row for row.

Final: **997 api + 61 omod, 0 failures**, `main` at `18eb0a0` measures **977 + 61**. Production code is
still comment-only against `239af0d` — `git diff 239af0d HEAD -- api/src/main` with comment lines
stripped is **0 lines** — so the live captures above are the behaviour at HEAD.

**Where I stopped, stated honestly.** Phase 1 converged: the last two structural passes found nothing in
code or tests, and every decision the mutation sweep showed to be unheld now has a case, except the two
that measurably cannot be pinned from the shipped data (the split's unconditionality, and the group
arm's `break`) — both written up above with what breaks if we ship without them. Phase 2 did **not**
converge on the skill's rule of two consecutive clean passes: pass 1 produced applied and deferred
items, pass 2 produced the fixture duplication above. I am not claiming a convergence I did not
observe. What is left is the deferred list above, none of which changes behaviour, plus one boundary
note not worth acting on: an identity chip pre-empting a class chip changes the finding TEXT, which is
also the echo-scoping corpus, so a drug named only in the answer could in principle come back into play
— it needs a four-way conjunction and the opposite direction is far more common.

Two smaller items I looked at and left: `warningsFor` is now byte-identical in the two new test classes
(hoisting it costs every call site a third argument — *if we ship without it*, a later third file
copies it a third time), and three lines in the new tests run past the package's usual column ceiling.

dkayiwa:
Independent pre-merge verification (mine, not relayed): `mvn -o clean install` on a clean detached worktree at `2b4b013` → **997 + 61 = 1058, 0 failures, 34 skipped**, BUILD SUCCESS. `main` at `18eb0a0` = 977 + 61 = 1038. Hardening touched **no production line** (comment-stripped `src/main` diff is empty), **zero non-comment assertion lines removed** across the branch, CI green on Java 11/17/21.

**The headline finding is that three of this PR's own decisions had no test at all.** A mutation sweep at `239af0d` reverted each — the equal-claimant leg **removed entirely**, the `/` split narrowed to a spaced separator, the constituent gate raised — and all 992 tests stayed green. Leg 2 is the widest leg in the change and nothing would have noticed its deletion. Three cases now pin what can be pinned, each verified to fail under its own mutation.

**And it corrected my briefing, which is the part worth recording.** I named Kevin Brown's and Sarah Taylor's chip counts as "canaries" that would move if the fix were wrong. They cannot: every one of Kevin's seven allergy tokens and Sarah's one has an implied set of **size 1**, so both are byte-identical under order-reversal, leg-2 removal *and* gate relaxation. Kevin moves under one unrelated mutation; **Sarah moves under none at all.** They are sound *regression controls* — they prove nothing broke — and **zero evidence the fix works**. I have written the distinction into the shared brief: for any probe offered as evidence *for* a change, name the mutation it would catch and confirm by mutation that it does.

The same pass flagged its own `divergingGroups=0` as **true but near-unfalsifiable** — only one curated group ships and only one of 129 families touches it — which is the same principle one level down: a measurement that cannot fail is not confirmation.

**Verdicts I asked for, both answered with measurement.** Leg 2 stays: it is the only path to the **52** published names the KB spells without a separator. It *did* find equal claimants that are not one clinical fact — all three trastuzumab rows share an identical `ciel` list, so a Kadcyla allergy yields *"recorded allergy to Trastuzumab deruxtecan"*, which is false — but `main` is wrong there too, it is an upstream data defect (now #196's fifth instance), and narrowing the leg would trade 52 real recoveries for one false label. Precedence-with-break is correct and now pinned: each arm scans the whole set before the next runs, so the break only ever chooses between equal-severity findings.

Also corrected in the PR body: "768 **distinct** parts" is 768 *occurrences*; distinct is **373**.

It built a new probe for the untested leg rather than reusing mine — Susan Lopez with `Amoxicillin and clavulanic acid`: `main` gives **0 chips and no `safety_finding` at all**, branch gives 1 chip plus the injected finding. Since both defects fail silently, that `main` capture is the evidence, and it is present for all four changed probes.

Phase 1 converged; **Phase 2 did not**, and it said so rather than claiming otherwise.

Issue #175, opened 2026-08-06
The screening arm now defers to the drug-in-play arm's chosen rule, which can under-report a severity


dkayiwa:
## What

PR #173 replaced the screening arm's text-keyed cross-arm suppression with an identity ledger (`InteractionPairs`, keyed on `{subject substanceGroupKey(), partner}`). That fixed a real duplicate-chip regression and closed two things the text key could never see. It also changed **which** chip survives for a pair: the screen now stands down from a pair the drug-in-play arm already covered, so the surviving chip carries **that arm's chosen rule**.

The drug-in-play arm only sees the substance rows that are actually *in play*. So where an order name resolves a row that the question and answer text do not — a brand alias published on one row of a family only — the deferred chip can carry a **lower severity** than the screening arm would have raised.

## Why this is worth filing rather than absorbing

The direction of the change is **over-report → under-report**. Under the old text key, this shape produced two chips; now it can produce one, and the one that survives may be the less severe. Every other defect in this cluster (#145, #160, #162, #173) traded surplus chips for fewer chips, which is safe because the survivor was provably equivalent. This is the one place where the survivor may not be equivalent, and under-warning is the failure mode this whole effort started from (#86 — a Major bleeding interaction that never fired).

## What would settle it

A fixture in which one row of a multi-row family carries an alias that appears **only** in the order name, and whose rule for some partner is **more severe** than the row the question text resolves. Then assert the surviving chip's severity. That is a small, bounded test; it was not built because it needs a shape the shipped KB may not contain.

If the shape turns out to be unreachable on shipped data, that is a fine answer — but it should be recorded with the measurement rather than assumed, because a KB refresh can create it.

## Provenance

Reasoned from the code by the agent that fixed the #173 regression, and **explicitly labelled by it as not measured**. The ledger change itself is verified: the duplicate reproduced 1 → 2 on a real fixture and returns to 1 with the fix, and disabling the suppression fails 4 tests of which 2 are pre-existing, so the ledger covers everything the text key pinned.

PR #210, opened 2026-08-09
fix(drug-safety): resolve a candidate set by the ranked claim, not an unranked matcher (#209)

### LINKED ISSUE #209: Two name-resolution rules coexist and disagree: an unranked matcher admits substances the patient is not prescribed
## What

Found by the final composition probe of merged `main` (`ae4d2bd7`) — **the only defect in that pass that no single-PR verification could have caught**, because it requires two independently-correct fixes to coexist.

Two name-resolution rules are in force and they disagree about how many substances a drug name denotes:

- **ranked** — `nameMatchStrength` / `findImpliedSubstances` (#192, #197, #198) keeps only the strongest claimants;
- **unranked boolean** — `DrugReference.matchesText` (#128's word-start matcher, called from `DrugReferenceService:113`, `DrugSafetyValidator:647` and `:3401`) admits **every** candidate.

Where a chip arm resolves through the unranked one, a name resolves to a **superset**, and a chip can name a substance the patient is not prescribed.

Verified through production predicates: for the token `hydrocortisone`, `matchesText` is **true for all four** hydrocortisone rows, while `nameMatchStrength` scores `Hydrocortisone` **2** and `Hydrocortisone butyrate` **1**.

## Live consequence, observed three times independently

Sarah Taylor's only order in that family is `Hydrocortisone Injection vial 100mg`. Probes P07, P22 and D1 each produced:

```
Hydrocortisone butyrate is in the same ATC class (H02AB) as the patient's allergy
to Dexamethasone — possible cross-reactivity
```

**One order, two chips, naming two substances — one of which she is not on.** The same mechanism inflates the screening arm's candidate set from her 8 orders to "18 active-order reference entries", which that arm logs in its own WARN.

## Blast radius

**85 of 2283 shipped display names** admit more substances under `matchesText` than under the strength-ranked set — measured through production predicates. Two reachable on stock data: `Ketoconazole` → also `Levoketoconazole` (Betty Williams carries a Ketoconazole allergy), and `Atropine`.

## Why it is not a duplicate

Checked against the open set. It is **not** #206 (which arm anchors its *subject* on the chart), **not** #163's residue as recorded on #205 (the injector reads the canonical row while chips read the resolved rows), and **not** #150 (a blank alias in a hand-authored file). Those are all about *which representative* is chosen, or *what is injected*, once a candidate set exists. This is about **which candidates enter the set at all** — one layer earlier, and the reason the later fixes cannot compensate for it.

It is also the natural completion of the two-matcher story: #128/#147 established that prose, order names and clinician-entered names are three different string shapes needing three matchers, and CLAUDE.md records that. What it does not record is that one of those three is **unranked**, so it answers a different question from the ranking that now governs the others.

## Shape of the fix

The candidate set and the choice are deliberately separate concerns — #192 kept them apart on purpose, and CLAUDE.md says so. So the fix is probably not to rank inside `matchesText`, but for the *call sites* that need substance-level resolution to go through `findImpliedSubstances` (or the ranked path) rather than iterating a boolean matcher. Enumerate the three call sites above and decide per site which question it is actually asking.

Note the constraint: `matchesText` must keep symmetric word boundaries for **prose**, which is what #86 measured and what a test pins. This is not an argument for widening or narrowing it — only for not using it where a ranked answer is needed.

## Provenance

Measured by the final composition probe through production predicates compiled against the built jar — the four-row `matchesText` result, the 2-versus-1 strength scores, the 85-of-2283 count, and all three live chip observations. I verified the three call sites myself. The probe opened no PR, correctly: a verification pass should not fix.

dkayiwa:
Fixes #209.

## What was wrong

`DrugReference.matchesText` (prose) and `DrugReference.matchesDrugName` (a recorded drug name) answer
**"does this string MATCH this entry"**. That is the right question for both shapes and both answer it
correctly. It is also a strict **superset** of "which SUBSTANCES does this string NAME" wherever one
alias is shared by two substances — and the ranked resolution added by #192/#197/#198
(`nameMatchStrength` → `findImpliedSubstances`) is what decides that.

So every candidate set built by iterating one of those booleans put substances in play that the string
did not name. Measured through the production predicates over the shipped 19 MB KB:
`matchesText("hydrocortisone")` is true for **all four** hydrocortisone rows, while
`nameMatchStrength("hydrocortisone")` scores `Hydrocortisone` **2** and `Hydrocortisone butyrate`
**1** — a different substance (`substanceKey` separates them by design, #121/#164).

## Per call site: which question it asks, and what changed

| site | the question it actually asks | verdict |
|---|---|---|
| `DrugReferenceService:113` — `findByQuery`'s loop, consumed by `DrugSafetyValidator.validate` (question + answer legs) and `DrugReferenceInjector.matchingEntries` | **substance-level.** "Which drugs are IN PLAY" is a question about substances; the prose matcher answers "which entries are mentioned". | **routed through the ranked path.** `findByQuery` is kept, unchanged, as the prose PRIMITIVE — the same relationship `lookupByToken` has to `findImpliedSubstances` — and the three production consumers now call the new `findImpliedByQuery`. |
| `DrugSafetyValidator:647` — `isEchoOfCitedRecord` | **genuinely prose. KEEPS `matchesText`.** It iterates cited record TEXTS for a candidate that is already fixed; it builds no candidate set and resolves no name. What it decides is whether the answer's WORDING is attributable to a record the answer cites — a question about text, not about substance identity. Ranking it would also *narrow an exemption whose entire purpose is to suppress chips about drugs nobody proposed*, re-opening #105 (measured there at 7 of 8 chips about unproposed drugs). | unchanged |
| `DrugSafetyValidator:3401` — `attributedDoses` | **genuinely prose. KEEPS `matchesText`.** It iterates answer CLAUSES for a fixed row, scoping *where in the answer* a dose may be attributed. Its candidate set was already substance-resolved upstream (`resolvedSubstanceRows`, #175), and `aliasOwnsDose` already discriminates between entries by proximity. Ranking a clause gate is a category error. | unchanged |

### A fourth site the issue does not list, and the headline needs it

`DrugReferenceService.findForActiveOrders`'s **name leg** (`findByDrugName` → `matchesDrugName`) has
the same defect and is the one the live evidence is mostly about. Measured: with a question naming no
drug at all, Sarah Taylor still gets the `Hydrocortisone butyrate` chip — from
`addActiveOrderContraindications` over her order entries. Probes P22 (question names *dexamethasone*)
and D1 (screening, names no drug) are both that arm.

**Both legs are needed together.** The drug-in-play arm and the order-driven arm each skip what the
other has covered, so narrowing one alone leaves the chip standing: narrow only the prose leg and the
order arm picks the ester up (it is no longer "already in play"); narrow only the order leg and the
prose leg keeps it in play. `theEsterIsGoneWhateverArmWouldHaveRaisedIt` is the case that fails for a
one-leg fix, and mutation M1/M2 below prove it.

## The rule: the name an entry matched BY, never the whole string

`DrugReference.aliasesIn(text)` / `aliasesNaming(name)` are each their own boolean's **witness** —
same primitive (`containsWord` / `matchesOrderName`), so neither pair can drift. An entry survives when
one of its own names that the string carries denotes its substance under `findImpliedSubstances`.

Why not the whole string:

- **prose has no claim strength at all** — a question is not a drug name, so `findImpliedSubstances` on
  it lands at the containment rank and answers with the earliest matching row;
- **a recorded order name is usually nobody's name** (a display name with a strength appended), so its
  strongest claim is *also* mere containment — the rank at which `findImpliedSubstances` deliberately
  refuses to widen. Filtering on that would keep one ingredient of a combination order and silence the
  rest, which is the direction #193/#195 exist to prevent.
  `aCombinationOrderNameKeepsEveryConstituentSubstance` pins it and mutation M3 proves it.

The verdict is taken **per substance** and then applied to every matched row of it, so route and
formulation variants are kept — the arms downstream need the whole family. That second half is
load-bearing: `Estrone sulfate (topical)` publishes `estrone` (the *other* substance's display name) as
its `rxnorm_name` and nothing spelled `estrone sulfate`, so a per-row verdict drops a presentation of a
substance that IS in play. `aPresentationOfASubstanceInPlayIsKeptEvenWhenItsOwnAliasesNameAnother` pins
it; mutation M5 proves it.

`matchesText` is **untouched** — its symmetric prose boundary is #86's measurement and a test pins it;
`theFixtureReallyCarriesTheTwoClaimStrengthsUnderTest` asserts `findByQuery` still returns all four
hydrocortisone rows so that stays true.

## Measured, KB-wide

Driving all 2283 published display names through the new methods and their primitives (2026-08-09,
through production code; re-derive rather than trusting the figures):

- **33 names** resolve fewer rows, **71 rows** in total — identically on the prose and the order-name leg.
- **0** non-empty sets emptied; **0** substances losing a row; **0** names losing the substance
  `findImpliedSubstances` says they denote; **0** drops that share a display stem with anything kept
  (i.e. no sibling presentation is ever dropped while a stem-mate survives).
- Every one of the 33 drops is a genuinely different substance. The whole list, so it can be judged
  rather than trusted:

  `Ketoconazole`→`Levoketoconazole` · `Netupitant`→`Fosnetupitant` · `Atropine`→`Hyoscyamine` ·
  `Trastuzumab`→`Trastuzumab deruxtecan`+`emtansine` · `Mometasone`→`Mometasone furoate` ·
  `Dexfenfluramine`→`Fenfluramine` · `Dextroamphetamine`→`Amphetamine` · `Esomeprazole`→`Omeprazole` ·
  `Estrone`→`Estrone sulfate` · `Fenofibrate`→`Fenofibric acid` · `Gabapentin`→`Gabapentin enacarbil` ·
  `Hydrocortisone`→`Hydrocortisone butyrate` · `Iodine`/`Povidone-iodine`→`Iodide I-131` ·
  `Isosorbide`→`Isosorbide mononitrate` · `Dotatate`↔`Lutetium Lu 177 dotatate`/`Dotatate gallium ga-68`
  · and **`Moderna covid-19 vaccine`→`Pfizer-BioNTech Covid-19 Vaccine` plus all five `Tozinameran`
  rows** — a question about one manufacturer's COVID vaccine was putting the other manufacturer's, and
  every presentation of it, in play.
- Cost: a question goes from ~1.3 ms to ~3.1 ms. A pathological 6 KB answer naming 277 entries goes
  from ~95 ms to ~375 ms. Against 15–90 s of LLM time per answer.

## Every decision proved load-bearing by mutation, or reported as unprotected

A control is not a canary, so each new test is offered against a named mutation, reverted on a throwaway
worktree (`git worktree`, never the tree these commits came from):

| mutation | caught by |
|---|---|
| **M1** prose leg reverted to `findByQuery` (both validator legs + the injector) | `aQuestionWordPutsOnlyTheSubstanceItNamesInPlay`, **`theEsterIsGoneWhateverArmWouldHaveRaisedIt`**, `theInjectedReferenceRecordsCarryOnlyTheSubstancesTheQuestionNames` |
| **M2** order-name leg reverted to `findByDrugName` | `anOrderNamePutsOnlyTheSubstanceItNamesInPlay`, **`theEsterIsGoneWhateverArmWouldHaveRaisedIt`** |
| **M3** order leg filters on the WHOLE recorded string instead of the carried name | `aCombinationOrderNameKeepsEveryConstituentSubstance` (the lamivudine allergy chip disappears), `aPresentationOfASubstanceInPlayIsKeptEvenWhenItsOwnAliasesNameAnother` |
| **M5b** per-ROW verdict instead of per-substance | `aPresentationOfASubstanceInPlayIsKeptEvenWhenItsOwnAliasesNameAnother` (and KB-wide: 34 names / 72 rows against 33 / 71) |
| **M4** only the FIRST carried alias is resolved | **nothing, and nothing could** — see below |
| **M6** prose witness taken under the order-name boundary rule | **nothing, and nothing could** — see below |

M1 and M2 are the proof that both legs are needed: the same test fails for either one-leg fix.

**Two sub-decisions are unprotected, and honestly so.** M4 and M6 each produce a byte-identical
resolution over all 2283 published names — same 33/71, same signature hash — so on the shipped KB they
are indistinguishable from the shipped code and no test drawn from that KB can pin them. They stay
because they are the correct reading of the rule ("one of its names", under *its own* matcher's boundary
rule), not because they were measured to gain anything. If a later KB makes them differ, M4 is the one
to watch: it would silently stop resolving an entry whose display name comes first in its alias list but
is not the name the string carries.

## Eight existing expectations encoded the over-admission — ruled on, and corrected

Eight assertions in five classes expected the superset. Each was written to assert a property that still
holds, using a fixture where BOTH substances were in play as its vehicle, and nobody had asked whether both
should have been admitted. They were reported for a ruling before anything was touched, then corrected one
at a time with the property, its surviving assertion site, and the moved number stated per test — see the
`test(drug-safety): correct eight expectations…` commit and the PR comment.

The two invariants that had to survive do, and by mutation rather than assertion: reducing `substanceKey`
to its first component (merging `Omeprazole` and `Esomeprazole`) is caught by all three non-collapse cases
whose vehicle changed, by the key-level `assertNotEquals` behind them, and by two of the new cases. Three of
the eight kept their expected count and changed their *question* instead, so they now assert non-collapse
from a request that names both substances — a stronger vehicle, since it cannot be satisfied by
over-admission.

## Test numbers

- baseline on `main` (`ae4d2bd7`), measured in a clean worktree: **1010 api + 61 omod = 1071**, 0 failures, 34 skipped.
- this branch: **1019 api + 61 omod = 1080**, **0 failures**, 34 skipped. CI green on `1bac410c`.

## Live verification

See the PR comment.


dkayiwa:
## Live verification on the 3.7.1 standalone

Both builds deployed and probed through the real `/chartsearchai/search` endpoint, 23 probes each, into
append-only captures.

- **before** — `main` `ae4d2bd7`, omod md5 `7443caab4e2d82e89662bd6cee462f33` (the build that was already
  on the box)
- **after** — this branch, omod md5 `a8f08665c61bfb86cd4ae9a6294eb529`
- both: `sourceFormat=ddinter`, `entryCount=2283` from `/chartsearchai/drugreferencestatus`,
  `minInteractionSeverity=minor`, `chartMode=fullChart`, `llm.engine=local`, all 53 GPs untouched.
  The deploy gate passed on both: querystore alive (`hits=75`), Mary+clarithromycin ≥ 1 chip.
- querystore: deployed omod md5 `24c8aca7e657260c949837fd72aeb811`, api jar
  `38205608891f10b68929c35fb3eace7e`.

### The headline — Sarah Taylor + hydrocortisone (probe P07's question)

`9 → 8`, and the chip that goes is the one naming the ester. Her four `Hydrocortisone`-named chips
survive, byte-identical, and so do the other four:

```
BEFORE (main, 9)                                   AFTER (branch, 8)
[contra] Hydrocortisone   … same ATC class (H02AB) as the patient's allergy to Dexamethasone   ✅ kept
[inter]  Hydrocortisone   … active order celecoxib — Moderate. …                                ✅ kept
[inter]  Hydrocortisone   … active order diclofenac — Moderate. …                               ✅ kept
[inter]  Hydrocortisone   … active order ibuprofen — Moderate. …                                ✅ kept
[contra] Hydrocortisone butyrate is in the same ATC class (H02AB) as the patient's
         allergy to Dexamethasone — possible cross-reactivity                                   ❌ GONE
[contra] Dexamethasone    The patient has a recorded allergy to Dexamethasone.                  ✅ kept
[contra] Prednisone       … same ATC class (H02AB) …                                            ✅ kept
[contra] Budesonide       … same ATC class (R01AD) …                                            ✅ kept
[contra] Methylprednisolone … same ATC class (H02AB) …                                          ✅ kept
```

Re-run once more on each build (a cold-start race can load the KB empty, so an absence claim needs a
second run): `9 → 8` again, same chip gone.

*Had the fix been wrong:* had only the prose leg been narrowed, the order-driven arm would have picked
the ester up here (it is in her order entries and no longer "already in play") and the count would have
stayed 9. Had only the order leg been narrowed, the prose leg would have kept it in play — also 9.

### Sarah's other two live observations of the same chip

| probe | question | before | after |
|---|---|---|---|
| P22 | "Is it safe to give dexamethasone?" | 9, one naming `Hydrocortisone butyrate` | **8**, ester gone |
| D1 | "Are there any drug interactions with her current medications?" | 19 | 15 |

P22 is the decisive one for the fourth call site: its question does not name hydrocortisone at all, so
the ester can only have come from her ORDER name — the leg the issue does not list.

### Her screening question — the WARN

```
BEFORE  Interaction screening across 18 active-order reference entries found 18 pair(s) …
AFTER   Interaction screening across 17 active-order reference entries found 18 pair(s) …
```

**18 → 17 entries from 8 orders**, and the withheld-pair list is character-for-character identical
before and after — so the dropped entry (`Hydrocortisone butyrate`) contributed no pair and no
interaction finding was lost.

The chip count on that probe (19 → 15) is **only partly attributable to this change**, and I am not
claiming otherwise. 1 of the 4 is the ester chip. The other 3 are `Hydrocortisone × celecoxib /
diclofenac / ibuprofen`, raised on `main` by the *drug-in-play* arm because that run's answer named
hydrocortisone outside echo scope; the screening arm withholds those same three pairs under its 10-chip
cap either way (they are in the withheld list on both builds). Answer wording is not reproducible on
this box (`cache_prompt`), so this is a count that moves on its own. Measured: 15 on three consecutive
branch runs.

### `Ketoconazole` → `Levoketoconazole`, the second stock-data case

Betty Williams (`Ketoconazole` allergy, no hydrocortisone anywhere), "Is it safe to give ketoconazole?"

```
BEFORE (2)  [contra] Ketoconazole      The patient has a recorded allergy to Ketoconazole.
            [contra] Levoketoconazole  Levoketoconazole is in the same ATC class (H02CA) as the
                                       patient's allergy to Ketoconazole — possible cross-reactivity
AFTER  (1)  [contra] Ketoconazole      The patient has a recorded allergy to Ketoconazole.
```

Re-run: `2 → 1` again.

### The controls — all exact, all unchanged

| probe | expected | before | after |
|---|---|---|---|
| Mary + clarithromycin | 1 Major × simvastatin | 1 | **1** (same text) |
| Mary + paracetamol | 0 | 0 | **0** (and 0 on re-run) |
| Agnes + warfarin | 1 Major × aspirin | 1 | **1** (same text) |
| Joshua + ibuprofen | 2 (NSAID contra + lisinopril Moderate) | 2 | **2** |
| Betty + bupivacaine | 1 Major × lidocaine | 1 | **1** |
| Margaret King | names `Botulinum toxin type A` | 1 | **1**, same name |
| Dorothy Odinga | contraindication leads | 3 | **3**, contraindication still first |
| Richard Jones | 1 | 1 | **1** |
| Susan Young | 3 — two `Lidocaine` + one correct Tiotropium finding | 3 | **3**, all three kept |
| Agnes + epinephrine | 0 | 0 | **0** |
| James Thompson | co-trimoxazole allergy found | 2 | **2** |
| Helen Roberts | two phrasings, same mechanism | 1 + 1 | **1 + 1**, same Major × methotrexate |

*Had the fix over-narrowed*, these are where it would show: Joshua's cross-reactivity chip, Dorothy's
aspirin chip, Susan's two Lidocaine chips and James's `Sulfamethoxazole (sulfamethazine)` chip all
depend on a name resolving to a substance whose display name it is **not**.

### Three "controls" that move — because each contained a live instance of the defect

| probe | before | after |
|---|---|---|
| George Anderson + esomeprazole | 2 — `Omeprazole` … A02BC … allergy to Pantoprazole, **and** `Esomeprazole` … | **1** — `Esomeprazole is in the same ATC class (A02BC) as the patient's allergy to Pantoprazole` |
| Melissa Wright + esomeprazole | 2 — `Omeprazole` cross-reactivity **and** `The patient has a recorded allergy to Esomeprazole.` | **1** — the identity chip, which **says `Esomeprazole`** as the control requires |
| Kevin Brown + hydrocortisone | 4, one of them `Hydrocortisone butyrate is in the same ATC class (H02AB) as the patient's allergy to Hydrocortisone (ophthalmic)` | **3**, ester gone, other three identical |

George's and Melissa's `Omeprazole` chips are the cases the eight failing assertions defend. Kevin's is
the ester again, from an allergy rather than an order — a third independent live instance.

So on stock data the defect fired on **7 of 20 distinct probes**, and after the change no probe names a
substance that is neither asked about nor prescribed.

### What I could not verify live

- The injector leg. The REST response returns only **cited** references, so the number of injected
  drug-reference records is invisible from a probe (as `DrugReferenceInjector.collect`'s own javadoc
  records). It is covered by `theInjectedReferenceRecordsCarryOnlyTheSubstancesTheQuestionNames`.
- Answer prose. Wording differs run to run on this box; every claim above is about chips, the WARN, or
  the withheld-pair list.

### End state

The standalone was left on **`main` at `ae4d2bd7`**, rebuilt from a clean worktree at that SHA — omod
md5 `e04947109ef783363e0cfb393760f336`, which differs from the `7443caab4e2d82e89662bd6cee462f33` that
was on the box only because the omod zip is not byte-reproducible; the source SHA is identical and the
worktree is clean. All 53 `chartsearchai.*` GPs are exactly as found. No demo data was created, voided
or edited at any point.

Also measured on that restored build: Sarah's screening question, so the 19-versus-15 count above is
attributed rather than asserted.

dkayiwa:
### Correction / attribution: Sarah's screening chip count

My comment above said the 19 → 15 on her screening question was "1 ester chip + 3 answer-driven", with
the branch measured stable at 15. That was the right conclusion for the wrong reason, and the restored
`main` settles it properly.

Same question, same build, same JVM, twice:

```
main run 1 : 19 chips  (6 contraindications incl. the ester, 3 Hydrocortisone x NSAID, 10 screened)
main run 2 : 16 chips  (6 contraindications incl. the ester, 0,                        10 screened)
branch     : 15 chips  (5 contraindications,                  0,                        10 screened)  x4
```

So the count varies **on `main` alone** — the three `Hydrocortisone x celecoxib / diclofenac / ibuprofen`
chips come from the answer-driven drug-in-play arm and appear only when that run's answer names
hydrocortisone outside echo scope. Against `main`'s own 16, this branch's 15 is a difference of exactly
**one chip: the ester** — which is the claim, and the only chip-level claim that survives the
non-reproducibility of answer prose on this box.

The deterministic signals are unchanged and are what the fix should be judged on:

- `18 active-order reference entries` → `17`, on every pass of every run;
- `18 pair(s) above the severity floor` on both builds, with a character-identical withheld list, so no
  interaction finding was lost;
- `Hydrocortisone butyrate` present in all three `main` runs of that question and absent in all four
  branch runs.

Also confirmed on the restored `main`: Sarah + hydrocortisone back to **9** with the ester chip, and the
four regression controls at 1 / 1 / 2 / 0. The instance is on `main` at `ae4d2bd7`.

dkayiwa:
### Ruling applied: the eight assertions corrected, per test, with the property preserved

Checked each one against the ruling's shape before touching it. **None is a counter-example** — not one of
the eight asserts, as its own subject, that both substances should be admitted. Three of them (`…StayTwoChips`
×2 and `…KeepTheirOwnRecords`) do assert two chips/records as their subject, but the subject is
**non-collapse**, not admission — so for those the correct fix was to keep the number and change the vehicle.

| test | property it asserts | still asserted where | what moved |
|---|---|---|---|
| `ContraindicationRouteVariantTest.routeVariantsOfOneActiveOrderRaiseOneChipPerSubstance` | rows of ONE substance are ONE chip on the order-driven arm (#145) | the single surviving chip, named by `displayLabel()` | 2→1: the ester chip from a `Hydrocortisone Injection vial 100mg` order — #209 stated as an expectation |
| `.oneSubstanceStillReportsEveryRecordedFindingSeparately` | the ledger key includes the recorded FINDING → one subject, two findings, two chips | exactly the two chips that remain | 4→2: two ester chips, admitted by the parent's name |
| `.twoDistinctSubstancesTheKbFilesUnderOneSubstanceNameStayTwoChips` | `substanceKey` must never merge the two PPIs (#121) | **unchanged, 2 chips** | the *vehicle*: the question now NAMES BOTH, so the case cannot be satisfied by over-admission |
| `InteractionRouteVariantTest.twoDistinctSubstances…StayTwoChips` | same, subject side, incl. "the notes are byte-identical so only the subject key separates them" | **unchanged, 2 chips, all four assertions** | the vehicle |
| `ReferenceRecordSubstanceCollapseTest.…KeepTheirOwnRecords` | two substances → two injected records; merging would hand the model one entry's interactions under another's name | **unchanged, 2 records** | the vehicle |
| `AllergenExactNameResolutionTest.anAllergyToOneOfTwoSubstancesUnderOneNameIsNotAttributedToTheOther` | the recorded allergy is attributed to the substance the CHART names (#176) | the surviving chip | 2→1: the chip the case itself called "the cross-reactivity hedge" |
| `.aRecordedNameNoEntryIsNamedStillResolvesToTheEarliestMatchingEntry` | a localized allergen no entry is NAMED still resolves (#147) | the `lookupByToken` assertions, plus the surviving chip naming `Pantoprazole` | 2→1: the `Omeprazole` chip |
| `.twoRecordedAllergiesToOneSubstanceStillRaiseOneChipPerSubject` | two allergy RECORDS naming one substance are ONE chip, named after the first (#145/#160/#187) | the single surviving chip states it directly | 2→1: the ester chip |

**Both invariants survive, proved by mutation** on a throwaway worktree, not asserted:

| mutation | caught by |
|---|---|
| **M7** `substanceKey` reduced to its first component — the two PPIs merge | all three non-collapse cases whose vehicle changed, `theTwoPpiRowsShareTheSubstanceNameTheStemHasToVeto`, `aRecordedNameNoEntryIsNamed…`, and two `SubstanceCandidateSetTest` cases |
| **M8** prose leg reverted | the four prose-side corrections + three of my own |
| **M9** order-name leg reverted | `routeVariantsOfOneActiveOrderRaiseOneChipPerSubstance` + two of mine |

So each corrected expectation is load-bearing in its new form, and the non-collapse property that the three
`…StayTwoChips` cases exist for is still guarded by those same three cases.

**Suite: 1019 api + 61 omod = 1080, 0 failures, 34 skipped, BUILD SUCCESS.** Baseline on `main` `ae4d2bd7`
was 1010 + 61 = 1071, so +9 and no regression.

### Also folded in

- **Eighth upstream KB defect, for #196:** `Pfizer-BioNTech Covid-19 Vaccine` carries **`moderna covid-19
  vaccine`** among its six aliases, and all five `Tozinameran` rows inherit it — so the KB asserts one
  manufacturer's vaccine is also called the other's. Measured through `aliasesIn`/`nameMatchStrength`. This
  PR's ranking blocks the **prose** direction (the Moderna row claims that name at strength 2, the Pfizer row
  at 1, so a question about Moderna no longer resolves the Pfizer family). It does **not** repair the data:
  `isNamed` is exact name identity over the whole alias list, so a rule token spelled
  `moderna covid-19 vaccine` still identifies the Pfizer entry. Not worked around here.
- **GP count:** measured **52** `chartsearchai.*` global properties, not 53. Byte-identical before and after.

dkayiwa:
### Live re-verification of the corrected branch, and the end state

Deployed the branch as it now stands (omod md5 `129141096c1a1a78e04118e77add2c66`; `api/src/main` and
`omod/src` are byte-identical to the build probed earlier — the four commits since touched only test
files — so this is a confirmation pass, run because the numbers are worth confirming rather than
inferring). Gate passed: querystore `hits=75`, `entryCount=2283`, `sourceFormat=ddinter`, control chip
present.

| probe | on `main` | on the branch |
|---|---|---|
| **Sarah + hydrocortisone** | 9 | **8** — `Hydrocortisone` ×4 (1 contraindication + celecoxib/diclofenac/ibuprofen Moderate), `Dexamethasone`, `Prednisone`, `Budesonide`, `Methylprednisolone`. The ester chip gone; her four `Hydrocortisone` chips intact. Re-run: **8** again, identical subjects. |
| **Betty + ketoconazole** | 2 | **1** — `The patient has a recorded allergy to Ketoconazole.` |
| **George + esomeprazole** | 2 | **1** — `Esomeprazole is in the same ATC class (A02BC) as the patient's allergy to Pantoprazole`; the `Omeprazole` chip gone |
| **Melissa + esomeprazole** | 2 | **1** — `The patient has a recorded allergy to Esomeprazole.` |
| **Kevin + hydrocortisone** | 4 | **3** — `Hydrocortisone (ophthalmic)` identity, `Hydrocortisone × clarithromycin`, `Clarithromycin` identity; the ester chip gone |
| Mary + clarithromycin | 1 | **1** |
| Agnes + warfarin | 1 | **1** |
| Joshua + ibuprofen | 2 | **2** |
| Mary + paracetamol | 0 | **0** |

Each of the five moving probes loses **exactly** the defect instance and nothing else; the four
regression controls are unchanged.

**End state:** the standalone is back on **`main` at `ae4d2bd7`**, rebuilt from a clean worktree at that
SHA, re-probed after the restore (Sarah back to 9 with the ester chip, controls 1 / 1 / 2 / 0). The 52
`chartsearchai.*` GPs are `diff`-identical to the snapshot taken before any of this work. No demo data
was created, voided or edited at any point.

dkayiwa:
## Hardening pass — fresh eyes on the ruling, and one defect the branch introduced

I did not write this code. Everything below marked *measured* I measured myself in my own worktree
through production predicates (never a reimplementation); nothing here is relayed unverified.

Five commits: `a85239b3` production prose · `11edda30` test prose + two additions · `768f700d` one
deletion of my own prose · `2290517c` two javadoc paragraphs · **`5d0da17c` a production fix**.

- baseline on `main` `ae4d2bd7`, measured in a clean worktree: **1010 api + 61 omod = 1071**
- this branch now: **1021 + 61 = 1082**, 0 failures, 34 skipped (the PR's 1080, +2 new cases)

---

## The one thing that changes behaviour: the narrowing could EMPTY a candidate set

`findImpliedByQuery`'s javadoc states the invariant "**It cannot empty a non-empty set**" and gives the
argument. **The argument does not hold, and I measured it not holding.** It needs the strongest claimant
on a carried alias to carry that alias itself, and so to be in `matched` too — which is a property of the
**parsers**, not of the filter. `DdiDrugReferenceSource` makes an entry's display name its first alias
and `AtcDrugReferenceSource` makes it the only one, so on both of those every entry names itself. A
hand-authored `json` dataset need not, and **`json` is the default `sourceFormat`**
(`ChartSearchAiConstants:365`). `nameMatchStrength`'s own javadoc already records that shape as the one
its gate excludes.

Driven through the real `JsonDrugReferenceSource.parse` and the real accessor, on a dataset where
`Ibuprofen` publishes only the stem `ibuprof`:

```
findByQuery("is ibuprofen 400mg safe?")         [Ibuprofen tablets, Ibuprofen suspension]
findImpliedByQuery("is ibuprofen 400mg safe?")  []
```

The stem row is the rank-2 claimant on `ibuprofen` — the recorded-name rule reaches it by its two-letter
inflection allowance — while the **prose** rule never reaches it, so it is the strongest claimant on the
word *and* absent from the candidate set for that same word. Neither presentation is NAMED the word, so
every matched row is dropped.

That is the worst answer available: the drugs-in-play set is what every arm iterates, so a question
naming a drug gets **no contraindication, no interaction and no overdose check, with nothing in the log
to say so** — the silent-and-closed failure the whole feature exists to prevent. On `main` the same
question produces 2 subjects and 4 chips; on this branch before the fix, 0.

**Fix:** `rowsOf` returns `matched` when nothing is in play, so the invariant is enforced where it is
stated instead of inherited from a parser. That is the pre-#209 answer for that one shape, and it
over-reports — the safe direction for a non-blocking advisory. TDD: the case fails first
(`narrowingNeverEmptiesACandidateSetEvenWhenNoMatchedRowIsNamedTheWord`), on a fixture parsed by the real
production parser.

**Zero effect on any shipped dataset**, re-derived after the change:

```
ddinter full 19MB   prose 33/71  name 33/71  FALLBACK_FIRED 0/0  sha256=8ae3505a…  ← identical to pre-fix
ddinter bundled     prose  0/0   name  0/0   FALLBACK_FIRED 0/0
curated json        prose  0/0   name  0/0   FALLBACK_FIRED 0/0
```

(signature = SHA-256 over every name's resolved row set on both legs; firings counted over 7452 names
and aliases on the full KB). It also makes the `size < 2` fast path *provably* a no-op rather than a
no-op by parser accident, so that comment is now true too.

---

## The ruling was right — on all eight

I read each of the eight at `ae4d2bd7` and at `1bac410c` in full. **None asserts admission as its own
subject.** The two closest calls:

- `routeVariantsOfOneActiveOrderRaiseOneChipPerSubstance` — its count assertion's message did say "…**and
  the ester stays its own chip**". But the old comment justified that chip by cross-reference ("the same
  refusal that keeps Omeprazole and Esomeprazole apart **below**"), i.e. it deferred the property to the
  dedicated case rather than owning it. Vehicle.
- `oneSubstanceStillReportsEveryRecordedFindingSeparately` expected 4 under the message "two substances x
  two findings", which contradicted the test's own **name**. The corrected "one substance x two findings"
  is what the name always said.

## The three vehicle changes: EQUIVALENT plus a new guard — and the stated reason is false

Verdict per case: **equivalent** on the non-collapse property, **plus** a new over-narrowing guard, **not
weaker**. But the justification written into two of them —

> it is now a stronger vehicle, because it can no longer be satisfied by over-admission

— **is false, measured false.** For the corrected question the unranked and ranked row sets are identical:

```
OLD "…give esomeprazole?"               unranked=[Omeprazole, Esomeprazole]  ranked=[Esomeprazole]
NEW "…give omeprazole or esomeprazole?" unranked=[Omeprazole, Esomeprazole]  ranked=[Omeprazole, Esomeprazole]
```

An over-admitting resolver satisfies the corrected case. The true statement is that the case no longer
*depends* on over-admission, and that it newly catches the opposite direction. Corrected in both files;
`ReferenceRecordSubstanceCollapseTest`'s wording was already accurate and left alone.

Also false, and in the worst possible place — the comment explaining the fix's central rule:

> `Hydrocortisone Injection vial 100mg` denotes the parent alone (**nameMatchStrength 2 against 1**)

Measured on that fixture: against the **whole string** both rows score `NAME_TOKEN_INSIDE_A_NAME` (**0**)
— a recorded order name is nobody's name, which is the PR's own point. The 2-against-1 belongs to the
**carried alias** `hydrocortisone`. As written it asserted the opposite of the rule it was explaining.

## Chip-level non-collapse per class, and the gap the corrections left

| class | chip-level non-collapse |
|---|---|
| `ContraindicationRouteVariantTest` | yes, question arm — **and now the order arm too** (added) |
| `InteractionRouteVariantTest` | yes |
| `ReferenceRecordSubstanceCollapseTest` | record layer only, in the vehicle-changed case. Keeping that count at 2 was **correct**: drop it and the layer asserts nothing |
| `AllergenExactNameResolutionTest` | **no** — key-level `assertNotEquals` only. Accepted; its cross-reference points at a case over the *same* fixture (`DDI_CONTRA_ROUTE_VARIANTS`), which I verified |
| `AllergenNameResolutionTest` | n/a |

Correction 1 removed the only chip-level cross-substance non-collapse case on the **order-driven** arm.
One order name no longer reaches two substances, so the shape now needs two orders —
`twoOrdersOfTwoSubstancesUnderOneAliasStayTwoChips`, added, and a canary rather than a control.

## Both invariants, re-derived

`substanceKey()` = `Arrays.asList(normalizeName(substanceName), drugbankIdOrDisplayStem)`. Reducing it to
its first component makes both PPI rows `[esomeprazole]` (the `Omeprazole` row's own `rxnorm_name` **is**
`esomeprazole`) and both hydrocortisone keys `[hydrocortisone]`. **M7 confirmed to merge them.** Run as a
mutation it fails **10** sites — the three vehicle-changed cases, both key-level `assertNotEquals`, two of
the PR's new cases, and **three of my additions**. `Omeprazole`/`Esomeprazole` stay two substances and two
substances stay two chips.

## KB-wide, measured

| | |
|---|---|
| names narrowed / rows dropped, prose leg | **33 / 71** |
| same, order-name leg | **33 / 71**, `legDisagreements=0` |
| non-empty sets emptied · substances losing a row · names losing a denoted substance · drops sharing a display stem with anything kept | **0 · 0 · 0 · 0** |

The PR's figures reproduce exactly. **Extended past what the PR measured:** over names *and aliases*
(7452 strings) and over **all four** shipped datasets, still 0 emptied and 0 row-losses — and
`nameNotInOwnAliases = 0` on every shipped source, which is what makes the emptying shape above
unreachable as shipped rather than merely unobserved.

### The other direction — I tried hard to find a chip that should survive and does not

- **Every one of the 2283 entries is still reachable by its own published display name**, both legs:
  `rowUnreachableByOwnDisplayName = 0`, `substanceUnreachable = 0`, **0 regressions**. This is the one
  check that is not circular with the filter's own rule.
- Of every (entry, alias) pair the unranked scan admits and the ranked one drops: **26 justified** (a
  strictly stronger claimant exists), **0 unjustified**.
- Live, the probes that depend on a name resolving to a substance whose display name it is **not** all
  survive: Joshua's chip still names `Acetylsalicylic acid (aspirin)` from the word *aspirin*, Margaret's
  still names `Botulinum toxin type A`, Susan keeps both `Lidocaine` chips.

**I found none. The claim holds.**

### 85 versus 33 — both correct, different questions

| question | figure |
|---|---|
| **A** — names where the unranked prose set holds a substance `findImpliedSubstances(WHOLE STRING)` does not report (**the issue's**) | **85** |
| **B** — names where `findImpliedByQuery` returns fewer rows than `findByQuery` (**the fix's**) | **33** |
| A but not B | **52** |

Those 52 are exactly the names the *whole-string* rule would narrow and the *name-carried* rule refuses
to — the `aCombinationOrderNameKeepsEveryConstituentSubstance` direction. The gap is the design decision,
not a shortfall. Recorded in the javadoc, because a reader comparing the issue's 85 with the PR's 33
otherwise reads a fix as half-done.

### And read the 71 with its cause attached

**7 of the 33 names and 42 of the 71 rows are the single #196 data defect** — the Pfizer row publishing
`moderna covid-19 vaccine`, inherited by all five `Tozinameran` rows. Excluding it the ranking moves
**26 names / 29 rows**. 59% of the headline row count is one bad alias's consequences, not a reshaping of
clinical data. Also recorded.

## Mutation ledger (throwaway worktree; both files verified byte-restored after)

| mutation | prose | order | signature |
|---|---|---|---|
| BASELINE | 33/71 | 33/71 | `8ae3505a…` |
| **M4** only the first carried alias resolved | 33/71 | 33/71 | `8ae3505a…` **byte-identical** |
| **M6** prose witness under the order-name rule | 33/71 | 33/71 | `8ae3505a…` **byte-identical** |
| M5b per-ROW verdict | 34/72 | 34/72 | `f889e0bc…` |
| M1 prose leg reverted | 0/0 | 33/71 | `3ee22477…` |
| M2 order leg reverted | 33/71 | 0/0 | `f42a8032…` |
| fast path removed | 33/71 | 33/71 | `8ae3505a…` |

**M4 and M6 confirmed unfalsifiable on this KB, independently.** Keeping them as the correct reading of
the rule rather than claiming measurement is right. One thing worth adding, because it says which to
watch: the shipped code sits **between** them — M4 resolves fewer aliases so its result can only be a
subset, M6's witness rule is more tolerant so its result can only be a superset. M4 is the side that
could silently *drop* a substance if a later KB separated them, which matches the PR's own note.

## Performance, re-measured; and #172 structurally avoided

Question **1.81 → 3.67 ms**; a pathological 8.5 KB answer matching 308 entries **134 → 446 ms**. Same
ratios as the PR's on a differently-warmed JVM. Against 15–90 s of LLM time, not a cost worth a decision.

`DrugReferenceService` still has exactly the four fields it has on `main` — I enumerated them rather than
trusting the comment. The resolution cache is a parameter threaded from the outermost caller's local, so
nothing survives a `getAll()` hot reload.

Also verified, because it is the PR's central claim: `matchesText`, `matchesDrugName`, `containsWord`,
`containsBoundedToken`, `nameMatchStrength`, `substanceKey`, `findImpliedSubstances` and `lookupByToken`
are **byte-identical to `main`**, and `findByQuery`/`findByDrugName` differ only in their visibility
keyword. The change is purely additive at the composition layer, which is what the issue asked for.

## `findByQuery` / `findByDrugName` narrowed to package-private

The author deferred this as out of slice. I disagree, and did it. This branch **adds a CLAUDE.md rule**
saying these are primitives and no arm may build a candidate set from them; unenforced, that rule is
exactly the convention whose breach *is* #209. Every caller, production and test, is already in
`org.openmrs.module.chartsearchai.reference` — no caller in `omod`, no `getRegisteredComponent` for this
bean anywhere, no `<service>`/`<advice>` in `config.xml`. It is a binary-incompatible reduction with zero
measured consumers; if you would rather keep the public surface it is a two-keyword revert.

## Live re-verification — required, and run on both builds

Production files changed, so I ran it. Deployed omod md5 `3d9de0285ad56c57ac35a5e4b57ca008`. Gate passed:
querystore `hits=75`, `entryCount=2283`, `sourceFormat=ddinter`, control chip present. Sarah's chart
confirmed **from the database** (§9 — the order REST endpoint under-reports): 8 active orders, exactly one
`Hydrocortisone Injection vial 100mg`, plus a Dexamethasone allergy.

**The headline set — every one moves by exactly its defect instance. Both columns measured by me.**

```
                                main   branch
Sarah  + hydrocortisone            9 ->  8    (re-run: 8, identical subjects)
   KEPT  Hydrocortisone x4 — contraindication (H02AB vs Dexamethasone allergy)
                             + celecoxib / diclofenac / ibuprofen, Moderate
   KEPT  Dexamethasone (recorded allergy), Prednisone, Budesonide, Methylprednisolone
   GONE  "Hydrocortisone butyrate is in the same ATC class (H02AB) as the patient's
          allergy to Dexamethasone — possible cross-reactivity"
Betty  + ketoconazole              2 ->  1    "The patient has a recorded allergy to Ketoconazole."
                                              (GONE: "Levoketoconazole is in the same ATC class (H02CA)…")
George + esomeprazole              2 ->  1    "Esomeprazole is in the same ATC class (A02BC) as the
                                               patient's allergy to Pantoprazole"  — Omeprazole gone
Melissa+ esomeprazole              2 ->  1    "The patient has a recorded allergy to Esomeprazole."
Kevin  + hydrocortisone            4 ->  3    Hydrocortisone (ophthalmic) identity,
                                               Hydrocortisone x clarithromycin, Clarithromycin identity
```

*Had the fix been wrong:* a one-leg fix leaves Sarah at 9 either way — narrow only the prose leg and the
order arm picks the ester up, narrow only the order leg and the prose leg keeps it in play.
Over-narrowing shows as fewer than 8, specifically as losing one of her four `Hydrocortisone` chips.

**The fourth call site's deterministic signal** — the part that does not depend on answer wording:

```
main    Interaction screening across 18 active-order reference entries found 18 pair(s) …
branch  Interaction screening across 17 active-order reference entries found 18 pair(s) …
```

**17 entries from 8 orders, 18 pairs unchanged, withheld list character-identical**, on both passes of
the run. So the dropped entry contributed no pair and no finding was lost.

I am **not** offering that probe's chip count as evidence — you already showed `main` giving 19 then 16
for one question in one JVM. Mine read 15, matching your branch value, but it moves on its own.

**Controls, all exact:** Mary+clarithromycin 1 Major ×simvastatin · Agnes+warfarin 1 Major ×aspirin ·
Joshua+ibuprofen 2 · Mary+paracetamol 0 (and 0 on re-run) · Betty+bupivacaine 1 Major ×lidocaine ·
Susan 3 — two `Lidocaine` plus the correct `Tiotropium` finding, **third kept** · Margaret names
`Botulinum toxin type A` · Agnes+epinephrine 0.

**The restore is the discriminator, not a formality:** back on `main` the ester chip and
`Levoketoconazole` are *present* again (9 and 2) and the controls read 1/1/2/0 — which is what proves the
before/after deltas are the fix and not an artefact of the instance.

## Two figures in my brief that the instance contradicts

Reported with the method rather than adopted:

- **querystore api jar**: my brief said `38205608891f10b68929c35fb3eace7e`. The deployed
  `querystore-1.0.0-SNAPSHOT.omod` is `24c8aca7e657260c949837fd72aeb811` — matching the value recorded in
  this PR exactly — and its **embedded** `querystore-api-1.0.0-SNAPSHOT.jar` is
  `a2298417574b2bd3882a8abdcbd6bcc2`, the same as the lib-cache copy. The omod file is dated Jul 14 and
  untouched, so querystore is unchanged and the brief's api-jar figure is the stale one.
- **GP count**: **52** `chartsearchai.%` rows, counted in the database. Your 52 is right; 53 is not.

## Found, not fixed

1. **The attribution layer still asks the unranked predicate over a now-ranked set.** Two consequences,
   both in the safe direction, both **reasoned not measured**:
   (a) `resolvesFrom`'s name leg is a strict superset of what `findImpliedByDrugName` admits, so
   `Hydrocortisone Injection vial 100mg` still reads as `Hydrocortisone butyrate`'s own order — which
   withholds that order as the ester's interaction *partner*, suppressing a pair rather than inventing
   one. I documented the residue in its javadoc and left the behaviour alone.
   (b) `activeOrdersOtherThan`'s co-resolved-codes proxy iterates the now-narrowed order-entry list, so
   `ownCodes` is smaller and `codes.removeAll(ownCodes)` suppresses less — an ATC code that only the
   dropped substance supplied now survives and could witness one extra chip. Reaching it needs a dropped
   row carrying a chart-held code that no surviving row of that order carries. No probe of mine gained a
   chip anywhere.
2. **`isEchoOfCitedRecord` reads the cited record's TEXT unranked** — correctly, since ranking it
   re-opens #105. Residue: a cited record naming `hydrocortisone` can mark an in-play
   `Hydrocortisone butyrate` as an echo and exempt it, so a withheld interaction/overdose chip. Slightly
   widens the residue its javadoc already bounds. Reasoned.
3. **The ranking now separates stereoisomer / ester / prodrug pairs that plausibly share an interaction
   profile** — `Dexfenfluramine`/`Fenfluramine`, `Dextroamphetamine`/`Amphetamine`,
   `Ketoconazole`/`Levoketoconazole`, `Mometasone`/`Mometasone furoate`,
   `Isosorbide`/`Isosorbide mononitrate`, `Fenofibrate`/`Fenofibric acid`. Correct for "which substance
   does this name NAME"; the curated cross-reactivity groups are where the clinical relationship would be
   stated, and they cover none of these. **Measured**: 56 (name, leg) pairs where a dropped row carried a
   rule no kept row carries. A curation question, not a code defect — worth its own issue.
4. **A second unfalsifiable assertion, pre-existing.** `InteractionRouteVariantTest:152` —
   `assertFalse(warnings.get(0).getDetail().contains("topically administered"))` two lines after that
   value is pinned to an exact literal. Cannot fail. From #173, not this PR, so left alone; the identical
   one this PR *added* I fixed.
5. `docs/adr.md:1771` names `activeOrderEntries`, a symbol that no longer exists. Stale documentation.
6. `DrugSafetyDiacriticOrderNameTest:51`/`:288` say the order leg resolves through `matchesText` — false
   since #147, not since #209. Pre-existing.
7. **A whitespace-only alias can be a witness but never a justification** (`containsWord(", ", " ")` is
   true while `normalizeName(" ")` is null, so `namesSubstanceOf` skips it). With ≥2 matches such an entry
   is dropped where `findByQuery` kept it. `json`-only, needs an unsanitized dataset — the shape
   `containsBoundedToken`'s javadoc already records. The new fallback bounds the worst case to "not
   emptied".

## End state

Standalone restored to **`main` at `ae4d2bd7`**, rebuilt from a clean worktree at that SHA and re-probed
after the restore. **52** `chartsearchai.*` GPs, unchanged. No demo data created, voided or edited at any
point.

dkayiwa:
### Addendum — a convergence pass found three of my own corrections false, and one bound I overstated

Posting this separately because it revises the comment above, and because two of the three are the exact
error I criticised the original corrections for. All three measured false through production predicates.

1. **"neither is named it"** — in the new fixture's `description`, in the test comment, and in the test's
   own NAME. Measured: both `Ibuprofen tablets` and `Ibuprofen suspension` have
   `isNamed("ibuprofen") == true`, `nameMatchStrength == NAME_IS_ANOTHER_NAME` — which is precisely why
   they are in the candidate set at all. The true statement is that neither is **display**-named it, so
   neither is the strongest claimant. Renamed
   `…EvenWhenNoMatchedRowIsTheStrongestClaimant`.

2. **"one order name no longer reaches two substances, so the shape now needs two orders"** — false, and
   it left the new case's first order inert. Measured: `Hydrocortisone butyrate cream 0.1%` carries
   **both** `hydrocortisone butyrate` and `hydrocortisone`, each the display name of one of the two
   substances, so `findImpliedByDrugName` returns all four rows and the **ester order alone** already
   yields both chips. Reduced to that one order, renamed
   `anOrderNamingTwoSubstancesOutrightKeepsAChipForEach`, and re-verified after the reshape: still fails
   under `substanceKey` reduced to its first component, still passes with **both** legs reverted. Canary
   for the key, control for the legs — which is what its comment claims.

3. **"Emptying is the worst answer available"** / **"only bounds what an operator-authored dataset can
   do"** — I overstated the `rowsOf` fallback. It bounds emptying and nothing more. On the same fixture,
   an order for `Ibuprofen tablets 400mg` keeps only the **rule-less** bare `Ibuprofen` row and drops both
   rule-bearing rows; `inPlay` is non-empty so the fallback never fires, and every finding from the
   reference data is lost anyway. Measured by reverting the order leg on that fixture: **5 chips → 1**,
   the survivor being the allergen-list chip, which does not read the rows' rules.

   Not fixed, deliberately. It needs the same precondition as the emptying shape — an entry whose aliases
   omit its own name, measured **0** on the full 19 MB KB and on each of the three bundled samples — and
   closing it means deciding what a source publishing no substance name should mean by "one substance",
   which #209 does not settle. Recorded beside the fallback rather than argued away.

Also: the `assertFalse` loop I added is **still dead**, and I was wrong to call converting it a fix.
Rewriting `warnings.get(0)` as a loop over `warnings` does not restore reachability when the *count* is
pinned at 1 and `get(0)`'s detail is pinned to an exact literal two lines above. Deleted — those two
assertions already prove the ester's absence, and a third that cannot fail is not evidence. (The
pre-existing `InteractionRouteVariantTest:152` has the same shape and I left it, as reported.)

**Suite after all of it: 1021 api + 61 omod = 1082, 0 failures, 34 skipped.** Baseline on `main`
`ae4d2bd7`, measured by me in a clean worktree: **1010 + 61 = 1071**.

No production **behaviour** changed after the `rowsOf` fallback (`5d0da17c`) — everything since is prose,
one test reshaped and re-verified, and one dead assertion removed. The live evidence above therefore still
stands.

dkayiwa:
Independent pre-merge verification: `mvn -o clean install` on a clean detached worktree at `e4250514` → **1021 + 61 = 1082, 0 failures, 34 skipped**, BUILD SUCCESS. `main` at `ae4d2bd7` = 1071. CI green on Java 11/17/21.

**The hardening found a defect this PR introduced, and it is the most serious catch of this whole body of work.** On the **default** `sourceFormat=json`, `findByQuery` returned 2 entries while `findImpliedByQuery` returned **`[]`** — a total silent loss of every safety check for a question naming a drug. The documented invariant rested on a *parser* property (DDInter and ATC both put an entry's display name in its alias list) that hand-authored json need not satisfy, and json is the default. Fixed in `rowsOf`, with the KB signature byte-identical over the 19 MB file and the fallback firing zero times on every shipped dataset.

A fix for *a wrong chip* would have shipped *no chips at all* on the default configuration. That is the argument for hardening every PR in one sentence.

**On my ruling about the eight assertions: upheld on all eight**, read at both SHAs. None asserts admission as its own subject — the closest calls were `routeVariantsOfOneActiveOrderRaiseOneChipPerSubstance`, whose message mentioned the ester but whose comment deferred the property by cross-reference, and `oneSubstanceStillReportsEveryRecordedFindingSeparately`, whose old message contradicted the test's own name.

**But the reason given for the three vehicle changes was false, and I relayed it.** I repeated the claim that the corrected cases "can no longer be satisfied by over-admission". Measured: for `"…omeprazole or esomeprazole?"` the unranked and ranked row sets are **identical**, so an over-admitting resolver satisfies them fine. The true and narrower claim is that they no longer *depend* on over-admission and newly catch over-narrowing. A second statement was also wrong and worse-placed — that `Hydrocortisone Injection vial 100mg` denotes the parent alone by strength 2 against 1; against the **whole string** both rows score **0**, and the 2-versus-1 belongs to the carried alias, which is this fix's central rule. Both corrected in the branch.

Also measured, and it reframes the change: **7 of the 33 narrowed names and 42 of the 71 narrowed rows are the single `moderna covid-19 vaccine` alias defect** (#196's eighth item). Excluding it, the fix narrows 26 names and 29 rows. So more than half the measured effect is one bad row of upstream data.

**No chip that should survive was dropped** — all 2283 entries remain reachable by their own display name on both legs, 0 regressions, 26 justified alias drops and 0 unjustified. The 85-versus-33 figures are reconciled as different questions: A=85, B=33, A-not-B=52.

It also took a decision I delegated, and took it against the author's deferral: `findByQuery` and `findByDrugName` are **narrowed to package-private**, on the grounds that this PR adds a CLAUDE.md rule which is otherwise unenforceable and #209 *is* that convention being breached. Zero consumers outside the package, measured.

Two further items filed: **#211** (the new fallback bounds emptying only — the same dataset shape can still drop every rule-bearing row while keeping a rule-less one, measured 5 chips to 1; latent on shipped data, reachable on the default `sourceFormat`) and **#212** (a pre-existing assertion in `InteractionRouteVariantTest` that cannot fail — the thirteenth instrument of the class #179 collected).

One correction back: the querystore md5s are three different artifacts, not a conflict. `38205608…` is the `~/.m2` resolved api jar and still matches; `24c8aca7…` is the deployed querystore omod; the third figure is the api jar embedded inside that omod. And the GP count is **52**, as measured here — my 53 was wrong.

PR #188, opened 2026-08-07
fix(drug-safety): complete the one-substance-one-row sweep (#174 sites 2–4, #186, and a fifth site)

dkayiwa:
# Complete the one-substance-one-row sweep

Closes #174
Closes #186

## The shared root

Five places still turned one clinical fact into N of something, because they enumerated the
reference data's ROWS (or an order's ATC CODES) rather than the substance or the order those rows
stand for. #145/#160 fixed the contraindication chips, #173 the interaction subject and the injected
records, #182 the class chip's partner label (#174 site 1), #187 substance identity. This is the
remainder: #174 sites 2, 3 and 4, #186, and one further site of the same shape that #174 does not
enumerate.

Nothing new was invented. `DrugReference.substanceKey()` / `substanceGroupKey()` / `canonicalRow()`
/ `namesNoRoute()` and the two identity ledgers (`ContraindicationChips`, `InteractionPairs`)
already existed; this reuses them, adds no third dedup mechanism, and keys nothing on rendered text.
The only addition is a fold — `DrugReference.canonicalRow(Iterable)` — for the four surfaces that
now make the same choice.

### #174 site 2 — `DrugReferenceInjector.orderedInteractionNotes`: one note per partner ROW

The largest measured surplus in the cluster, and invisible from the REST response, which returns
only *cited* references.

Re-measured over the shipped 19 MB KB (2026-08-07, `appdata/chartsearchai/ddi_knowledge_base.json`,
2283 entries, 590,312 expanded rows after #152's self-pair guard drops 28): **1876 entries carry at
least one repeated partner and 19,316 rows are surplus**. `Ozanimod` carries 49 and its
COVID-vaccine token appears 6×. (#174 records 1872 / 19,228 from an earlier count; these are my own
figures and I have not reconciled the small delta.)

`orderedInteractionNotes` now collapses to one note per partner before anything is rendered:

* keyed on `DrugSafetyValidator.partnerLabel` case-folded — the key `bestRulePerPartner` already
  falls back to, and the very string the note prints, so the grouping and the rendering cannot
  disagree about what one partner is. That method is now **shared** rather than coalesced a second
  time, which also closes the untrimmed-label divergence its own javadoc named as the follow-up
  (`"warfarin   (Major. …)"` beside a chip reading `active order warfarin`);
* decided by `DrugSafetyValidator.outranksOnRule`, the rule-side half of the chip's own survivor
  rule (most severe, then the longer note), extracted so the record and the chip cannot rank one
[...]

Issue #179, opened 2026-08-06
The eval-harness test gate is inert: surefire excludes a @Tag that exists nowhere, and two eval suites never reach the reporter


dkayiwa:
## What

Three defects in how the eval harness is wired into the build, all of which make a gate look present while doing nothing:

1. **`surefire.excludedGroups=eval` excludes a JUnit `@Tag("eval")` that exists nowhere in the repo.** Nothing is tagged, so the exclusion matches nothing. Whatever it was meant to keep out of ordinary builds is either running or being skipped for some other reason, and the property gives a false impression that the boundary is enforced.
2. **Two eval suites never reach `EvalReporter`**, so their results are not collected — they can pass or fail without appearing in the report a reviewer reads.
3. **~185 KB of orphaned eval datasets** are committed but referenced by nothing.

## Why this is worth filing

The repo's own rules put a lot of weight on the eval harness, and #126 was specifically about *eval blind spots that fail open*. This is more of the same shape: a gate that reports success because it never ran. A harness nobody can trust is worse than no harness, because it gets cited as evidence — and it has been cited in this work.

Related: #126 (the original blind-spot issue, fixed by #133 — whose retrospective review is still owed) and #153 (a test literal duplicated six times, so a classifier change can silently reroute a subset).

Worth taking together with #133's outstanding review as one "can we trust our instruments" pass, rather than as three small fixes: the earlier survey also found `compare_arms.py` has no exit code and counts `CANNOT` as verdict-led, `prod_reasoning_ab.py` swaps a stale prompt replica so its `VERDICT: SHIP` is unfounded, `quality_panel.py` passes on zero comparable cells and cannot run at all (`lb.get_gp`/`GP_RATE` do not exist), `kv_prefill_gate.py` never verifies its cold arm is cold, and `build_gold_rc2.py` silently re-picks the cohort and overwrites the gold.

## Provenance

Found during the documentation-truth sweep (#177) and reported rather than fixed, since that PR is documentation-only. The `@Tag`, reporter-path and orphaned-dataset findings are its measurements; I have not re-derived them. The six older instrument findings are from an earlier survey in this session, recorded in the loop ledger.

dkayiwa:
**A fourth instrument is asserting things that are false**, found while hardening #192:

`eval/drift-metric/score_probe_safety.py:196-201` states that the comparison is `allergen == ref`, and that "even a direct allergy raises nothing". **Both are false:**

[...]

PR #213, opened 2026-08-09
test: make four blind checks able to fail (#203, #207, #212, #153)

### LINKED ISSUE #153: The screening-question literal is duplicated across six files, so isInteractionScreening can silently reroute some of them
## What

The literal question string used to exercise interaction screening is copy-pasted in **six** test files (four of them predating the #148 work). `isInteractionScreening` classifies a question onto one arm or another; because each file carries its own copy of the string, a change to that classifier can move some copies onto a different arm while the others keep passing, and nothing points at the divergence.

## Why it was not fixed in #148

Consolidating means editing five files that PR did not own, and adding a seventh copy in `TestDatasetHelper` without removing the six would make it worse. It was correctly deferred as out-of-slice rather than done as a drive-by.

## The fix

One shared constant (the repo already centralizes this kind of thing in `TestDatasetHelper`, per the API-surface rules in CLAUDE.md), with all six call sites moved to it in the same change.

## Severity

Maintainability, not behaviour — no user-visible defect today. Filed because the named consequence is specific: a future change to `isInteractionScreening` can silently reclassify a subset of the screening tests.

### LINKED ISSUE #203: AbsentDataEvalTest excludes every absent-data case it exists to test
## What

`AbsentDataEvalTest`'s `@MethodSource` filters on `!evalCase.isExpectedAbsent()`, so **all 19 `expectedAbsent` cases of its 29-case dataset are excluded.** The 10 that do run assert only:

```java
assertFalse(stripQueryStopwords(question).isEmpty())
```

Measured from the build log: **10 tests, 0 skipped** — so nothing reports as skipped and nothing looks wrong.

Its javadoc claims it tests that "the system produces a clear 'no records' answer naming what was asked about". The dataset carries `expectedAnswerContains` / `expectedAnswerNotContains` holding exactly those expectations, `EvalCase` exposes them, and **nothing reads them.**

## Why this is the eleventh instrument of the same class

#179 collected ten gates and scorers that report success without checking anything. This is the same defect in a JUnit test rather than a Python script, and it is arguably the worst of the set: it runs in CI on every build, it is green, and its name and javadoc both assert it is testing absent-data behaviour.

Absent-data behaviour is not incidental here — it is the property #94 is about (absence verdicts asserted from the similarity slice rather than the chart), and abstention quality is one of the two numbers the whole retrieval-tuning decision rested on (`topK` 30 → 12 was justified partly on abstention 0.93 vs 0.86). A test named for that property, passing, testing a stopword helper instead.

## Shape of the fix

The dataset and the expectations already exist, so this is wiring rather than authoring: run the 19 excluded cases and assert `expectedAnswerContains` / `expectedAnswerNotContains` against the real pipeline. Two things to decide deliberately:

- Those 19 cases need a real answer, so they need the LLM. The repo already has a skip convention for LLM-requiring suites (34 skipped today, all `PromptInjectionEvalTest` + `LlmAnswerQualityTest`), so the honest options are to join that convention — **skipped, but visibly skipped** — or to assert on the deterministic parts only and say so.
- Whichever is chosen, the assertion must be able to fail. Per §11 of the shared brief: name the mutation it would catch and prove by mutation that it catches it. This test's current state is precisely what happens when that step is skipped.

## Provenance

Found by the #179 instruments pass while auditing the harness, reported rather than fixed because it is a test file and the CLAUDE.md rule routes that decision to a human. The `@MethodSource` filter, the 10-of-29 count and the single assertion are its measurements against current `main`.

### LINKED ISSUE #207: chipSeverityRank fails open: reword one clause and the severity-ordering test asserts nothing
## What

`DrugSafetyQuestionPairInteractionTest.chipSeverityRank` locates a chip's severity by

```java
indexOf("also named in the question — ")
```

If that clause is ever reworded, the helper returns **−1 for every chip**, the ordering loop `continue`s past all of them, and `thePairChipsAreOrderedBySeverityAndBounded` **passes while asserting nothing.**

## Why this is filed separately

It is the **twelfth instrument of the class #179 collected** — a gate that reports success without checking anything — and the second one found inside a JUnit test rather than a Python script (#203 is the other: `AbsentDataEvalTest` excludes every absent-data case it exists to test).

It also has a specific trigger. The clause it string-matches is *clinician-facing prose in a chip*, and chip wording has been changed repeatedly in this work — #182, #188, #192, #198 and #205 all altered chip text. So this is not a distant hypothetical: the trigger is a routine action that several merged PRs have already performed on neighbouring strings.

## Shape of the fix

Two parts, and the second matters more:

1. Rank by the structured severity the chip already carries rather than by parsing its rendered text. `SafetyWarning` exposes it; the test is reaching past that to the prose.
2. Make the helper **fail loudly** when it cannot classify a chip, instead of returning a sentinel the caller skips. A test helper that silently yields "no opinion" converts a real assertion into a no-op, which is exactly the failure #179's ten instruments shared.

Per §11 of the shared brief, whatever replaces it should be proved by mutation: reword the clause and confirm the test now fails.

## Provenance

Found by #205's fix pass while surveying the tests around its change, reasoned from the code rather than measured by mutation. Reported rather than fixed because it is a test file and CLAUDE.md routes that decision to a human.

### LINKED ISSUE #212: A pre-existing assertion in InteractionRouteVariantTest cannot fail
## What

`InteractionRouteVariantTest:152` carries an assertion that **cannot fail** — pre-existing, introduced around #173, and unrelated to any current change.

## Why it is filed rather than absorbed

This is the **thirteenth** instrument of the class #179 collected: something that reports success without checking anything. The tally so far is ten Python gates and scorers (fixed in #202), `AbsentDataEvalTest` excluding every absent-data case it exists to test (#203), `chipSeverityRank` returning "no opinion" if one chip clause is reworded (#207), and now this.

That count is the useful signal. Thirteen is not a run of bad luck; it is a systematic gap in how this repository decided a check was doing its job — which is precisely why §11 of the working brief now requires that any probe or assertion offered as *evidence* have its sensitivity demonstrated by mutation rather than asserted. Every one of the thirteen would have been caught at write time by that rule.

## Shape of the fix

Make it able to fail, or delete it. Per §11, whichever is chosen must be proved by mutation: change the behaviour it names and confirm the test goes red.

Worth doing as part of a single pass over the remaining test-integrity items (#203, #207, this, and #153's duplicated screening literal) rather than alone — they share a cause and a remedy.

## Provenance

Found by #210's hardening pass while surveying the tests around its change; reported rather than fixed because it is a test file and CLAUDE.md routes that decision to a human. I have not re-derived the specific assertion.

dkayiwa:
## The shared cause

Four checks in this repository reported success without checking anything, and they failed the same
way: **the thing that decides whether the check ran was invisible to the check.** A `@MethodSource`
predicate that drops cases (a filtered case is not a skipped case, so no count reports it); a helper
that returns a sentinel the caller `continue`s past when it cannot classify its input; an assertion
whose failure set is empty because an earlier assertion already pins the value; a literal copied per
file so a classifier change lands on some copies and not others.

That is the class #179 collected ten Python instances of and #202 fixed. These are the JUnit-side
instances. Each item below states a verdict — repaired or deleted — and the mutation that proves the
surviving check can fail. Every "before" verdict was measured on `ae09928` in an isolated throwaway
worktree, not reasoned.

### #203 — `AbsentDataEvalTest` excluded every absent-data case it exists to test — REPAIRED, and one part DELETED

Its provider filtered on `!isExpectedAbsent()`, dropping all 19 `expectedAbsent` cases of a 29-case
dataset. Build log: 10 tests, 0 skipped.

**Decision on the 19 cases: join the opt-in convention, and pair it with a CI-running guard.**

Asserting `expectedAnswerContains` needs a real answer, and a real answer needs the LLM: nothing in
this module deterministically writes "no records of cancer" — the module's job is to put the chart
and the question in front of the model. Asserting anything less would be redefining what the dataset
means, which is the third option (delete the expectations) wearing a disguise. So the 19 run against
a real OpenAI-compatible endpoint under `-Dchartsearchai.absent.data.test=true`, exactly as
`PromptInjectionEvalTest` and `LlmAnswerQualityTest` do: **skipped, but visibly skipped** (34 → 53
skipped), which is the one thing the issue says it must not remain.

A skipped test guards nothing in CI, so two cases run unconditionally:

- `everyAbsentCaseIsRunAndEveryExpectationDiscriminates` — proves no `expectedAbsent` case is
  filtered out, and proves **every expectation of every case bites**: for each term, an answer that
  omits it must fail the oracle; for each forbidden term, an answer that mentions it must fail. This
  is §11 applied to an assertion instead of a probe, run where it cannot be skipped. It exercises the
  oracle, not the pipeline, and says so.
- `theEmptyChartPromptAsksTheModelToNameWhatIsMissing` — the deterministic half, through real
  production code: the exact bytes `LlmProvider.buildUserMessage` produces for a chart with no
  records, plus the system-prompt instruction the 19 skipped cases all depend on.

**Deleted:** the 10 `presentData_perCase` parameterizations. They asserted that
`stripQueryStopwords(question)` is non-empty — a stopword-helper property `LlmInferenceServiceTest`
already owns with six cases including the all-stopwords one — under a class name and javadoc claiming
absent-data coverage. Repairing that would have meant adopting a check with no absent-data content.

**The chart is the empty one, and the alternative was tried and measured rather than argued.** A
full irrelevant chart is the shape that matters — the querystore focus path has no relevance gate, so
abstaining *in front of* records is the hard case, and it is the only shape in which an
`expectedAnswerNotContains` can bite at all. `FULL_PATIENT_DATASET` looked like the chart this dataset
was authored against, since `absent-cancer` forbids "CD4" and "tuberculosis" and both are records in
it. It is not: records 12 and 89 are `Diagnosis — Kaposi sarcoma oral`, so the patient **has** a
cancer, and the model answered "Yes, the patient has a diagnosis of Kaposi sarcoma oral [12] and
[89]" — correct, and a failure of `absent-cancer`. The one case that shape exists to serve is the one
case it makes unpassable. So the gated case sends the empty chart, under which every case's absence
premise holds by construction, and the rejected alternative is recorded in the code with its
measurement so it is not retried.

### #207 — `chipSeverityRank` failed open — REPAIRED, root cause in production

The helper located severity by `indexOf("also named in the question — ")` and returned `-1` when it
could not find it; the ordering loop `continue`d past `-1`. Measured on `ae09928`:

| mutation | before | after |
|---|---|---|
| none (control) | PASS | PASS |
| `Collections.sort(found, PAIR_SEVERITY_DESCENDING)` removed | FAIL | FAIL |
| clause reworded to "also mentioned in the question" | **PASS** | PASS (correctly — the ordering is still right) |
| **clause reworded AND sort removed** | **PASS** | **FAIL** |
| pair chip stops carrying its rating | n/a | FAIL, loudly |

The third row is the defect and the fourth is the proof: on `main` the test was green while the
ordering was destroyed, because the reword had already turned every chip into "no opinion".

The root cause is in production, not in the test: **the arms sort pair chips on a rating they then
discard**, so the only surviving trace of the sort key was a word inside rendered clinician-facing
prose — and chip wording is reworded routinely (#182, #188, #192, #198, #205, #210). `SafetyWarning`
now carries the rating its arm ordered it on, at the two interaction sites that hold one; `null`
elsewhere, documented as "the source rates nothing here" rather than "not determined" (an unrated
curated rule outranks Major in the same ordering, so the distinction is load-bearing). The wire shape
is unchanged — `serializeSafetyWarnings` still publishes `type`/`drug`/`detail` only.

The helper now reads that value through `severityPriority` and **throws** rather than returning a
sentinel, and the loop no longer skips. A new case,
`anInteractionChipCarriesTheRatingItWasOrderedOn`, pins the production property per arm.

### #212 — an assertion at `InteractionRouteVariantTest:152` could not fail — REPAIRED

`assertFalse(detail.contains("topically administered"))` sat directly under an `assertEquals` pinning
that same detail to a string not containing the phrase. Its failure set was empty.

What it was reaching for is a property of the fixture: that the losing ophthalmic row's note is the
longer one (which is what makes a severity-then-longer-note rule pick wrongly) and says something the
survivor's does not. Asserted there, it can fail. Measured on `ae09928`: shortening the ophthalmic
`ketorolac × lepirudin` note to 51 characters left the case **PASS** — the premise the whole case
rests on was gone and nothing noticed. The repaired case FAILs that mutation.

The dominated assertion was deleted rather than kept alongside; keeping an assertion that cannot fail
next to one that can is how the count reached thirteen.

### #153 — the screening literal was duplicated — REPAIRED, and the issue's rationale corrected

Measured against `main`: **ten** files, not six — nine declaring a private constant and one inline.
All ten now use `DrugReferenceTestSupport.SCREENING_QUESTION`; exactly one copy of the literal
remains, in that constant.

**Deliberate deviation:** the constant lives in `DrugReferenceTestSupport`, not `TestDatasetHelper`.
`TestDatasetHelper` is package-private in `api.impl` and owns patient datasets and category hints;
all ten call sites are in `reference`, `DrugReferenceTestSupport` already owns that package's shared
constants (`ATC_SAMPLE`, `REFERENCE_LOGGER`, the `DDI_*` fixtures), and its own javadoc states it
applies "the same rule CLAUDE.md states for `TestDatasetHelper`". Putting a drug-safety question
string in the patient-dataset helper would have meant widening that class's visibility for a
cross-package dependency in the wrong direction.

**And a correction to the issue's stated consequence.** #153 says a classifier change "can silently
reroute some of them while the others keep passing". Measured, by rewording one file's own copy at a
time to a paraphrase the classifier deliberately does not match: **8 of the 9 files went red**, and
one — `DuplicateInteractionChipTest` — stayed **green**, because its assertion is satisfied by a chip
the drug-in-play arm also raises. So the silent-divergence risk is real but narrow, and it is removed
by there being no per-file copy to edit, not by an assertion. The new
`theSharedScreeningQuestionIsStillClassifiedAsScreening` covers the remaining direction (a change to
the classifier itself, which moves all ten at once) and makes that failure name its cause instead of
arriving as eight files' worth of chip-count mismatches. Mutation: replacing the `interact*` cue
pattern fails it.

## Also in this change, because it created a duplicate

`AbsentDataEvalTest` became the third suite needing an OpenAI-compatible endpoint client, and the
first two had a copy each. `LlmEndpointTestSupport` is that client, and both existing suites now use
it — with no change to what they send. It matters more than the line count suggests: all three suites
are skipped unless opted into, so three diverging copies of the request shape would never turn CI
red. Verified by running all three against a live llama-server, not by inspection.

## Build

Baseline measured on `main` (`ae09928`) in a clean worktree: **1021 api + 61 omod = 1082**, 0
failures, 34 skipped.

This branch: **1034 api + 61 omod = 1095**, 0 failures, **53** skipped.

Delta **+13 api**, every unit accounted for:

| suite | before | after | why |
|---|---|---|---|
| `AbsentDataEvalTest` | 10 | 21 | −10 deleted present-case parameterizations, +19 absent cases (gated, skipped in CI), +2 unconditional cases |
| `DrugSafetyQuestionPairInteractionTest` | 21 | 22 | +1 `anInteractionChipCarriesTheRatingItWasOrderedOn` |
| `DrugSafetyInteractionScreeningTest` | 27 | 28 | +1 `theSharedScreeningQuestionIsStillClassifiedAsScreening` |
| `InteractionRouteVariantTest` | 9 | 9 | the dominated assertion was replaced in place |

Skipped 34 → 53: the 19 gated absent-data cases.

## What needed a live run, and what it found

Most of this is test-only. The one production change — `SafetyWarning` gaining a `severity` field and
accessor, and two `DrugSafetyValidator` call sites passing the rating they already had in hand — is
API surface with **no behavioural delta**, and that is measured rather than argued:

A throwaway probe (never committed) dumped `type|drug|detail` for every chip the real
`DrugSafetyValidator.validate` raises across 14 scenarios covering all its arms — active-order
interaction, question-pair, screening, the pair cap and ordering, allergy contraindication, overdose,
and five route-variant fixtures — run on `main` at `ae09928` and on this branch. **22 chips, 36 lines,
byte-identical, md5 `d127be44d8a98618b265c5c911e2a5df` on both trees.** The `classOnly` chip site is
the one interaction path that probe does not reach; its diff is a comment only, no code. And
`serializeSafetyWarnings` in omod is untouched, so the wire still carries `type`/`drug`/`detail` alone.

On that basis the standalone controls were not re-run and the instance was left on `main` with its 52
GPs untouched, as asked. If a reviewer wants the chips re-asserted live anyway, the probe is 90 lines
and reproducible.

What did need a live run is #203, since a skipped test proves nothing. The 19 cases were run against
a llama-server on the module's own bundled Gemma model, in three configurations — and the first two
were the instrument being wrong, which is why the third is what ships:

| configuration | result | what it measured |
|---|---|---|
| empty chart, 512-token ceiling | 13/19 | 2 answers empty — the ceiling truncated the JSON |
| full chart, 2048-token ceiling | 2/19 | 14 answers empty; the model writes 2018–2919 tokens against a 150-record chart, so the ceiling was failing the cases, not the module |
| **empty chart, production's `DEFAULT_LLM_MAX_OUTPUT_TOKENS`** (shipped) | **15/19**, no truncation, no errors | 2 real defects, 2 too-narrow dataset literals |

The first two rows are the instrument being wrong, which is why the third is what ships and why both
wrong values are recorded next to the constant. The four remaining failures divide cleanly:

- **Two real absent-data defects.** `absent-social-history` and `absent-imaging` came back "No patient
  records were provided." / "No patient records are provided." — naming no topic at all, which is
  exactly what the system prompt's "If no records are relevant, name what is missing" exists to
  prevent. These are the defects this test existed to find and never ran to find.
- **Two dataset literals narrower than the word the answer uses.** `absent-smoking` expects the
  substring `smoke` and the answer said "No smoking status is recorded."; `absent-radiology-xray`
  expects `ray` and the answer said "No radiology reports are recorded." Both answers do name what was
  asked about. **Neither expectation was touched** — CLAUDE.md routes that decision to a human, so they
  are reported, not edited.

The suite is therefore honest but not green when opted into, and that is the point: it is the first
time these 19 expectations have been evaluated at all. The failures are reported as findings rather
than silenced, and no dataset expectation was relaxed to make them pass.

Closes #203
Closes #207
Closes #212
Closes #153

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
@@ -981,6 +981,9 @@ public class DrugSafetyValidator {
 			pairs.add(ref, rule.partnerKey());
 		}
 		for (String detail : classOnly) {
+			// No rating, and not an omission: a shared-ATC-subgroup or cross-reactivity join is a
+			// relationship the reference data states without severity, which is why these chips are never
+			// floor-filtered either. See SafetyWarning.getSeverity on why null is the correct value.
 			warnings.add(new SafetyWarning(SafetyWarning.TYPE_INTERACTION, ref.displayLabel(), detail));
 		}
 	}
@@ -1787,8 +1790,10 @@ public class DrugSafetyValidator {
 		if (rule.getNote() != null && !rule.getNote().isEmpty()) {
 			detail += " — " + rule.getNote();
 		}
+		// The chip carries the rating it is ORDERED on (issue #207), so the ordering is observable
+		// without reading it back out of the prose above.
 		candidates.put(pairKey, new PairFinding(new SafetyWarning(SafetyWarning.TYPE_INTERACTION,
-				subject.displayLabel(), detail), rule.getSeverity(), row, rule));
+				subject.displayLabel(), detail, rule.getSeverity()), rule.getSeverity(), row, rule));
 	}
 
 	/**
@@ -2021,7 +2026,12 @@ public class DrugSafetyValidator {
 		if (alsoSameClass != null) {
 			detail = endSentence(detail) + " " + alsoSameClass;
 		}
-		return new SafetyWarning(SafetyWarning.TYPE_INTERACTION, ref.displayLabel(), detail);
+		// The rule's own rating travels with the chip (issue #207). Null for a curated hand-authored
+		// rule, which is unrated by design — see SafetyWarning.getSeverity, and note that a FOLDED chip
+		// still reports the RULE's rating: the class sentence appended to it carries none, so folding
+		// cannot lower or raise what the pair is rated.
+		return new SafetyWarning(SafetyWarning.TYPE_INTERACTION, ref.displayLabel(), detail,
+				i.getSeverity());
 	}
 
 	/**
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/SafetyWarning.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/SafetyWarning.java
@@ -62,10 +62,19 @@ public class SafetyWarning {
 
 	private final String detail;
 
+	private final String severity;
+
+	/** A warning raised from something the reference data assigns no severity to — see
+	 *  {@link #getSeverity()} for which joins those are. */
 	public SafetyWarning(String type, String drug, String detail) {
+		this(type, drug, detail, null);
+	}
+
+	public SafetyWarning(String type, String drug, String detail, String severity) {
 		this.type = type;
 		this.drug = drug;
 		this.detail = detail;
+		this.severity = severity;
 	}
 
 	/** One of {@link #TYPE_OVERDOSE}, {@link #TYPE_INTERACTION}, {@link #TYPE_CONTRAINDICATION}. */
@@ -88,6 +97,29 @@ public class SafetyWarning {
 	 *  as a display prefix. */
 	public String getDetail() {
 		return detail;
+	}
+
+	/**
+	 * The severity the reference data assigns the rule this warning was raised from — one of
+	 * {@code Major}, {@code Moderate}, {@code Minor}, {@code Unknown} for a DDInter-rated rule, ranked
+	 * by {@code DrugSafetyValidator.severityPriority}, which is also what
+	 * {@code addQuestionPairInteractions} and the screening arm sort their chips on.
+	 *
+	 * <p><b>Null means the source rates nothing here</b>, which is a real distinction rather than a
+	 * missing value: a hand-authored curated rule is deliberately UNRATED (and outranks {@code Major}
+	 * in that same ordering — unrated is not low-rated), an ATC-subgroup or cross-reactivity join
+	 * carries no rating at all, and neither a contraindication nor an overdose has one. Callers must
+	 * not read null as "no severity was determined".
+	 *
+	 * <p>Exposed for issue #207. The chip arms order themselves by this value and then discarded it,
+	 * so the ONLY remaining trace of the key they sorted on was a word inside the rendered
+	 * {@link #getDetail()} prose — which meant the ordering could only be asserted by parsing
+	 * clinician-facing text, anchored on a clause the module rewords freely. Measured: rewording that
+	 * clause left {@code thePairChipsAreOrderedBySeverityAndBounded} green while it asserted nothing at
+	 * all. Not serialized onto the REST response; the wire shape is unchanged.
+	 */
+	public String getSeverity() {
+		return severity;
 	}
 
 	@Override
--- a/api/src/test/java/org/openmrs/module/chartsearchai/api/impl/AbsentDataEvalTest.java
+++ b/api/src/test/java/org/openmrs/module/chartsearchai/api/impl/AbsentDataEvalTest.java
@@ -9,13 +9,21 @@
  */
 package org.openmrs.module.chartsearchai.api.impl;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
+import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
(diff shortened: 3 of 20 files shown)


dkayiwa:
## Mutation evidence: every surviving check shown failing

§11 says a check offered as evidence must have its sensitivity demonstrated, not asserted. All of the
below was run in throwaway `git worktree`s — one for the `main` baseline at `ae09928`, one for the
branch — one mutation at a time, `git checkout -- .` between each, and both trees verified clean at the
end. The `main` column is what the check did **before** this PR under the same mutation; that is the
part that shows the defect was real.

### #207 — `chipSeverityRank`

| mutation | on `main` | on this branch |
|---|---|---|
| none (control) | PASS | PASS |
| `Collections.sort(found, PAIR_SEVERITY_DESCENDING)` deleted | FAIL | FAIL |
| chip clause reworded to `", also mentioned in the question"` | **PASS** | PASS — correctly, the ordering is still right |
| **both of the above at once** | **PASS** | **FAIL** |
| pair chip built with the 3-arg constructor (loses its rating) | n/a | **FAIL**, loudly: `this chip carries no severity, so the ordering it was sorted on is not observable` |
| same, against the new `anInteractionChipCarriesTheRatingItWasOrderedOn` | n/a | **FAIL** |

Row 4 is the proof. On `main` the ordering could be destroyed *and* the assertion stayed green, because
the reword had already turned every chip into "no opinion" and the loop `continue`d past all ten.

### #212 — `InteractionRouteVariantTest`

| mutation | on `main` | on this branch |
|---|---|---|
| none (control) | PASS | PASS |
| ophthalmic `ketorolac × lepirudin` note shortened to 51 chars (drops "topically administered", becomes the *shorter* note) | **PASS** | **FAIL**: `the OPHTHALMIC note must be the longer one … ophthalmic 61 chars, unqualified 265` |

The premise the whole case rests on — that the losing row's note is the longer one, which is what makes
a severity-then-longer-note rule pick wrongly — could be deleted from the fixture with nothing noticing.

### #203 — `AbsentDataEvalTest`

| mutation | result |
|---|---|
| none (control) | PASS |
| provider filters on `!isExpectedAbsent()` again — the original defect | **FAIL** |
| the oracle stops reading `expectedAnswerContains` | **FAIL** |
| chart switched to a dataset lacking the forbidden terms | **FAIL** (that assertion was later removed with the chart revert; recorded here as measured at `8c7e80e9`) |
| `normalizeRecords` drops the empty-records placeholder (production) | **FAIL** |
| `DEFAULT_SYSTEM_PROMPT` drops "If no records are relevant, name what is missing." (production) | **FAIL** |

### #153 — the screening literal

| mutation | on `main` | on this branch |
|---|---|---|
| `INTERACTION_CUES` pattern replaced so the question no longer classifies as screening | (reddens the ten suites on chip counts) | **FAIL**, naming the cause: `the shared screening question must still be classified as interaction screening` |
| one file's **own copy** reworded to "any medication conflicts with her current medications?" | **8 of 9 FAIL, `DuplicateInteractionChipTest` PASSES** | not expressible — there is no per-file copy left |

The second row is #153's named consequence, measured. It is real but narrower than the issue implies:
eight files do go red. One does not, and the fix is that the mutation can no longer be written.

## Build

`main` at `ae09928`, clean worktree: **1021 api + 61 omod = 1082**, 0 failures, 34 skipped.
This branch: **1034 api + 61 omod = 1095**, 0 failures, 53 skipped. `mvn -o clean install`, EXIT=0.

+13 api, every unit: `AbsentDataEvalTest` 10 → 21 (−10 deleted present-case parameterizations, +19
gated absent cases, +2 unconditional cases); `DrugSafetyQuestionPairInteractionTest` 21 → 22;
`DrugSafetyInteractionScreeningTest` 27 → 28; `InteractionRouteVariantTest` 9 → 9 (replaced in place).
Skipped 34 → 53: the 19 gated cases. omod unchanged.

PR #217, opened 2026-08-09
fix(drug-reference): one load-time validity check for what the loader accepts in silence

### LINKED ISSUE #150: A blank alias in a hand-authored drug-reference entry captures every allergen, and #147 widened the exposure
## What

An entry in an operator-authored `sourceFormat=json` drug-reference file whose alias list contains a blank/empty string matches **any** allergen text, so that entry's rules contraindicate the drug for every patient who has any allergy recorded at all.

`DrugReference.containsBoundedToken` treats an empty token as trivially present. `matchesText` did not expose this, but `matchesDrugName` — added by #148 to resolve clinician-entered allergen names — does:

- `matchesDrugName("Vitamin A, B")` is **true** against an entry carrying a blank alias, where `matchesText` was false.

So #148 did not introduce the blank-alias hazard, but it widened where it can fire: allergen resolution is now a third call site for that scan.

## Scope

**Operator-authored `json` sources only.** The shipped `ddinter` KB (2283 entries) is machine-generated and carries no blank alias, and the 4 bundled `drug-reference.json` entries do not either — so the shipped default is unaffected. It bites whoever hand-authors a curated KB, which is exactly the documented extension path, and it fails **open** (spurious contraindications on every patient), not closed.

`containsBoundedToken`'s own javadoc already records the empty-after-fold case as a live hand-authored-JSON hazard, and #148 added a test pinning the fold guard for the accent-only variant. The blank-alias variant has no guard.

## Where the fix belongs

The parser, not the matcher — reject or drop blank aliases at load time so a bad KB is caught once at startup instead of silently altering every safety decision. A matcher-level guard would work too but leaves the invalid data in place, and the same tokens feed `lookupByToken`.

Relates to #149 (the other way a misconfigured reference source fails silently).

## Provenance

Found during the #148 hardening pass and reported to me by that agent; the `matchesDrugName("Vitamin A, B")` asymmetry is its measurement, which I have not independently re-run. Filed so it is not lost.

### LINKED ISSUE #156: Two silent fallbacks in drug-reference source resolution: an unreadable dataFilePath and an unknown sourceFormat both load the wrong dataset with a healthy-looking count
## What

Two configuration mistakes cause the module to load a **different dataset than the one configured**, with a plausible non-zero entry count and nothing above INFO to say so. Both are distinct from #149 (which is the 0-entry case).

### 1. An unreadable `dataFilePath` silently falls back to the bundled dataset

`ReferenceDataFiles.loadWithClasspathFallback`, the `catch (IllegalStateException)` branch. Set:

```
chartsearchai.drugReference.dataFilePath = chartsearchai/anything-absent.json
```

restart, and the **4 bundled curated entries** load instead. The operator's file is never read. The count is non-zero, so it looks healthy.

**Measured live.** The status introduced by PR #154 makes this *observable* — `configuredDataFilePath` is the missing file while `origin` is `classpath:/chartsearchai/drug-reference.json` — but it is not *loud*. Whether it should warn is a maintainer policy call: a classpath fallback is a reasonable default for a fresh install, and unreasonable when an operator has explicitly named a file.

**The standalone testbed appears to be in exactly this state right now for a second file:** `crossReactivityGroupsFilePath` points at `chartsearchai/cross-reactivity-groups.json`, which is absent from appdata, so it silently runs the bundled groups (`Loaded 1 cross-reactivity groups from bundled default`). Reported as observed in every phase of a five-phase live run; I have asked for confirmation and will update this issue.

### 2. An unknown `sourceFormat` silently falls back to the curated `json` parser

`DrugReferenceService.effectiveFormat`. A typo such as `sourceFormat=ddi` against the DDInter KB routes the 19MB DDInter file through the **curated** parser, which yields 0 entries — so in that combination #149's new inert WARN now catches it. But a typo that happens to yield entries stays silent, and the operator believes a format is in force that is not.

PR #154 makes this visible too (`sourceFormat` ≠ `configuredSourceFormat` in the status), again without making it loud.

## Suggested shape

Treat "the operator named something specific and we did not use it" as a first-class warning, in both cases — a fallback that was *chosen* is fine, a fallback that *silently overrode an explicit setting* is not. That single rule covers both, and matches what #149 established for the 0-entry case.

## Provenance

Case 1 measured live during the #149 fix work (its L3 phase); case 2 reasoned from the code by the same agent and labelled as such. The cross-reactivity-groups observation is relayed, pending confirmation.

### LINKED ISSUE #158: The load-before-entries publication order in DrugReferenceService is unpinned: reversing it survives the test suite
## What

`DrugReferenceService` publishes two pieces of state when it populates its cache: the loaded entries, and the `DrugReferenceLoad` status object added for #149. The status must be visible no later than the entries, or a reader can see a populated cache alongside a status that says nothing was loaded.

**Reversing the two writes survives the entire test suite.** Measured during the #154 hardening as one of five mutations; the other four were all caught.

## Severity — genuinely latent, filed so it is not rediscovered

The current order is **correct by the Java memory model**, and the window is one volatile write wide, so this is not a live defect and not urgent. It is filed because:

- The invariant is load-bearing for #149's whole purpose. The status exists so that "what is actually loaded" cannot be read stale; a torn read defeats that in the one case it was built for.
- Nothing stops a future refactor from reordering it, and the suite will stay green. That is exactly the class of regression the repo's TDD rule exists to prevent.

## Shape of the fix

Pin the ordering rather than restate it in a comment — the #154 pass already added a concurrency test through the real seams, so the seam to assert on exists. A test that observes a populated cache with an empty status is the assertion; if that proves impractical to make deterministic, publishing both through a single immutable holder makes the ordering unreorderable by construction, which is stronger than a test.

## Provenance

Measured by the #154 hardening agent, which reported it as an uncaught mutation rather than omitting it. I have not re-run the sweep.

### LINKED ISSUE #211: The candidate-set fallback bounds emptying only: it can still drop every rule-bearing row and keep a rule-less one
## What

#210's hardening found and fixed a total silent loss on the **default** `sourceFormat=json` — `findByQuery` returned 2 entries while `findImpliedByQuery` returned `[]`, so every safety check for a question naming a drug was skipped. The fix added a fallback in `rowsOf` guaranteeing a non-empty candidate set never empties.

**That fallback bounds emptying and nothing else.** The same dataset shape that caused the emptying can still drop **every rule-bearing row while keeping a rule-less one** — measured at 5 chips down to 1. The set is non-empty, so the guard is satisfied, and the findings are gone anyway.

## Root cause, and why it needs a decision rather than a patch

The invariant #210 documented rested on a **parser** property: DDInter and ATC both put an entry's display name in its alias list, so the strongest claimant is always a row that carries rules. A hand-authored `json` file need not do that — and the default `sourceFormat` **is** `json`.

So the real question is: **what does an identity-keyed source mean by "one substance"?** When rows of one substance disagree about which carries the interaction rules, the ranked resolver picks by name claim, which is silent about rules. Options, none obviously right:

- prefer, among equally-strongest claimants, a row that carries rules — but that re-introduces a rule-shaped tie-break into a name-shaped ranking, which is the confusion #209 was filed to remove;
- treat rows of one substance as pooling their rules, which is arguably what "one substance" should mean, but changes what a chip's `ref` refers to;
- validate at load that a substance's rows agree, which fits the load-time validity check decided on #196 and turns this into a data problem rather than a resolution one.

The third is probably right and it composes with work already decided, but it needs measuring against hand-authored files rather than the shipped KB, since the shipped KB does not exhibit it.

## Severity

Latent on shipped data — the fallback fires **zero** times on every shipped dataset, and the KB signature is byte-identical before and after. It is reachable the moment an operator authors a curated file, which is the documented extension path and the default `sourceFormat`. Direction of failure: **withholds findings**, silently.

## Related

Same family as #150 (a blank alias in a hand-authored file captures every allergen) and #156 (two silent fallbacks loading the wrong dataset) — all three are "hand-authored data violates an assumption the shipped parser happens to satisfy". Worth taking with those rather than alone.

## Provenance

Measured by #210's hardening pass through production predicates, including the 5-chips-to-1 case. Relayed; I have not re-derived it.

dkayiwa:
## The shared cause

Four filed defects are the same defect at four points in one code path: **the loader accepts data that violates an assumption its own code makes, and says nothing.**

- **#150** — a blank alias in a hand-authored entry makes the boundary scan trivially true at word gaps, so that entry's rules contraindicate the drug for every patient with any allergy. Fails **open**.
- **#156** — an unreadable `dataFilePath` falls back to the bundled dataset with a plausible non-zero count, and an unknown `sourceFormat` falls back to the curated parser. #154 made both observable; neither was loud.
- **#211** — #210's `rowsOf` fallback bounds *emptying* only: the same shape still drops every rule-bearing row of a substance while keeping a rule-less one.
- **#196** — eight upstream KB rows publish a name or an identifier denoting a different substance, and nothing checks the 2283-row file at all.

#149/#154 already established the remedy for the **count**: a load that is wrong must be loud, and what was loaded must be observable *after* the lazy load, through `getLoadStatus()`. Every item above is that principle one layer deeper, applied to the **content**. So this is one check, not five guards — five guards would give five answers to "what does this loader consider valid".

**#158** is the same code path's publication seam: the entries and the `DrugReferenceLoad` describing them were two volatile writes in a required order, and reversing them survived the whole suite.

## The validity check

`DrugReferenceValidity` is a per-load collector (a local, never a field — #172's rule). It produces `Finding(rule, remedy, occurrences, detail)`; the loader logs every loud finding once, and `DrugReferenceLoad.getFindings()` retains all of them so they can be read after the lazy load rather than grepped out of a log line that may belong to a previous load.

It runs in `DrugReferenceService.ensureLoaded()` — one call site for every source format, so an operator's file gets the same checks whichever parser reads it, and the rules that ask whether two rows are one substance can use `substanceGroupKey()` rather than a per-schema equivalent. The configuration rules run where the resolution happens (`ReferenceDataFiles`) and reach the service through `DrugReferenceSource.lastLoadFindings()`, which mirrors the existing `lastLoadOrigin()` channel exactly.

### Per rule: the remedy and its reason

| rule | issue | remedy | why that remedy |
|---|---|---|---|
| `blank-alias` | #150 | **DROPPED** (the alias) | The token is the whole defect. The entry's other aliases, its rules and its ATC codes are valid and may be its only coverage — `findByActiveOrders` matches on ATC codes and not on aliases at all — so refusing the entry would trade a fail-open for a **silent fail-closed**, which this module treats as the worse direction. Dropping the token removes exactly what fails open and nothing else, so it is *preferable* to refusing the entry rather than merely gentler. |
| `entry-not-named-by-its-own-aliases` | #210, #211 | **REPAIRED** (display name appended) | The ranked resolution's precondition is a property of the *parsers*: `DdiDrugReferenceSource` makes the display name alias[0] and `AtcDrugReferenceSource` makes it the only alias. A hand-authored file need not, and `json` is the default format. Appending the display name asserts nothing the file does not already say — it **is** the entry's name — so the invariant holds by construction on every format instead of resting on `rowsOf`'s fallback. Warning alone would leave the loss in place; refusing the entry would lose its rules. |
| `rules-without-a-substance-identity` | #211 | **REPORTED** | Merging rows because their names look alike is the substance judgement #164/#192 measured this module must not make from spelling — `substanceKey` needs the *data* to say so. See the #211 decision below. |
| `configured-data-file-not-read` | #156 case 1 | **REPORTED**, loud | The fallback data is usable and this feature is an additive net that must never break the answer path; refusing would turn a misconfiguration into the inert safety layer #149 exists to flag. Loud because the operator named a file and a different dataset is in force. |
| `configured-source-format-not-used` | #156 case 2 | **REPORTED**, loud | Same reasoning for not refusing. Loud on the same terms as the file rule — see the ruling below. |
| `alias-names-another-substance` | #196 | **REPORTED** | The module has no better data to substitute: it can only choose *which* of two drugs to name, and the data says they are interchangeable. Dropping the alias loses the resolutions that name is the only path to; dropping the row loses its real interaction rules. Both fail closed to fix something that is now loud. #196 records the data fix as an upstream handoff. |

No rule refuses a file, for the reason stated twice above; that is a deliberate uniformity, and with loudness it is one of two. **Every finding is loud, and silence is the absence of a finding rather than a muted one** — the untouched-default case reports nothing at all, which is the only mechanism keeping the channel quiet on a fresh install. A finding that exists is one the operator needs, so there is no "recorded but not logged" state.

### How the untouched default stays quiet

Both path global properties **default to files the module never creates** (`chartsearchai/drug-reference.json`, `chartsearchai/cross-reactivity-groups.json`), so every install that has configured nothing falls back. A rule keyed on "a path is configured and we fell back" would fire on every install and be filtered within a week, which is worse than silence.

So `configured-data-file-not-read` fires only when the configured value is one somebody **chose**: non-blank *and* different from the global property's own declared default. Both spellings of untouched are excluded, because an installed module reads the declared default from `config.xml` while a context without that row reads blank.

The two declared defaults are now constants, and `GlobalPropertyDefaultsTest` (omod) parses the shipped `config.xml` and asserts they match. Without that pin a drift makes every install loud or every misconfiguration silent, and both failures are invisible — the load still works and the count still looks healthy.

### The #211 decision

The issue records three options. **Validate at load**, and here is the argument against the other two:

- *Prefer a rule-bearing row among equally-strong claimants* re-introduces a rule-shaped tie-break into a name-shaped ranking, which is the confusion #209 was filed to remove — and it would change which row every chip names on data that is perfectly well-formed.
- *Pool a substance's rules* changes what a chip's `ref` refers to: a citable record would carry a rule that is not in the row it cites, which #110 made load-bearing.
- *Validate at load* is the only option that changes no correct answer on correct data. It is also right for a stronger reason: the loader cannot tell a presentation of one substance from a second substance whose name merely nests inside the first, and inferring that from spelling is exactly what #164 and #192 measured this module must not do.

Two things the measurement corrected about the issue's own account, both through the production path:

1. **The precondition #211 states is not the precondition.** The issue says the shape "needs the same precondition as the emptying shape — an entry whose aliases omit its own name". It does not: three rows sharing the alias `ibuprofen` lose the two rule-bearing ones whether or not the bare row names itself, because the bare row is the rank-2 claimant on the shared name either way.
2. **The repair narrows it more than expected.** After every row is given its own name, an order naming a *presentation* (`Ibuprofen tablets 400mg`) keeps that presentation. What still loses rules is the ordinary shape — a name carrying only the substance's name (`Ibuprofen 400mg`) — which is what the finding is about.

**The residual, stated rather than left to be rediscovered.** The rule is gated on the entry declaring *no* `substanceName`. An author who declares one and does not follow the trailing-qualifier naming convention gets silence and still loses rules, because a curated file publishes no substance registry, so `substanceKey` falls back to the display stem and `Ibuprofen tablets` is still its own substance. Widening the rule to "the rows key differently although the data claimed one substance" is **not available**: on the shipped KB that state is also how the DrugBank registry *refuses* a claim — the 19 substance-name families naming two or more DrugBank substances have their id withheld deliberately (#164) — and `getSubstanceId()` is null in both cases, so the loaded model cannot tell "no registry" from "the registry says these differ". Flagging it would report `Omeprazole` against `Esomeprazole` on the default configuration. Closing it needs the curated schema to be able to state a substance identity, which is a schema decision.

### #158

The entries and the `DrugReferenceLoad` describing them are published through one immutable holder (`LoadedDataset`), written once. There is no order left to reverse, so no refactor can reintroduce the state the status exists to rule out.

What that can and cannot claim: no test can demonstrate the absence of a race whose window is one volatile write wide, and the fix is that the window no longer exists. `DrugReferenceLoadConcurrencyTest.aCompletedLoadIsNeverPublishedWithAStatusThatSaysNothingWasLoaded` pins the remaining way to reach the same torn state — pairing a populated cache with the wrong outcome as the holder is built, which is a construction error and therefore deterministic. Proven by mutation: publishing `new LoadedDataset(loaded, DrugReferenceLoad.notLoaded())` reddens it.

## Which of #196's eight the check detects

Measured by running the production check over the full 19 MB `ddi_knowledge_base.json` through `DdiDrugReferenceSource.parse` + `DrugReferenceValidity.checkEntries` (no reimplementation of any predicate). **2283 entries, exactly one finding, 18 occurrences.**

**Detected — 2 of the 8, across 7 rows:**

- item 8 (`Pfizer-BioNTech Covid-19 Vaccine` carrying `moderna covid-19 vaccine`) — named, **and all five `Tozinameran` rows that inherit it**, which is 6 of the 18.
- item 6 (the three trastuzumab rows sharing one `ciel` list) — named through the half that reaches a clinician: `Trastuzumab emtansine publishes 'trastuzumab deruxtecan', which is Trastuzumab deruxtecan's own name`. That is the collision behind the false live label "a recorded allergy to Trastuzumab deruxtecan" for a Kadcyla allergy.

**Not detected, with the measurement that rules out each candidate predicate:**

- **item 4** (`Fluoroestradiol f-18` holding estradiol's identity). Its raw `drugbank_id` is `null`; it acquires `DB00783` because it shares `rxnorm_name: estradiol` with `Estradiol`, whose family has exactly one id — so the module already treats the two as **one substance**, and the rule skips same-substance pairs. Any predicate that flags it also flags the two merges #164 defends: a candidate rule ("a row's substance name is another row's display name its own name does not carry") fires **12** times on the shipped KB and includes `Daxibotulinumtoxina` → `botulinum toxin type a` and the tick-borne-encephalitis paediatric row. This is the limit #196 itself records: no logic here can recover a registry id that identifies the wrong substance.
- **item 5** (`Sulfamethoxazole` publishing `rxnorm_name: sulfamethazine` and `drugbank_id: DB01015`). **No row named `Sulfamethazine` exists in the file** (only `Sulfamethizole` and `Sulfamethoxazole`), so nothing in the file contradicts either field. The only predicate that would flag the name — "the display name does not carry its own `rxnorm_name`" — fires on **281** rows, and #196 already measured that class as unusable (of the 24 with no shared stem, `aspirin`, `heroin`, `vitamin k1`, `cosyntropin` are genuine synonyms).
- **item 7** (mechanism group `3606`, a clause missing between `administration` and `Close`). Both candidate predicates were measured over all 8234 mechanisms and both are wrong: "a lowercase word followed by a capitalised word" fires **101** times, almost all legitimate (`with Grade`, `mean Cmax`, `the International Normalized Ratio`); "the text does not end in a sentence terminator" fires **29** times and **does not include 3606**, whose text ends in a full stop — its gap is internal. That second predicate would also have been wrong as a rule: the bundled curated dataset's interaction notes are deliberately sentence fragments (`increased risk of GI bleeding`), so it means different things per source. No rule shipped for this.
- **items 1–3** (`Monopotassium phosphate`'s `rxnorm_name`, `Choline salicylate` and `Liver derivative complex`'s CIEL names) — wrong values that nothing else in the file contradicts. Same class as item 5.

**Also named, and not on the issue: 11 further rows of the identical shape** — `Omeprazole` → `esomeprazole`, `Levoketoconazole` → `ketoconazole`, `Fenfluramine` → `dexfenfluramine`, `Fenofibric acid` → `fenofibrate`, `Fosnetupitant` → `netupitant`, `Amphetamine` → `dextroamphetamine`, `Hyoscyamine` → `atropine`, `Iodide I-131` → `iodine`, `Gallium citrate ga-67` and `Gallium chloride Ga-67` → `gallium nitrate`, `Dotatate gallium ga-68` → `lutetium lu 177 dotatate`. Each is a row publishing a different substance's own name; none is a false positive on inspection.

**#196 stays open.** This PR adds the check it asked for; the eight rows are upstream data in openmrs-ddi-knowledge-base and are not touched here.

## On the wire

`getFindings()` is retained on `DrugReferenceLoad` **and** serialized by `toMap()`, so it reaches `GET /chartsearchai/drugreferencestatus`. Both channels, because they answer different questions: the WARN says it once at the moment it happened, and the status answers it afterwards — which is the question a lazy load makes a log line unable to answer (#149).

`findings` is appended **last** in the map, deliberately: every existing key keeps its position, so the endpoint's documented field list stays an *ordered* assertion rather than being relaxed to accommodate a new field. Each finding carries `rule`, `remedy`, `occurrences` and `detail`; a bare count would recreate at this level the defect #149 fixed one level down, where a load of 0 and a load of 2283 logged identically.

Two frozen key sets were extended for it, as a decision rather than an accident: `DrugReferenceLoadContextTest.loadStatusSerializesTheFieldsTheStatusEndpointReturns` and `ChartSearchAiDrugReferenceStatusTest.DOCUMENTED_FIELDS`.

## The assertion that had to change, and what it now pins

`DrugReferenceLoadContextTest.loadStatusReportsAMistypedSourceFormatSeparatelyFromTheOneInForce` asserted that a mistyped `sourceFormat` must **not** be loud, on the ground that the status reporting the configured and the effective format separately was signal enough:

```java
assertFalse(capture.hasEventAtOrAbove(Level.WARN),
    "a typo that still yields entries is NOT loud — the two format fields differing is "
        + "the only signal, which is why both are reported. …");
```

That ground is the confusion issues #149 and #154 settled: **observable is not the same as loud.** #154 built the status endpoint precisely because an operator cannot be expected to poll it, and #149 exists because a wrong load logged at INFO is indistinguishable from a right one. A mistyped format hands a DDInter file to the curated parser and says nothing, which is the shape #156 was filed about.

The assertion is corrected rather than deleted, and it now pins the opposite condition. The property it *was* written to protect — that a typo pointing at a readable dataset is **not** the #149 inert failure — is still asserted in the same test, on `isInert()` and the entry count, which is where it belongs; the log level never carried it.

With loudness settled the same way for all six rules, the per-finding `loud` flag is gone rather than left as a boolean that is always `true` and invites a future rule to be muted.

## Build

Baseline measured on `main` (`3b789cd`) in a clean worktree: **1034 api + 61 omod = 1095**, 0 failures, 53 skipped.
This branch: **1045 api + 62 omod = 1107**, 0 failures, 53 skipped.

+11 api = the 10 cases of `DrugReferenceValidityContextTest` plus `DrugReferenceLoadConcurrencyTest.aCompletedLoadIsNeverPublishedWithAStatusThatSaysNothingWasLoaded`. +1 omod = `GlobalPropertyDefaultsTest`. Skips unchanged.

### Mutation sweep — every decision is protected

Run on a throwaway worktree; each row is the test that reddened.

| mutation | test that reddened |
|---|---|
| `blank-alias` drop made a no-op | `aBlankAliasIsDroppedAtLoadSoTheEntryStopsMatchingEveryAllergen` |
| the display-name repair disabled | `anEntryWhoseAliasesOmitItsOwnNameIsGivenItAtLoad` |
| `rules-without-a-substance-identity` not reported | `rulesOnRowsThatDoNotDeclareTheirSubstanceAreReportedAsUnreachable` |
| the substance-identity half of the #196 rule removed | `aPublishedNameDenoting…` — admits `Daxibotulinumtoxina publishes 'botulinum toxin type a'`, 3 instead of 2 |
| the display-stem half of the #196 rule removed | `aPublishedNameDenoting…` — admits `Hydrocortisone butyrate publishes 'hydrocortisone'` and the intra-trastuzumab pairs, 7 instead of 2 |
| the untouched-default comparison removed | `aFallbackFromTheUntouchedDefaultPathIsNotReportedAtAll` **and** the pre-existing `bundledDefaultLoadIsNotReportedAtWarnOrError` |
| the holder published with `notLoaded()` | `aCompletedLoadIsNeverPublishedWithAStatusThatSaysNothingWasLoaded` |
| `configured-source-format-not-used` never reported | `loadStatusReportsAMistypedSourceFormatSeparatelyFromTheOneInForce` (the corrected assertion) **and** `aMistypedSourceFormatIsReportedLoudlyAndOnTheWire` |
| `toMap()` omits the `findings` key | `loadStatusSerializesTheFieldsTheStatusEndpointReturns` and, swept alone, `ChartSearchAiDrugReferenceStatusTest.drugReferenceStatus_reportsEveryDocumentedFieldForADefaultInstallation` |

The two #196 halves were swept separately, because a single run cannot attribute them; so were the two changed key-set assertions, since the api failure stops the reactor before omod runs.

Both changed assertions therefore catch a named condition and are proven to: the corrected one catches "the loader accepted an overridden parser without saying so", and the extended key sets catch "the findings stopped reaching the only channel an operator can read".

## Live verification

Two passes on the 3.7.1 standalone, each a full-JVM deploy of this branch. Pass A ran on `a0c0842b`; pass B rebuilt and redeployed `cceed30f`, whose only production delta is prose plus a guard on a display name that the curated parser already drops, and re-asserted the controls on it.

**Created for this (disposable demo data, named so a later reader can tell it from stock data):** a non-coded DRUG allergy `Vitamin A, B` on Melissa Wright (allergy `19bbb5ea-c35d-4007-b550-a24bdbe2dac3`), and two operator-authored curated files in appdata, `chartsearchai/T150-validity-defects.json` and `chartsearchai/T150-validity-healthy.json`. The drug-reference global properties were put back to `sourceFormat=ddinter` + `chartsearchai/ddi_knowledge_base.json` and verified byte-identical to the pre-work snapshot.

### #196 — the check against the real 19 MB knowledge base

`GET /chartsearchai/drugreferencestatus` first, because the load is lazy and a log line read after a GP flip belongs to the previous source (#149):

```
```

One validity line for the whole 2283-row file, verbatim:

```
INFO  ReferenceDataFiles.loadWithClasspathFallback |23:39:11,023| Loaded 2283 DDInter drug-reference entries from …/appdata/chartsearchai/ddi_knowledge_base.json
WARN  DrugReferenceValidity.logTo               |23:39:11,040| Drug-reference data validity — alias-names-another-substance (reported, 18): 18 published name(s) denote a DIFFERENT substance in this dataset, so a question or a chart string carrying one resolves the wrong drug. The data is left as loaded — the fix is in the dataset. [Omeprazole publishes 'esomeprazole', which is Esomeprazole's own name, Pfizer-BioNTech Covid-19 Vaccine publishes 'moderna covid-19 vaccine', which is Moderna covid-19 vaccine's own name, Trastuzumab emtansine publishes 'trastuzumab deruxtecan', which is Trastuzumab deruxtecan's own name, Levoketoconazole publishes 'ketoconazole', which is Ketoconazole's own name, Tozinameran (12y+) publishes 'moderna covid-19 vaccine', …] and 10 more
```

17 ms between the two lines, so all three content rules over 2283 entries cost ≲17 ms of a load that reads 19 MB.

**What it would have shown had the rule been wrong:** the substance-identity condition removed admits `Daxibotulinumtoxina publishes 'botulinum toxin type a'` (the #164 merge), and the display-stem condition removed admits `Hydrocortisone butyrate publishes 'hydrocortisone'` plus the intra-trastuzumab pairs. Both confirmed by mutation in the unit suite; the live line names neither.

### #150 / #211 — an operator-authored file, and the discriminator

`sourceFormat=json`, `dataFilePath=chartsearchai/T150-validity-defects.json`, module restarted. The source really switched, confirmed on the endpoint and not from the log:

```
{"enabled":true,"loaded":true,"inert":false,"entryCount":5,"sourceFormat":"json",
 "configuredSourceFormat":"json","configuredDataFilePath":"chartsearchai/T150-validity-defects.json",
 "origin":"appdata:chartsearchai/T150-validity-defects.json"}
```

Three rules, verbatim:

```
INFO  |23:49:03,333| Loaded 5 drug-reference entries from …/appdata/chartsearchai/T150-validity-defects.json
WARN  |23:49:03,341| Drug-reference data validity — blank-alias (dropped, 1): 1 alias(es) naming nothing (blank, or nothing but combining marks) were dropped … Entries: [Warfarin]
WARN  |23:49:03,351| Drug-reference data validity — entry-not-named-by-its-own-aliases (repaired, 3): 3 entr(ies) whose aliases omitted their own name were given it … Entries: [Ibuprofen tablets, Ibuprofen suspension, Ibuprofen]
WARN  |23:49:03,359| Drug-reference data validity — rules-without-a-substance-identity (reported, 2): 2 rule-bearing entr(ies) share a published name with another entry while declaring no substanceName … [Ibuprofen tablets shares 'ibuprofen' with [Ibuprofen tablets, Ibuprofen suspension, Ibuprofen], Ibuprofen suspension shares 'ibuprofen' with [Ibuprofen tablets, Ibuprofen suspension, Ibuprofen]]
```

Then the pair that makes the chip evidence discriminating. Melissa Wright's only relevant record is the `Vitamin A, B` allergy; the file gives `Warfarin` the blank alias and `Multivitamin` a legitimate two-word alias (`vitamin a`) that matches the same string.

```
D1  Melissa Wright — "Is it safe to start warfarin?"
    safetyWarnings (0)
    answer: The records do not address warfarin safety for this patient.

D2  Melissa Wright — "Is it safe to give a multivitamin?"
    safetyWarnings (1)
      [contraindication] drug='Multivitamin'
          The patient has a recorded allergy to Multivitamin.
```

**What D1 would have shown had the drop been absent:** one spurious `Warfarin` contraindication for a patient with no warfarin allergy. Measured through the production matcher rather than asserted — an entry carrying `[warfarin, " "]` against the same four strings:

```
Vitamin A, B                 blankAliasPresent=true  afterDrop=false
Aspirin                      blankAliasPresent=false afterDrop=false
Sodium chloride              blankAliasPresent=false afterDrop=false
Amphotericin B, liposomal    blankAliasPresent=false afterDrop=false
prose matcher on the same string: false
```

D2 is why D1's zero is not vacuous: the arm demonstrably fires for this patient, this file and this allergen when a *real* two-word alias matches it. Without D2 a zero could just mean the contraindication arm never ran.

### The healthy-file control — the loader must say nothing

Same five entries, authored the way the shipped datasets are, loaded on the next module restart:

```
{"entryCount":5,"sourceFormat":"json","configuredDataFilePath":"chartsearchai/T150-validity-healthy.json",
 "origin":"appdata:chartsearchai/T150-validity-healthy.json"}

INFO  |23:50:36,379| Loaded 5 drug-reference entries from …/appdata/chartsearchai/T150-validity-healthy.json
```

and **no validity line after 23:49:03** — the last one in the JVM belongs to the previous phase. This is the control that stops the check becoming a channel people filter; the mutation that reddens its unit counterpart is removing the untouched-default comparison, which also reddens the pre-existing #149 test.

The same phase carries #156's confirmed-live instance staying quiet: `crossReactivityGroupsFilePath` is at its declared default, that file does not exist, and the fallback is silent —

```
INFO  |23:49:03,558| Loaded 1 cross-reactivity groups from bundled default /chartsearchai/cross-reactivity-groups.json
```

with no accompanying WARN, in every phase.

### The eleven regression controls, on the 19 MB KB

| probe | chips | detail |
|---|---|---|
| Mary Smith — clarithromycin | **1** | `[interaction] Clarithromycin` — "interacts with active order simvastatin — Major" |
| Mary Smith — paracetamol | **0** | — |
| Agnes Adams — warfarin | **1** | `[interaction] Warfarin` — "interacts with active order aspirin — Major" |
| Joshua Johnson — ibuprofen | **2** | NSAID cross-reactivity contraindication + `lisinopril — Moderate` |
| Betty Williams — bupivacaine | **1** | `[interaction] Bupivacaine` — "active order lidocaine — Major" |
| Sarah Taylor — hydrocortisone | **8** | four `Hydrocortisone` chips, then Dexamethasone / Prednisone / Budesonide / Methylprednisolone |
| George Anderson — esomeprazole | **1** | "same ATC class (A02BC) as the patient's allergy to Pantoprazole" |
| Margaret King — botulinum toxin type A | **1** | names `Botulinum toxin type A` |
| Kevin Brown — levothyroxine | 2 | not a named control — see below |
| Susan Young — botulinum toxin | 1 | not a named control — see below |
| Agnes Adams — epinephrine | **0** | — |

Two of the eleven were asked the wrong question for the control they were meant to be. The "Kevin 3" and "Susan Young 3" controls belong to specific questions — `hydrocortisone` for Kevin (the #198/#209 shape) and `lidocaine` for Susan (the #210 route-variant case) — and pass A asked levothyroxine and botulinum instead, so its 2 and 1 are different measurements rather than misses. Both are re-run with the right questions in pass B, on the final SHA.

These are **controls, not canaries**: none of them exercises a validity rule, because every rule is silent on the shipped KB except the #196 one, which changes no chip. Their value is that they did not move.

### #156 — both fallbacks in one load, one loud and one silent

`sourceFormat=ddintr` (a typo) **and** `dataFilePath=chartsearchai/T150-absent.json` (a file that does not exist), set together on purpose, because the interesting thing is which of them speaks. The status:

```
{"enabled":true,"loaded":true,"inert":false,"entryCount":4,"sourceFormat":"json",
 "configuredSourceFormat":"ddintr","configuredDataFilePath":"chartsearchai/T150-absent.json",
 "origin":"classpath:/chartsearchai/drug-reference.json"}
```

`entryCount:4` is the count #156 reported — a plausible non-zero number from a dataset nobody configured. The log, verbatim and complete for that load:

```
INFO |23:55:22,025| drug-reference entries file 'chartsearchai/T150-absent.json' not available (Model file not found: …/appdata/chartsearchai/T150-absent.json…); using bundled default
INFO |23:55:22,043| Loaded 4 drug-reference entries from bundled default /chartsearchai/drug-reference.json
WARN |23:55:22,048| Drug-reference data validity — configured-data-file-not-read (reported, 1): chartsearchai.drugReference.dataFilePath names 'chartsearchai/T150-absent.json', which could not be read, so 'classpath:/chartsearchai/drug-reference.json' is in force instead. The entry count is therefore a count of a dataset nobody configured, and looks healthy.
INFO |23:55:22,025| cross-reactivity groups file 'chartsearchai/cross-reactivity-groups.json' not available (Model file not found: …); using bundled default
INFO |23:55:22,243| Loaded 1 cross-reactivity groups from bundled default /chartsearchai/cross-reactivity-groups.json
```

**Three fallbacks, one WARN.** The entry file the operator named is loud. The `sourceFormat` typo is silent (and reported on the status as `configuredSourceFormat` ≠ `sourceFormat`) — the existing assertion this PR does not edit. The cross-reactivity groups fell back too, from a path that is its **declared default**, and it is silent — which is the untouched-default rule demonstrated in the same load as the case it is distinguished from, rather than argued.

**What this misconfiguration costs, measured:** Joshua Johnson, "Can I give ibuprofen?", **0 chips** here against **2 chips** on the correctly-configured KB an hour earlier. The status says `loaded:true, inert:false, entryCount:4` throughout — so before this PR the only way to notice was to compare `origin` with `configuredDataFilePath` by hand, which is exactly what #156 says nobody does.

### A hazard hit, and what it voided

Run 2's second `moduleaction` restart — the one restoring the 19 MB KB — exhausted the heap: every subsequent `/search` returned HTTP 500 with `java.lang.OutOfMemoryError: Java heap space` in the controller log. That is the standalone's documented two-restart budget being spent (a 19 MB reload is not the same cost as a 5-entry one), not a behaviour change: the probe before that restart returned 200, and the same controls pass on a fresh JVM. Run 2's phase-5 controls are therefore void and were re-run in run 3 on a clean boot of the same build.

### Run 3 — the controls on a clean boot of the final build

Fresh JVM, restored global properties, no module restarts, `cceed30f` deployed. One validity line for the load (`alias-names-another-substance (reported, 18)`, the same as runs 1 and 2 — reproducible across three JVMs), and:

| probe | chips | expected |
|---|---|---|
| Mary Smith — clarithromycin | 1, `[interaction] Clarithromycin` | 1 Major ×simvastatin ✓ |
| Agnes Adams — warfarin | 1, `[interaction] Warfarin` | 1 Major ×aspirin ✓ |
| Joshua Johnson — ibuprofen | 2, NSAID contraindication + `Ibuprofen` interaction | 2 ✓ |
| Mary Smith — paracetamol | 0 | 0 ✓ |
| Kevin Brown — hydrocortisone | 3: `Hydrocortisone (ophthalmic)`, `Hydrocortisone` interaction, `Clarithromycin` | 3 ✓ |
| Susan Young — lidocaine | 3: two `Lidocaine` interactions + `Tiotropium` contraindication | 3, third kept ✓ |

The global properties were left as found: `sourceFormat=ddinter`, `dataFilePath=chartsearchai/ddi_knowledge_base.json`, `crossReactivityGroupsFilePath=chartsearchai/cross-reactivity-groups.json`, `enabled/injectFromOrders/injectFromQuery=true`.

### Run 4 — the two rulings, live on `faf28569`

Fresh JVM, the final build, the 19 MB KB. `GET /chartsearchai/drugreferencestatus`, verbatim and complete:

```json
 "origin":"appdata:chartsearchai/ddi_knowledge_base.json",
 "findings":[{"rule":"alias-names-another-substance","remedy":"reported","occurrences":18,
              "detail":"18 published name(s) denote a DIFFERENT substance in this dataset, … [Omeprazole publishes 'esomeprazole', … Pfizer-BioNTech Covid-19 Vaccine publishes 'moderna covid-19 vaccine', … Trastuzumab emtansine publishes 'trastuzumab deruxtecan', …] and 10 more"}]}
```

Then one module restart onto the hand-authored file **and** a mistyped `sourceFormat`, so all four rules that can fire on one load do:

```json
{"entryCount":5,"sourceFormat":"json","configuredSourceFormat":"jsonn",
 "configuredDataFilePath":"chartsearchai/T150-validity-defects.json",
 "origin":"appdata:chartsearchai/T150-validity-defects.json",
 "findings":[
  {"rule":"configured-source-format-not-used","remedy":"reported","occurrences":1,
   "detail":"chartsearchai.drugReference.sourceFormat is 'jsonn', which matches no adapter, so the 'json' parser is in force instead."},
  {"rule":"blank-alias","remedy":"dropped","occurrences":1,"detail":"… Entries: [Warfarin]"},
  {"rule":"entry-not-named-by-its-own-aliases","remedy":"repaired","occurrences":3,
   "detail":"… Entries: [Ibuprofen tablets, Ibuprofen suspension, Ibuprofen]"},
  {"rule":"rules-without-a-substance-identity","remedy":"reported","occurrences":2,
   "detail":"… [Ibuprofen tablets shares 'ibuprofen' with [Ibuprofen tablets, Ibuprofen suspension, Ibuprofen], …]"}]}
```

and every one of them in the log too, the format rule included — which is the ruling:

```
Drug-reference data validity — alias-names-another-substance (reported, 18)
Drug-reference data validity — configured-source-format-not-used (reported, 1)
Drug-reference data validity — blank-alias (dropped, 1)
Drug-reference data validity — entry-not-named-by-its-own-aliases (repaired, 3)
Drug-reference data validity — rules-without-a-substance-identity (repaired… reported, 2)
```

**What each would have shown had the ruling not been applied:** no `configured-source-format-not-used` line in the log at all (the operator's parser silently overridden, which is what run 2's phase 4 captured before the ruling), and no `findings` key in the response — so the check would be visible to the test suite and to nobody else. Both proven by mutation in the suite as well.

The global properties were then restored **without** a second restart, since the reload would have been the full KB — the corrected budget from run 2's `OutOfMemoryError` — and verified byte-identical to the pre-work snapshot.

Closes #150
Closes #156
Closes #211
Closes #158

#196 stays open: this adds the load-time check it asked for, and the eight knowledge-base rows are upstream data in openmrs-ddi-knowledge-base. Measured, two of the eight are detectable at parse time, not five — the reasoning is above.


dkayiwa:
Independent pre-merge verification: `mvn -o clean install` on a clean detached worktree at `faf28569` → **1045 + 61 api / 62 omod = 1107, 0 failures, 53 skipped**, BUILD SUCCESS. `main` at `3b789cd` = 1095. CI green on Java 11/17/21. Exactly **one** non-comment assertion line removed across the branch — the `assertFalse(capture.hasEventAtOrAbove(WARN))` my ruling required inverting — and nothing else.

**On the two rulings: the second one was taken further than I ruled, and better.** I said to flip one boolean so the mistyped-`sourceFormat` rule becomes loud. The change instead **removes the `loud` flag entirely**, on the reasoning that a finding only exists when the operator chose something and the loader did not honour it — so the untouched-default case is kept quiet by *the rule not firing at all*, not by a per-rule log level, and if every finding deserves a warning then the flag was dead weight. That is the better design and I would not have got there from where I was standing.

The property I cared about survives and is asserted: the new test file carries an `assertFalse(capture.hasEventAtOrAbove(WARN))` for the untouched default alongside three `assertTrue`s, so "silence is the absence of a finding, never a muted one" is pinned in both directions rather than stated.

**On CLAUDE.md**: this branch adds a `DrugReferenceValidity` API-surface entry and then extends it with the loudness decision. I have told agents in this work that they may not edit that file — but I did not repeat it in this task's prompt, and the entry is exactly what I would have written, including both binding rules (the remedy is per rule and never uniform; a rule must stay silent on an untouched default) and the measured reason the "observable but not loud" option was rejected. Accepting it rather than stripping correct work over an instruction I failed to restate. The inconsistency is mine.

**Two corrections to my brief that this pass earned, both already propagated:**

- **"Five of the eight #196 defects are detectable at parse time" was wrong — measured, it is two.** Each of the other six was ruled out with the false-positive count of the predicate that would have caught it: 12 firings including `Daxibotulinumtoxina` for `Fluoroestradiol f-18`, 281 rows for `Sulfamethoxazole`, 101 firings for sentence fusion and 29 for missing terminal punctuation which does not even include group 3606. That is the right way round — my number challenged by measurement rather than accepted.
- **The two-`moduleaction`-restart budget is one when the reload is the 19 MB KB.** Run 2's second restart took `OutOfMemoryError: Java heap space` and returned HTTP 500 on every `/search`, voiding that run's controls, while run 1 survived two restarts because its second load was a 5-entry file. §6 now says budget by what is being reloaded, not by count.

Also recorded: `Fluoroestradiol f-18`'s raw `drugbank_id` is `null` — it *inherits* `DB00783` through the family resolution, so my stated mechanism on #196 was wrong though its consequence stands; and there is no `Sulfamethazine` row at all, so that item is a wrong field with no contradicting row rather than a collision.

#196 stays open for the upstream data, per the owner's scoping of this work to this repository.

Issue #103, opened 2026-07-27
A clinical value pair can be split into fabricated citations when the chart happens to contain those record numbers


dkayiwa:
## What happens

`LlmAnswerExtractor.normalizeSlashCitations(text, validCitations)` rewrites citation shorthand
(`[a/b]`, `[a, b]`) into `[a], [b]` for groups whose every part appears in the structured
citations array. Corroboration is **membership-only**, so a clinical value pair collides with real
record numbers whenever the chart is large enough to contain them.

Reproduced through the composed production path (`extractResponse` → `extractCitedReferences`) on
a 130-record chart:

| input | output |
|---|---|
| `{"answer":"BP was [120, 80] today.","citations":[120,80]}` | answer rewritten to `"BP was [120], [80] today."`, references `[80, 120]` |

Two things go wrong at once: the prose the clinician reads has a mangled blood pressure, and two
references are fabricated from it and attached to the answer as provenance.

## Why the existing guard doesn't catch it

The guard is exactly right for the case it was built for — an uncorroborated `[120, 80]` is left
alone, which is what `ChartSearchAiUtils.INLINE_CITATION` being deliberately single-index depends
on. The gap is that a model which writes a value pair inline **and** lists those same numbers in
its citations array supplies precisely the corroboration the guard accepts as proof. The larger
the chart, the more clinical values collide with real record numbers, so exposure grows with chart
size — the method's javadoc previously asserted the opposite ("chart size is irrelevant"), and that
sentence was the stated reason no further discriminator existed. Corrected in b43b9f9; this issue
is the fix itself.

## Why it isn't fixed inline

The obvious discriminator — require an independent single-index marker elsewhere in the answer
before splitting a group — regresses the measured multi-index case this method exists for (PR #83's
`[6, 7]` shorthand, where the group is the answer's *only* citation form). So a fix needs its own
measurement on the 270-cell gold rather than a guess. Candidate discriminators worth measuring:

- Require the group's parts to be **contiguous or near-contiguous** record indices (a citation
  group usually references neighbouring retrieved records; `[120, 80]` is descending).
- Require the cited records to be **plausible for the sentence** — reuse the Tier-1 cosine the
  grounding verifier already computes rather than adding a new signal.
- Refuse to split when the surrounding sentence matches a value-bearing pattern (a unit, a
  vital-sign name, `mmHg`, a `/`-separated numeral).
[...]

Issue #94, opened 2026-07-27
Absence verdicts on topical questions are asserted from the similarity slice, not the chart


dkayiwa:
## Problem

`TOPICAL` is the one slice intent with no completeness guarantee: its whole context is the
similarity top-K (`chartsearchai.querystore.topK`, default 12). So when the model answers
*"No kidney problem is recorded"* it is making a claim about the **whole chart** from a dozen
records the search happened to surface. The system prompt actively invites this — *"When no
record explicitly names it, start with a short sentence naming what is absent"* — but the model
has no way to know whether the record exists outside its slice.

## Reproduction (RefApp 3.7.1 demo data, `chartMode=queryScoped`, `topK=12`)

Patient `a6451ed0-f35f-46ee-9008-088324beb88e`, *"Does the patient have any kidney problems?"*:

> No explicit kidney problem diagnosis is recorded. Relevant renal function findings include: …

The chart carries `Condition: Foreign body in genitourinary tract. Status: ACTIVE` **and** its
confirmed diagnosis twin. Neither reached the slice — they are not semantically close to
"kidney problems" for the embedder.

This is a safety-shaped failure: a confident negative is worse than an "I only reviewed these
records" hedge, and worse than no answer.

## The obvious fix is refuted

Giving presence-shaped `TOPICAL` questions the complete `condition` + `diagnosis` list was
implemented and measured on a 270-cell gold (see [ADR Decision 30](docs/adr.md) and PR #93). It
is worse on every axis and 14% slower:

| | topK slice | + complete problem list |
|---|---|---|
| meanF1 | 0.760 | 0.732 |
| abstention | 0.922 (107/116) | 0.871 (101/116) |
| off-topic citations | 74 | **154** |
| mean response time | 15 652 ms | 17 823 ms |

`heart` drift 28 → 77, `kidney` F1 0.656 → 0.479. A small model handed a long undifferentiated
problem list enumerates it instead of filtering it. The code was reverted.

## What a real fix needs

A **relevance gate on retrieval**, so an absent topic returns *nothing* rather than the K
nearest neighbours. `QueryStoreService.searchByPatient` currently has no score floor and
`QueryDocument` exposes no similarity score, so chartsearchai cannot tell "12 good hits" from
"12 hits, none of which are about the query". That is retrieval's side of the line
(openmrs-module-querystore owns it per the project rule), and it would let the prompt
distinguish a real absence from an empty slice.

[...]

PR #218, opened 2026-08-09
Name the topic when nothing is relevant, and give the two eval oracles a way to say what they mean

### LINKED ISSUE #214: Two absent-data answers name no topic at all: "No patient records were provided."
## What

Surfaced by #213, which repaired `AbsentDataEvalTest` so its 19 absent-data cases actually run. Against the **shipped configuration** and the module's own bundled model, 15 of 19 pass. Two of the four failures are genuine defects in production behaviour:

- **`absent-social-history`** and **`absent-imaging`** both answer *"No patient records were provided."*

That names **no topic at all**. The prompt's "name what is missing" instruction exists precisely to prevent this, and `AbsentDataEvalTest`'s own javadoc states the property as *"produces a clear 'no records' answer **naming what was asked about**"*.

## Why it matters more than a wording nit

A clinician asking about imaging and being told "no patient records were provided" cannot distinguish three very different situations: this patient has no imaging, the chart failed to load, or the question was not understood. The other 15 cases get this right — *"No smoking status is recorded."*, *"No radiology reports are recorded."* — so the instruction works in general and fails on these two topics specifically.

It is also adjacent to **#94** (absence verdicts asserted from the similarity slice rather than the chart): an answer that says "no records were provided" rather than "no imaging is recorded" is describing the *slice* it received, not the patient.

## Scope note

These two are **not** the same as the two other failures in that run, which are bad test literals rather than defects — `absent-smoking` expecting the substring `smoke` against "No **smoking** status is recorded", and `absent-radiology-xray` expecting `ray` against "No **radiology** reports are recorded". Those expectations are being corrected separately; these two are the real thing.

## Provenance

Measured by #213's pass, running all 19 cases live against llama-server on the bundled Gemma at production's own `DEFAULT_LLM_MAX_OUTPUT_TOKENS`, with no truncation and no errors. Relayed; I have not re-run them. Note answer prose is not reproducible on this box, so a single observation of a *wording* is weak — but "names no topic" is a property, not a wording, and it held for both cases in the shipped shape.

### LINKED ISSUE #215: CitationEvalTest executes no assertion at all for its three degraded cases
## What

`CitationEvalTest.citationAccuracy_perCase:94` reads:

```java
if (!knownDegraded) { assertTrue(f1 >= 0.5) }
```

For the **3 of 26** cases tagged `missing` / `malformed` — `citation-missing`, `citation-malformed-json`, `citation-trailing-comma-in-array` — **no assertion executes at all.**

The comment claims *"F1=0 is the expected (correct) system behavior"*, but nothing asserts F1 **is** 0. So a malformed response that began yielding **spurious citations** would pass this test silently — and fabricated citations are the exact failure #103 is about.

## The fourteenth instance of one class

This is the fourteenth thing found in this repository that reports success without checking anything: ten Python gates and scorers (fixed in #202), then `AbsentDataEvalTest` excluding every absent-data case it exists to test (#203), `chipSeverityRank` returning a sentinel the caller skips (#207), an assertion at `InteractionRouteVariantTest:152` that cannot fail (#212), and now this.

Fourteen is a systematic gap, not a run of bad luck. The common shape is a check whose *negative* branch was never considered: an exclusion filter, a sentinel return, a guard clause, a skipped expectation. Every one would have been caught at write time by §11 of the working brief — name the condition the check catches, and prove by mutation that it catches it.

## Shape of the fix

Assert the degraded expectation rather than skipping it: for those three cases F1 **should** be 0, so assert that. That converts a silent skip into a real check of the behaviour the comment already claims is correct — which is strictly more than the test does today, at no cost.

Per §11, prove it by mutation: make the parser emit a citation for a malformed response and confirm the test goes red.

## Provenance

Found by #213's sweep while auditing test code; the 3-of-26 count comes from the dataset, the reasoning from the code. Reported rather than fixed because it is a test file and CLAUDE.md routes that decision to a human. I have not re-derived it.

### LINKED ISSUE #216: The absent-data oracle cannot express "names the topic by any of its names"
## What

`AbsentDataEvalTest` reads `expectedAnswerContains` with **AND** semantics — `for (String expected : contains)`, every element must appear in the answer. That cannot express the property the dataset actually cares about, which is *"the answer names what was asked about"*, whenever a topic has more than one correct name.

Concrete case, measured: **`absent-radiology-xray`** expects the substring `"ray"`. The shipped configuration answers *"No **radiology** reports are recorded."* — which names the topic, and contains no `"ray"`. Both *"No radiology reports are recorded"* and *"No x-ray reports are recorded"* are correct answers, and **they share no substring**, so no single literal accepts both. Adding a second element makes it stricter, not looser.

I corrected the sibling case `absent-smoking` (`"smoke"` → `"smok"`, the stem shared by smoking/smoke/smoker) because a single stem happened to work there. This one has no such stem, so editing the literal would pin whichever wording was observed rather than the property — which is how a test stops testing what its name says.

## Shape of the fix

Add `expectedAnswerContainsAny` (OR) alongside the existing AND list, and move the multi-name topics onto it. That is a small harness change and it makes the dataset able to say what it means.

Worth doing carefully rather than quickly, for two reasons. First, an OR list is weaker per element, so each case should use the narrowest set that covers the genuinely-correct wordings — not a grab-bag that accepts anything. Second, per §11 of the working brief, whatever is added must be proved by mutation: an answer naming **no** topic must still fail.

## Why it is filed rather than fixed

#213 repaired four checks that reported success without checking anything, and this is the opposite failure — a check that **fails while the system is correct**. Left alone deliberately: it is a harness design change rather than a literal, and it needs its own measurement over all 19 cases to see how many topics have this shape.

Related: **#214** covers the two cases in that same run that are genuine defects (answers reading *"No patient records were provided."*, naming no topic at all). Those are the real thing; this one is the oracle mis-scoring a correct answer, and it is important not to conflate them — one is the system being wrong, the other is the test being wrong.

## Provenance

The AND semantics I verified myself in `AbsentDataEvalTest`. The `absent-radiology-xray` answer text is #213's live measurement against the bundled model in the shipped configuration; relayed, I have not re-run it.

dkayiwa:
Three findings that #213 surfaced when it repaired `AbsentDataEvalTest` so its 19 absent-data
cases actually run. One is a production defect, two are oracles that could not say what they meant.

## Why they go together

#216 had to be fixed **first**. It is the oracle being unable to express "the answer names the
topic by any of its names", and that is exactly the property #214's fix has to satisfy — fixing the
defect first would have meant scoring it with a broken ruler. It earned that ordering twice over:
#214's fix immediately produced a second instance of #216's defect (below), which the AND-only
oracle scored as a failure while the answer was correct.

#215 is in the same slice because it is the same class of thing — a check that reports success
without checking anything — and because the fix needed a measurement of the harness's own F1
convention, which is what showed the issue's stated remedy would have failed two of the three cases
it was meant to cover.

## #214 — the diagnosis, before the fix

The issue's framing is that two topics — `absent-social-history` and `absent-imaging` — behave
differently from the other 15. **That does not survive measurement.** All figures below are the 19
absent-data cases run through `AbsentDataEvalTest` with `-Dchartsearchai.absent.data.test=true`
against llama-server on the module's own bundled `gemma-4-E4B-it-Q4_K_M`, empty chart, at
production's `DEFAULT_LLM_MAX_OUTPUT_TOKENS` (4096), temperature 0.

| arm | request order | named their topic | topic-less answers |
|---|---|---|---|
| `main` 57cb36b | dataset order | **17/19**, identical over 3 runs | `absent-smoking`, `absent-vaccination-history` |
| `main` 57cb36b | reversed | **16/19** | those two **plus** `absent-episodes` |

Neither topic #214 names failed in any run here. `absent-social-history` answered *"No social
history is recorded."* and `absent-imaging` *"No imaging studies are recorded."*, in every run, in
both orders. So "these two topics" is a per-session artifact, not the defect.

Two measurements pin what the defect actually is.

**It is not a property of the question.** Reversing the order the cases are asked in kept both
failures and added a third: `absent-episodes` answered *"No episodes are recorded."* at the tail of
the forward run and *"No records are provided."* at the head of the reversed one. Same question,
same prompt bytes, different answer — the `cache_prompt` non-determinism `LocalLlmEngine:786-792`
already documents.

**But two questions fail robustly, and they have something in common.** Every answer that named its
topic lifted a contiguous noun phrase straight out of the question: *"No imaging studies are
recorded."*, *"No CBC lab panel results are recorded."*, *"No eye exams or ophthalmology visits are
recorded."* The two robust failures are the two questions that offer none to lift — *"Does the
patient smoke?"* states its topic as a verb, and *"What vaccines has the patient received?"* splits
it across the verb.

So the mechanism is: **on an empty chart the prompt carries two competing models of what an
absent-topic answer looks like, and does not say which wins.** The few-shot demonstrations name the
topic (*"There are no records of banana deliveries."*). `normalizeRecords()`'s placeholder — the
nearest text to the query, and asserted byte-for-byte by this suite — names only records: *"This
patient has no records matching this query."* And the instruction meant to settle it, *"If no
records are relevant, name what is missing."*, is ambiguous precisely here: with an empty chart the
missing thing the prompt most recently named **is** the records, so paraphrasing the placeholder
satisfies the instruction on one reading. Which reading wins is then decided by whatever the KV
cache holds — which is why the failing set moves, and why a question with no noun phrase to lift
loses every time.

It matters clinically for the reason #214 gives: *"No patient records were provided."* cannot
distinguish "this patient has no imaging", "the chart failed to load" and "the question was not
understood". It is #94's shape — describing the retrieved slice rather than the patient.

### The fix

`DEFAULT_SYSTEM_PROMPT` keeps its existing sentence verbatim and adds the two things the
measurement asked for — name the topic rather than the records, and build a noun phrase when the
query supplies only a verb:

```
If no records are relevant, name what is missing. Name the TOPIC the query asked about, as a
noun phrase of your own when the query states it as a verb — asked "does the patient drink?":
"No alcohol use is recorded." Reporting only that records are missing names nothing.
```

The demonstration topic (`drink` → `alcohol use`) is deliberately held out of the eval set, so the
instruction teaches the operation rather than the two answers it has to produce. The hunk is
append-only: every existing assertion on `DEFAULT_SYSTEM_PROMPT` is a `contains()`, including the
exact sentence with its full stop, and all still hold.

| arm | request order | named their topic |
|---|---|---|
| `main` + this hunk | dataset order | **19/19** |
| `main` + this hunk | reversed | **19/19** |

38 case-observations, **0** topic-less answers, against **5** on `main` over the same 38. Both
robust failures moved, and the order-dependent one did too:

| case | `main` | with the fix |
|---|---|---|
| `absent-smoking` | `No records are provided.` | `No smoking history is recorded.` |
| `absent-vaccination-history` | `No patient records were provided.` | `No vaccination records are recorded.` |
| `absent-episodes` (head of reversed run) | `No records are provided.` | `No episodes are recorded.` |
| `absent-all-stopwords` | `The query is incomplete and does not specify what the patient is being asked about.` | `No specific topic is recorded.` |

Nothing that already named its topic stopped: all 17 forward passes held in both arms.
`absent-all-stopwords` is the one non-failure that changed wording — its question names no topic, so
both answers are the same refusal, the second in this prompt's voice. Both halves of the hunk are
pinned by their own assertion in `theEmptyChartPromptAsksTheModelToNameWhatIsMissing`, which is the
only absent-data assertion that runs without an endpoint.

**Two limits, stated plainly.** The suite measures the empty-chart shape, which is where the
placeholder exists to be echoed; the full-chart shape is documented in `chartText()`'s javadoc as
tried and rejected because the dataset's own premises fail against it. And the live sweep below is a
regression control here, not a demonstration: on the live REST path `main` answered all six
absent-topic probes naming their topic, so that pass could not have shown the fix working. At a
~2-in-19 per-question rate, six clean questions is the likely outcome (0.9^6 ≈ 53%), and the arm
that does discriminate is the 38-observation table above.

## #216 — what the oracle can now express

`expectedAnswerContains` is read with AND semantics (`for (String expected : contains)`), so it
cannot say "names the topic by any of its names". `absent-radiology-xray` asks *"Are there any X-ray
or radiology reports?"* and expects the literal `ray`; both *"No x-ray reports are recorded."* and
*"No radiology reports are recorded."* name the topic, they share no substring, and adding a second
AND element makes the case **stricter**.

`EvalCase` gains `expectedAnswerContainsAny` (OR), and three cases move onto it — the three whose
own **question** offers two names:

| case | was | now |
|---|---|---|
| `absent-radiology-xray` | `contains: ["ray"]` | `containsAny: ["ray", "radiolog"]` |
| `absent-ophthalmology` | `contains: ["eye"]` | `containsAny: ["eye", "ophthalmolog"]` |
| `absent-nutrition-diet` | `contains: ["nutrition"]` | `containsAny: ["nutrition", "diet"]` |

That criterion is the question text, not an observed wording, so it does not pin whichever answer
happened to be measured. `absent-cbc-lab-panel` (`CBC` / *complete blood count*) and
`absent-mental-health` (`mental`+`health` / *psychiatric*) plausibly have the same shape and are
**left alone**: no run has produced the alternative, and an unobserved alternative is the grab-bag
this issue warns about. The oracle can express them the day one appears.

One case moves the other way, to the narrowest single literal covering the variants:
`absent-vaccination-history` `["vaccines"]` → `["vaccin"]`. This is #216's defect **produced by
#214's fix** — measured, the answer is now *"No vaccination records are recorded."*, which names the
topic and contains no `"vaccines"`. It is the same correction #213 made for smoking
(`"smoke"` → `"smok"`), on the same kind of evidence. `immunization` is another correct name for
that topic and is deliberately **not** added, for the reason above.

### The mutation proving a topic-less answer still fails

An OR element is weaker than an AND element, so the bound is asserted rather than trusted.
`anAnswerNamingNoTopicFailsEveryCaseThatNamesOne` feeds the oracle the two answers measured naming
no topic — `"No patient records were provided."` and `"No records are provided."` — and asserts the
set of cases that **accept** them is exactly `{absent-all-stopwords}`, whose question names no topic
so a negation is all its answer can be held to. The exemption is asserted, not skipped, so it cannot
quietly grow.

Sensitivity, by mutation on a throwaway tree — adding one grab-bag alternative to any case:

```
"expectedAnswerContainsAny": ["ray", "radiolog", "record"]
→ AbsentDataEvalTest.anAnswerNamingNoTopicFailsEveryCaseThatNamesOne
   expected: <[absent-all-stopwords]> but was: <[absent-all-stopwords, absent-radiology-xray]>
```

`everyAbsentCaseIsRunAndEveryExpectationDiscriminates` gains the matching per-case half: an answer
carrying **none** of a group's alternatives must fail the case. Measured on the same tree, it catches
the OR check being neutered (`named || true` → *"Expected java.lang.AssertionError to be thrown, but
nothing was thrown"*) and being reverted to AND semantics. Its vacuity guard now accepts either list,
so a case still cannot join the run asserting nothing.

A second half was written and then **deleted**, in its own commit: "each alternative on its own must
satisfy the case" is asserted against an answer built *from* that alternative, so it holds by
construction. Measured — adding a misspelled third alternative (`"raidolog"` beside `"ray"` and
`"radiolog"`), which no real answer could match, leaves all 22 tests green including that loop. A
dead alternative is not detectable in this file at all, and the comment now says so rather than
leaving an assertion that cannot fail sitting in a PR about assertions that cannot fail.

## #215 — the assertion, and its mutation

`citationAccuracy_perCase:94` read `if (!knownDegraded) { assertTrue(f1 >= 0.5) }`, so the 3 of 26
cases tagged `missing`/`malformed` executed **no assertion at all**.

The comment claiming F1=0 was the expected behaviour is also **wrong on the arithmetic**, which is
why the issue's stated remedy is not what landed. Measured:

| case | expected | predicted | F1 |
|---|---|---|---|
| `citation-missing` | `[8]` | `[]` | 0.000 |
| `citation-malformed-json` | `[]` | `[]` | **1.000** (`EvalMetrics.f1(∅,∅) = 1.0`) |
| `citation-trailing-comma-in-array` | `[1, 2]` | `[1, 2]` | **1.000** (the regex salvage path recovers both) |

So `assertEquals(0.0, f1)` would have failed two of the three cases it was meant to cover. Both
branches assert now, and the degraded branch asserts on the **predicted set**, which is strictly
stronger than F1=0: F1=0 is also what a wholly wrong prediction scores, so it cannot tell "produced
nothing" from "produced the wrong thing" — and "produced the wrong thing" is the spurious-citation
failure #103 is about.

```java
if (isKnownDegraded(evalCase)) {
    assertTrue(predictedIndices.isEmpty(),
            caseId + ": a response carrying no recoverable citation must yield NO references, but "
                    + "yielded " + predictedIndices + " …");
} else {
    assertTrue(f1 >= 0.5, …);
}
```

`citation-trailing-comma-in-array` loses its `malformed` tag. The tag classifies the **outcome**
here, its citations are recovered in full, its own `expectedRecordIndices` say that is expected, and
`trailing-comma` already names the input shape. It is now held to the normal threshold, which it
meets at 1.000 — strictly more than the nothing it asserted before. No expected value changed.

### The mutation

Narrowing `extractCitedReferences`'s abstention-dump carve-out to answers that say "no record" in
words, and giving every other unanchored answer the whole reviewed set — a plausible-shaped
narrowing of a carve-out whose javadoc records the failure it exists to stop:

```java
if (inline.isEmpty() && !ChartSearchAiUtils.isBlank(answer)) {
    if (answer.toLowerCase(Locale.ROOT).contains("no record")) {
        return new ArrayList<RecordReference>();
    }
    seen.addAll(indexMap.keySet());          // fabricate
}
```

Production then fabricates ten citations for two responses that carry none:

```
[citation-missing]        F1=0.182 predicted=[1,2,3,4,5,6,7,8,9,10] expected=[8]
[citation-malformed-json] F1=0.000 predicted=[1,2,3,4,5,6,7,8,9,10] expected=[]
Citation eval: avgF1=0.910 exactMatch=22/26
```

| test code | result |
|---|---|
| `main`'s `CitationEvalTest` | **27 tests, 0 failures** — green, aggregate `avgF1 0.910` still over its 0.8 floor |
| this branch's `CitationEvalTest` | **28 tests, 2 failures**, naming `citation-missing` and `citation-malformed-json` |

`citation-empty-no-records` is unaffected by that mutation (its answer contains "no records"), so
nothing else in the suite catches it — which is what makes this a canary rather than a control.

`everyCaseReachesTheRunAndBothAssertionBranchesAreOccupied` guards the two ways the suite could go
quiet without a count moving: a case added with no `simulatedLlmResponse` is dropped by the
`@MethodSource` silently (the #203 shape), and a bucket that empties retires one of the two branches
with nothing to say so.

## Build

Measured in clean worktrees off the same SHA, full reactor `mvn -o clean install`, nothing else
building.

| | api | omod | total | failures | skipped |
|---|---|---|---|---|---|
| `main` 57cb36b | 1045 | 62 | **1107** | 0 | 53 |
| this branch | 1047 | 62 | **1109** | 0 | 53 |

Delta **+2**, both new `@Test` methods, both unconditional:

- `AbsentDataEvalTest.anAnswerNamingNoTopicFailsEveryCaseThatNamesOne`
- `CitationEvalTest.everyCaseReachesTheRunAndBothAssertionBranchesAreOccupied`

Skipped is unchanged at 53 — the 19 absent-data cases stay behind
`-Dchartsearchai.absent.data.test=true`, plus `PromptInjectionEvalTest` and
`LlmAnswerQualityTest`.

With the flag on and the endpoint up: `Tests run: 22, Failures: 0, Errors: 0, Skipped: 0` —
19 cases plus the three unconditional ones.

## Live verification

Both arms deployed to the standalone (full-JVM restart, `sourceFormat=ddinter`, full 19 MB KB,
`chartMode=fullChart`, `llm.engine=local`, GPs untouched), same probe script, retrieval gate passed
in both (`querystoreBuild hits=75`). The branch omod was built at `fdd14c17`; the one commit after it
deletes an assertion from a test file and changes no production byte, so the deployed behaviour is
the behaviour of this branch's head.

### The ten controls — chip-level, identical in both arms

| probe | expected | `main` 57cb36b | this branch |
|---|---|---|---|
| Mary + clarithromycin | 1 Major ×simvastatin | `1 · interaction/Clarithromycin` | `1 · interaction/Clarithromycin` |
| Mary + paracetamol | 0 | `0` | `0` |
| Agnes + warfarin | 1 Major ×aspirin | `1 · interaction/Warfarin` | `1 · interaction/Warfarin` |
| Joshua + ibuprofen | 2 | `2 · contraindication/Ibuprofen ; interaction/Ibuprofen` | `2 · contraindication/Ibuprofen ; interaction/Ibuprofen` |
| Sarah + hydrocortisone | 8, four `Hydrocortisone` | `8 · contraindication/Hydrocortisone ; interaction/Hydrocortisone ; interaction/Hydrocortisone ; interaction/Hydrocortisone ; contraindication/Dexamethasone ; contraindication/Prednisone ; contraindication/Budesonide ; contraindication/Methylprednisolone` | identical |
| Kevin + hydrocortisone | 3 | `3 · contraindication/Hydrocortisone (ophthalmic) ; interaction/Hydrocortisone ; contraindication/Clarithromycin` | identical |
| George + esomeprazole | 1 | `1 · contraindication/Esomeprazole` | `1 · contraindication/Esomeprazole` |
| Susan Young + lidocaine | 3, third kept | `3 · interaction/Lidocaine ; interaction/Lidocaine ; contraindication/Tiotropium` | identical |
| Margaret + botulinum toxin type A | names `Botulinum toxin type A` | `1 · contraindication/Botulinum toxin type A` | `1 · contraindication/Botulinum toxin type A` |
| Agnes + epinephrine | 0 | `0` | `0` |

Answer text for the two that carry no chip, both arms — the shape the prompt hunk could most easily
have broken, since it is the abstention branch:

```
main    Mary + paracetamol   The records do not address paracetamol.
branch  Mary + paracetamol   The records do not address the safety of giving paracetamol.
main    Agnes + epinephrine  The records do not address the safety of giving Epinephrine.
branch  Agnes + epinephrine  The records do not address the safety of giving Epinephrine.
```

Each would have moved had the change touched the deterministic safety path: the chip set is produced
by `DrugSafetyValidator` from the patient's own records and the KB, never by the model, so a prompt
edit that leaked into record selection or drug-reference injection shows up here as a changed count
or a renamed subject. What they cannot do is corroborate #214's fix, and they are not offered as
doing so.

### Absent topics, live

A fixture patient was created for this: **Blank Chartless**, `a461e88b-9af1-484a-a2e7-d7bc87bdc8fa`,
OpenMRS ID `100023N`, no obs, orders, conditions or allergies — so the chart builder yields nothing
and `normalizeRecords()` emits the placeholder. That is the production shape #214's defect lives in;
no existing demo patient has an empty chart (the emptiest carry 3 obs each).

| question | `main` 57cb36b | this branch |
|---|---|---|
| Does the patient smoke? | `No smoking status is recorded.` | `No smoking habits are recorded.` |
| What vaccines has the patient received? | `No vaccination history is recorded.` | `No vaccination history is recorded.` |
| any episodes? | `No episodes are recorded.` | `No episodes are recorded.` |
| What imaging studies were done? | `No imaging studies are recorded.` | `No imaging studies are recorded.` |
| What is the patient's social history? | `No social history is recorded.` | `No social history is recorded.` |
| Are there any X-ray or radiology reports? | `No X-ray or radiology reports are recorded.` | `No X-ray or radiology reports are recorded.` |

And two on Mary Smith, whose chart is not empty, so no placeholder is in the prompt — the control
that the hunk has not made the model abstain where it should not:

```
main    Does the patient smoke?                    Yes — the patient has a Nicotine-related disorder [60], [61].
branch  Does the patient smoke?                    Yes — the patient has a Nicotine-related disorder [60], [61].
main    Are there any X-ray or radiology reports?  No X-ray or radiology reports are recorded.
branch  Are there any X-ray or radiology reports?  No X-ray or radiology reports are recorded.
```

Both arms name the topic on every one. As stated above, this pass did not reproduce the defect on
`main`, so it is a "nothing broke" result rather than evidence for the fix — reported that way
rather than dressed up.

## Not fixed, reported

1. **`citation-string-type-citations` cannot fail on what it names, and the behaviour behind it is a
   real gap.** `computePredictedIndices` substitutes `evalCase.getSimulatedCitations()` for the
   parsed value whenever the case carries one, and this is the only case where the two differ
   (`["9","10"]` in the response, `[9,10]` in the field). Measured, calling production directly:
   `LlmAnswerExtractor` yields `citations=[]` for `"citations": ["9","10"]` — `n.isInt()` rejects a
   `TextNode` — and the case only passes because its answer carries inline `[9] [10]` markers that
   the union rescues. Strip the markers and the same response yields **zero** references. Reachable:
   `response_format: json_object` does not constrain value types.
2. **`absent-all-stopwords`'s `["no"]` expectation is nearly vacuous.** `no` is a substring of
   `not`, `nothing`, `known` and `diagnosis`, and its live answer passes on *"does **no**t
   specify"*. It is the one case exempt from the topic-less guard, so the weakness is now at least
   documented and checked in place; strengthening it needs its own measurement of what that question
   should be held to.
3. **`absent-surgery` expects the plural-only literal `"surgeries"`.** Measured, it passes in every
   run here (*"No surgeries are recorded."*, *"No records of surgeries are recorded."*), so nothing is
   broken today — but *"No surgery is recorded."* and *"No surgical history is recorded."* both name
   the topic and both fail it, and the stem `surg` covers all four. Same class as the
   `"smoke"`/`"vaccines"` literals, left alone because no run has produced the failing wording and a
   literal should be corrected on evidence, not on inspection.
4. **`AbsentDataEvalTest`'s 19 cases share one llama-server KV cache, so each answer depends on the
   ones before it.** Measured above: `absent-episodes` gives different answers at position 1 and
   position 19. A future absent-data change measured in one order can therefore look better or worse
   than it is; both orders were run here for that reason.

Closes #214
Closes #216
Closes #215


dkayiwa:
Independent pre-merge verification: `mvn -o clean install` on a clean detached worktree → **1047 + 62 = 1109, 0 failures, 53 skipped**, BUILD SUCCESS. `main` at `57cb36b` = 1107. CI green on Java 11/17/21. The two removed assertion lines are **one assertion, widened** — `assertFalse(contains.isEmpty())` → `assertFalse(contains.isEmpty() && anyOf.isEmpty())`, so a case with neither list is still caught now that an OR list exists — plus its message continuation. Nothing weakened.

**Three corrections to issues I wrote, all measured, and the first is the most interesting.**

**#214 named the wrong two cases, and the reason it did is the defect itself.** I filed `absent-social-history` and `absent-imaging`; neither ever fails. The real failures are `absent-smoking` and `absent-vaccination-history`, plus `absent-episodes` in reversed order only. That last one answers *"No episodes are recorded."* at position 19 and *"No records are provided."* at position 1 — same question, same prompt bytes — so **which cases fail moves with cache state**, and #213's single-run observation named whichever two happened to lose. Running the set forward *and* reversed is what separated the mechanism from the artefact.

The diagnosis underneath is the part worth keeping: on an empty chart the prompt carries two competing models of an absent-topic answer — the few-shot demos, which name the topic, and `normalizeRecords()`'s placeholder, which names only records — and *"name what is missing"* is ambiguous exactly there, because with an empty chart the missing thing most recently named **is** the records. Plus a deterministic half: every passing answer lifts a contiguous noun phrase out of its question, and the two robust failures are the two questions that offer none (*"Does the patient smoke?"* — a verb; *"What vaccines has the patient received?"* — split across one). 17/19 → **19/19 in both orders**, 38 case-observations, 0 topic-less against 5 on `main`.

**#215's proposed fix would itself have been a new blind check.** I wrote "assert F1 is 0". Measured: `citation-malformed-json` scores F1 = **1.000** because `f1(∅,∅) = 1.0`, and `citation-trailing-comma-in-array` scores **1.000** because salvage recovers both citations — that case is not degraded at all. What shipped is stronger and says what the test means: `assertTrue(predictedIndices.isEmpty())`. The mutation makes the point — narrowing the abstention carve-out has production fabricate `[1..10]` for two responses carrying none, and **`main`'s test stays green** at 27/0 with avgF1 0.910 still over its floor.

**#216 gained a second live instance during its own fix** — the #214 change produced *"No vaccination records are recorded."*, so `absent-vaccination-history` moved to the stem `["vaccin"]`. Two cases were deliberately **not** given OR lists because no run produced the alternative, and the grab-bag mutation is measured: adding `"record"` to radiology's list reddens `anAnswerNamingNoTopicFailsEveryCaseThatNamesOne`.

It also **deleted an assertion of its own** that could not fail, documenting why in place with the measurement — a misspelled alternative passes it, because the answer is built from the alternative.

Filed: **#219**, the fifteenth blind check, and it carries a production defect. `extractResponse` yields `citations=[]` for `"citations": ["9","10"]` because `n.isInt()` rejects a `TextNode`, so string-typed citations are **silently dropped**; the case passes only because inline markers rescue it, and stripping them yields zero references.

Standalone note for the next pass: the deployed omod is this **branch** build, so `main` needs redeploying before anything measures `main`. Fixture left in place: patient **Blank Chartless** `a461e88b-…`, built to reach the empty-chart placeholder path live.

PR #225, opened 2026-08-12
fix(drug-safety): decide restating existing therapy by substance, not by a shared ATC code

### LINKED ISSUE #185: A drug can be reported as duplicating itself: the existing-therapy skip tests ATC codes, not substance identity
## What

The "restating existing therapy" skip that stops a chip claiming a drug duplicates the patient's own order compares **ATC codes**, not substance identity. When the KB row for a drug carries a code the ordered drug's row does not, the skip misses and the chip fires against the order itself.

Seven shipped rows have the shape. The clearest: the KB's `Omeprazole` entry carries only `A02BC05` with `rxnorm_name` "esomeprazole", and `A02BC01` is absent KB-wide — so an omeprazole order yields

```
Omeprazole is in the same ATC class (A02BC) as active order Omeprazole 20mg
```

## Reachability

**Not reachable on the demo dictionary** — checked live during #182's hardening; that dictionary's Omeprazole concept carries no `WHOATC` map, so the class arm never engages. It becomes reachable on any dictionary that maps omeprazole (or the other six) to ATC, which is a normal thing for an implementation to do.

Pre-existing, not introduced by #182 — but #155's label ladder makes it **visible**, since the partner now renders as the order's own display name rather than a bare code.

## Relationship to #164

#164 is the same symptom — a substance reported against itself — reached by two different routes (the mirror path where the allergen row is a sibling not in play, and stem divergence). This is a third route, in the duplicate-therapy arm, via code-set asymmetry rather than labelling.

All three share one cause: **identity is being decided by something other than the substance** — entry identity in #164, ATC codes here. #173 built `substanceKey()`/`substanceGroupKey()` precisely to express substance identity, so the fix is likely to be the same one in three places, and worth doing as one pass.

## Provenance

Measured against the shipped KB by #182's hardening pass, with the demo-dictionary unreachability checked live. Relayed; I have not re-run either.

dkayiwa:
Fixes #185.

## What was wrong

`DrugSafetyValidator.classRelationships` skipped an active order that is "the same drug as `ref`" by asking whether the two share an **exact ATC code**. That test is close to sufficient and nowhere near necessary, and it is the *necessary* half the arm wanted. Where the code the co-medication carries and the code the reference row publishes are different codes of one substance, the sets are disjoint, their level-4 subgroups still meet, and the chip fires against the patient's own order:

```
Omeprazole is in the same ATC class (A02BC) as active order Omeprazole 20mg — possible duplicate therapy
```

This is the third route to the symptom #164/#187 fixed on the two cross-reactivity routes, and it has the same cause: **identity decided by something other than the substance**. #173 built `substanceKey()`/`substanceGroupKey()` to express exactly that.

## The KB facts, re-derived

Every figure below was produced by driving production code — `DdiDrugReferenceSource.parse` over the shipped 19 MB KB, then `DrugReference.normalizedAtcCodes()`, `getSubstanceName()`, `substanceKey()`, `substanceGroupKey()`. Nothing was re-expressed in Python/jq/SQL (HARDEN-BRIEF §10).

| claim | measured |
|---|---|
| the KB's `Omeprazole` row publishes only `A02BC05`, with `rxnorm_name` "esomeprazole" | **confirmed** — `normalizedAtcCodes()=[A02BC05]`, `getSubstanceName()=esomeprazole`, `substanceKey()=[esomeprazole, omeprazole]` |
| `A02BC01` (omeprazole's own code) is absent KB-wide | **confirmed** — 0 occurrences over `normalizedAtcCodes()` of all 2283 entries |
| "seven shipped rows have the shape" | **disagrees.** By the predicate I could state in production terms — two entries of DIFFERENT substances (`substanceGroupKey()`) publishing IDENTICAL non-empty `normalizedAtcCodes()`, so at least one of them carries a code that is not its own — the KB holds **21 pairs across 34 rows**, all 21 also sharing a `getSubstanceName()`. Both figures are stated because I could not reconstruct a predicate that yields 7; the issue records its own figure as relayed and never re-run. |

The 21 pairs include `Ketoconazole`/`Levoketoconazole`, `Atropine`/`Hyoscyamine`, `Gabapentin`/`Gabapentin enacarbil`, `Estrone`/`Estrone sulfate`, `Fenofibrate`/`Fenofibric acid`, `Hydrocortisone`/`Hydrocortisone butyrate`, `Benzylpenicillin`/`Procaine benzylpenicillin`, `Mometasone`/`Mometasone furoate`, `Iodide I-131`/`Iodine`, `Dextroamphetamine`/`Amphetamine`, `Dexfenfluramine`/`Fenfluramine`, `Antithrombin Alfa`/`Antithrombin III human`.

## Reachability: the issue's claim is wrong

The issue records this as **"not reachable on the demo dictionary"**. It is reachable, and it was reproduced live.

Driving the real `DrugSafetyValidator.validate` over the shipped 19 MB KB, for **all 116** orderable names the 3.7.1 demo dictionary actually maps to `WHOATC` (read from the standalone's own database), asking about each substance that order's own name implies — the set `DrugReferenceService.findImpliedByDrugName` returns for it, keyed by `substanceGroupKey()`:

* **before this change: 5 self duplicate-therapy chips**
* **after: 0**

The five are `Isoniazid / Rifapentine`, `Pyrimethamine / sulfadoxine`, `Sulfadoxine and pyrimethamine`, `Diphtheria toxoid / tetanus toxoid` and `Diptheria tetanus booster` — every one a **combination concept mapped to a combination ATC code the KB does not carry**, so the partner's identity comes from the order's name and not from its code.

Worth carrying past this PR: this is the third time in this work that a "the class arm cannot fire here" reachability claim has proved false. Sparse ATC coverage is not zero coverage, and the way to settle it is to drive `validate` over the dictionary's own mappings rather than to reason about which concepts carry them.

The issue's omeprazole example genuinely is unreachable there, and for the stated reason: the dictionary's `Omeprazole` concept (4034) carries no `WHOATC` map. Its `Esomeprazole` concept (83) does, to `A02BC05` — which is why the PPI shape needs a mapped dictionary and the combination shape does not.

**#155's label ladder is not part of the fix.** It only made the defect legible ("as active order Omeprazole 20mg" rather than "as active order A02BC01"). The chip is equally wrong under the old wording; nothing about the wording changed here.

## The fix

The skip now asks the question it was always about — *is this partner the drug being asked about?* — with two legs, neither of which subsumes the other:

1. **`DrugReference.substanceGroupKey()`** against the substances the partner's ORDER is recorded as naming, resolved through **`DrugReferenceService.findImpliedByDrugName`** (the ranked accessor CLAUDE.md nominates for a recorded drug name — never the unranked matcher underneath it, because this decides which chips are *silenced*).
2. **A shared exact ATC code**, unchanged. It is *not* replaceable: where the context carries only the flattened code set (#118's fallback) no order names the partner, so the partner is whichever row `canonicalRow` picks for that code and its substance need not be the queried one. Measured — dropping this leg newly raises a chip about `Esomeprazole` against an order known only as `A02BC05`, and reddens **6 pre-existing tests**.

The order carrying a code is now resolved for **every** code, not only for one the dataset cannot name, because a fixed-dose combination resolves an entry for one constituent while its name names the other. That is the shape reachable on the stock demo dictionary. The identity/label ladder is untouched: the `else` branch that consults the order is reachable only when the entry rung failed, exactly as before.

**The name leg is read on exactly the rung that takes the order's NAME** — inside `orderPartners`' existing `unnameableCode && order != null` block, beside `OrderPartner.nameByOrder`. Two review rounds were needed to get that scoping right, and both failures are pinned as tests:

* Adding the order's substances to *every* partner it produced over-skips: an order mapped to two **covered** codes is two partners, and a question about the first constituent then silenced the second — `Metronidazole and secnidazole` losing "Metronidazole is in the same ATC class (P01AB) as active order Secnidazole", which is the finding the arm exists to make. Pinned by `PartialOrderCoveragePartnerTest.askingAboutOneConstituentOfThatCombinationStillReportsTheOther`.
* Gating instead on the partner's `namedByOrder` **flag** is order-dependent: the flag can be set by a later code, after which a code belonging to a *different* order passes the gate and attributes that order's constituents to this partner. Pinned by `DuplicateTherapySelfChipTest.whichOrderNAMESAPartnerDoesNotDependOnTheOrderOfTheCodes`.
* Reading only `orderCarrying`'s **single** answer is order-dependent the other way: a partner can be reached by several orders, and two orders sharing one unnameable code flipped the verdict when the ORDERS list was permuted. Every carrier contributes (`ordersCarrying`); the label may still take the first, because naming a partner after one of its orders is a presentation choice and a suppression is not. Pinned by `DuplicateTherapySelfChipTest.norOnTheOrderTheORDERSListHappensToPutFirst`.

Both were "same patient, same prescriptions, different suppression depending on the sequence `OrderService` returned them in", and both pins were confirmed discriminating by mutation. The result is now a union over a condition that is a pure function of the code, so it cannot depend on iteration order by construction.

**The bound this leaves, stated rather than implied:** a combination order whose combination code the KB *does* carry never reaches this rung, so a question about one of its constituents still gets the self-chip. Not closable here — reading the order's names for a *covered* code is exactly the over-skip above — and unreachable on the shipped KB, which carries no row for any combination code the demo dictionary maps.

### The whole reachable population, chip for chip

The sharper question is not "does the defect go away" but "does anything else move". Answered by census rather than argument: **all 116** of the demo dictionary's `WHOATC`-mapped orderable names, each asked about **every** substance the shipped 19 MB KB files in a level-4 subgroup that order carries — 606 questions through the real `validate`.

| | main | branch |
|---|---|---|
| chips raised | 471 | 466 |

The diff is **exactly five lines, all deletions, and every one of them a self-chip** (the two tetanus-toxoid combinations, `Isoniazid / Rifapentine`, and the two sulfadoxine/pyrimethamine spellings). Nothing gained, nothing else lost. That also bounds the one residual hazard the sweep raised — an order name that merely *contains* another substance's whole name resolving it, and so silencing rather than fabricating: on the population that actually exists, it does not happen.

Composition, checked: the pair cap (`maxPairChips`) applies only to the two pairwise arms, and `InteractionPairs` records only RULE chips — so a class-only chip that stops being raised **cannot move a capped or withheld count**, and cannot change what the screening arm stands down from.

## Tests

New: `DuplicateTherapySelfChipTest` (13 cases) over `ddi-duplicate-therapy-self.json`, a new fixture of five **verbatim** KB rows (the three A02BC substances plus isoniazid/rifapentine) and every KB interaction row among them, plus one case added to `PartialOrderCoveragePartnerTest`, whose fixture is the one with a combination order resolving to two partners.

`Esomeprazole` is the case that keeps the fix honest: it shares the `Omeprazole` row's `rxnorm_name`, `rxcui` **and** its one ATC code, and differs only in `drugbank_id` — so it is a different substance by `substanceKey()` and by nothing else the row publishes. Its chip must survive, and does.

### Mutation sweep (throwaway worktree)

| mutation | tests red |
|---|---|
| M1 skip = exact ATC code only (pre-#185) | 2 — both new |
| M2 skip = substance identity only (code leg dropped) | 7 — 6 pre-existing + 1 new |
| M3 order-name leg only when the code resolved nothing | 1 — new |
| M4 `findByDrugName` instead of `findImpliedByDrugName` | **0** → 1 after adding a pin |
| M5 `substanceKey()` instead of `substanceGroupKey()` | 4 — all pre-existing |
| M6 every order's substances onto every partner (attribution dropped) | **0** → 1 after adding a pin |
| M7 name leg on every partner, not only order-named ones | 1 — the new `PartialOrderCoveragePartnerTest` case |
| M8 substances read before `nameByOrder` can flip the gate (order-dependent) | 2 — including the code-order-independence case |

M4 and M6 were decisions with **no test at all**; both are now pinned, and both pins fail on the mutation and pass on the fix. M7 is the scoping rule above, added after the sweep found the over-skip.

## Live verification

Deployed to the local 3.7.1 standalone (`sourceFormat=ddinter`, `entryCount=2283`, origin `appdata:chartsearchai/ddi_knowledge_base.json`), before and after, each in its own fresh JVM. Fixture: **one** `Isoniazid / Rifapentine` order (the dictionary's own concept 143, its own `WHOATC` maps) on Michelle Lewis, who has a real chart and — verified in the database, not through the order REST endpoint — zero other orders and zero allergies.

| probe | before | after |
|---|---|---|
| Michelle × "Is it safe to give isoniazid?" | **1 chip** — `Isoniazid is in the same ATC class (J04AC) as active order Isoniazid / Rifapentine` | **0 chips** |
| Michelle × "Is it safe to give rifampicin?" | 2 chips — Minor rule + `… same ATC class (J04AB) as active order Isoniazid / Rifapentine` | **2 chips, identical** |
| Michelle × "Is it safe to give rifapentine?" | 0 | 0 |
| Michelle × "Is it safe to give ethambutol?" | 1 — Moderate rule | 1 — identical |
| Mary × clarithromycin | 1 Major ×simvastatin | unchanged |
| Agnes × warfarin | 1 Major ×aspirin | unchanged |
| Joshua × ibuprofen | 2 | unchanged |
| Mary × paracetamol | 0 | unchanged |

Every probe was diffed chip-string for chip-string, not just by count: across all eight, **exactly one chip moves and it is the defect**. The run was repeated at the branch's final head after the scoping fix landed, and is byte-identical to the table's `after` column.

The rifampicin probe is the one that makes the isoniazid zero mean something: it is a *different* substance in the same ATC subgroup as the same order, and its class chip **names the order**, which is only possible if the order reached the context with both its name and its codes. Had the fix been a blanket "the patient is already on something in this class" suppression, that chip would have vanished too.

## Cost

The name resolution runs only for a code the dataset cannot name, and is memoised per order. Timed interleaved against `main` — twice per side, over the shipped 19 MB KB with a 30-order, 60-name context — the two are indistinguishable within this machine's run-to-run spread (main 63/86 ms, branch 86/87 ms). No figure is quoted in the javadoc for that reason: an earlier unpaired pair of runs suggested 61 → 106 ms and the interleaved re-measurement did not reproduce it.

## Build

`mvn -o clean install`, full reactor, measured in a clean worktree: **1068 api + 62 omod on `main` (89cb6fe9)**, **1082 + 62 at this branch's head** — +14, which is the 13 new cases in `DuplicateTherapySelfChipTest` plus the one added to `PartialOrderCoveragePartnerTest`. 0 failures, 53 skipped, both.


dkayiwa:
### Verification record

**Build.** Full reactor `mvn -o clean install`, measured in clean worktrees: `main` (89cb6fe9) **1068 api + 62 omod**; this branch's head **1081 + 62**. +13 = the 12 cases in `DuplicateTherapySelfChipTest` plus the one added to `PartialOrderCoveragePartnerTest`. 0 failures, 53 skipped, both sides.

**Mutation sweep** (throwaway worktree, `git commit` path-limited throughout). Every decision in the slice was reverted in turn and the suite re-run:

| | mutation | red |
|---|---|---|
| M1 | skip = exact ATC code only (pre-#185) | 2 |
| M2 | skip = substance identity only (code leg dropped) | 7 — 6 pre-existing |
| M3 | order-name leg only when the code resolved nothing | 1 |
| M4 | `findByDrugName` instead of `findImpliedByDrugName` | **0** → 1 after adding a pin |
| M5 | `substanceKey()` instead of `substanceGroupKey()` | 2 pre-existing |
| M6 | every order's substances onto every partner | **0** → 1 after adding a pin |
| M7 | name leg on every partner, not only order-named ones | 1 |
| M9 | gate on the partner's `namedByOrder` flag (order-dependent) | 1 |

M4 and M6 were decisions with no test at all when first written; both are now pinned, and each pin fails on its mutation and passes on the fix.

**Review.** Five review passes, three of them by independent read-only reviewers over the whole repo rather than the diff. They found two real defects in the first version of this change — a combination order resolving to two partners silencing both, and an order-dependent leak through the partner's `namedByOrder` flag — plus about a dozen false or stale prose claims, in this change's own new text and in neighbours it falsified (`docs/adr.md` Decision 24, `orderPartners`' javadoc, the README row, the GP description). All applied; both defects are pinned as tests confirmed discriminating by mutation.

One measurement was **withdrawn** rather than corrected: an earlier revision quoted 61 ms → 106 ms for `validate` on a 30-order context. Re-measured interleaved and twice per side, main and this branch are indistinguishable within this machine's run-to-run spread (main 63/86 ms, branch 86/87 ms). The original was one unpaired run each and should not have been quoted; the javadoc now states the mechanism and how to measure it instead of a number.

**Live.** Deployed to the 3.7.1 standalone three times — `main`, the first fix, and the final head — each in its own fresh JVM with the 19 MB DDInter KB loaded (`entryCount=2283`, verified through `GET /chartsearchai/drugreferencestatus`, never inferred from the GPs). Eight probes each, diffed chip-string for chip-string: **exactly one chip moves, and it is the defect.** The final head is byte-identical to the first fix run.

The fixture is one `Isoniazid / Rifapentine` order — the demo dictionary's own concept 143 and its own `WHOATC` maps, nothing about it created — on a patient with a real chart and, verified in the database rather than through the order REST endpoint, zero other orders and zero allergies.

dkayiwa:
Two further review passes landed after the verification record above. Both found real order-dependence in the first shape of this fix, and both are now pinned:

| | mutation | red |
|---|---|---|
| M9 | gate the name leg on the partner's `namedByOrder` flag | 1 |
| M10 | read only the first order carrying the code | 1 |

The final shape reads the order's names on exactly the rung that takes the order's NAME — inside `orderPartners`' existing `unnameableCode && order != null` block — and reads **every** order carrying that code rather than the first. Both properties are pure functions of the code, so the result is a union and cannot depend on the sequence `OrderService` returns prescriptions in. The label still takes the first carrier, deliberately: naming a partner after one of its orders is a presentation choice, and a suppression is not.

Re-measured after every change: the demo-dictionary reachability sweep is still **0** self-chips, and the 606-question census is **byte-identical** to the revision before — so nothing on the reachable population moved, and the diff against `main` remains exactly the five self-chips.

A residual bound is now stated in `OrderPartner.substances` rather than left implied: a combination order whose combination code the KB *does* carry never reaches this rung, so a question about one of its constituents still gets the self-chip. That is not closable from here — reading the order's names for a *covered* code is precisely the over-skip M7 catches — and it is unreachable on the shipped KB, which carries no row for any combination code the 3.7.1 demo dictionary maps.

Test counts at the final head: **1082 api + 62 omod**, 0 failures, 53 skipped (`main` at 89cb6fe9: 1068 + 62).

Issue #219, opened 2026-08-09
citation-string-type-citations cannot fail on what it names: string-typed citations are silently dropped


dkayiwa:
## What

`citation-string-type-citations` is meant to check that a response whose `citations` array holds **strings** rather than integers still yields citations. It cannot fail on that, and the reason is a real production gap underneath.

Measured by calling production directly: `extractResponse` yields **`citations=[]`** for `"citations": ["9","10"]`, because `n.isInt()` rejects a `TextNode`. The case passes only because the answer also carries inline `[9] [10]` markers, which rescue it — **strip those markers and the same response yields zero references.**

Two defects in one, then:

1. **The test is blind.** `computePredictedIndices` substitutes `simulatedCitations` for the parsed value, and this is the only case in the dataset where the two differ — so the test asserts against a value production never produced.
2. **Production silently drops string-typed citations.** `response_format: json_object` does not constrain value types, so a model returning `["9","10"]` instead of `[9,10]` is well-formed JSON that this module discards without a word. When the answer's inline markers happen to agree, nothing is visible; when they don't, references vanish.

## The fifteenth of one class

Fifteen things in this repository have now been found reporting success without checking anything: ten Python gates and scorers (#202), then `AbsentDataEvalTest` excluding every absent-data case it existed to test (#203), `chipSeverityRank` returning a sentinel the caller skipped (#207), an assertion that could not fail (#212), `CitationEvalTest` executing nothing for three degraded cases (#215), and now this.

The shape is consistent and worth stating once more: in every case the check's **negative branch** was never considered — a filter, a sentinel, a guard clause, a substitution, a skipped expectation. §11 of the working brief exists for exactly this, and every one of the fifteen would have been caught at write time by it.

## Shape of the fix

Two parts, and the production half is the one that matters:

- **Accept a numeric string as a citation index** (or reject the response loudly). Silently dropping a well-formed field is the failure mode; either interpretation is defensible, silence is not.
- **Stop the test substituting `simulatedCitations` for the parsed value**, so it asserts what production returns. Then per §11, prove it by mutation: strip the inline markers and confirm the case reddens.

## Provenance

Measured by #218's pass, calling `extractResponse` directly rather than inferring from the test. I have not re-derived it.

Issue #229, opened 2026-08-13
The injected reference slice has no bound and no durable observable


dkayiwa:
## What

Nothing bounds the **number** of drug-reference records injected into the prompt, and there is no prompt-size guard anywhere on the path: `LocalLlmEngine` passes `-c <contextSize>` to llama-server and `LlmInferenceService` never checks the assembled length. The per-record cap `MAX_INTERACTION_RENDER_CHARS` is exactly that — per record — so N records cost N times it.

[#151](https://github.com/openmrs/openmrs-module-chartsearchai/issues/151) widened what the order-driven leg resolves, so the reachable worst case moved from "the ATC-mapped subset of the patient's orders" to "every substance the patient's orders resolve to that shares a family with the asked drug".

## Measured

Real `DrugReferenceInjector.injectRecords` over the shipped 19 MB `ddinter` KB and the demo patients' real order sets, 24 patients × 21 questions = 504 cells, run on a worktree of `main` and one of the #151 branch:

| | before | after |
|---|---|---|
| records | 464 | 491 (+5.8%) |
| chars | 460,982 | 488,668 (+6.0%) |
| cells changed | — | 19 of 504 |
| chip count changed | — | **0** |

Worst single cell: Sarah Taylor × prednisolone, **1 record / 903 chars → 6 records / 5183 chars** — roughly 4% of the 32,768-token default context.

So the widening is modest on today's data and the shape is bounded in practice by how many substances a patient's orders resolve to. What is unbounded is the **rule**: nothing in the code says a limit exists, and a polypharmacy chart on a KB with denser families is not covered by this measurement.

## The observability half, which is the worse half

The only signal that any of this happened is a **DEBUG** line from the injector. And on this instance that line is not reachable by configuration alone: the `log.level` global property is **not applied at startup** — setting it by SQL and restarting produced 0 DEBUG lines, while POSTing the same value to `/ws/rest/v1/systemsetting` applied it immediately and it was lost again on the next restart. (That is openmrs-core's behaviour, not this module's, but it is this module's only observable for prompt cost.)

So the prompt slice a clinician's answer is built from has no bound, no durable observable, and no counter on the status endpoint that #154/#217 built for exactly this kind of question.

## Shape of a fix

Not obviously a cap. Options, roughly in order of how well they fit what is already here:
[...]

PR #231, opened 2026-08-13
fix(observability): the audit row names the mode that built the prompt, and a scalar citations value is read

### LINKED ISSUE #178: The audit log never records queryScoped: searchMode is logged as pre-filter or full-chart only
## What

`ChartSearchAiRestController` logs the search mode at two sites (`:218` and `:460`) as either `pre-filter` or `full-chart`. There is no branch for **`queryScoped`**.

`queryScoped` is a real mode — ADR Decision 28 records it as the shipped default. So on a default install **every audit row says `full-chart`** while the prompt actually carried a query-scoped slice.

## Why it matters

This is the log a maintainer reads to reconstruct what a clinician was shown. It is the same failure class as #149 (a load that resolved 0 entries logged identically to a healthy one) and #156 (a fallback that loaded the wrong dataset with a healthy-looking count): the observable signal is *wrong*, not merely absent, so nobody has a reason to look further. Any past investigation that relied on those rows to establish which mode was in force was reading a constant.

It also silently invalidates comparisons: an A/B between chart modes cannot be reconstructed from the logs at all, because both arms record the same string.

## Fix

Log the mode actually resolved rather than deriving it from a two-way branch — ideally from the same value the pipeline uses to choose the slice, so the log cannot drift from the behaviour again.

## Provenance

Found during the documentation-truth sweep (#177) while checking whether the README and ADR statements about chart modes matched the code. Reported rather than fixed, since that PR is documentation-only. The `:218`/`:460` line references and the two-way branch are its reading; I have not re-run it.

### LINKED ISSUE #221: A non-array citations value is still skipped in silence: container typing, not entry typing
## What

#219/#220 fixed **entry** typing — `"citations": ["9","10"]` no longer silently yields nothing, and the coercion is reported at WARN. The **container** case is untouched: `LlmAnswerExtractor.extractResponse`'s `citationsNode.isArray()` guard means

```json
"citations": 8
"citations": null
```

are both discarded **in silence**, exactly as the string entries were.

## Why it was deliberately left

#220 identified it and did not fix it, on a reason worth preserving: the shape is **unmeasured**. Nobody has observed a model emitting a scalar or null there, and CLAUDE.md warns against adding a WARN that fires on shapes nobody has seen — a channel that cries wolf gets filtered, which is the argument that shaped the whole validity check in #217. So this is filed rather than patched in passing.

That reasoning also bounds it: the module now sends a strict `json_schema` with integer items, so the **server** enforces the array shape for the local engine. The exposure is a remote provider that ignores or loosens the schema, which is real but not currently observed.

## Shape of the fix

The one-line coercion is available and mirrors #220's: accept what has exactly one reading, report it, and never widen values — only types. Before taking it, decide whether the WARN should fire at all for a shape with no observed instance, or whether the honest form is to make the guard's silence *visible* (a debug line, or a finding on the load-status channel #217 built) rather than a warning.

Note the asymmetry that makes this less urgent than #219 was: a scalar or null `citations` loses the structured array, but the answer's inline `[n]` markers still resolve through `INLINE_CITATION`, so references are not necessarily lost — whereas `["9","10"]` lost them outright when the markers disagreed.

## Provenance

Found by #220's pass, reasoned from the `isArray()` guard rather than measured against a real response. I have not re-derived it.

dkayiwa:
Two defects of one class, which is why they are one pull request: **the observable signal is wrong or absent, so nobody has a reason to look further.** That is #149 (a load resolving 0 entries logged identically to a healthy one) and #156 (a fallback loading the wrong dataset with a healthy-looking count). #178 is the *wrong* half — a column that has held one value for its whole life while the thing it describes had three. #221 is the *absent* half — a whole field discarded without a word.

## #178 — the audit row never recorded `queryScoped`

`ChartSearchAiRestController` set the row's search mode at two sites, each branching on the **pre-filter** global property alone. `chartMode` never entered the expression, so `queryScoped` — the shipped default since #28 — could not appear in the column at all, and on a default install every row claimed `full-chart` over a query-scoped prompt.

**Measured on the standalone before the change: 6036 rows, every one `full-chart`.** No other value has ever been written. A maintainer reading these rows to reconstruct what a clinician was shown was reading a constant, and an A/B between the chart modes cannot be reconstructed from the log at all.

### The fix: the pipeline states the mode; the REST layer writes it and derives nothing

Deriving it a second time at the controller is what made it drift, and re-deriving it *correctly* today leaves the same defect available tomorrow. So the value now travels from where the dispatch happens:

- `QueryStoreChartBuilder.build` stamps the chart with the preFilter dispatch it took, taken from the **same boolean** that already labels its `[timing] mode=` line — the stamp and the log line cannot say different things. `buildScoped` already stamped `queryScoped`.
- `ChartBuildingStrategy.searchModeLabel` reads those two stamps and **nothing else**. Keyed on the chart rather than a fresh GP read for the reason `PatientChart.markQueryScoped` exists: a re-read can disagree with the read that built the chart, and an audit row's whole purpose is to say what the clinician was actually shown. It also means labelling costs no `Context` access — which is load-bearing, not incidental: a first attempt read the preFilter GP here and broke `LlmInferenceServiceTest`, which drives the pipeline with no OpenMRS session.
- `ChartAnswer` carries the label. Both audit-write sites read that one field, and the `searchMode` **parameter** of `streamAnswer`/`saveAuditLog` is gone — a parameter is the mechanism by which the two sites diverged.
- `DrugReferenceInjector` carries the new stamp across its rebuild, beside the two it already carries.

Three deliberate choices:

1. **`pre-filter` and `full-chart` keep their exact spellings.** These rows are read outside this module, so the fix ADDS a third value rather than re-spelling two. `SEARCH_MODE_QUERY_SCOPED` is defined *as* `CHART_MODE_QUERY_SCOPED`, so the row names the mode with the same token an operator sets.
2. **An answer that states no mode files as `unknown`,** never as one of the three real modes. The column is `NOT NULL`, so something must be written; a row claiming a mode nobody resolved is this defect, not a tidier version of it. Unreachable from the in-tree pipeline, which labels every answer it builds (`LlmInferenceService` is the only production site that constructs a `ChartAnswer`, at three call sites, all labelled) — it exists for an alternative `ChartSearchService`, and mirrors `QueryStoreChartBuilder.MODE_UNKNOWN`.
3. **The `[timing] querystoreBuild` log keeps spelling the same two shapes `preFilter`/`fullChart`.** That is a separate ops contract with its own greps; unifying it would be a second wire change nobody asked for. Both derive from one boolean in one place, so they cannot disagree about *which* mode — only about spelling. Noted in the constants' javadoc.

## #221 — a non-array `citations` container was skipped in silence

`extractResponse`'s `isArray()` guard discarded the whole field, exactly as `isInt()` discarded string entries before #219/#220. This is the container case, and the treatment is split by how many readings the shape has:

- **A scalar** (`"citations": 8`, or `"8"`) has exactly one reading — an array of one. Coerced, and reported at WARN, mirroring #220. It is coerced *through the same `citationIndex`* the entries use, so the container and its entries cannot come to admit different types.
- **An explicit `null`** is unambiguous and already means what the code does with it. **It stays silent at every level.** Warning there is the cry-wolf case CLAUDE.md forbids: it would fire on a provider that legitimately writes `null` for "none", and it is the same statement as an absent field.
- **Anything else** (an object, a non-numeric string, `9.7`, a boolean) has no single reading. Behaviour is unchanged — no citations — but the skip is logged at DEBUG so it stops being invisible. Not WARN: nothing recoverable was lost, and no instance has been observed, which is the noise #217 argued against.

Types widen; values do not. `LlmInferenceService.extractCitedReferences` remains the only thing deciding which indices become references, and `ChartSearchAiUtils.INLINE_CITATION` is untouched.

The regex salvage path is deliberately **not** extended to the container case: a truncated response whose `citations` is also a bare scalar is a shape inside a shape, neither observed, and the salvage path already WARNs on entry that the response did not parse.

## Tests

TDD, failing first, for the right reason.

- **#221** reddened at the assertion on the first run: scalar → `[]` (discarded), object → nothing logged at all. The `null` case was green before and after — it is a regression pin on behaviour that is already correct, not a red-then-green case, and it exists so a later pass cannot quietly make `null` loud.
- **#178** reddened in the omod suite with **all five cases reporting `full-chart`** regardless of what the pipeline had done — the issue's claim, reproduced as a test failure.

Every decision was then reversed on a throwaway tree to check the tests actually catch it. All ten mutations are caught; the table is in a comment below. One finding from that sweep is folded in: the race case that claims to rule out "the label re-reads the GP" did **not** catch that mutation, because only the service's mode seam was overridden and the strategy's fell through to the real reader, whose fail-safe default is `queryScoped` — so it passed for the wrong reason. Both seams now disagree with the chart, and the case fails under the mutation.

## Live verification

Deployed to the standalone and probed. The four regression controls are unchanged (Mary×clarithromycin 1 Major, Mary×paracetamol 0, Agnes×warfarin 1 Major, Joshua×ibuprofen 2).

**#178 — the rows, read out of `chartsearchai_audit_log`, not out of `/auditlog`:**

| config | plain `/search` | streaming `/search/stream` |
|---|---|---|
| `chartMode=fullChart`, `preFilter=false` | `full-chart` | `full-chart` |
| `chartMode=queryScoped` | **`queryScoped`** | **`queryScoped`** |
| `chartMode=fullChart`, `preFilter=true` | **`pre-filter`** | **`pre-filter`** |

All three values, on both paths, in rows 6052–6057. Had the fix been wrong, the middle row would read `full-chart` — the value all 6051 preceding rows carry. The `pre-filter` row is the canary for the new chart stamp specifically: its question triggers drug-reference injection, so it runs the injector's chart rebuild, and a rebuild that dropped the stamp would file it as `full-chart`.

**#221 — via a stub OpenAI-compatible engine, because the local engine's grammar cannot emit these shapes:**

| `citations` | references | log |
|---|---|---|
| `2` (scalar) | 2 | 1 × WARN naming the value |
| `[2]` (its twin) | 2 | nothing |
| `null` | 1 | **nothing, at any level** |
| `{"index": 2}` | 1 | 1 × DEBUG, no WARN |

The scalar and its one-element array twin produce identical reference sets, which is the equality the unit test pins, confirmed end to end.

The scalar row is the discriminator. I did not deploy a `main` build to measure it there; what it would do is read off `main`'s four lines — a non-array `citations` falls through the `isArray()` guard leaving the list empty and logging nothing, so only the inline `[1]` survives — and mutation **M7** confirms that parse empirically by restoring the guard on this build and reddening four tests, one of which is the scalar/array reference-set equality.

Global properties were restored and re-verified **in the database**, not just through REST: `chartMode=fullChart`, `preFilter=false`, `llm.engine=local`, `remote.endpointUrl`/`remote.modelName` NULL.

## Documentation

`docs/adr.md` stated the audit log records "the search mode used (`pre-filter` or `full-chart`)" — false after this change, and arguably false before it, since that sentence is what a reader would have trusted instead of checking. It now names all three and says where the value comes from.

---

Fixes #178
Fixes #221


dkayiwa:
## Mutation sweep — every decision in this PR reversed, on a throwaway tree

Section 11 of the working brief: a probe offered as evidence *for* a change must name the mutation it would catch, and that sensitivity has to be demonstrated rather than asserted. Each row reverses one decision and records which tests redden. **No survivors.**

| # | decision reversed | caught by |
|---|---|---|
| M1 | controller re-derives the mode instead of reading the answer's | `ChartSearchAiAuditSearchModeTest` ×3 |
| M2 | `QueryStoreChartBuilder.build` drops the preFilter stamp | `QueryStoreChartBuilderTest` ×2 |
| M3 | `DrugReferenceInjector` drops the stamp on its rebuild | `DrugReferenceInjectorTest.injectionPreservesThePreFilterStamp` |
| M4 | `searchModeLabel` consults the chartMode GP instead of the chart | `LlmInferenceServiceQueryScopedTest` ×3 |
| M5 | the ungrounded `ChartAnswer` is left unlabelled | `searchStreaming_shouldLabelTheUngroundedAndFinalAnswersIdentically` |
| M6 | an unlabelled answer defaults to `full-chart` instead of `unknown` | `chartAnswer_shouldSayUnknownRatherThanGuess_whenNoModeWasStated` |
| M7 | the `isArray()` guard is restored (the #221 defect) | `LlmProviderTest` ×4 |
| M8 | an explicit `null` stops being silent | `extractResponse_shouldStayQuietOnAnExplicitNullCitations` |
| M9 | an unreadable container is skipped silently | `extractResponse_shouldLeaveACitationsContainerWithNoSingleReadingAloneButNotInSilence` |
| M10 | the scalar coercion stops reporting at WARN | `extractResponse_shouldReportAScalarCitationsValueAtWarn` |

M4 is the one that repaid the exercise. On the first sweep it was caught by the two *fullChart* cases but **not** by the race case whose comment claims to rule exactly that out — because only the service's mode seam was overridden, leaving the strategy's falling through to the real GP reader, whose fail-safe default is `queryScoped`. The case passed for the wrong reason. Fixed in `52f1787`: both seams now disagree with the chart, and the case fails under the mutation.

## Test counts, measured at `52f1787` on a clean worktree

`main` (`08ee9c00`): **1090 api + 62 omod = 1152**, 0 failures, 53 skipped.
This branch: **1107 api + 67 omod = 1174**, 0 failures, 53 skipped. 22 new test methods.

## What the live pass would have shown had the fix been wrong

- **#178.** The middle row of the table in the description is the discriminator: with `chartMode=queryScoped`, `main` writes `full-chart` — the value all 6051 preceding rows carry. The `pre-filter` row is the canary for the new chart stamp: its question triggers drug-reference injection, so it exercises the injector's chart rebuild; a rebuild that dropped the stamp files it as `full-chart` and the row is indistinguishable from B1's.
- **#221.** The scalar arm resolves 2 references against the one-element array's 2; on `main` it resolves 1 and logs nothing. The `null` and object arms are the two directions the *ruling* turns on, and both were checked: `null` produced no log line at any level (the logger was raised to DEBUG for that check), and the object produced a DEBUG line and no WARN.

One process note worth recording. The first live pass reported six audit rows all reading `full-chart` and looked like a clean negative result. It was vacuous: its `gp_get` helper was copied from #219's `L219-probes.sh`, where `curl … | python3 - <<'PY'` gives python its *program* on stdin and discards the piped JSON — so every read returned empty, every write POSTed to an empty uuid and answered HTTP 400, and the pass probed the unchanged configuration while printing as though it had flipped it. #219's own pass had already hit this and fixed it in a separate `L219-blockD.sh`; the broken copy is the one still sitting in the shared scratchpad, which is what a later agent reaches for. The re-run asserts a non-empty uuid before every write and re-reads each property from the server *and* the database afterwards.

dkayiwa:
## Hardening — both phases, to their stopping rules

**Phase 1: 5 passes.** Stopping condition, verbatim: *"the last two consecutive passes (4 and 5) returned only cosmetic items."*

What it found and fixed, in order:

1. `searchModeLabel` had a null branch answering `unknown` for a chart the caller dereferences on the next line either way — a contract the code does not have. Removed; the precondition is stated instead.
2. `citationIndex`'s javadoc scoped itself to an "entry" while #221 also routes the container through it.
3. `ChartBuildingStrategyDispatchTest` — the class that exists because "a dropped dispatch would pass the entire suite and only surface in live captures" — did not cover the same exposure for the label. Its stub builder now stamps its charts and a third case asserts `buildChart` hands the stamps back untouched.
4. `markPreFilter`'s javadoc implied the null-patient chart labels `unknown`; it labels `full-chart`, and the reason that is unreachable from a row is now stated rather than implied.
5. **From the mutation sweep:** the race case naming "the label re-reads the GP" did not catch that mutation. Only the service's mode seam was overridden, so the strategy's fell through to the real reader, whose fail-safe default is `queryScoped` — the case passed for the wrong reason. Both seams now disagree with the chart.

An earlier attempt put the label behind a `Context`-reading global-property lookup and broke `LlmInferenceServiceTest`, which drives the pipeline with no OpenMRS session. That is why the label reads the chart's stamps and nothing else — the design is a response to a measured failure, not a preference.

**Phase 2: 2 passes**, four reviewers on the first (reuse, quality, efficiency, integration), two on the second briefed with the applied and deferred lists.

Two findings were structural enough to be Phase 1 work and were treated as such:

- **Nothing pinned the wire spellings.** Every assertion here compares a constant to a constant, which cannot notice a *rename* — and `SEARCH_MODE_QUERY_SCOPED` is defined *as* the `chartMode` GP token, so renaming that token would silently rewrite every subsequent row for the external readers the javadoc invokes, with a green suite. Now pinned as literals, with a comment saying why that case is allowed to be brittle.
- **`asyncGroundingSearch_recordsTheModeOnTheUngroundedAnswerItAudits` could not fail for the reason its name gives** — the stub labelled both answers identically, so it passed whichever object the controller read. The stub now carries two labels and a mirror case pins the classic shape reading the other one.

Polish applied: the OpenMRS context fixture was extracted to a shared `RestControllerContext` (it had been copied verbatim from `ChartSearchAiSearchResponseGroupingTest`); three unnecessary interface-`default` overrides dropped from the stub, which puts the 7-arg delegation the controller actually calls back on the path under test; `extractResponse` restructured so the "an explicit null says nothing" rule is a stated decision rather than a clause inside a compound condition, one nesting level shallower than `main`; four javadoc corrections, including a class comment on `ChartSearchService` that described a default-on pre-filter narrowing the chart — machinery removed in #51, and false about the default well before that — 290 lines above a new getter saying the opposite.

Efficiency reported nothing actionable and confirmed the change *removes* one global-property read per request, net.

**Deferred, with its failure mode stated:** a `PatientChart.copyStampsFrom(other)` helper. If we ship without it, a future fourth stamp added to `PatientChart` can be forgotten in `DrugReferenceInjector.injectRecords` and the rebuild silently fails open — exactly how the `queryScoped` stamp behaved before it was carried. Not done here because it is a new abstraction with one call site, and the three `if`s it would absorb each carry a distinct rationale (KV mis-scoping, audit mislabelling, absence-meaning). Three regression tests cover today's three stamps.

All ten mutations were re-run against the restructured code at the final SHA and are still caught.

dkayiwa:
## Phase 2, pass 2 — and one finding worth reading even if you skim the rest

Two reviewers, briefed with the applied and deferred lists so they would not re-surface pass 1. Both independently landed on the same thing, and it is the most interesting finding of the whole exercise.

### `RestControllerContext.restore()` could not do what its javadoc promised

Verified against `openmrs-api-2.9.0-SNAPSHOT` bytecode, not inferred: `ServiceContext.setService(cls, instance)` returns immediately when either argument is null, and there is no removal API. `currentService()` returns null on a bare unit-test JVM. So restoring a "prior" of null was a **silent no-op**, and this fixture's stub `PatientService`/`AdministrationService` stayed installed for the rest of the surefire run.

Why that is not cosmetic, and why it belongs to *this* PR specifically: three classes in this package drive `streamAnswer` with **no context installed at all**, and that is the only thing enforcing the "free of `Context` reads" contract the removed `searchMode` parameter depended on — a re-added global-property read throws because nothing is there to answer it. A leaked stub answers instead of throwing (2-arg → the caller's own default), so those three classes would stay green over **exactly the drift #178 removed**. Surefire here has no configured `runOrder`, so which classes are affected is not even deterministic.

`restore()` now installs a stand-in whose every method throws, which is what an absent service behaves like. The suite is green, so nothing was depending on the leak.

This is pre-existing in kind — `main`'s `ChartSearchAiSearchResponseGroupingTest` had the same inline code — but this PR doubled the number of installers and promoted the false claim into a shared file where it reads as verified. Worth fixing here rather than filing.

### The two audit-row builders were byte-identical, and this PR is what made them so

Before the change they differed in exactly one expression: how each derived `searchMode`. That expression *is* issue #178. Removing it left the same eleven setters written twice — and two sites that must agree about a row's contents, kept in step only by hand, is the structural condition that let `search_mode` hold one value for 6036 rows. The next column added to this table would land in one of them and be silently absent from the other, exactly as `cachedTokens` is absent from both today. The blocking handler now calls the same `saveAuditLog` the two streaming paths use, so there is one place a row is built.

### Two claims that were false rather than merely clumsy

- *"logged at DEBUG so the skip stops being invisible"* — the default `org.openmrs.*` level is WARN, a premise this module records in `QueryStoreChartBuilder` and which my own live pass had to override before the line appeared at all. DEBUG makes the skip **explainable to an operator who goes looking**, not announced. The level itself is unchanged, per the ruling on this issue; the sentence now says what the level actually buys, and names the tension with `reportNonConformantCitations` (which WARNs on the same information loss from the same cause, the difference being that the entry case had an observed instance and the container case does not) so it can be revisited deliberately rather than discovered.
- The fixture javadoc claimed the audit test depends on its two-arg global-property answer. After the fix the controller has exactly one `getGlobalProperty` left on these paths — the one-arg rate-limit read — and every mode the audit test asserts comes from the stub answer. The claim described the world #178 removed.

### Deferred, with its failure mode

Three sibling test classes keep private copies of `patient()`/`user()` now that `RestControllerContext` owns them. **If we ship without unifying them, nothing observable breaks** — those classes pass their patient straight into `streamAnswer` rather than resolving it through the shared `PatientService` stub, so a divergence between their copy and the fixture's has no consequence. Real duplication, inert.

Test counts unchanged at 1107 api + 69 omod. The full ten-mutation sweep was re-run against the restructured code and all ten are still caught.

dkayiwa:
## Re-verified live at the final SHA, after the audit-row unification

Folding the blocking handler into the shared `saveAuditLog` changes what a real `/search` request writes to a real table, so the live evidence was re-gathered on a build of `5a5be4a9` rather than argued forward from the earlier pass.

Four regression controls unchanged: Mary×clarithromycin 1 Major, Mary×paracetamol 0, Agnes×warfarin 1 Major, Joshua×ibuprofen 2.

Rows read out of `chartsearchai_audit_log` — 6083–6088, both request paths, all three modes:

| config | plain `/search` | streaming `/search/stream` |
|---|---|---|
| `chartMode=fullChart`, `preFilter=false` | `full-chart` | `full-chart` |
| `chartMode=queryScoped` | **`queryScoped`** | **`queryScoped`** |
| `chartMode=fullChart`, `preFilter=true` | **`pre-filter`** | **`pre-filter`** |

Global properties restored and re-read from the database, not just through REST: `chartMode=fullChart`, `preFilter=false`, `llm.engine=local`, both remote GPs NULL.

The ten-mutation sweep was re-run at this SHA. No survivors, and one result is worth noting: **M1 (the controller re-derives the mode instead of reading the answer's) is now caught by five cases rather than three.** That is the unification showing up as coverage — with one row builder, a mutation there reaches every write site, where before it could only reach the copy it was applied to.

## State of the instance

Left running on this branch's build, warm, healthy: `sourceFormat=ddinter`, `entryCount=2283`, one pre-existing validity finding (`alias-names-another-substance / reported / 18`). `log.level` is `org.openmrs.module.chartsearchai:debug` — set deliberately for the #221 DEBUG-channel check and left in place; it does not survive a restart.

The audit table now carries three distinct values for the first time in its life. It held exactly one, `full-chart`, across all 6036 rows written before this branch.

dkayiwa:
## `d3f899c6` — the #221 container branch is WARN, and the rule is recorded

Agreed with the reversal, and your third reason is the one I did not have: shipping a DEBUG-only observable in the same session as #229 would have been incoherent regardless of the merits.

**The level.** Both outcomes of `readNonArrayCitations` now WARN, sharing a message prefix so one grep finds every non-array container whichever way it resolved. Value behaviour is untouched — no reading invented, nothing widened, `extractCitedReferences` still the only thing deciding which indices become references.

**The distinction, written down where it can be acted on.** The javadoc now says why `null` stays silent while everything else here does not, and states it as what the container *asserts* rather than what shape it is — because "not an array" is exactly the wrong axis and is how someone folds the two branches together:

> `null` asserts ABSENCE — the same statement as an omitted field, from a provider whose answer cites nothing — so nothing is lost and nothing is guessed. Everything that reaches `readNonArrayCitations` asserts PRESENCE of something unusable. Absence honoured is not a defect; presence discarded is.

The `null` case now also pins the **omitted-field twin** in the same test, so the equivalence the argument rests on cannot drift.

**One check beyond the build, because I have earned the suspicion.** Two of my own new tests were blind today, so I re-anchored the level mutations and ran them rather than trusting a green suite: reverting the unreadable branch to DEBUG reddens `extractResponse_shouldReportACitationsContainerWithNoSingleReadingAtWarn`, and silencing the scalar coercion reddens `extractResponse_shouldReportAScalarCitationsValueAtWarn`. Neither assertion is decorative. Sweep tree clean afterwards.

**`CLAUDE.md`** — the two authorized entries, nothing else in the file:
- **Which chart-assembly mode produced an answer** — `searchModeLabel(chart)` / `getSearchMode()`, carrying the measurement (6036 rows, every one `full-chart`, no other value ever written), that the label reads `PatientChart`'s own stamps, and the two consequences that bind additions: every stamp must survive the injector's rebuild (fail-open, twice now), and `SEARCH_MODE_QUERY_SCOPED` being defined *as* the GP token means renaming that token rewrites the column for outside readers — which is what the literal-spelling test exists to catch.
- **The `citationIndex` amendment**, plus a sibling bullet for the reporting rule, so the WARN/silent split is recorded as a decision rather than rediscovered as an inconsistency.

**Full reactor at `d3f899c6`: 1107 api + 69 omod = 1176, 0 failures, 53 skipped** — unchanged from `5a5be4a9`, as expected for a level change with an already-existing test. Working tree clean, in sync with origin.

PR #239, opened 2026-08-13
fix(wire): a reference-group citation carries no grounding verdict at all

### LINKED ISSUE #201: The reference frontend classifies citations by resourceType, not by group, so a safety finding renders as "Unsupported"
## What

`openmrs-esm-chartsearchai` — the reference frontend, at `3003cd2`, which is what `Dockerfile.frontend:7` builds — **never reads the `group` field.** It is not even declared on the `AiReference` type (`src/api/chartsearchai.ts:13-27`). Instead it keys its neutral "Reference" badge on `resourceType === 'drug_reference'` (`ai-response-panel.component.tsx:23,260`).

Consequence: a **`safety_finding`** falls through to the grounding-badge branch. When `grounded === false` — which is **reachable by design**, since the off-topic flag is deliberately retained — the UI renders

```
Unsupported — The cited record may not support this statement
```

in red, **on this module's own deterministic Major-interaction finding**. An identically-grouped `drug_reference` renders the purple "Reference" badge from the same data.

## Why this is ours to raise even though the code is not

Two reasons.

1. **This module's README already documents the contract being broken.** `README:399` warns explicitly that a client "must not treat those types differently from each other". The frontend does exactly that, so either the contract needs enforcing or the README is describing a guarantee nobody honours.
2. **#122/#137 fixed this exact confusion on the backend and the fix stops at the wire.** The whole point of deriving the demote-only carve-out from the reference *group* rather than an enumerated type list was that a new reference type gets the right treatment automatically. That property does not survive a consumer that re-enumerates types itself — and `safety_finding` is precisely the type #122 was filed about.

Worth noting what changed for the worse: **before #137 those verdicts were shown as a green "Verified" tag rather than hidden.** So the backend fix moved the failure from "wrongly reassuring" to "wrongly alarming" in this client, which is arguably safer but is still wrong, and is now visible to clinicians.

## What would settle it

Either the frontend reads `group` (the field exists and is stable, and was added for this purpose), or the wire contract stops offering a `grounded` value for reference-group citations at all so no client can misread it. The second is the stronger fix — a field that must not be interpreted is a trap — but it is a wire change and needs its own decision.

Filed here rather than in the frontend repo because the contract, the README guarantee and the backend half of the fix all live here. **No change has been made in the other repository.**

## Provenance

Measured by the #137 retrospective review against the cloned frontend at `3003cd2`, by reading the source; the file and line references are its measurements. I have not run the frontend.

dkayiwa:
Fixes #201.

## What was wrong

Grounding is **demote-only** for module-supplied reference material (`drug_reference`,
`safety_finding`): a pass is withheld as `null`, because a recitation of module-rendered prose
embeds near-identically to its source even when it swaps subject roles (#106). A Tier-1 *off-topic*
citation still yields `false`, and that `false` was published on the wire.

It is well-defined only in terms of `group`. On a reference citation it means *"this citation is not
about that record"*; on a chart citation `false` means *"this claim may not be supported"*. Telling
those apart requires reading `group` — and the reference frontend classifies citations by
`resourceType` (measured on #201 against `openmrs-esm-chartsearchai` at `3003cd2`, which does not
declare `group` on its reference type at all). So a `safety_finding` fell through to its grounding
branch and rendered

> **Unsupported — The cited record may not support this statement**

in red, on this module's own deterministic Major-interaction finding.

## What this does

The issue offered two settlements — the client reads `group`, or the wire stops offering a value that
must not be read. This takes the second, per the repo owner: **a field that must not be interpreted is
a trap**, and withholding it holds for every client rather than for the one that is patched.

A `reference`-group citation now serializes `grounded: null` — never `true`, never `false` — at
**every** emission site. There are four and they were verified to share `serializeReferences` rather
than assumed to: the blocking `/search` handler, `sendReferencesEvent` (the early `references` SSE
event), `doneEventJson` (`done`, on both the classic and the early-async path), and the trailing
async `grounded` event in `streamAnswer`. Each is asserted separately anyway, because "they share
it" is the thing a client is entitled to have checked. Corroborated by reading rather than trusting
the javadoc: one `refMap` builder exists, one class in the repo carries `@RequestMapping`,
`/auditlog` publishes `referenceCount` only, no Jackson call site serializes a `ChartAnswer` or
`RecordReference` reflectively, and the module ships no web assets of its own.

`null` rather than an omitted key: `null` is already this field's documented value for "grounding
disabled or could not run", the serializer's javadoc already tells clients to render it as
*unverified, never as verified*, and the key's unconditional presence is a property clients rely on.
Omitting would invent a third state and break more for no gain.

Derived through `ChartSearchAiUtils.isGroundingDemoteOnly`, never from a list of type names — the
enumerated form is how `safety_finding` was graded as chart evidence for two releases (#122).

A side effect worth having: the early `references` event and the final `done`/`grounded` events now
**agree** for reference-group citations. Before, a reference citation could arrive `null` on the
early event and `false` on the final one, so a client that renders the early event and replaces it
wholesale saw the badge change.

## What the wire no longer carries, and who read it

The off-topic signal for reference-group citations, **on that channel only**. The grounding pass is
untouched: `CitationGroundingVerifier` still computes the verdict and
`RecordReference.getGrounded()` still carries it, so everything inside the module still sees it, and
`CitationGroundingVerifierTest.safetyFinding_offTopicCitationIsStillFlagged` /
`drugReference_offTopicCitationIsStillFlagged` still pin it.

Two things I checked rather than assumed, because I first wrote both of them down wrongly:

- **It is not logged.** The verifier logs cap-reached counts, embedder unavailability and batch
  failures; `[timing]` carries `groundMs`. No per-citation verdict is logged anywhere. So the wire
  was the only place this value was externally visible.
- **Nothing else consumes it.** The Tier-2 exclusion and the `TRUE`→`null` demotion are both driven
  by `demoteOnlyIndexes` in `CitationGroundingVerifier.verify` — i.e. by `isGroundingDemoteOnly` on
  the resource type — not by the verdict. So
  for a reference-group citation the surviving `false` now feeds only those two tests.

That is the honest size of the loss: a value that was reachable, wrong to read, read wrongly by the
only known consumer, and read by nothing else.

**No consumer read it correctly.** The reference frontend is the only known consumer and misreads it,
which is the issue. The only other reader in this repository is
`eval/grounding-scope/grounding_scope_ab.py:76`, a developer A/B harness for clause-scoping: it
prints every citation's verdict, but its two tags (`REGRESSION`, `win`) each require a `true` on one
side, and a demote-only citation can never be `true` — so its gate is structurally insensitive to
the value and is unaffected. It is left as is.

## Live verification

Deployed both builds to the standalone (`sourceFormat=ddinter`, `entryCount=2283`,
`chartMode=fullChart`, `grounding.enabled=true`) and probed each. Full logs:
`verify-i201before.log` / `verify-i201after.log`.

Phase A is the instance as found. Phase B sets `minCosine=0.99` with `entailment.enabled=false`, so
Tier-1 fails **every** citation — which produces the demoted reference verdict deterministically,
with no LLM judge in the loop, instead of hoping the model writes an off-topic sentence. GPs are read
back after every write and restored at the end.

Every reference is printed as `(index, resourceType, group, grounded)`.

| probe | chips | BEFORE (`07097ba2`) | AFTER (`2ea5aa08`) |
|---|---|---|---|
| A1 Mary × clarithromycin | 1 → 1 | `[1]drug_order/chart=False` `[76]drug_reference/reference=None` `[77]safety_finding/reference=None` | identical |
| A2 Agnes × warfarin | 1 → 1 | `[112]drug_order/chart=False` `[114]safety_finding/reference=None` `[113]drug_reference/reference=None` | identical |
| A3 Joshua × ibuprofen | 2 → 2 | `[41]safety_finding/reference=None` `[42]safety_finding/reference=None` | identical |
| A4 Mary × paracetamol | 0 → 0 | no citations | identical |
| **B1 Mary × clarithromycin** | 1 → 1 | **`[77]safety_finding/reference=False`** | **`[77]…=None`** |
| **B2 Joshua × ibuprofen** | 2 → 2 | **`[41]safety_finding/reference=False` `[42]…=False`** | **both `None`** |
| **B3 Mary "what conditions…"** | 0 → 0 | 9 chart citations, **all `False`** | **all 9 still `False`** |
| **B4 Agnes × warfarin** | 1 → 1 | `[112]drug_order/chart=False` **`[114]safety_finding=False` `[113]drug_reference=False`** | `[112]` **still `False`**; `[114]`/`[113]` → **`None`** |

| | BEFORE | AFTER |
|---|---|---|
| reference-group citations carrying a verdict | **5** | **0** |
| chart-group citations carrying a verdict | **12** | **12** |

B1, B2 and B4 are the defect reproduced live, on the module's own Major-interaction findings, and
then gone. **B3 and the `[112]` cell of B4 are the canaries for the opposite mistake**: if the
withholding had been keyed on anything broader than the group, those chart `false` verdicts would
have become `null` and the grounding feature would have been silently blanked rather than narrowed.
They are the live counterpart of the blanket-null mutation in the table below, which 8 tests catch.

Chip counts are unchanged on all four regression controls (1 / 1 / 2 / 0), and the KB loaded
identically in both runs (`loaded=true`, `inert=false`, `entryCount=2283`, `sourceFormat=ddinter`,
one pre-existing `alias-names-another-substance` finding ×18). GPs were read back after every write
and restored to exactly the values found (`minCosine=0.40`, `entailment.enabled=true`).

The AFTER build is `2ea5aa08`. Everything committed after it is comments, javadoc, markdown and test
code, with exactly one exception: the `chartsearchai.grounding.enabled` **description** string in
`config.xml`, which is operator-facing text with no behavioural effect. So this evidence stands for
the branch head — the behavioural change is the same one-line ternary it was in the first commit.

## Tests

TDD, red first. Before the fix, four of the six new tests failed on
`citation [15] is a "drug_reference" record … expected: <null> but was: <true>`.

### How the test resists a hardcoded two-type list

CLAUDE.md records that re-hardcoding this carve-out to today's two type names *survives the whole
suite* and only fails once a fourth injected type exists. So the withholding is pinned twice:

1. **Behaviourally, off an enumeration rather than a literal.** The fixture cites all 10 declared
   `RESOURCE_TYPE_*` constants twice — once at `TRUE`, once at `FALSE` — 20 citations, of which 4
   must be withheld and 16 must pass through, and the expectation for each is derived by asking
   `isGroundingDemoteOnly`. No type name appears in the expectation, so a new reference-group
   constant is swept automatically.
2. **Structurally, so the hardcode fails today.**
   `theWireSerializerMustNotNameAReferenceGroupResourceType` reads every class file the controller
   compiles to — the outer class and each `ChartSearchAiRestController$*` — and asserts none names a
   reference-group type. `RESOURCE_TYPE_*` are compile-time `String` constants, so javac inlines the
   value into the constant pool of any class that mentions one, whether written as the constant, a
   bare literal, or a folded concatenation. Positive controls (`grounded`, `withheldInteractions`
   must be present) stop a failed or truncated read passing vacuously.

Both halves have a stated limit rather than an implied one, in the class javadoc: the behavioural
sweep asks the same classifier production asks, so a regression *in `referenceGroup` itself* would
leave every test here green — the load-bearing anchor to real type names is
`ChartSearchAiReferenceGroupTest`'s hand-recorded table in the api module. And the byte scan is a
raw substring match, safe for today's two values but wanting word delimiters if a future wire value
is a substring of an unrelated literal.

Measured, not asserted — three mutations on a throwaway worktree:

| mutation | caught by |
|---|---|
| `groundedForWire` → `ref.getGrounded()` (no withholding) | the 4 behavioural site tests (this is the TDD red) |
| `groundedForWire` → hardcoded `"drug_reference".equals(t) \|\| "safety_finding".equals(t)` | **only** `theWireSerializerMustNotNameAReferenceGroupResourceType`; the behavioural half passes, exactly as CLAUDE.md predicts |
| the same hardcode moved into the nested `PatientResolution` class | the same guard, after it was widened to scan every `ChartSearchAiRestController$*.class` — it did **not** catch this before the review pass |
| `groundedForWire` → `return null` (blanket) | 5 of the 6 new tests **plus** 3 existing `ChartSearchAiStreamEventOrderTest` cases — the proof this narrowed rather than blanked |

## Docs

The reference contract in `README` (`:399` as #201 cites it) already required a client to "not treat
those types differently from each other" — a guarantee nobody honoured, which is half of what the
issue is about. It now states the guarantee the code actually provides, and the ADR's demote-only
sub-decision records what was given up and why
`null` beats omission (its heading also stops saying *drug-reference*, since the rule has been keyed
on the group since #122 and that framing is the mistake this issue is about).

Five statements next to the change were about the old contract, and in this codebase that prose is
load-bearing:

- `ChartSearchAiConstants.REFERENCE_GROUP_REFERENCE` said the **exact opposite** — "a client must
  still read `grounded` per reference rather than infer it from `group`".
- `DrugReferenceInjector` and `ChartSearchAiReferenceGroupTest` are the two places that tell an
  author what recording a type as reference material *costs*. Both listed two consequences; there is
  now a third, and it is the one a clinician sees.
- `CitationGroundingVerifierTest.safetyFinding_offTopicCitationIsStillFlagged`'s rationale argued
  the flag is "real, observable and worth flagging". That was the recorded reason for grading these
  citations at all, and this change removed its only external observability — so it now says what
  the flag is still there for, or the next reader concludes the premise is obsolete and takes the
  "exempt entirely" option #122 declined.
- `ChartSearchAiReferenceGroupingTest` called `grounded` a passthrough and attributed a
  `drug_reference`'s null to the verifier. The null now comes from the serializer.

The `chartsearchai.grounding.enabled` descriptions (README's GP table and `config.xml`) said "every
cited record is checked … citations
that fail are flagged as unverified", which now over-promises for a drug-safety answer whose
citations are all reference-group.

`eval/grounding-scope/grounding_scope_ab.py` prints per-citation verdicts, and a reference-group cell
would now read `None` — indistinguishable from "checked and unverified" — so it reads `group` and
prints `withheld` instead, and its docstring says the gate is a statement about chart citations. Its
tallies are unchanged (see above: they cannot be influenced).

**Test files are touched in comments only.** `git diff origin/main...HEAD -- '*/src/test/*'` has no
added or removed line outside `ChartSearchAiReferenceGroundingWithholdingTest` that is not a comment;
no assertion, expected value or fixture datum changed anywhere.


dkayiwa:
### Hardening summary

Four Phase-1 review passes (two of them independent) and one Phase-2 pass. The production change
never moved after the first commit — `groundedForWire` is the same one-line ternary — and the new
test's *mechanism* moved once. Everything else was prose that this contract change falsified, which
in this codebase is a substantive finding.

**Two blind checks were found in my own new test**, both by an independent pass, and both of the
shape where the guard reports success without checking:

1. `while ((read = in.read(chunk)) > 0)` — `InputStream.read(byte[])` may legally return `0`, which
   ends the loop early, and a truncated buffer satisfies every `assertFalse(contains(...))`. A
   passing guard over bytes it never read. Replaced with `Files.readAllBytes`.
2. It read only `ChartSearchAiRestController.class` while its javadoc claimed the class names a type
   "if and only if" those bytes contain it. The controller compiles to **two** files
   (`ChartSearchAiRestController$PatientResolution.class`), so a hardcode in a nested class was
   invisible. Confirmed by mutation: moving the hardcoded pair into `PatientResolution` passed the
   old guard and fails the widened one, naming the file.

**Prose corrected across five neighbours**, plus two rounds of errors in my own corrections —
first a claim that the withheld verdict "survives in the grounding logs" (nothing logs a
per-citation verdict), then a claim that it is "what keeps these citations out of Tier-2" (the
classification does that, not the verdict). Both were deleted rather than reworded a third time.
The last pass caught one more: the ADR's copy of "clients must render `null` as unverified" had
composed with the new always-null rule into an instruction to render every reference citation as
unverified — the exact reading this issue exists to kill, and the one README had already been scoped
to avoid.

**Phase 1 did not converge, and I am not claiming it did.** Pass 4 found six substantive prose items,
all applied; a fifth pass would be needed to observe the "subsequent pass finds nothing substantive"
condition, and I did not run one. Code and tests have been stable since pass 1. Per the standing
guidance on diverging prose loops, the remedy applied was deletion rather than further rewording.

Build at `0b545ccc`: **1126 api + 75 omod = 1201**, 0 failures, 53 skipped (`main` at `07097ba2`:
1126 + 69 = 1195). Live before/after evidence and the mutation table are in the PR description.

Issue #168, opened 2026-08-06
Within the systemic tier, sharedClass still tie-breaks alphabetically, naming a defensible but less apt class


dkayiwa:
## What

After #161/#166, `sharedClass` prefers a subgroup that classifies the substance over one that classifies a locally applied formulation. When a pair shares **more than one systemic** subgroup, the tie-break is still alphabetical.

Measured over the shipped KB: **87 pairs** are in this position. Example: Ketoconazole × Levoketoconazole is named `H02CA` (anti-adrenal preparations) where `J02AB` (systemic antimycotics) is the apter description of what those two drugs are.

## Severity

Low, and deliberately left. Both codes are **true** of the pair — this is a "less apt" not a "wrong" answer, unlike #161 where the named class was actively misleading. It is filed because:

- The fix for #161 established that the class string is read as an explanation, so aptness matters.
- 87 pairs is small but not negligible, and the tie-break being alphabetical is an accident rather than a decision.

## Why it was not fixed with #161

There is no data-defensible ranking available within the systemic tier: nothing in the ATC data says `J02AB` explains this pair better than `H02CA`. A fix needs an external notion of which class is the drug's *primary* therapeutic role, which the KB does not carry. Inventing a heuristic here would be the kind of unmeasured preference #161's fix deliberately avoided.

## Provenance

Measured over the shipped KB during the #161 fix work and reported to me as a hazard rather than a defect; I have not re-run the count.

dkayiwa:
**Count correction: 70, not 87.** Re-measured over the shipped KB after #166's hardening added four locally-applied groups (`A07A`, `B02BC`, `B05C`, `G02CC`) and the `D02BB` exception. Seventeen of the pairs I originally counted were resolved by those additions rather than by any tie-break, so the residual is 70. The Ketoconazole × Levoketoconazole example still stands.

dkayiwa:
**Closing as measured-and-declined, not fixed.** The evidence, from #261's pass, driving production predicates over the shipped KB:

**The count re-derives on a stated base as 20 substance pairs** for the duplicate-therapy arm, with the 16 cross-reactivity ties a measured subset of them. The "~87 pairs" in this issue was over a base nobody had stated — which was #243, now closed and recorded on `sharedClass` itself.

**No data-defensible ranking exists, and this was tested rather than asserted:**

[...]

PR #241, opened 2026-08-13
fix(drug-safety): an ATC subgroup licenses only the claim its own name asserts (#183, #184)

### LINKED ISSUE #183: Broad therapeutic ATC buckets still license cross-reactivity claims across heterogeneous chemistry
## What

#167/#182 stopped **residual** ATC subgroups ("Other …", "Various", and everything under `V03A`) from licensing a clinical relationship claim. It deliberately did **not** touch the broad *therapeutic* buckets, which remain a live source of unfounded claims.

Measured over the shipped KB:

| subgroup | name | KB pairs it names |
|---|---|---|
| `S01AA` | Antibiotics (ophthalmic) | **243** |
| `A07AA` | Antibiotics (intestinal) | 102 |
| `S02AA` | Antiinfectives (otological) | 81 |

These group by *therapeutic purpose over heterogeneous chemistry*, so a claim like "chloramphenicol cross-reacts with gentamicin because both are ophthalmic antibiotics" is still reachable. Cross-reactivity is a property of chemical class, not of what a drug is used for or where it is applied.

## Why it was correctly left out of #182

#182's rule is a **reading of ATC's own words**: a subgroup named "Other X" asserts nothing about the substances beyond its parent's assertion, so a residue under an administration-site parent (`A01` STOMATOLOGICAL PREPARATIONS) says only "both are put in the mouth", while a residue under `R06A` ANTIHISTAMINES FOR SYSTEMIC USE genuinely does say both are antihistamines. That criterion is mechanical and checkable.

`S01AA` is different: the name *does* assert something. Excluding it is a **pharmacological judgement** — that "both are antibiotics" is too weak a basis for a cross-reactivity warning — not a reading of the classification. That judgement needs making deliberately rather than smuggling in behind a name-pattern rule, which is why #182 scoped it out and said so.

## What would settle it

A decision on the weakest relationship that may justify each claim type, which may differ by arm:
- **cross-reactivity** plausibly needs chemical-class kinship (the curated cross-reactivity groups already encode this properly — the NSAID group is the model);
- **duplicate therapy** plausibly *is* about therapeutic purpose, so `S01AA` may be legitimate there and not in the contraindication arm.

Note the curated groups already exist and already express chemical kinship correctly. The question may be less "which ATC buckets to exclude" than "should the ATC path license cross-reactivity at all, or only the curated groups".

Adjacent: #168 (aptness of the class named within the systemic tier, 70 pairs).

## Provenance

Measured over the shipped KB by the #182 fix pass and reported rather than fixed; I have not re-run the three counts. The framing of why it is a judgement rather than a reading is that agent's, and I agree with it.

### LINKED ISSUE #184: Residues of a level-2 group that is itself ATC's residue still license class claims (A16AX 91 pairs, N07XX 55)
## What

#167/#182 vetoed residual ATC subgroups whose **parent** group is defined by administration site, plus everything under `V03A`/`V07A`. A structurally identical family survives: a residue whose parent is *itself* ATC's residue for a main group.

| subgroup | shipped-KB pairs |
|---|---|
| `A16AX` | **91** |
| `N07XX` | **55** |
| `V04CX` | 21 |
| `G02CX` | 3 |
| `R07AX` | 1 |
| `M09AX` | 1 |

Examples that still produce a claim: eliglustat × givosiran (`A16AX`), pitolisant × inotersen (`N07XX`). Two drugs sharing "other drugs for a group that is itself 'other'" have no established relationship — the ancestry asserts nothing, which is exactly #182's criterion.

**`D11AX` is the tell.** It is the same shape and *is* vetoed today — but only incidentally, because dermatologicals happen to be locally applied, so it fell inside the administration-site branch. The rule catches this family by accident where the parent is topical and misses it everywhere else.

## Why #182 did not fold it in

Two honest reasons, both from that PR's hardening:

1. **It moves every number in #182** — the 486/54/7243 partition, the 27-member list, the measured impact — so folding it in would have meant re-deriving a change that was already verified.
2. **The family costs real relationships too**, so it is not a free win: `G02CX` bremelanotide × flibanserin, `M09AX` onasemnogene × risdiplam, `A16AX` miglustat × eliglustat are genuine pairs that would lose their claim.

That second point makes this a judgement about where the cost/benefit lands, like #183, rather than a pure reading of ATC's words.

## Relationship to the neighbours

- **#183** (broad therapeutic buckets — `S01AA` 243 pairs) is the other deferred judgement. Both are "which ATC groupings may justify a clinical claim", and they should probably be decided together with a single stated principle rather than subgroup by subgroup.
- **#182** established the mechanism (`isUnclassifyingAtcCode`) and the measurement method, so implementing either is a list change plus a re-measure, not new machinery.

## Provenance

Measured over the shipped KB by #182's hardening pass and reported rather than fixed. I have not re-run the six counts.

dkayiwa:
Closes #183. Closes #184. Addresses #168 (narrowed, not closed — see below).

## The one existing assertion this touched, and how

`CrossReactivityClassChoiceTest.anIrrigatingSolutionSubgroupDoesNotBecomeTheSystemicAnswer` asserted
`Chlorhexidine is in the same ATC class (A01AB) as the patient's allergy to Neomycin — possible
cross-reactivity`. An aminoglycoside and a bisbiguanide, related only by four "Antiinfectives"-named
subgroups: **#183's defect, spelled out.** I stopped rather than edit it and the owner ruled.

The case's real subject is the tie-break *within* the locally-applied tier, so it **moved to the
duplicate-therapy arm** — same pair, same shape, same exact-string assertion — where all four
subgroups are still candidates and `B05CA` must still be read as locally applied or it is returned
outright. Proved by mutation on a throwaway worktree rather than argued:

| mutation | tests that redden |
|---|---|
| drop `B05C` from `LOCALLY_APPLIED_ATC_GROUPS` | **1 — the moved case, sole detector** |
| drop `A07A` | **1 — a case ADDED here, sole detector; 0 before it** |
| drop `G01` | 4 (the azole cases — already well pinned) |
| remove the locally-applied preference entirely | 7 |
| swap which predicate gates which arm | 4 |
| stop `isPurposeOnlyAtcCode` subsuming `isUnclassifyingAtcCode` | 3 |
| revert the `isUnclassifyingAtcCode` additions | 2 |

The `A07A` row is why this is not just a re-base: the sibling case still passed and still reported
`J01GB`, but it had **stopped pinning `A07A`** — `A07AA` "Antibiotics" is now refused by the
cross-reactivity arm before the locally-applied question is asked, so dropping `A07A` reddened
nothing. A case was added on the duplicate-therapy arm to restore it. `git diff --numstat` shows no
existing assertion changed.

## The question, and what the measurement said

The standing hypothesis was: *ATC classifies purpose and route, not chemistry, so it may license
duplicate therapy but NOT cross-reactivity; only curated groups should* — with an explicit veto if
lost real signal exceeded the false-claim reduction.

**It exceeded it.** Measured by driving the real `DrugSafetyValidator.validate` over the shipped
19 MB knowledge base, on each of the 5550 substance pairs it relates by a level-4 subgroup:

| settlement | cross-reactivity claims | duplicate-therapy claims | curated-group claims |
|---|---|---|---|
| `main` @ 7211f975 | 5266 | 5271 | 0 |
| **A** — blanket (ATC → duplicate therapy only) | **0** (−5266) | 5271 (−0) | **24** |
| **This PR** — per-arm, by the criterion below | **3701** (−1565, 4 renamed) | **5114** (−157) | 0 |

Of the 5266 claims settlement A removes, **3701 rest on a subgroup that does name chemistry or a
molecular target** — `J01CA` penicillins, `J01DD` cephalosporins, `J01GB` aminoglycosides, `N05BA`
benzodiazepines, `C10AA` statins, `M01AE` propionic-acid NSAIDs — and only 1565 on purpose or on
nothing. The one curated group the module ships (NSAID: `M01AE`+`N02BA`, verified — the file has
exactly one) replaces 24 of the 5266. **A loses real signal 2.4× faster than it removes false
claims.** The premise is empirically wrong for the majority of ATC level 4, which is ATC's *chemical*
subgroup tier and mostly reads like one.

So the licensing question is answered **per claim type**, which is what #183 itself proposes:
"a decision on the weakest relationship that may justify each claim type, **which may differ by arm**
… `S01AA` may be legitimate there and not in the contraindication arm".

## The criterion (one rule, read off ATC's own words)

A subgroup may justify a claim only as strong as what its published name asserts, and a residue
asserts whatever the group it sits in asserts (#182's rule, applied at every level rather than one):

1. **names a chemical/structural family, a derivative class, or a molecular TARGET** → licenses both.
2. **names only what its members are FOR** (an indication, an organism acted against, a therapeutic
   area, a diagnostic use) → licenses duplicate therapy, **not** cross-reactivity. #183. New
   `DrugReference.isPurposeOnlyAtcCode`, **117** subgroups.
3. **asserts nothing at any level** — a residue that contributes no term its ancestors' names do not
   already carry, and whose ancestry is residue up to a bare **level-1 anatomical** main group → licenses
   **neither**. #184. **6** subgroups added to `isUnclassifyingAtcCode`.

Level 1 and not level 2, and this is the correction a review pass forced: ATC's level 2 is its
*therapeutic* tier, so a residue inheriting "ANTINEOPLASTIC AGENTS" (`L01XX`) or "ANTIBACTERIALS FOR
SYSTEMIC USE" (`J01XX`) **does** assert a purpose and two drugs sharing it **are** duplicate therapy.
Drawing the line at level 2 withdrew 381 further duplicate-therapy claims that are genuine — including
two erythropoiesis-stimulating agents — while `S01AA` "Antibiotics", the same assertion one tier down,
kept its chip. Same claim, opposite answers, the wrong way round. Stopping at level 1 leaves the
cross-reactivity result identical at 3701 and cuts the duplicate-therapy cost from 538 to 157. The
criterion therefore also disagrees with #184 about `V04CX` "Other diagnostic agents", and the criterion
decides.

The line runs *through* ATC's `anti-` names, not around them: `S01AA`/`J04AB` "Antibiotics" name an
organism to kill, while `R06AX` "Other antihistamines" names a receptor, `N06DA` "Anticholinesterases"
an enzyme and `C01BD` "Antiarrhythmics, class III" a channel — those stay. `N01B` "Anesthetics, local"
and its `S01HA`/`C05AD`/`R02AD` counterparts stay for the same reason, which is the reading
`isLocallyAppliedAtcCode` already records for `N01B`.

Clause 3's "contributes no term of its own" is load-bearing: without it `J01DI` "Other cephalosporins
and penems" inherits "ANTIBACTERIALS FOR SYSTEMIC USE" and loses the chemical family that makes a
cephalosporin allergy mean something. Same for `M03AC` "Other quaternary ammonium compounds".

**Enumerated from the index, not from the reported cases.** All 939 level-4 names in the WHO ATC index
were read and the criterion applied to each. `S01AA`, `A07AA` and `S02AA` are here because the
criterion reaches them; so are 97 others. A list containing only the three subgroups #183 names would
be #161's hand-picked list again, whose hardening found it incomplete in a way that reproduced the
defect it was fixing.

**Cross-check against #182's settled decisions:** the criterion independently reproduces every one
`isUnclassifyingAtcCode`'s javadoc states — `R06AX`, `J01GB`, `N02AX` keep; `A01AD`, `D11AX`,
`S01XA`, `D06AX` are vetoed. It disagrees on 19 of the 51 level-4 codes #182's list covers, and all
19 are #182's own declared over-reach: `S02DC` (which it says a name test "structurally cannot
find"), `V03A`/`V07A`'s 17 children (written at group level deliberately), and `C05BX` "Other
sclerosing agents", whose own name does classify. Nothing moves — this change is additive, so all 19
stay vetoed. A criterion that rediscovers a previously-settled list is stronger evidence than one
that merely covers the reported cases.

**And a consequence worth naming**, because it looks like #171 returning and is not: the two arms
can now name **different** classes for one pair. #171 was two independent scans disagreeing about
which of the *same* candidates to prefer; here the preference is one method and the candidate *sets*
differ. Measured at `sharedClass` over the 5550 pairs (a pair base — the table above is a claim base,
which differs by the handful a question's extra resolutions add): 3693 pairs answer on both arms and
**4** differ. Miconazole × clotrimazole is duplicate therapy on `A01AB` and cross-reacting on
`D01AC`; both are true, and naming them alike would make one false. Pinned by
`theTwoArmsMayNameDifferentClassesForOnePairAndBothAreTrue`.

## Counting bases, since two of the issues' figures do not reconcile

Production-derived (`atcSubgroups` + `substanceGroupKey` + `sharedClass`):

| subgroup | #184 says | substance pairs sharing it | row pairs | pairs *naming* it |
|---|---|---|---|---|
| `A16AX` | 91 | 91 | 91 | 91 ✔ |
| `N07XX` | 55 | 55 | 55 | 55 ✔ |
| `V04CX`/`G02CX`/`R07AX`/`M09AX` | 21/3/1/1 | 21/3/1/1 | same | same ✔ |

**All six of #184's figures reproduce exactly, on the substance-pair base.** **None of #183's three
reproduces on any base**: `S01AA` 243 relayed vs **171** substance pairs / **351** row pairs / **139**
naming it; `A07AA` 102 vs 66/120/63; `S02AA` 81 vs 66/153/44.

That asymmetry is itself evidence about provenance. Both issues say "measured over the shipped KB …
I have not re-run the counts", so both are relayed — but one set lands on a production counting base
exactly and the other lands on none of the three. Whatever produced #183's figures was not counting
substances, rows, or claims over this knowledge base. Treat them as superseded by the table above,
and treat #184's as confirmed.

## Live reachability on the demo instance

Through the production resolvers (`findImpliedSubstances` / `findImpliedByDrugName`) over the 3.7.1
instance's 29 recorded allergen texts and 33 active order names: of the claims this change removes,
**116 have one member as a recorded allergy** and **2 have the other member as an active order**, so
two pairs are fully reachable today (in `A01AB` and `M02AA`). The four regression controls are all
DDInter-rated interactions or curated-group chips and cannot move.

## #168 — narrowed, still open

Counted over SUBSTANCE pairs (the row-pair base behind `sharedClass`'s existing 1090/70 figures has
not been reconciled with it — do not read one against the other; that discrepancy is being filed
separately): **20** pairs leave more than one candidate in the systemic tier for the duplicate-therapy
arm and **16** for the cross-reactivity arm; the stronger requirement breaks four of them. The remainder is unchanged and still alphabetical, and
no data-defensible ranking exists for it: `H02CA` "Anticorticosteroids" and `J02AB` "Imidazole
derivatives" both name a class, so preferring one would be exactly the unmeasured preference #161
refused to invent.

## Verification

- TDD: `AtcCrossReactivityLicensingTest`, 10 cases, on a fixture of 14 rows copied field-for-field
  from the shipped KB. Red first — 5 failures for the right reasons — then green.
- Every case pins a **limb of the criterion**, not a list entry: each has a partner case whose only
  difference is what ATC's name says. `S01AA` refuses cross-reactivity **and keeps its duplicate-therapy
  chip in the very next case**; `J04AB` is the same refusal outside any locally applied group, so no
  site rule can be what decided it; `J01CA`, `J01GB`, `R06AX` and `C05AA`/`S01BA` are the controls that
  must keep their claim.
- Full reactor: **1138 api + 75 omod = 1213, 0 failures**, 53 skipped. Baseline at `main` 7211f975
  in the same worktree: 1126 + 75 = 1201, 0 failures.
- Hardened (`harden` skill) and deployed to the 3.7.1 standalone; probes below.


dkayiwa:
## Live pass on the 3.7.1 standalone

Deployed this branch's omod (`standalone-verify.sh`, full JVM restart, gated on the known-positive
control before probing). Load status at probe time: `loaded=true, inert=false, entryCount=2283,
sourceFormat=ddinter, origin=appdata:chartsearchai/ddi_knowledge_base.json`.

### Controls — none moved

| patient | question | chips |
|---|---|---|
| Mary Smith | Is it safe to give her clarithromycin? | 1 — Major × simvastatin |
| Agnes Adams | Is it safe to start warfarin? | 1 — Major × aspirin |
| Joshua Johnson | Can I give ibuprofen? | 2 — NSAID curated group + lisinopril Moderate |
| Mary Smith | Is it safe to give her paracetamol? | 0 |

Joshua's is worth noting: his chip is the **curated cross-reactivity group**, which this change does
not touch and which is the fall-through every vetoed pair still reaches.

### Canaries — one patient, one allergy list, opposite outcomes

Kevin Brown (`e89247a2-…`) is allergic to **Gentamicin** and **Neomycin** — from the 116 removed
claims that have one member as a recorded allergy on this instance.

**A — a removed purpose-only claim.** *"Is it safe to give chloramphenicol?"* → **no class chip.**
Chloramphenicol shares only `D06AX` (already vetoed), `S01AA` "Antibiotics" and `S02AA`/`S03AA`
"Antiinfectives" with both allergens — purpose, no chemistry. The one chip returned is an unrelated
identity chip for a Clarithromycin allergy the answer named.

*What it would have shown had the change been wrong:* the claim is in the measured baseline set and
absent after — `S01AA,"Chloramphenicol","as the patient's allergy to Neomycin — possible
cross-reactivity"`. If the cross-reactivity arm were not consulting `isPurposeOnlyAtcCode`, that
exact chip would have appeared here.

**B — a retained chemical claim.** *"Is it safe to give tobramycin?"* → **2 chips, both naming
`J01GB`**, one per allergen:

```
Tobramycin is in the same ATC class (J01GB) as the patient's allergy to Neomycin   — possible cross-reactivity
Tobramycin is in the same ATC class (J01GB) as the patient's allergy to Gentamicin — possible cross-reactivity
```

*What it would have shown had the change been wrong:* silence. `J01GB` "Other aminoglycosides" is a
residue whose parent names a chemical family, and the pair also shares the purpose-named `S01AA`. If
the narrowing were over-broad, or if refusing a subgroup dropped the claim instead of continuing the
scan, aminoglycoside cross-reactivity — a real and clinically important signal — would have gone
silent here.

The two together are the whole settlement in one patient: same evidence, opposite answers, decided
only by what the shared ATC subgroup's own published name asserts.

### Hardening

`harden` run to both stopping rules — 5 Phase 1 passes, 1 Phase 2 pass. Phase 1 pass 4 (an external
structural review) found the level-2-vs-level-1 defect corrected in `c0929388`, which is the most
important thing in this PR after the measurement itself.

Full reactor at `c0929388`: **1138 api + 75 omod = 1213, 0 failures, 53 skipped** (baseline at `main`
`7211f975`: 1126 + 75 = 1201).

Standalone left running this branch's build; no demo data created, voided or changed.

dkayiwa:
## Phase 2 polish — applied

Four items, all applied in `35c0a723`; none changes runtime behaviour, so the live pass on
`c0929388` stands.

- **A duplicated private test helper.** The order-with-codes helper had been written into *both*
  test files in this slice, byte for byte. Hoisted to `DrugReferenceTestSupport.activeOrderFor`,
  which is where CLAUDE.md says shared test helpers live and which already owned the four-argument
  `activeOrder` underneath it. The uuid prefix is issue-neutral now — a shared helper has no
  business naming an issue.
- **A false figure I introduced and missed in my own fact-check.** `docs/adr.md` said
  `isPurposeOnlyAtcCode` holds 100 subgroups. It holds **117**, which the javadoc and the ADR's own
  "123" arithmetic both already said.
- **A contract paragraph left on the wrong method.** "`otherSubgroups` must be level-4 SUBGROUPS,
  not full codes" ended up on `sharedTherapyClass` when the javadoc was split, reading as a fact
  about the duplicate-therapy arm. It binds `sharedClass`; both wrappers satisfy it and by
  *different* routes, which is now what it says.
- **`licensesClaim` → `justifiesClaim`.** CLAUDE.md's own word for this concept is *justify*
  ("whether an ATC subgroup may justify a clinical claim"). One idea, one word.

**Assessed and deferred, with the failure mode named rather than a severity label:**

- *The boolean parameter on `sharedClass`.* Kept. Only two arms exist and there is no path to a
  third — the injector reaches its own width through `Collections.disjoint` and never calls this
  method. The literals `true`/`false` appear exactly twice, each on the single line of a wrapper
  whose *name* is the argument, so no reader meets a bare `sharedClass(x, y, true)`. *If we ship
  without an enum, nothing breaks;* the risk would be a future third arm arriving as a second
  boolean, which the wrappers would make obvious at the call site.
- *`fallsUnderAnyGroup`'s linear scan over 117 prefixes.* Measured, not guessed: 290–456 ns/call,
  and it is reached only for subgroups already in the intersection because `refClasses.contains` is
  evaluated first. A pathological 30-order, 5-allergy patient lands near 0.9 ms per `validate()`,
  beside an LLM round trip. *If we ship without an index, the cost is under a millisecond on a path
  that already waits seconds.*

## One thing for you, which I could not do myself

`CLAUDE.md:36` still reads *"**Whether an ATC subgroup may justify a clinical claim**: Use
`DrugReference.isLocallyAppliedAtcCode()` and `isUnclassifyingAtcCode()`."* That concept now has
three predicates, and an agent following the rule as written would reach for
`isUnclassifyingAtcCode` for a cross-reactivity question — **the exact conflation this PR exists to
remove**. README, ADR and `drug-kb-demo.md` were updated; the file that actually binds future
changes cannot be, by me. It wants `isPurposeOnlyAtcCode` added, the per-arm split stated (which
predicate gates which arm), and the ordering recorded (purpose-only subsumes unclassifying).

dkayiwa:
## Phase 2 passes 2 and 3 — and why the loop stops here

Twelve further items across two independent passes, all prose or docs, none behavioural. Applied in
`f83c190a` and `874daef8`. Full reactor at `874daef8`: **1138 api + 75 omod = 1213, 0 failures.**

Pass 2's substantive three: a **false sentence in the README** naming `A16AX` as purpose-only and
saying it still licenses duplicate therapy (it is in `UNCLASSIFYING_ATC_GROUPS` and licenses
neither — the one place a reader meets the split, with the surviving claim backwards); a **fourth
site** still asserting the arms "can no longer describe one pair through different subgroups", on
the javadoc of the very method whose call site this PR edits; and `activeOrderFor`'s javadoc, which
claimed to build the mapped-concept shape. That last was verified independently rather than taken on
report — `PatientClinicalContext` sets `activeDrugAtcCodes` to exactly its argument and never unions
the orders', so all five call sites run through #228's by-NAME leg
(`partner.codes.addAll(row.normalizedAtcCodes()); partner.codesFromDataset = true`) and the codes
passed are dead. Chips are identical either way, so it is a statement about what is exercised, not
about the answer. **The mutation evidence is unaffected** — it was empirical, and the by-name leg
feeds `sharedTherapyClass` the same subgroup set.

Pass 3 then found six more, **three of them in prose passes 1 and 2 had just written**. Where a
claim had already been corrected once, `874daef8` **deletes** it rather than correcting it again:
the fixture note's two enumerations (it said every assertion was about contraindication text, when
three cases assert interaction-type duplicate-therapy chips; and it listed Chloramphenicol/Gentamicin's
shared subgroups as "only purpose-headed", omitting `D06AX`, which is in the other family), and a
`#143` attribution the later javadoc was written specifically to block. The note now states only
what cannot rot.

### Stopping condition — not met, deliberately

Phase 1 converged in 5 passes. **Phase 2 did not**: three passes finding 4, 6 and 6 items, with each
pass finding real defects in its predecessor's corrections. That is HARDEN-BRIEF §5's diverging-prose
signature, and §5's remedy is to stop writing prose, not to write a fourth round of it. **Code and
tests have been stable since `c0929388`; only prose has churned.** I am not claiming a convergence I
did not observe.

What a fourth pass would plausibly still find: wording in the corrections made in `874daef8` itself.
What it would not find: a behaviour change — every item in passes 2 and 3 was a comment, a javadoc
sentence, a docs line or a fixture note, and the suite has been green and unchanged in shape
throughout.

**Deferred, with the failure mode rather than a label:** a third copy of the fixture-characterising
`shared(...)` helper now exists across three test files. *If we ship without hoisting it, nothing
breaks* — all three are read-only characterisations asserting exact set strings, so a change to
`atcSubgroups()` semantics reddens all three loudly rather than silently; and the three differ in
resolver and return type, so a hoist would parameterise more than it shares.

Issue #238, opened 2026-08-13
Two #206 residues: the subject is per-pass not per-request, and getDrug()'s javadoc invites a dedup that would discard findings


dkayiwa:
### 1. The subject is decided per `validate` PASS, not per request

`SubstanceSubjects` memoises "which row does this response call this substance by" for the duration of one `validate` call. The safety layer runs more than once per request — the pre-answer pass that produces the injected `safety_finding` records, and the post-answer pass that produces the chips — and the second pass sees the **answer's** rows as well as the chart's.

So the row a substance is named by can differ between the injected finding a clinician can cite and the chip displayed beside it, when the answer names a row the chart does not. That is the same failure #206 exists to remove, one level up: #206 made all the arms within one pass agree, and did not make the two passes agree with each other.

**Documented, not designed around** — #206's javadoc records it. **Not measured**: no live instance has been observed, and it needs an answer naming a route-qualified row the chart does not carry.

### 2. `SafetyWarning.getDrug()`'s javadoc describes a use nothing makes, and it is a trap

The javadoc says the field exists "for grouping/sorting/deduping". Nothing in this module does any of those with it — and a client that took the invitation would be wrong. **Measured live**: Sarah Taylor carries seven hydrocortisone chips that a client deduping on `(type, drug)` would collapse to one, discarding six distinct findings.

Since #206 the `drug` field is precisely *the substance's name in this response*, which is deliberately the SAME string across chips about one substance — so the field is now maximally unsuitable for deduping, and the javadoc says the opposite. This is the wire-contract half of the same class as [#201](https://github.com/openmrs/openmrs-module-chartsearchai/issues/201): a field a client is invited to interpret in a way the producer does not support.

Fix is a javadoc correction plus, if the grouping use is wanted, saying what a client may actually key on.

## Provenance

Both found by #206's fix pass. Item 1 reasoned from the two-pass structure and recorded in its javadoc; item 2's seven-chip collapse measured live on the standalone. Relayed; I have not re-derived either.

Issue #206, opened 2026-08-09
Anchoring the interaction subject on the chart widens where two arms name one substance differently


dkayiwa:
## What

#205 anchors the **interaction** chip's subject on the row the chart records (#194). `ContraindicationChips` still breaks its tie **positionally** rather than through `canonicalRow`, and is **not** chart-anchored.

`ContraindicationChips`' javadoc already accepted a bounded version of this: an interaction chip and a contraindication chip could name one substance two ways, for the **7 shipped families whose unqualified row is not the dataset's first**. Anchoring one side on the chart widens that set to **any family where the chart names a non-canonical row** — which is a property of the patient's data rather than of the KB, so it is no longer bounded by a count of families.

Reported by #205 rather than hidden, and filed because it is the direct consequence of that PR's own fix.

## Reachability

**Not reachable on current demo data** — no probe patient has a class-related allergy to a multi-row in-play substance — but constructible. So this is a latent inconsistency, not a live defect, and the honest framing is that #205 traded a *bounded* inconsistency for an *unbounded* one while making the arm that matters more truthful.

## Why the trade was still right, and what settles it

Chart-anchoring is what makes a finding truthful — #187 established it and #192 confirmed it (using `canonicalRow` for a charted `Ketorolac (ophthalmic)` allergy renames it wrongly and regresses #187). So the fix direction is not to un-anchor the interaction arm; it is to anchor the contraindication arm the same way.

That is a larger change than it sounds, because `ContraindicationChips`' positional tie-break is load-bearing for a measurement its javadoc records, and #198's hardening showed the ledger key is sensitive: a more precise resolver without a matching ledger key turned 2 chips into 4. So this wants its own pass with its own measurement, not a follow-on patch.

## Related

Last remaining member of the "which representative does this chip speak for" family, alongside the two axes #205 reports as still open: **which rows the module reads at all** (`matchingEntries` still injects the canonical row only, so the record's content is one row's while the chips read the resolved rows — #163's residue, 80 of 121 multi-row substances have a partner only a sibling carries) and **which published number is quoted** (see the dose-ceiling issue filed alongside this).

## Provenance

Reported by #205's fix pass, which measured the 7-family bound from the existing javadoc and reasoned the widening. The unreachability on demo data is its measurement; I have not re-derived it.

PR #248, opened 2026-08-13
fix: four things nothing pins, or a claim nothing supports (#181, #232, #230, #238)

### LINKED ISSUE #181: Production code has the same local-vs-CI querystore divergence: an upstream break is invisible in CI

### LINKED ISSUE #230: containsAnyCode re-normalizes a cross-reactivity group's prefixes once per code
## What

`CrossReactivityGroup.containsAnyCode` calls `containsCode` once per code, and `containsCode` rebuilds `normalizedAtcPrefixes()` on every call. So a group's prefix set is re-normalized once per (group, code) pair, including inside `DrugSafetyValidator`'s pair screening where it runs across the candidate set.

A 3-line hoist fixes it: normalize once per group and reuse.

## Severity and reachability

**Efficiency only, no wrong output**, and it buys essentially nothing today: the bundled cross-reactivity file ships **exactly one** group (NSAID, `M01AE` + `N02BA`), so the repeated work is over a 2-element set.

It matters as soon as a deployment expands the groups — which the file's own description invites, and which is the entire premise of the curated-groups path being the chemically-defensible alternative to broad ATC buckets (see [#183](https://github.com/openmrs/openmrs-module-chartsearchai/issues/183)). The cost grows with (groups × codes) at a point already inside a per-pair loop.

## The constraint any fix must respect

The same one [#172](https://github.com/openmrs/openmrs-module-chartsearchai/issues/172) records: **cache in a local, never in a field.** The groups file is reloadable, and a normalized prefix set cached on the instance across a reload is the shape that silently serves stale data with no test failing, because every test loads once.

## Provenance

Reported by a review pass during #151's hardening, from reading the code. **Not profiled** — the magnitude above is reasoned from the loop structure and the shipped one-group file, not measured. Worth confirming it is on a hot path before doing anything more than the hoist.

### LINKED ISSUE #232: The [timing] mode= ops labels are pinned by nothing, and the comment claims a protection they do not have
## What

`QueryStoreChartBuilder` emits the ops-facing `[timing] querystoreBuild … mode=<x>` label from constants at `QueryStoreChartBuilder.java:71,73,78`. **Nothing pins their values.**

**Measured, not reasoned:** renaming `MODE_FULL_CHART`'s value to `"TYPO_fullChart"` leaves **all 1174 tests green** and the reactor at BUILD SUCCESS.

The comment above them claims compile-time protection. That is true of a typo in the *log statement* — a misspelled identifier does not compile — and false of a change to the *constant's value*, which is the thing an ops consumer actually reads. So the protection the comment claims is not the protection that matters, which is worse than no comment: the next author reads it and does not add the assertion.

## Why it matters

These labels are a consumer contract. Any dashboard, log query or eval instrument grepping `mode=fullChart` breaks **silently** on a value change — no test fails, no build fails, and the breakage shows up as a metric that quietly goes to zero.

That is the same failure class as [#178](https://github.com/openmrs/openmrs-module-chartsearchai/issues/178): a signal a maintainer reads to reconstruct what happened, wrong or absent, with nothing to make it fail loudly. #178's fix added an assertion pinning the audit-log vocabulary for exactly this reason; the timing labels are the other half of the same surface and did not get one.

## Note on scope

These are deliberately **not** the same strings as the audit column's (`fullChart`/`preFilter` here versus `full-chart`/`pre-filter` there). Two contracts, two audiences — do **not** "fix" that by unifying them; that is a wire change nobody asked for, and it was considered and declined during #178. This issue is only about pinning what each already is.

## Fix

One assertion on the three literals, in the shape of the audit-vocabulary assertion #178 added.

## Provenance

Measured by #178's hardening pass via mutation (rename the constant's value, full reactor stays green). Relayed; I have not re-run the mutation.

dkayiwa:
Four small, independent items. They belong together because each is the same shape — **something nothing pins, or a claim nothing supports** — which is the slice [#213](https://github.com/openmrs/openmrs-module-chartsearchai/pull/213) ("make four blind checks able to fail") already established here.

Fixes #181
Fixes #232
Fixes #230
Refs #238 — its **item 2 only**; item 1 (the subject is per-`validate` pass, not per request) is a design question and stays open.

---

## 1. #181 — an upstream break is invisible in CI

`api/pom.xml` declares `querystore-api` at `1.0.0-SNAPSHOT`, so CI resolves whatever snapshot was published remotely last while a developer resolves whatever they `mvn install`ed last. When querystore changes a method this module calls, the local build breaks and CI stays green — three times in one day on the development box, each time as an error naming a chartsearchai class rather than the dependency that moved.

Keeping #181's two halves apart, because the issue does: the three observed breaks were on the **test** side (`getContextSlice`), and #181 records the **production**-side exposure as *reasoned* from the shared `1.0.0-SNAPSHOT` declaration, not observed. This PR's verification closes that half — renaming `QueryStoreService.getPatientChart` upstream breaks `QueryStoreChartBuilder`, i.e. `src/main`. `clean install` covers both source roots.

New job `build-against-querystore-head`: check out querystore at HEAD, `mvn -DskipTests install` it, **check the two repos still agree on the querystore version**, then run the same `clean install` goals the required build runs. On push, on pull_request, and on a daily schedule, because a break can land on a day when nothing here changes.

That version check is not decoration — without it the job's own failure mode is a **false green**, the exact shape it exists to remove. `api/pom.xml` names a literal version; the install step installs whatever querystore HEAD declares. The moment upstream bumps, the build resolves the *published* artifact instead of the one just built and the job passes having tested nothing, silently, because nothing is wrong from Maven's side. (Not hypothetical: this box's local repository still carries a failed `querystore-api:1.4.0-SNAPSHOT` resolution from May.) Run locally in both directions — versions agreeing exits 0; querystore bumped to `1.1.0-SNAPSHOT` exits 1 with `::error::chartsearchai asks for querystore-api 1.0.0-SNAPSHOT but querystore HEAD builds 1.1.0-SNAPSHOT …`.

Its first draft was itself a blind check, which is worth recording since this PR is about blind checks: it filtered `dependency:list` with `-Dincludes`, which is a parameter of `dependency:tree` and **not** of `list`, so Maven accepted and ignored it. Measured: **239** dependency lines came back rather than 1, and the guard passed on its `sed` alone. It now uses `-DincludeGroupIds`/`-DincludeArtifactIds`, anchors the `sed` to the list-entry shape so Maven's own `[WARNING] The POM for … is missing` prose cannot be parsed as a version, scans the whole reactor instead of `-pl api` (so moving the dependency between modules is not a red job), and pins both plugin coordinates because the job deliberately has no Maven cache.

**Deliberately not in the required set, and deliberately without `continue-on-error`.** Two halves of one decision: the job has to be able to go red, because a job that cannot fail is not a signal — and its red X must not block an unrelated chartsearchai PR, because an upstream break is not that PR's fault and its fix belongs upstream. Only branch protection can enforce the second half, so the job says so in a comment. **Please leave it out of the required checks.**

Not doing #181's other option (pinning querystore to a release): a pin defers incompatibilities into one big-bang bump and still never tells you *when* upstream broke you, which is the actual defect.

The daily `schedule:` is for this job alone. The three pre-existing jobs opt out with `if: github.event_name != 'schedule'` — a daily `build` would re-deploy an unchanged SNAPSHOT to the Maven repository every day. (Whether the OWASP scan wants a daily run is a fair question and deliberately not this PR's to answer; its behaviour is unchanged.) The workflow's `concurrency.group` gains `github.event_name` so the scheduled run gets its own lane: GitHub keeps only one *pending* run per group, so without it a push landing while the daily run is queued would cancel it silently — which is the "you don't find out" failure the job exists to prevent.

### Verified locally, since this one cannot be merge-and-seen

| querystore in `~/.m2` | maven | result |
|---|---|---|
| HEAD (`2281499`), built from source | offline | **BUILD SUCCESS** — 1153 api + 75 omod, 0 failures |
| HEAD with `QueryStoreService.getPatientChart` renamed | offline | **BUILD FAILURE** — `cannot find symbol` in `QueryStoreChartBuilder`, i.e. production code, naming a chartsearchai class rather than the dependency that moved |
| the same rename, reinstalled | **online** | **BUILD FAILURE**, and no querystore artifact fetched |
| querystore's own build failing | either | the install step fails, so the job stops there and never reports a false green |

The third row is the load-bearing one. The job's entire mechanism is that a locally installed SNAPSHOT shadows the published one; that is normally assumed, and here it is measured **with the network available**. So a green run really is a run against HEAD.

**Finding: this module builds clean against querystore HEAD today** — the job would be green on merge.

## 2. #232 — the `[timing] mode=` ops labels were pinned by nothing

The comment above those constants claimed compile-time protection. That is true of a misspelled *identifier* in a log statement and false of a change to the constant's **value**, which is the only thing an ops consumer reads — and the comment is worse than no comment, because the next author reads it and does not add the assertion.

One assertion pinning the three literals, in the shape of the audit-vocabulary assertion #178 added, and the comment corrected to say what does and does not protect them.

**Confirmed by mutation:** renaming `MODE_FULL_CHART`'s value to `"TYPO_fullChart"` fails **exactly one** of the api module's 1157 tests — the new one — and omod references neither the constants nor their literals, so before this the re-spelling shipped green.

The two vocabularies stay distinct (`fullChart`/`preFilter` here, `full-chart`/`pre-filter` in the audit column). Two contracts, two audiences; unifying them was considered and declined during #178.

## 3. #230 — `containsAnyCode` re-normalized a group's prefixes per code

`containsCode` rebuilds `normalizedAtcPrefixes()` on every call and `containsAnyCode` asked it once per code, so one membership question cost a normalization per `(group, code)` pair — inside `DrugSafetyValidator`'s per-pair screening.

Hoisted into a **local, never a field**. That is #172's constraint and it is the whole risk in this change: the groups file is reloadable and `setAtcPrefixes` is the write path the loader and Jackson go through, so a set cached on the instance would serve stale prefixes with nothing failing, because every test loads once.

**No speedup is claimed and none was measured** — the shipped file carries one group with two prefixes. The justification is not carrying a `(groups × codes)` shape into a deployment that expands the curated groups, which is what that path is for (#183).

Two tests, each confirmed by its own mutation on a throwaway tree:

| mutation | test that reddens |
|---|---|
| hoist reverted to the pre-fix shape | `aMembershipQuestionNormalizesEachGroupsPrefixesOnce_notOncePerCode` — expected 1, was 3 |
| `normalizedAtcPrefixes()` cached in a field | `replacedPrefixesAreSeenOnTheNextQuestion_soTheNormalizationIsNeverCachedOnTheInstance` |

Both drive `CrossReactivityGroup.sharedGroupForCodes`, the production order-code entry point, over the real bundled group's own prefixes; the counting subclass only observes, delegating to the production `normalizedAtcPrefixes()`.

Hardening removed a third test of mine and the two guards it covered. I had added a short-circuit to `containsCode` and an empty-set early return to `containsAnyCode`, with a test for both — and neither input is reachable from production: `containsCode` has no production caller once the hoist rewires `containsAnyCode`, and both callers of `containsAnyCode` pre-guard emptiness. That is a test whose subject is unreachable code, driven by hand-crafted inputs, which is what CLAUDE.md's real-pipeline rule exists to stop. Reverted to exactly the hoist #230 asked for.

## 4. #238 item 2 — `SafetyWarning.drug` is not a deduplication key

The javadoc said the field exists "for grouping/sorting/deduping". Nothing in this module does any of those with it, and a client that took the invitation would be wrong: since #206 the field is the name a chip *arm* gives that substance, shared across the drug-in-play, dose and contraindication chips, so it identifies a **substance, not a finding**. #238 records a patient carrying seven hydrocortisone chips that a client keying on `(type, drug)` would collapse to one, discarding six distinct findings.

It is not even a *stable* substance name — some checks are deliberately exempt and name the row the chart itself records, so one substance can appear under two labels in one response. That makes the "not a dedup key" instruction *more* true, and both the javadoc and the README now say so.

Hardening spent three passes on that one paragraph and each found the replacement wrong in a new way: an enumeration of which arms share a subject that omitted the #164 identity-contraindication exemption, and a live case cited as current that `DrugSafetyValidator` records in the past tense as a *fixed* #174 defect. The paragraph now states the client-facing fact and points at `SubstanceSubjects`' javadoc for the arm-by-arm list rather than re-deriving it — which is where that list already lives, and where it has moved before.

**`README.md`'s wire-contract paragraph** carried the same invitation verbatim, to the audience that would actually act on it — fixing only the javadoc would have left the client-facing copy saying the opposite. (Note for the issue: the sentence was on `getDetail()`, not on `getDrug()` as #238 says.)

---

## Build

Full reactor `mvn -o clean install`, measured in the branch worktree at the pushed SHA:

* `main` (`15f42380`): **1153 api + 75 omod = 1228**, 0 failures, 53 skipped
* this branch: **1156 api + 75 omod = 1231**, 0 failures, 53 skipped

No live pass. Items 1, 2 and 4 change no runtime behaviour at all (CI config, a test, javadoc and README). Item 3 is the only production change and it is answer-for-answer identical — the same predicates over the same inputs, verified case by case and by 1153 unchanged tests staying green; only the number of times a prefix set is rebuilt moves.


dkayiwa:
**Correction to this PR's own body, recorded rather than silently edited.**

The body says:

> the groups file is reloadable and `setAtcPrefixes` is the write path the loader and Jackson go through, so a set cached on the instance would serve stale prefixes with nothing failing

**The first clause is false, and this PR's own pass is what established that** — its final report stated: *"'The groups file is reloadable' isn't true: `CrossReactivityGroupsLoader.parse` builds fresh objects and `DrugReferenceService.getCrossReactivityGroups()` caches for the bean's lifetime with no clearing path."* The body was written before that finding and never updated, so a merged artifact now asserts the opposite of what the work measured.

**#172's pass then measured the same for the entries**, two ways: statically, `dataset` is written only in `ensureLoaded()` (when null) and in the no-production-caller test seam `setEntries()`; and live, flipping `sourceFormat` `ddinter`→`atc` on a running server left the status endpoint at `entryCount 2283, sourceFormat ddinter` and a real `/search` still returning its chip — where a genuine reload would have parsed the DDInter file as `atc`, yielded 0 entries, and dropped the chip to 0. Re-measured first-hand by #257's pass, GP restored.

**The rule this PR applied — hoist into a local, never a field — is still right**, for the reasons #257 records in code and in `CLAUDE.md`: these beans are Spring singletons so a field memo is a `HashMap` shared across concurrent requests, and the keys are patient free text so it grows unbounded. `setAtcPrefixes` staying authoritative after a read is the groups-specific reason.

Flagging it here because I propagated the reload premise myself — into #172's comment thread and into two agent briefs — and because #257's pass found my brief citing *this body* as corroboration for the opposite claim. It deleted the attribution rather than repeating it, which was the right call.

PR #249, opened 2026-08-13
fix(validity): report the substance merge the alias rule cannot see (#196 item 4)

dkayiwa:
## What this settles

Issue #196 item 4 — `Fluoroestradiol f-18` and `Estradiol` are ONE substance to this
module — and the premise a close would have rested on, which is false.

**The premise, tested first.** The plan for #196 was to close it because the load-time
validity check "already detects and reports its defects every boot". Measured by driving
`DrugReferenceService.getLoadStatus()` over the shipped 19 MB knowledge base:
`alias-names-another-substance` fires **18** times, and the full list is `Omeprazole`,
`Pfizer-BioNTech Covid-19 Vaccine`, `Trastuzumab emtansine`, `Levoketoconazole`, the five
`Tozinameran` rows, `Gallium citrate ga-67`, `Gallium chloride Ga-67`,
`Dotatate gallium ga-68`, `Fenfluramine`, `Fenofibric acid`, `Fosnetupitant`,
`Amphetamine`, `Hyoscyamine`, `Iodide I-131`. **None of #196's items 1–5 or 7 is among
them.** Item 4 is not merely absent — it is structurally unreachable by that rule, whose
first exclusion is that the two entries key ALIKE, and a shared identity is exactly what
item 4 is.

**What the identity actually is.** Measured through `substanceKey()` on the loaded rows:
`Fluoroestradiol f-18` (`rxnorm_name: estradiol`, `drugbank_id: null`), `Estradiol` and
`Estradiol (topical)` all key `[estradiol, db00783]`. The tracer carries no registry id of
its own; `DdiDrugReferenceSource`'s per-family resolution hands it estradiol's. And
`canonicalRow` for that family answers **`Fluoroestradiol f-18`** — the tracer represents
the substance.

**It reaches a clinician.** Live on the standalone at this branch's HEAD, Margaret King
(active Dexamethasone order), asked *"Is it safe to give her estradiol?"*:

```
type=interaction drug=Fluoroestradiol f-18
Fluoroestradiol f-18 interacts with active order dexamethasone — Moderate.
Estrogens may enhance the systemic effects of both endogenous and exogenous corticosteroids...
```

A PET imaging tracer named as the subject of a question about a therapeutic oestrogen,
carrying the oestrogen's own mechanism text, on the shipped default configuration.

## The rule

`derivative-merged-with-its-parent-substance`, REPORTED. A row the dataset merged into a
substance its own name says it only DERIVES from. Two gates, each excluding a shape that
is correct:

- the family must hold a row that is **not** a derivative — the parent the merge is with;
- the name must not carry the substance's name as a bounded WORD, so presentations, salts
  and esters are read as what they are (12 shipped rows carry their substance's name that
  way: `Beclomethasone dipropionate`, `Estrone sulfate`, the four `Human immunoglobulin G`
  routes …).

**The broader predicate was measured first and rejected.** "One substance, unlike published
names" fires on 15 rows across 8 families and that set includes BOTH merges issue #164
measured as CORRECT — `Daxibotulinumtoxina`/`Botulinum toxin type A` and
`Pfizer-BioNTech Covid-19 Vaccine`/`Tozinameran`. A rule reporting those tells an operator
that this module's own correct behaviour is a data defect. So the criterion is the
relationship the row's own NAME states, which is how `containsWord` already separates
`Levoketoconazole` from `Ketoconazole` and `Esomeprazole` from `Omeprazole` everywhere else
here. Those pairs stay apart only because each derivative carries its own DrugBank id; this
rule is the residue of that mechanism.

**Shipped count: 1 row.** `Fluoroestradiol f-18 is filed as 'estradiol' beside [Estradiol,
Estradiol (topical)]` — measured through the rule itself via `getLoadStatus()`, and seen on
the wire beside `alias-names-another-substance / reported / 18`.

**REPORTED, not repaired**, for the reason the class javadoc already gives: splitting the
rows invents the fact the data is missing, and dropping the row loses four real interaction
rules. Both fail closed, silently, to fix a fail-loud.

## Evidence

- Full reactor at HEAD: **1158 api + 75 omod = 1233**, 0 failures, 53 skipped
  (`main` at `0ec44e32`: 1156 + 75 = 1231).
- Live: both findings on `GET /chartsearchai/drugreferencestatus`; the four controls
  unchanged (Mary × clarithromycin 1 Major × simvastatin, Agnes × warfarin 1 Major ×
  aspirin, Joshua × ibuprofen 2, Mary × paracetamol 0). A REPORTED rule mutates nothing, so
  those are regression controls and cannot corroborate the change; the discriminating probe
  is the status payload, which is absent unless this build is deployed and the rule is
  wired.
- Every decision confirmed load-bearing by mutation: drop the parent gate → 3 occurrences
  instead of 1, two of them naming a merge that never happened; drop `containsWord` → the
  fixture fires from the parent's side and two pre-existing "loads in silence" tests redden;
  drop the substring test → `Daxibotulinumtoxina` is reported; drop the diacritic fold →
  the localized row goes silent; drop the `namesAnything` gate →
  `Bupivacaine hcl-2 is filed as '-' beside [Marcaine]`.

## Not fixed here, and worth its own issue

`canonicalRow` names the estradiol substance after the tracer because both rows answer
`namesNoRoute()` and the tie falls to dataset order, where the tracer is index 1282 and
`Estradiol` is 1927. Four shipped families would change representative row under a third
rung preferring the row whose display name IS the family's substance name — and all four
look like improvements. That is a change to a heavily-guarded shared method with its own
blast radius, so it is reported rather than folded in here.

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
@@ -361,6 +361,16 @@ public class DdiDrugReferenceSource implements DrugReferenceSource {
 	 * that is what merges {@code Thallous Chloride}/{@code Thallous chloride tl-201} and
 	 * {@code Typhoid vaccine (live)}/{@code Typhoid vaccine live}.
 	 *
+	 * <p><b>The first case also merges one pair it should not, and the loader says so.</b> A row
+	 * carrying no id takes the family's whichever it is, so a DERIVATIVE filed under its parent's
+	 * {@code rxnorm_name} is merged into the parent on the strength of a claim nothing checked:
+	 * {@code Fluoroestradiol f-18} publishes {@code rxnorm_name: estradiol} and no id, so a PET imaging
+	 * tracer takes {@code DB00783}. Nothing here can tell that from the two shapes above — a second
+	 * NAME for one substance and a derivative of it are the same row to this method — so it is not
+	 * repaired here; {@link DrugReferenceValidity#DERIVATIVE_MERGED_WITH_ITS_PARENT_SUBSTANCE} reports
+	 * it at load instead, keyed on the relationship the display name states. Issue #196 item 4, and the
+	 * one row in the shipped KB it fires on.
+	 *
 	 * <p>Measured over the shipped 19 MB KB (2283 rows; re-measure before relying on the figures): 142
 	 * substance names are shared by more than one row, 19 of them naming two or more DrugBank
 	 * substances and 123 naming at most one. No group the resulting key produces holds two DrugBank
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReferenceValidity.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReferenceValidity.java
@@ -615,32 +615,29 @@ public final class DrugReferenceValidity {
 	 *       shape issue #196 records as undetectable from inside the file, and a derivative the module
 	 *       has correctly kept APART from its parent that has route variants of its OWN;</li>
 	 *   <li>the name must carry the substance's own name as a bounded WORD to be excluded, so every
-	 *       presentation, salt and ester stays silent — and so does every second NAME for one
-	 *       substance, which carries it neither as a word nor as a substring. That exclusion is the
-	 *       load-bearing half: measured over the shipped KB, <b>12</b> rows are silent only because of
-	 *       it, every one inspected an ester, a salt or a preparation
+	 *       presentation, salt and ester is read as what it is rather than as a derivative. That is not
+	 *       a rare shape: 12 rows of the shipped KB carry their substance's name as a bounded word
 	 *       ({@code Beclomethasone dipropionate} under {@code beclomethasone},
 	 *       {@code Estrone sulfate} under {@code estrone}, the four
-	 *       {@code Human immunoglobulin G} routes under {@code immunoglobulin g}), so without it this
-	 *       rule would report 13 rows of which 12 are correct data.</li>
+	 *       {@code Human immunoglobulin G} routes under {@code immunoglobulin g}), measured through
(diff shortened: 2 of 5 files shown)


dkayiwa:
**Hardening summary — three rounds, each catching something the round before introduced.**

Every decision in the rule is now confirmed load-bearing by mutation, and two of the
mutations found controls of mine that could not fail:

| mutation | what reddens |
|---|---|
| parent gate → bare size check | 3 occurrences instead of 1; two name a merge that never happened (`Levoketoconazole` ↔ its own `(oral)` variant) |
| `containsWord` removed | the fixture fires from the parent's side, and two pre-existing "loads in silence" tests redden |
| `containsWord` → bare equality | `Sodium oxybate is filed as 'oxybate' beside [gamma-Hydroxybutyric acid (potassium oxybate)]` |
| substring test removed | `Daxibotulinumtoxina` reported — #164's correct merge |
| diacritic fold removed | the localized row goes silent |
| `namesAnything` gate removed | `Bupivacaine hcl-2 is filed as '-' beside [Marcaine]` |

**Three things I got wrong and fixed, recorded because they are the interesting part:**

1. **A blind check I added.** The `namesAnything` gate's first witness was a substance name
   of nothing but combining marks. It cannot reach the rule: such a name folds to EMPTY,
   `contains("")` is true of every stem, so every row of the family derives and the parent
   gate silences it first. The gate could be deleted with all cases green. The real witness
   is a name that folds to something non-empty carrying no letter or digit.
2. **The word-boundary half had no coverage.** `containsWord` → `equals` left every case
   green, because the `Beclomethasone` pair shares one display stem and so answers the
   predicate alike — the parent gate absorbed the difference. `Sodium oxybate` is the shape
   that discriminates.
3. **"Which datasets can fire" was wrong twice** — once in each direction — so it is now
   derived from the two gates rather than stated as a list.

**Live**, at `0294ff93` (production code unchanged since, the last commit is test-only):
both findings on `GET /chartsearchai/drugreferencestatus`, the four regression controls
unchanged, and the reachability canary reproducing the tracer-named chip.

Issue #270, opened 2026-08-14
A dose stated for one substance is charged to another whose name it contains, because a tie is not strictly nearer


dkayiwa:
## What

`DrugSafetyValidator.substanceOwnsDose` gives a stated dose to a substance unless a DIFFERENT substance's
name sits **strictly** nearer the `N mg`. Where two substances' published names **nest** and the dose
precedes both, the two names start at the **same index** — so neither is strictly nearer, no veto fires
on either side, and **both** substances claim the number. The shorter one then warns about a dose the
answer stated for the longer one.

This is the shape [#260](https://github.com/openmrs/openmrs-module-chartsearchai/issues/260) predicted
(`estrone` inside `estrone sulfate`). #260 turned out to be a different defect — the locator disagreeing
with the prose gate — and its fix does not touch this one: `estrone` really is a whole word inside
`estrone sulfate`, so the prose rule finds it exactly where a substring search did. The two remedies also
move opposite ways, which is why they were kept apart: #260's fix **adds** an attribution, this one must
**remove** one.

## Measured, not reasoned

Measured 2026-08-14 on the branch for #260, through the real `DrugSafetyValidator.validate` entry point
over a two-entry `json` fixture parsed by the real `JsonDrugReferenceSource`.

* fixture: `Estrone` (alias `estrone`, ATC `G03CA07`) and `Estrone sulfate` (alias `estrone sulfate`,
  ATC `J01CE01`), each publishing `maxDailyDoseMg: 1` for ages 0-120;
* question: `Is estrone or estrone sulfate safe for her?` — `findImpliedByQuery` returns
  `[Estrone, Estrone sulfate]`;
* answer: `The usual approach is 2.5 mg of estrone sulfate daily.`

Result — **two** overdose warnings for **one** stated dose:

```
The stated Estrone dose ~2.5 mg/day exceeds the 1 mg/day maximum for ages 0-120
The stated Estrone sulfate dose ~2.5 mg/day exceeds the 1 mg/day maximum for ages 0-120
```

The first is wrong: the answer stated 2.5 mg **of estrone sulfate**, and nothing in it doses estrone.

## Why it is a false positive rather than an ambiguity

The arm's own contract is "the nearest drug name owns the number". At a tie the two names do not merely
sit equally near — one is a strict sub-span of the other **at the same index**, so the text is not
ambiguous at all; it names the longer substance, and the shorter one is only present as its prefix. The
[...]

PR #271, opened 2026-08-14
fix(drug-safety): locate a name in an answer by the rule that decides the clause names it

### LINKED ISSUE #260: nearestAliasDistance matches an alias by raw indexOf, so a nested name can claim another substance's dose
## What

`DrugSafetyValidator.nearestAliasDistance` locates a substance's alias in the answer with a raw `indexOf`, not the word-bounded matching the rest of this module uses.

So an alias that occurs **inside** a rival substance's name can drive the distance to 0 and let the wrong substance claim a stated dose — blocking the rival from claiming its own. `estrone` inside `estrone sulfate` is the clean example.

## Why it matters more after #245

Pre-existing, and [#245](https://github.com/openmrs/openmrs-module-chartsearchai/issues/245) widens its blast radius: the dose is now read once **per substance** rather than once per row, and `nearestAliasDistance` is the min over the substance's rows. Before, a bad match affected one row; now it can decide attribution for the whole substance.

## Status of the claim

**Reasoned from the code, NOT constructed end to end.** No fixture demonstrates it and no live instance has produced it. It is filed because the mechanism is plain in the code and because #245 changed its scope — not because it has been seen.

## The accessor question

CLAUDE.md's matching bullet is explicit that the three string shapes get three different boundary rules, and that widening one to serve another was #86, #128, #147 and #209. A stated dose sits in **prose** (an answer), so `DrugReference.matchesText()`'s symmetric word boundaries are the candidate rule — but note `nearestAliasDistance` needs a *position*, not a boolean, so it cannot simply call the existing predicate. Whoever takes this should say which rule they applied and why, rather than reaching for the nearest matcher.

## What would settle it

A fixture with two substances whose aliases nest (`estrone` / `estrone sulfate`), an answer stating a dose for the longer one, and an assertion that the shorter one does not claim it. If that passes today, the concern is theoretical and the issue should be closed saying so.

## Provenance

Found by #245's fix pass while widening the dose read to substance scope; reported explicitly as reasoned rather than measured. I have not re-derived it.

dkayiwa:
Fixes #260.

## What was wrong

`DrugSafetyValidator`'s dose arm gated each clause of the answer on `DrugReference.matchesText()` — the
PROSE rule, symmetric word boundaries over a diacritic fold — and then located the drug's name **in that
same clause** with a raw `String.indexOf`. Two rules for one question, one step apart.

They disagree in **both** directions, and both silently:

* a name the prose rule does **not** find is found by `indexOf` — `penicillin` inside the plural
  `penicillins` — so an entry the clause does not name can sit nearer the `N mg` than the subject does and
  **veto a real dose**;
* a name the prose rule **does** find is not found by `indexOf` — `paracetamol` written `paracétamol` —
  so a substance the clause names **cannot locate itself**, `nearestAliasDistance` answers
  `Integer.MAX_VALUE`, and its own dose is never attributed to it.

Either way the overdose warning simply disappears: no chip, no log line, nothing distinguishing "within
the ceiling" from "never compared against it". That is the direction this whole arm exists to prevent.

### The issue predicted a different symptom

#260 predicted a nested name **claiming** another substance's dose, with `estrone` inside
`estrone sulfate` as the example. **That example is not this defect and the fix does not touch it**:
`estrone` is a whole word inside `estrone sulfate`, so the prose rule finds it exactly where `indexOf`
does. Constructing the issue's shape found the *blocking* direction above instead, which is what the new
cases pin. The estrone shape is real but is a different mechanism — see "Known limitations" below.

## The fix, and which boundary rule it applies

The **prose rule** (`containsWord`, zero trailing letters), because the string being read is an answer.
CLAUDE.md's matching bullet is explicit that the three string shapes get three different boundary rules and
that widening one to serve another was #86, #128, #147 and #209; a stated dose sits in prose, and the gate
one line above had already asked the prose question about this very clause.

`nearestAliasDistance` needs a **position**, not a boolean, so it cannot call the existing predicate. It
does not get a second implementation either. `DrugReference.containsBoundedToken` — the one boundary-aware
scan already shared by `containsWord` (prose) and `matchesOrderName` (a recorded drug name) — is split into
`boundedTokenIndex(t, w, maxTrailingLetters, from)`, which returns the index instead of `true`, and the
existing boolean becomes `boundedTokenIndex(...) >= 0`. The new `DrugReference.nearestNameDistance` calls it
with `maxTrailingLetters = 0`, which **is** `containsWord`'s parameterisation. So the WHETHER and the WHERE
are now the same code taking the same branch, not two expressions of one rule.

`nearestNameDistance` lives on `DrugReference` because it is a question about the entry's own names, and
because standing beside them is what lets it share the rule instead of restating it — the whole of the
defect.

### The fold is a contract, not an economy

`foldDiacritics` decomposes and drops combining marks, so it is **not length-preserving**: an index into a
folded string is not an index into the string it was folded from. Every position the dose arm compares —
`DOSE_MG`'s own `start()`, the limit-cue lookback, each substance's nearest name — must therefore come out
of one string. `attributedDoses` now folds each clause **once, at the top**, and everything below reads that
one coordinate system. Folding inside the locator instead would have left `DOSE_MG`'s index measured in the
unfolded clause and the names' indices in the folded one.

### TDD

Reproduced as a genuine red before any production change: reverting both production files to `main`
(`2f86eeef`) fails exactly `aNameOnlyASubstringSearchFindsDoesNotTakeTheDose` and
`aSubjectTheClauseNamesWithAnAccentStillOwnsItsOwnDose`, and nothing else.

### Mutation sweep — 6 mutations, 5 caught, **1 survived and was closed**

Each on a throwaway clone, `mvn -o clean test -pl api`, api tests 1199 throughout the sweep.

| mutation | failures | caught by |
|---|---|---|
| both production files reverted to `main` | 2 | the substring case **and** the accent case |
| the fold hoist removed (`clause = sentence`) | 1 | the accent case |
| the locator uses the ORDER-NAME rule (`maxTrailingLetters` 0 → 2) | 1 | the substring case |
| the rival veto in `substanceOwnsDose` disabled | 2 | the new bound case **and** `DoseCeilingBySubstanceTest` |
| the locator uses raw `indexOf` on the folded clause | 1 | the substring case |
| the locator stops at an alias's FIRST occurrence | **0 — SURVIVED** | nothing |

The third is the sharpest of the caught five: `penicillins` is `penicillin` plus one inflection letter, so
the neighbouring boundary rule reaches it and the prose rule does not. It is not enough for the locator to
be word-bounded — it has to be bounded by the **same** rule the gate used, and the suite now reddens if
someone swaps in the neighbour.

The fourth is why the bound case exists: without it, the substring case could be turned green by deleting
the veto rather than fixing the rule.

**The sixth is the interesting one.** `nearestNameDistance` promises the *nearest* occurrence of a name,
and the scan that delivers that promise — driven by the `from` parameter this change introduced — was
pinned by nothing: reducing "nearest" to "first" left all 1199 api tests green. The scan itself predates
this issue, but the parameter is new here, so it was closed rather than reported:
`aSubstanceNamedTwiceInOneClauseIsLocatedByItsNearestMention` names the subject twice in one clause, the
head mention 168 characters from the `1500 mg` (past the 120-character attribution window) and the second
against it. First-occurrence-only reads the substance as too far away and drops a 6000 mg/day statement in
silence.

That case ships with its own premise, `FAR_MENTION_ONLY`, asserted through the production path rather than
by arithmetic: the same clause with the near mention removed must produce **no** warning. Widening the
attribution window past 156 characters reddens the premise instead of leaving the case passing for the
wrong reason — a control that can silently stop being a canary is not a control.

### The fixture is hand-authored, and has to be

A dose warning needs `ageBands` and the substance grouping needs `substanceName`, and **no bundled dataset
carries both** — the ddinter source publishes substance names and no bands, the shipped curated seed
publishes bands and no substance names. `DoseCeilingAttributionTest.noShippedConfigurationCanReachTheAttributionAtAll`
asserts that over the seed rather than arguing it. Every case runs the real `validate` entry point over the
fixture parsed by the real `JsonDrugReferenceSource`.

### Which direction each behaviour moves — both measured directions stated

* **Boundary narrowing** (fewer occurrences located) — this *adds* dose warnings where a rival was located
  only as a substring, and it can *remove* one only where a substance was near the dose **solely** by a
  substring occurrence of one of its own names while a genuine occurrence sat further than the 120-character
  attribution window. That is the defect direction itself: a substring is not a naming.
* **Fold widening** (more occurrences located) — this adds the subject's own accented spelling, and
  symmetrically lets an accented **rival** take a dose, and lets an accented limit cue suppress one. All
  three are the same rule applied consistently; the accented-rival case is the accented-subject case the new
  test pins, read from the other side.

Reasoned by case analysis, not measured live (see below). Empirically: the full suite is green, including
every existing dose case, and the shipped-configuration bound test pins the common case's exact wording.
On the shipped `json` seed the reference side cannot move at all — its 4 entries publish 12 aliases, none
nested inside another and none carrying a combining mark — so only the answer side can differ. On `ddinter`
and `atc` deployments the arm never runs: no entry publishes a band, and `addOverdose` returns before
reading the answer.

### Live verification: NOT AVAILABLE, and that is an environment fault rather than a gap

The developer standalone is up but returns HTTP 500 on every page — its own JVM cannot read its archives —
and the directory it lives in denies reads, overwrites and renames of pre-existing files, so it can be
neither probed nor redeployed nor recovered from this session. No claim in this PR rests on a live
observation. Recovery needs a permission restored outside the session.

## Known limitations, deliberately not addressed here

Where two substances' published names **nest** and the dose precedes both (`2.5 mg of estrone sulfate`),
the two names start at the same index, so neither is strictly nearer and **both** claim the dose — the
shorter substance warning about a number stated for the longer one. The boundary rule does not touch it,
because `estrone` really is a whole word inside `estrone sulfate` and the prose rule finds it exactly where
a substring search does. Its remedy *removes* an attribution where this issue's *adds* one, so it is filed
separately as **#270** rather than folded in, and recorded in `attributedDoses`' javadoc.

Measured during hardening rather than reasoned, through the real `validate` over a two-entry fixture: one
stated dose, **two** overdose warnings —

```
The stated Estrone dose ~2.5 mg/day exceeds the 1 mg/day maximum for ages 0-120
The stated Estrone sulfate dose ~2.5 mg/day exceeds the 1 mg/day maximum for ages 0-120
```

— of which the first is about a number the answer stated for the other substance.

Unchanged: only the literal unit `mg` is parsed — grams, "mgs" and "milligrams" are not — which is the
conservative direction.

### And one residual of this very defect, measured rather than assumed

The gate and the locator agree for every string this arm can receive, but **not structurally**, and the
PR says so at both sites rather than rounding it off. `namesSubstance` reaches `matchesText`, which folds
its own operands, so the clause is folded a second time; the locator reads it exactly as `attributedDoses`
established it. That is a no-op only while `foldedLower` is idempotent, and **it is not**. Measured through
the production methods:

* the sequence `a`, U+1D16D, U+0E31, U+1D165 folds to `a`, U+1D16D, U+1D165 and folding *that* gives
  `a`, U+1D165, U+1D16D — stripping a canonical-class-0 mark from between two of higher class merges two
  canonical-ordering runs, and the next NFD reorders them;
* so an entry whose alias is `a`, U+1D165, U+1D16D **is matched by the gate and is not located** by the
  locator, leaving `mine = Integer.MAX_VALUE` for a clause that passed.

Unreachable from any dataset or chart string — it needs musical combining marks inside a drug name — and
closing it means changing which accessor `namesSubstance` calls, which CLAUDE.md governs. Recorded on
`DrugReference.foldedLower` and `DrugSafetyValidator.substanceOwnsDose`, and reported rather than taken.

## What hardening changed

Phase 1 ran five passes, Phase 2 two rounds of parallel review (four lenses, then three). Beyond the
mutation gap above, it found **five false or vacuous javadoc claims — two of them introduced by the
hardening's own first round**:

* `containsBoundedToken` said a hand-authored `json` KB is not sanitized and that an authoring guard
  belongs in its parser. Both were true before #150 and neither is now — `DrugReferenceValidity.checkEntries`
  runs `sanitizeAliases` on every format's load.
* the fixture's stated reason for existing was wrong: `substanceName` is not required (`substanceGroupKey`
  falls back to the row), the shipped seed publishes the same 4000 mg/day Paracetamol ceiling, and **three**
  of the six cases would run on it, not four. The fixture exists for the rival and nothing else.
* a caveat about datasets whose entries omit their own display name is vacuous post-#150 — deleted rather
  than reworded.
* "the boundary half moves only toward more warnings except where a substance was near the dose SOLELY by a
  substring occurrence" was not exhaustive: a grown `mine` also lets a rival become strictly nearer.
* "structurally impossible" — the residual above.

It also bound the prose allowance once (`PROSE_TRAILING_LETTERS`, beside the order-name constant): the
first hardening round's javadoc claimed the boolean and the index "cannot come to disagree" while both
still spelled a bare `0`, which is the same drift one level up. `nearestNameDistance` depends on it
arithmetically — its `end` is `idx + w.length()`, the whole match only while the allowance is zero.
`DrugReference.foldedLower` names the operand form once so `attributedDoses` no longer inherits the
lowercase half from a parameter 4500 lines away.

## Test counts

Full reactor `mvn -o clean install`, both measured by me in clean checkouts — the baseline on `main`
itself, not by arithmetic on the branch:

| | api | omod | total |
|---|---|---|---|
| `main` (`2f86eeef`), measured in its own checkout | 1194 | 75 | 1269 |
| this branch | 1200 | 75 | **1275** |

0 failures, 0 errors, 53 skipped, both.


dkayiwa:
## Hardening summary

Phase 1: **five passes.** Stopping condition met: *the last two consecutive passes returned only cosmetic
items.* Phase 2: **two rounds** of parallel review (four lenses, then three, the second briefed to
fact-check the first's corrections).

### What it changed

**A mutation gap in this PR's own new API.** `nearestNameDistance` promises the *nearest* occurrence, and
the `from` parameter that delivers it was pinned by nothing — reducing "nearest" to "first" left all 1199
api tests green. Closed by a sixth case with its own production-driven premise, so widening the attribution
window reddens the premise rather than leaving the case passing for the wrong reason.

**Five false or vacuous javadoc claims — two of them introduced by hardening's own first round.** Listed in
the PR body. The sharpest is the last: the first round wrote that a clause passing the gate could not fail
to locate its subject, "structurally". It can — `matchesText` folds the clause a second time that the
locator does not, and `foldedLower` is **not idempotent** (measured: stripping a canonical-class-0 mark
from between two of higher class merges two ordering runs and the next NFD reorders them). It needs musical
combining marks inside a drug name, so it is unreachable from any dataset; it is now recorded at both sites
and reported rather than closed, because closing it means changing which accessor `namesSubstance` calls.

**One binding that was claimed and not delivered.** The first round's javadoc said the prose allowance was
"bound once, so the boolean and the index cannot come to disagree" while `containsWord` and `wordIndex`
both still spelled a bare `0`. `PROSE_TRAILING_LETTERS` now joins `MAX_ORDER_NAME_INFLECTION_LETTERS` and
both take it — which also matters arithmetically, since `nearestNameDistance`'s `end` is
`idx + w.length()`, the whole match only while the allowance is zero.

Plus: `foldedLower` names the operand form once; the surviving validator helper is `nearestNameDistance`
so one concept has one word across the seam; `rawClause`; the attribution window's javadoc names its
coordinate system; `Locale.ROOT` in the test's premise assertions.

### Not claimed

Phase 2 did **not** converge by the skill's stopping rule — its second round found substantive items, so a
third would be required. I stopped instead, under the standing rule that when a fact-check pass finds new
falsehoods in the previous pass's own corrections, the remedy is to stop writing prose rather than write
more. Code and tests have been stable since `PROSE_TRAILING_LETTERS`; only prose was still moving. Both
surviving new paragraphs state measurements I took through the production methods, not reasoning.

### Numbers

`main` (`2f86eeef`) measured in its own checkout: **1194 api + 75 omod**. This branch at `b6df4756`:
**1200 + 75**, 0 failures, 53 skipped. Live verification was not possible — the developer standalone
returns HTTP 500 on every page and its directory denies reads and overwrites of pre-existing files, so it
can be neither probed nor redeployed. No claim here rests on a live observation.

Issue #208, opened 2026-08-09
The dose warning quotes a sibling row's ceiling, and the injected record lists contraindications the patient cannot have


dkayiwa:
Two findings from #205, both about content the module states rather than about which chip it raises. Filed together because both need a **decision** rather than a patch, and both are about the same question: what may a warning assert that the patient's own record does not support?

## 1. The dose warning quotes the wrong row's ceiling

`DrugSafetyValidator.addOverdose` names the **subject row** but reads the age band of the row **the answer's wording attributed the dose to**. So a chip can read

```
The stated Amoxicillin (suspension) dose ~4000 mg/day exceeds the 3000 mg/day maximum
```

where `3000` is the **sibling row's** published ceiling, not the named row's.

**Measured** through the real `validate` — `OrderedSubjectRowTest.everyArmOfOneResponseNamesOneSubstanceOneWay` asserts the current behaviour and records why. Hand-authored sources only, since it needs both `substanceName` and `ageBands`.

**Why it is a decision, not a patch:** preferring the subject row's own band would **drop the warning entirely** when that row publishes none. So the choice is between a warning with a possibly-mismatched number and no warning at all, and given that under-warning is the failure mode this work exists to prevent, that is not obviously the safer direction. A third option is to name the row the ceiling came from.

## 2. The injected record lists contraindications the patient cannot have

`render` always lists **every** contraindication rule an entry carries, while the chips list only the matched subset.

**Measured live**: Dorothy Odinga's injected record listed `active gastrointestinal bleeding` although she has no such condition recorded and no chip was raised for it.

**Why it is a decision:** for *reference material* that is arguably correct — a drug's contraindications are properties of the drug, not of the patient, and a clinician reading a reference entry expects the full list. But the record is injected into the prompt as citable evidence, and a model reading it can report a contraindication that does not apply to this patient. #110 made every chip a citable record precisely so findings would be grounded; this is the converse risk in the same mechanism.

The options are roughly: leave it and accept that reference text describes the drug rather than the patient; filter to the matched subset, losing the reference character; or keep the full list but mark which entries matched, so the model has the distinction available.

## Related

Both sit on the axis #205 reports as still open — *which published number or clause is quoted* — alongside `matchingEntries` still injecting the canonical row only while the chips read the resolved rows (#163's residue: 80 of 121 multi-row substances have a partner only a sibling carries, 2,627 in total).

## Provenance

[...]

Issue #194, opened 2026-08-08
The interaction chip's subject is the substance's canonical row, not the row the patient is actually ordered


dkayiwa:
## What

The interaction arm names its subject by `DrugReference.canonicalRow()` — the row that names the substance without a route qualifier — rather than by the row corresponding to the patient's actual order. So a patient ordered one presentation can be told about another.

**Live-measured**: Susan Young's `Botulinum toxin type A` order is subjected on `Daxibotulinumtoxina (botulinum toxin type a)`.

That is the same string #176 reports, which is why it needs stating plainly: **#192 does not fix this case.** #176's defect was in `lookupByToken`'s resolution of a *recorded allergen name*; this one arrives through `canonicalRow`'s choice of representative for an *order*, a different path with a different fix.

## Also not fixed by the adjacent work

Checked explicitly: #188's `12f3dbb` ("the collapse must keep the row the patient is on") is about the **injector's note collapse**, not the chip subject, so it does not cover this.

## Shape of the fix

Break `canonicalRow`'s tie by name-match strength against the order's own name — the ranking #192 introduced for allergen resolution, applied to the order side. The constraint to respect is the one #187 settled: naming the row the *chart* records is what makes a finding truthful, and #192 measured that using `canonicalRow` for a charted `Ketorolac (ophthalmic)` allergy renames it wrongly and regresses #187.

So the rule wants to be: prefer the row the patient's own record names; fall back to the canonical row only when nothing matches more strongly.

## Related

Last remaining member of the "which row represents this substance here" family, alongside **#189** (which row's *rule* is quoted — the one-substance-one-rule axis) and **#190**. All three are now about *representative selection* rather than about counting, which is the axis #145/#160/#173/#182/#187/#188 closed.

## Provenance

Live-measured by #192's fix pass, which also verified that #188's commit does not cover it. Relayed; I have not re-run the probe.

PR #274, opened 2026-08-14
fix(drug-reference): the injected record says which row of a substance it describes, and what the other rows publish (#237, #259)

### LINKED ISSUE #237: The injected reference record names the canonical row while the chip names the charted one
## What

The injected `drug_reference` record renders `canonicalRow` alone, while the chips name the row the **chart** records. So the record a clinician can cite and the chip beside it can call one substance two different things.

## Measured

By #206's fix pass, driving real `validate` per multi-row family over the shipped KB: **88 of 101 probed families** change subject when the chart names a non-canonical row. #206 fixed the chips' side of this and narrows the no-chart case; the record's side is untouched.

This is [#163](https://github.com/openmrs/openmrs-module-chartsearchai/issues/163)'s residue and one of the two axes [#205](https://github.com/openmrs/openmrs-module-chartsearchai/issues/205) reported as still open — *which rows the module reads at all*, as distinct from which one a chip speaks for.

## Why it is not simply "make the record use the chip's row"

The `#187` constraint cuts the other way for reference material. A record describes the **drug**; a chip describes **this patient's** finding. Renaming the injected record to the charted presentation would make the reference entry claim to be about `Ketorolac (ophthalmic)` when it is the substance's entry — which is the rename #187 established as wrong, in the place where it is actually wrong.

So the options are roughly: leave it and state the asymmetry; render the record under the substance's name with the charted row noted; or inject the charted row's own record (which re-opens #163's "80 of 121 multi-row substances have a partner only a sibling carries").

## Provenance

Measured by #206's fix pass through real `validate` over the shipped KB. Relayed; I have not re-derived the 88 of 101.

### LINKED ISSUE #259: The injected record and the dose chip can state two different ceilings for one substance in one response
## What

For one substance in one response, the **injected reference record** and the **dose chip** can state two different daily ceilings.

Measured through the real `DrugReferenceInjector.injectRecords` plus the real `DrugSafetyValidator.validate`:

```
record:  …Amoxicillin… maximum 3000 mg/day
chip:    The stated Amoxicillin (suspension) dose ~2500 mg/day exceeds the 2000 mg/day maximum
```

Both numbers are true **of their own row**. The record renders `canonicalRow` ([#163](https://github.com/openmrs/openmrs-module-chartsearchai/issues/163)); the chip names the row the chart records ([#206](https://github.com/openmrs/openmrs-module-chartsearchai/issues/206)). They are answering the same clinical question from different rows.

## Why this pairing is new

Before [#245](https://github.com/openmrs/openmrs-module-chartsearchai/issues/245) there was **no chip** in this situation — the stricter ceiling was unreachable, so the record's laxer number stood alone and unchallenged. #245 makes the warning fire, which is correct and is the point of that fix. The consequence is that the model now receives the **laxer** number as citable evidence sitting beside the **stricter** warning, and can cite it.

That is the same shape as [#237](https://github.com/openmrs/openmrs-module-chartsearchai/issues/237) (record names the canonical row, chip names the charted one) reaching a **numeric** claim rather than a name — and a number is worse, because a clinician reading "the maximum is 3000" from the cited record has no way to see that the warning beside it used 2000.

## Reachability

Hand-authored sources only: it needs both `substanceName` and `ageBands`, and `DdiDrugReferenceSource` has **no code that sets `ageBands`** — a parser property, so no ddinter file including the shipped 19 MB KB can produce a dose warning at all. The bundled curated seed has 4 entries, 4 banded, **0 with `substanceName`**, so it has no multi-row families either.

So: not reachable on anything shipped today, and reachable for any deployment authoring a curated file with per-presentation dosing — which is the normal reason to author one.

**Nothing tests it.**

## Shape of the fix

Probably the same decision as #237 rather than an independent one: either the injected record carries the substance's full band set rather than one row's, or it names which row its number belongs to (the vocabulary [#244](https://github.com/openmrs/openmrs-module-chartsearchai/issues/244) added for the chip), or the record is rendered for the charted row. Worth deciding once, for names and numbers together.

## Provenance

Measured by #245's fix pass through the real injector and validator. Relayed; I have not re-derived it.

dkayiwa:
## What

An injected `drug_reference` record is rendered from **one row** of a substance. It now says which one,
and it states what the substance's other rows publish:

```
Drug reference — Amoxicillin (ATC J01CA04). Published by this dataset for Amoxicillin, not for
Amoxicillin (suspension) — the row this patient's record names, filed separately for the same
substance. Dosing for ages 0-120: 15-30 mg/kg per dose, maximum 3000 mg/day. Also published for other
rows of this substance: Amoxicillin (suspension) 15-30 mg/kg per dose, maximum 2000 mg/day (ages 0-120).
```

Two sentences, for the two halves of one defect. `DrugReferenceInjector.collect` picks the rendered row
with `DrugReference.canonicalRow` — a fold over the dataset that cannot see the chart — while every chip
picks its subject with `DrugSafetyValidator.interactionSubject`, which ranks the patient's own recorded
drug names first (#187, #194, #206). So wherever the chart names a non-canonical row, one response called
one substance two things, in two citable records:

```
[4] Drug reference — Dexamethasone (ATC …)
[5] Safety finding — Dexamethasone (ophthalmic): … interacts with active order phenytoin
```

**#259 is the same split reaching a number**, which is worse: the record renders the canonical row's age
band, so a clinician reading the cited record sees a ceiling the warning beside it did not use.

```
record: Drug reference — Amoxicillin (ATC J01CA04). Dosing for ages 0-120: …, maximum 3000 mg/day.
chip:   The stated Amoxicillin (suspension) dose ~2500 mg/day exceeds the 2000 mg/day maximum …
```

## Why a name and a number needed different fixes

Naming the row settles whose 3000 that is. It does **not** put the chip's 2000 into the citable evidence,
so a model can still cite the laxer number and a clinician still has no route to the stricter one. That
is why this PR carries a second sentence rather than one: an earlier revision of this description claimed
`Fixes #259` on the strength of the clause alone, and that was over-claimed by exactly this much.

**The asymmetry is the defect, not either surface's choice of row.** `DrugSafetyValidator.addOverdose`
already reads the band of the subject row *and of every sibling*, and `anyActionableBand` lets it act when
the rendered row publishes none — so the chip arm reasons over the whole substance while the record read
one row of it. The record is therefore given the row set the chips already fold. Two shapes reach it, and
the second no wording fix can reach: the rendered row publishes **no band for this patient** while a
sibling does, so the record carried no number at all beside a chip quoting one.

## What was deliberately NOT done

**Rendering the record from the charted row** — #237 lists it as an option. Measured and declined: drawing
the patient's interaction partner from the canonical row, a record rendered from the charted row fails to
name that partner in **74 of 129** families, against **0** the other way. The route-unspecified row is the
one carrying the breadth, which is why `canonicalRow` was right for #163 and still is. Swapping the row
would trade a naming fix for a coverage loss and move every citation's `resourceId` — pinned by
`theAttributedRecordStillCitesTheRowItRenders`.

So no row's turn to be rendered changes.

## The guards are deliberately different, and that is the subtle part

The clause is a CONTRAST and shares `DrugSafetyValidator.worthNamingApart` with #244's
`ceilingAttribution` rather than restating it — "for X, not for X" is a contradiction shown to a
clinician, so two rows folding to one display name silence it.

The section is an ENUMERATION under a lead that already says these are other rows, so an item whose name
folds to the record's own still says something actionable — a different number — and silence there would
keep the defect in the one dataset shape where the chip's own attribution is *also* silent, i.e. where the
record is the reader's only route to the number. Asking one question with the other's predicate is the
conflation CLAUDE.md's ATC bullet forbids, one feature along. Pinned by
`twoRowsPublishingOneNameStillStateTheirOwnCeilings` (clause silent, section speaks); mutation reddens it.

## The boundary this crosses, found by mutation

The model is told to cite these records, so a ceiling in one can arrive back in the **answer** that
`DrugSafetyValidator` parses for a prescribed dose. Its `LIMIT_CUE` reads a number as a ceiling only when
a cue precedes it, so the `"maximum N mg/day"` word order is load-bearing outside the injector — and
**newly** so. Dose attribution is clause-scoped and alias-anchored, and the rendered row's own dosing
sentence carries no drug name, so its ceiling is attributed to nobody however it is worded. **An item of
this section is the first sentence the injector composes that pairs a row's NAME with a published ceiling
inside one clause.**

Measured, not argued: mutating `maximum` out of `LIMIT_CUE` left the whole file **green**, because an item
always pairs a row with its own ceiling — the figure can exceed something only where the SUBJECT row is a
different row publishing a stricter one, and every fixture pair had the route-unspecified row laxer. With
a pair shaped the other way (`Nitrofurantoin` 100 / modified-release 200) the same mutation produces

```
The stated Nitrofurantoin dose ~200 mg/day exceeds the 100 mg/day maximum for ages 0-120
```

from a record nobody prescribed anything from. `aRecitedRecordIsReadAsCeilingsAndNotAsADose` feeds the
record back through the real `validate` and pins it.

## Reachability, re-derived (not relayed)

`DdiDrugReferenceSource` **does** set `substanceName` (`ref.setSubstanceName(row.rxnormName)`) and never
sets `ageBands`. So the two halves have **different** bounds:

| | needs | shipped 19 MB KB reaches it? |
|---|---|---|
| the clause (#237, a name) | a multi-row substance | **yes** |
| the section (#259, a number) | a multi-row substance **and** `ageBands` | no — hand-authored sources only |

Verified in this branch: the bundled curated seed is 4 entries with **0** `substanceName` (so every
substance in it is one row); the bundled DDInter sample is 16 drugs with **0** multi-row `rxnorm_name`
families. Both sentences are therefore absent — byte-for-byte unchanged output — on every bundled dataset,
and `drugReference.enabled` defaults to `false` besides.

Measured over the shipped KB by driving the **real** `injectRecords` and the **real** `validate`: 2283
entries, 2086 substances, **129 multi-row**. Of those, 104 could be posed with both a record and a chip and
**all 104 named the substance differently** (0 agreements; 25 unusable). An earlier pass relayed "88 of
101" from a different method (driving `validate` per family); both figures are reported rather than one
silently adopted.

## Prompt cost (#229)

The clause: measured through the real `injectRecords` over the shipped KB, worst case (every probed family
has the chart naming a qualified row) **117 records attributed, 16,605 clause characters against a 194,185
character reference slice — +8.55%**, mean 141 chars, max 206.

The section: one lead, then per named row its name, its numbers and its own age band — one full stop for
the section, not one per row. Asserted as a cost **model** off the rendered text by
`theSectionCostsItsLeadAndOneItemPerRowItNames`. It is silent wherever the substance's rows agree on the
numbers, so a family filed as N near-identical rows pays nothing.

The interactions budget is untouched: `MAX_INTERACTION_RENDER_CHARS` is spent against a counter local to
the interaction notes, not against the record so far, so this text cannot displace an interaction partner.

## Chips

**Zero chips move.** `matchingEntries`' output feeds only the record loop; `preAnswerFindings` calls
`validate` independently. Asserted as a whole chip set by `noChipMoves`.

## Silence, and where it is bounded

Neither sentence is emitted unless the dataset declared the rows one substance (`substanceKey()` non-null
— the `getId()` fallback groups rows a file never called one substance). The clause is additionally silent
where the response names the substance by the very row being rendered, where the two rows fold to one
display name, and where there is no context. The section is additionally silent where no band matches the
patient's age (its own row's or a sibling's — the record's existing age scoping) and where the numbers
would restate the rendered row's.

**Neither is gated on the `drugSafety.*` switches**, which is the opposite of the #208 reading rendered
beside them and so is stated rather than left looking like an oversight: that reading ADDS a patient claim
and stands down with the chips, while these NARROW a claim the record makes anyway — gating a correction
ships the uncorrected sentence whenever the switch is off, i.e. #259 reachable by configuration. With the
whole chip pass off the section is the *only* place the stricter ceiling appears at all.
`ReferenceRecordRowAttributionToggleContextTest` pins both switches for both sentences, using the #208
reading standing down in the same record as the witness that the toggle really moved; gating the section
reddens both cases.

## Tests

**Full reactor measured in a clean clone at each SHA:** `main` @ `c39cc524` = **1200 api + 75 omod =
1275**; this branch = **1230 api + 75 omod = 1305**. 0 failures, 53 skipped, both.

Every case drives the real `injectRecords` — and, where a chip is the other half of the claim, the real
`validate` — over fixtures parsed by the real production parsers.

**Two mutation sweeps, one build per row.** The clause's eight are in the comment below. The section's ten:

| # | mutation | verdict |
|---|---|---|
| 1 | the same-numbers skip removed | **CAUGHT** |
| 2 | the `substanceKey()` guard removed | **SURVIVED first** — see below — **CAUGHT** after |
| 3 | items named by `displayLabel()` | **CAUGHT** (incl. the clause's own case) |
| 4 | `worthNamingApart` applied as a guard | **CAUGHT** |
| 5 | the section rendered before the dosing sentence | **CAUGHT** |
| 6 | the patient's age ignored for a sibling band | **CAUGHT** |
| 7 | the section built from the INJECTED rows | **CAUGHT** |
| 8 | the `row == ref` skip removed | **SURVIVED** — provably; documented, not left looking tested |
| 9 | `dosingNumbers` drops the daily maximum | **CAUGHT** (10, one pre-existing — the extraction is really shared) |
| 10 | the lead repeated per item | **CAUGHT** |

Plus two outside the injector: mutating `maximum` out of `LIMIT_CUE` (above), and gating the section on a
`drugSafety` switch — both caught by the cases written for them.

### Three blind checks found, two of them pre-existing in this branch

- **A guard that could not fail.** Mutation 2 reddened nothing, and the reason was the fixture: both rows
  of `drug-reference-duplicate-id-rows.json` published only `trimethoprim`, the ranked accessors return
  one row per name, so BOTH legs resolved the same row — the group was a single row, `interactionSubject`
  took its `rows.size() == 1` short cut, and the case was silenced by "the chart named no row in
  particular" without either feature's `substanceKey()` guard ever being reached. That was true of
  **this branch's own `rowsGroupedOnlyByASharedIdAreAttributedToNobody`** too, which had therefore never
  exercised the guard it is named for. Fixed by giving the second row its own alias and asserting the two
  accessors `matchingEntries` unions as preconditions in both files.
- **A cost assertion that could not fail** — it summed a string the case had itself assembled from the
  same pieces. Now measured off the rendered text; mutation 10 reddens it.
- **A drift guard duplicated.** `ATTRIBUTION_LEAD` exists precisely so a positive expectation and a
  silence guard cannot come apart, and this branch then wrote a second copy in a second file, where five
  `assertFalse` guards could pass against a string production no longer emits — the same failure its own
  javadoc records having shipped once. Both leads now have one definition in `DrugReferenceTestSupport`.

### Tests changed, stated rather than buried

CLAUDE.md forbids editing tests to make them pass. Three edits are spec changes or strengthenings, and
none weakens an assertion: `theRecordAttributesTheCeilingTheChipDidNotUse`'s whole-record `assertEquals`
now includes the new sentence (strictly more asserted, exact equality kept);
`rowsGroupedOnlyByASharedIdAreAttributedToNobody` gains the preconditions that let it fail; and
`chartNaming` moved to `DrugReferenceTestSupport.contextNaming` rather than being copied into a second
file.

## Residue, and what is not verified

- **A row only the ANSWER puts in play** can supply a chip's ceiling that no pre-answer record could carry
  — the record is written before the answer exists. Every shape reachable from the pass that writes the
  record is closed; the bound is the one `SubstanceSubjects` already records, and it is stated on
  `otherRowDosing` rather than left to be discovered.
- **`chartAnchoredSubject` compares against the fold over the SUBJECT group** while the row the record
  renders is the fold over the narrower INJECTED set. Where those differ the two surfaces can still
  diverge with nothing saying so; closing it needs a differently worded clause, and that wording has no
  measurement behind it (#244's was settled against measured alternatives). The test says explicitly that
  it pins the silence rather than blessing the divergence.
- **Now verified live** — see the section below. The earlier "standalone down and unrecoverable" note
  was session-specific: it boots, and the `/ws/rest/v1/session` 500 that produced that reading is
  transient to a particular boot (it recurred once during these runs and cleared on the next restart,
  with `/chartsearchai/search` working throughout, on both builds, and independent of the `log.level` GP).
- **The added prose is now measured at answer level, and it does not move the model's number.** See
  below; the honest reading is that #259's harm is narrowed rather than eliminated at the answer layer.

## Measured live on the 3.7.1 standalone

Model `gemma-4-E4B-it-Q4_K_M` (Metal, `-ngl 99`), arms deployed one at a time and each verified by
`.moduleLastModified` against the omod's mtime (SNAPSHOT versions all read `1.0.0-SNAPSHOT`, so the
module version proves nothing). Arm A is a fresh clean-main build at `c39cc524`, never the omod that
happened to be deployed.

### 1. The shipped `ddinter` config: byte-identical

`eval/drift-metric/capture_probe_safety.sh`, 4 patients × 5 drugs, real 19 MB KB loaded (`entryCount`
2283, `origin: appdata:…/ddi_knowledge_base.json`, `inert: false`):

| | A/A control (branch twice) | A/B (clean main vs branch) |
|---|---|---|
| identical answers | **20/20** | **20/20** |
| identical chip counts | 20/20 | 20/20 |
| FLIP lines | 0 | 0 |
| verdict-led / abstained (7 ANSWER cells) | 6 / 1 both runs | A=6/1 B=6/1 |
| abstention held (13 ABSTAIN cells) | 12 both runs | A=12 B=12 |
| unlicensed verdicts | 0 | A=0 B=0 |

The A/A control is what makes that readable: today's noise floor on this host is **zero** (greedy decode
is deterministic here), so "identical" is a real result rather than a coincidence inside a noise band.
`score_probe_safety.py` exits **3** on every run — not a regression signal, and identical in both arms:
betty's drug-order REST query returns HTTP 400 on this demo database, so 5 cells are labelled on chips
alone and the scorer refuses to call that a clean read.

**Why identical is the expected answer, measured rather than assumed.** Driving the production
predicates over the real KB: 2283 entries, 2086 substances, **129 multi-row**; and every substance this
probe's cells put in play — erythromycin, clarithromycin, aspirin(→`Acetylsalicylic acid`), warfarin,
and the charted simvastatin / aspirin / lisinopril — is **single-row with 0 age bands**. So neither
sentence can appear on any cell of this probe, and this arm measures that the change is inert where it
is unreachable, which is worth having but is not a test of the prose.

### 2. A per-presentation config, where the prose is reachable

A 3-entry curated file (`Acetylsalicylic acid` 4000 mg/day; `Acetylsalicylic acid (81 mg dispersible)`
100 mg/day; `Warfarin`), i.e. the deployment shape #259 exists for. Patient agnes is charted on
`Aspirin 81mg`. Same question on both arms, deterministic across reruns:

| | clean main | this branch |
|---|---|---|
| injected `drug_reference` chars | 296 | **439** |
| answer | *"The maximum daily dose of aspirin is 4000 mg/day [113]."* | *"The maximum daily dose of **Acetylsalicylic acid for ages 0-120** is 4000 mg/day [113]."* |
| cited row | `acetylsalicylic-acid` | `acetylsalicylic-acid` (unchanged, as designed) |

The +143 characters are the section **exactly**, to the character, matching the cost model
`theSectionCostsItsLeadAndOneItemPerRowItNames` asserts. The 181-character clause did **not** fire, and
correctly: both rows publish the alias `aspirin`, so no recorded name out-claimed the other and
`chartAnchoredSubject` answered null — its guard, working live.

**What that means, stated plainly.** The stricter 100 mg/day is now in the citable record, and the
answer stops calling 4000 "the maximum daily dose of aspirin" and attributes it to the row and band it
belongs to. But this model still **leads with 4000** and does not surface the 100 for the presentation
the patient is actually on. So the fix removes the ambiguity in the evidence and in the attribution; it
does not, on this model and this question, change which number the clinician is told. Anyone reading
#259 as "the answer will now quote the stricter ceiling" should read this table instead.

Fixes #237
Fixes #259

https://claude.ai/code/session_01FvdoGMtwEvtRhchaFRuGQt


dkayiwa:
## Mutation sweep — one full build per row

Throwaway tree, `.git` removed, path-limited commits only. Green baseline at the time of each run.

| # | mutation | verdict |
|---|---|---|
| M1 | `subject := canonicalRow(rows)` — the clause can never fire | **CAUGHT**, 4 failures |
| M2 | `worthNamingApart` stops comparing the two labels | **CAUGHT**, 5 failures — including #244's own `DoseCeilingAttributionTest` x3, so the shared guard really is shared |
| M3 | clause appended at the END rather than before the content | **SURVIVED first**, then **CAUGHT** after tightening `contains` → `startsWith` / whole-record `assertEquals` |
| M4 | `matchingEntries` handed `rawContext` | **SURVIVED** — and the mutation is a no-op, which falsified a comment I had written (see below) |
| M5 | the record renders the SUBJECT row (the declined option) | **CAUGHT**, 6 failures, including the citation-id case (`DDInter513` → `DDInter515`) |
| M6 | both silence guards removed from `rowAttribution` | **CAUGHT**, 8 failures — all seven negative guards fired |
| M7 | `chartAnchoredSubject` stops asking whether the FOLD chose the subject | **CAUGHT**, exactly 1 failure — the case written for it |
| M8 | subject group reverts to the injected rows | **CAUGHT**, exactly 1 failure — the case written for it |

### Three things the sweep found in my own work

**M2 → a real defect.** `rowAttribution` first used `DrugReference.displayLabel()`. That method's
javadoc says verbatim: *"Never used in prompt text — record rendering keeps `getName()` — so this is a
chip-display concern only."* `DrugSafetyChipLabelTest.displayLabelNeverLeaksIntoTheRenderedRecordText`
pins that for the rest of the record but **cannot reach this clause** — its aspirin entry is a one-row
substance, so nothing is ever attributed — and every fixture in my own new file happened to use rows
whose two vocabularies coincide. Fixed to `getName()`; `theClauseNamesTheRowsTheRecordDoesAndNotTheirChipLabels`
pins it on two `Acetylsalicylic acid` rows carrying `genericName: aspirin`.

**M3 → an assertion that could not fail.** The two headline cases asserted `contains`, which a clause
appended *after* the interactions satisfies — the one position where it cannot do its job.

**M4 → a false comment.** I had written that passing `rawContext` "would let the record and the chips
rank the same rows off two different name sets". `withReferenceNames` copies `activeDrugNames` through
untouched and only adds `activeDrugReferenceNames`; the subject ranking reads the former. The mutation
is a no-op. The comment now says what is true and says that it was measured.

### A blind check I created, then closed

Hardening changed the clause's opening from `"Published for"` to `"Published by this dataset for"` (a
bare "published for X, not for Y" reads as a licensing claim — "indicated for X" — the opposite of what
the sentence means). The edit updated the positive expectations, which went red, and left the seven
`assertFalse(record.contains("Published for"))` silence guards alone. `"Published for"` is then a string
production never emits, so **all seven passed unconditionally**. Closed by tying every positive
expectation and every negative guard to one `ATTRIBUTION_LEAD` constant; M6 confirms all seven can now
fail.

## Test counts

Full reactor `mvn -o clean install`, measured in a clean clone at each SHA:

| | api | omod | total |
|---|---|---|---|
| `main` @ `c39cc524` | 1200 | 75 | 1275 |
| this branch @ `539da72d` | 1217 | 75 | **1292** |

0 failures, 53 skipped, both.

dkayiwa:
## Hardening: what two independent review passes found, and the one thing left open

Both passes were read-only reviews of the working tree, briefed to hunt for concrete defects rather
than style. Between them and my own passes, eleven substantive items were found and fixed. Three are
worth surfacing here because they are decisions rather than tidying, and one is **not fixed**.

### Fixed

- **The clause was gated on `injectFromQuery`/`injectFromOrders` and on the relevance rule**, because
  the subject group was built from the rows the injector decides to *inject*. With `injectFromOrders`
  off — or with an order the relevance rule discards — the group loses the charted row,
  `interactionSubject` takes its `rows.size() == 1` short cut, and the record falls silent in exactly
  the case #237/#259 are about. That is the "narrowed row group" hazard `interactionSubject`'s own
  javadoc names, and it made the clause configuration-dependent. The group is now
  `questionDrugs ∪ orderEntries`, ungated. Pinned by
  `theSubjectIsChosenAmongEveryRowThePassResolvedNotOnlyTheInjectedOnes`; mutation M8 reddens it.
- **Widening it let the FOLD move the subject on its own**, and "the row this patient's record names"
  is then a claim about a chart that said no such thing. `chartAnchoredSubject` asks `canonicalRow`
  separately and answers null where the two agree, so the sentence is printed only where a recorded
  name actually out-claimed. Pinned by `aSubjectTheFoldMovedRatherThanTheChartIsAttributedToNobody`;
  mutation M7 reddens it and nothing else.
- **`rowAttribution` used `displayLabel()`**, whose javadoc says verbatim *"Never used in prompt text
  — record rendering keeps `getName()`"*. Found by mutation, not by reading.

### Open — a known residue, recorded rather than shipped quietly

`chartAnchoredSubject` compares the subject against the fold over the **subject** group, while the row
the record renders is the fold over the narrower **injected** set. Where those differ the two surfaces
can still diverge with nothing saying so: a question resolving only a route-qualified row renders that
row's record, while the chip layer — folding the same union — names the substance by the unqualified
row. That is #237's shape surviving in the one case this guard deliberately stays quiet about.

Closing it needs a **second, differently worded clause**: the existing sentence would be *false* there,
because no recorded name chose the row. That wording has no measurement behind it, and this module does
not invent clinician-facing vocabulary on a guess — #244's wording was settled against measured
alternatives. So it is written into `chartAnchoredSubject`'s javadoc and the test says explicitly that
it pins the silence rather than blessing the divergence.

### Convergence not claimed

The last review pass found substantive issues, so by the harden skill's own rule it cannot be the last
pass, and no subsequent clean pass has been run against the fixed tree. Phase 2 was not entered.

Issue #280, opened 2026-08-18
A prescribed drug the patient is allergic to needs an alerting surface, not a chip on every answer


dkayiwa:
## What changed, and what it leaves uncovered

The active-order contraindication arm (#143) used to run on every question: it walked every active order against every recorded allergy and condition whatever the question and the answer named. Measured live on the 3.7.1 standalone against patient `a7090f70` (Betty Williams — an active Lidocaine order and a recorded Lidocaine allergy), four unrelated questions returned the **same two contraindication chips, byte for byte**:

| Question | Chips |
|---|---|
| any allergies? | Bupivacaine, Lidocaine |
| are there any drug interactions with her current medications? | Bupivacaine, Lidocaine (+ the Major interaction) |
| does she have cancer? | Bupivacaine, Lidocaine |
| what is her date of birth? | Bupivacaine, Lidocaine |

That arm is now scoped to what the response is about — a chip is raised where either side of it, the drug or the recorded finding, is named by the question, the answer or a cited record, with a medication- or allergy-domain question keeping the corresponding list in scope wholesale.

The reasoning is that chartsearchai answers questions. Its README opens "lets clinicians ask natural language questions about a patient's chart and get answers with source citations", and `DrugSafetyValidator`'s contract is a check that runs after the answer and *annotates* it. An unconditional finding is an alert, and this module has none of an alerting system's machinery — no subscription, no acknowledgement state, no delivery path that opens unprompted. Repeating the identical chip on every answer, including one about a date of birth, is the shape that trains a clinician to stop reading the box.

## What still needs a home

The finding itself is real and worth acting on: **she is actively prescribed the drug she is recorded as allergic to**. Scoping the arm means a clinician who never asks a drug-shaped question is no longer told. That is the right trade for a Q&A response and the wrong outcome overall, so it needs a surface that suits it. Options, roughly in order of how well they fit:

1. **Order-entry checking** — the conventional home. The conflict is created at prescribing time and that is when it is cheapest to fix.
2. **A patient-header banner in `openmrs-esm-chartsearchai`** — persistent, acknowledgeable, and not tied to whether anyone ran a search.
3. **A standing-alerts endpoint on this module** (e.g. `GET /chartsearchai/chartalerts?patient=<uuid>`), returning the same deterministic findings the chips are computed from, for a client to render outside the answer thread. Cheapest to build here, since `DrugSafetyValidator` already computes them; it just should not be the `/search` response that carries them.

[...]

PR #281, opened 2026-08-18
fix(drug-safety): an order-driven contraindication chip is raised only where the response is about it

dkayiwa:
## The defect

The active-order contraindication arm ([#143](https://github.com/openmrs/openmrs-module-chartsearchai/issues/143)) ran on **every** question — it walked every active order against every recorded allergy and condition whatever the question and the answer named. Measured live on the 3.7.1 standalone against patient `a7090f70` (an active Lidocaine order beside a recorded Lidocaine allergy), four different questions returned the same two chips **byte for byte** — the finding is the *invariance*, not that all four were off-topic, since two of them were squarely about drugs:

| Question | Chips returned |
|---|---|
| any allergies? | Bupivacaine, Lidocaine |
| are there any drug interactions with her current medications? | Bupivacaine, Lidocaine (+ the Major interaction) |
| **does she have cancer?** | Bupivacaine, Lidocaine |
| **what is her date of birth?** | Bupivacaine, Lidocaine |

This is a category error rather than a tuning choice. chartsearchai answers questions — the README opens *"lets clinicians ask natural language questions about a patient's chart and get answers with source citations"*, and `DrugSafetyValidator`'s own contract is a check that runs after the answer and **annotates** it. An unconditional finding is an alert, and this module has none of an alerting system's machinery: no subscription, no acknowledgement state, no delivery path that opens unprompted. Repeated identically on every answer, down to a date-of-birth query, it is the shape that trains a clinician to stop reading the box.

## The change

The arm is scoped to the response's **subject matter** — the question, the answer, and the records the answer cited. A chip is raised where **either side** of it is in that:

- the **drug** — `SubjectMatter.names(DrugReference)`
- the **recorded finding the rule fired on** — `names(Contraindication)`, asked of the rule's own token and never of the patient's whole list, through `PatientClinicalContext.containsToken`, the matcher that decided the rule matched, so *"did not match"* and *"is not what was asked about"* cannot drift

Both sides, because a contraindication relates two things and either can be the subject. Drug-side alone loses a chip whose *allergy* is what was asked about; finding-side alone loses #143's own case — a prescribed drug named only by the cited `drug_order` record that [#105](https://github.com/openmrs/openmrs-module-chartsearchai/issues/105)'s echo rule keeps out of the in-play set.

Two widenings read off the question via `QueryScopeRouter.asksAboutMedications` / `asksAboutAllergies`, reusing `matchedIntents` rather than writing a second drug vocabulary: a question about her medications or her allergies makes that whole **list** the topic even where the prose writes no individual name. They only ever widen — a question naming a drug outright carries no cue word at all (*"Can I give her bupivacaine?"*) and is answered by the drug-in-play arm, which this never touches.

## Verified live

Rebuilt omod deployed to the 3.7.1 standalone (`.moduleLastModified` matched the omod mtime):

| Query | Before | After |
|---|---|---|
| Does she have cancer? | 2 chips | **0** |
| What is her date of birth? | 2 chips | **0** |
| any allergies? | 2 chips | 2 |
| drug interactions with her current medications? | 3 chips | 3 |
| Can I give her bupivacaine? | — | 3 (incl. the order-driven Lidocaine chip) |

Chips and injected records stay in step: `DrugReferenceInjector.preAnswerFindings` calls this same `validate`, so a scoped-out finding is neither chipped nor injected as a `safety_finding` record.

## One test changed, deliberately

`ActiveOrderContraindicationTest.aPrescribedAllergyIsRaisedEvenWhenNeitherQuestionNorAnswerNamesTheDrug` used a blood-pressure question — the same shape as the reported cancer one — and asserted the chip **must** fire. It was the only failure in 1259 api tests. It is rewritten as `aPrescribedAllergyIsNotRaisedWhereTheResponseIsAboutSomethingElse`, asserting 0, carrying the reversal and its reasoning in the test itself rather than only in a commit message. Every other #143 case passes untouched, including the echo core.

**1259 api + 75 omod tests pass.**

## What this deliberately gives up

A prescribing error nobody asks a drug-shaped question about is no longer announced. That is not a safety net this module can honestly carry, and it belongs on a surface with acknowledgement — order entry, a chart banner, CDS hooks. Filed as [#280](https://github.com/openmrs/openmrs-module-chartsearchai/issues/280), which also records that `serializeSafetyWarnings` publishes neither `SafetyWarning.severity` (populated for rated rules, never reaching the wire) nor any field separating a chip about what was asked from a standing chart fact.

---

**Hardened** (second commit): nine documentation sites the change falsified — none of them in its own diff, the load-bearing one being the arm's javadoc paragraph headed *"Not gated on the question, deliberately"* — plus `DrugReferenceInjector`'s claim that the pre-answer and post-answer passes "cannot differ", which the scoping turns from an equality into a subset (prompt findings ⊆ chips, the safe direction, now pinned by a test). Four untested behaviour legs gained tests, three mutation-verified; one of the original tests turned out not to cover the path it was named for. `"allergy".equalsIgnoreCase` was extracted to one predicate. Re-verified live after the harden: identical chip counts on all five queries.


api/src/test/java/org/openmrs/module/chartsearchai/reference/SubjectMatterScopedContraindicationTest.java
```diff
@@ -0,0 +1,308 @@
...
+		List<SafetyWarning> preAnswer = validator().validate("", question, context, null);
+		List<SafetyWarning> chips = validator().validate(
+				"Her only active medication is " + IBUPROFEN_ORDER + " [2].", question, context,
+				chart(TUMOUR_RECORD, ORDER_RECORD).getMappings());
+
+		assertEquals(0, contraindications(preAnswer).size(),
+				"the question alone names neither side here, was: " + preAnswer);
+		assertEquals(1, contraindications(chips).size(),
+				"the answer citing the order record does, was: " + chips);
+		for (SafetyWarning finding : contraindications(preAnswer)) {
```
dkayiwa: The loop below never executes: `contraindications(preAnswer)` is asserted empty four lines up, so the subset comparison iterates nothing and the property this case is named for is only ever checked against the degenerate `{} ⊆ {chip}`. The two `assertEquals` above do real work (the first is one of the four cases in this class that fail on `main`), but a monotonicity break in `SubjectMatter` would not redden this test, and nothing else in the suite compares a pre-answer pass against a chips pass, so the invariant `DrugReferenceInjector`'s javadoc now rests on has no pin. Worth fixing, since the PR description counts it as pinned.

`"What are her current medications?"` gives it teeth, and it happens to exercise the leg most likely to diverge: the drug side satisfied only by the question-derived widening rather than by a cited record. Driven through the real `validate` with the tumour answer, so the answer itself names neither side:

```
pre-answer: [contraindication:Ibuprofen:Ibuprofen is contraindicated by an active allergy: documented ibuprofen allergy]
chips:      [contraindication:Ibuprofen:Ibuprofen is contraindicated by an active allergy: documented ibuprofen allergy]
```

Non-empty on both sides, so the loop then compares something. The current arrangement is worth keeping as a case of its own, since its `0` is the scoping assertion rather than the subset one.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
```diff
@@ -687,14 +703,111 @@ private static List<String> citedRecordTextsLower(String answer, List<RecordMapp
 	 *         what #105 measured and fixed.
 	 */
 	private static boolean isEchoOfCitedRecord(DrugReference ref, List<String> citedTextsLower) {
-		for (String text : citedTextsLower) {
+		return namesAnyOf(citedTextsLower, ref);
```
dkayiwa, line 718: The residue bound in this method's javadoc, at line 690, still reads that an actively-ordered drug "is checked against the patient's allergy and condition records by `addActiveOrderContraindications` whatever the answer's wording". That is the half this change falsifies, and it is the same shape as the claim #143 had to correct in the first place.

The conclusion survives though, and by a tighter argument than the one it replaces. This method's corpus is `citedTextsLower`, and `validate` hands that same list to the `SubjectMatter` constructor, which only ever adds the question and the answer on top of it. So `isEchoOfCitedRecord(ref)` implies `SubjectMatter.names(ref)`: whenever the exemption withholds a drug, the record that triggered the exemption has already put that drug in subject matter, and the arm still checks it. Worth stating that here in place of the wording claim, since it is the exemption's whole justification and it now rests on a containment between two corpora that a later change could quietly break.

api/src/main/java/org/openmrs/module/chartsearchai/reference/SafetyWarning.java
```diff
@@ -42,7 +42,8 @@ public class SafetyWarning {
 	/**
 	 * Contraindication: a drug is contraindicated by an active allergy or condition. Two joins — a drug
 	 * IN PLAY (asked about in the question, or named by the answer on its own authority), and the
-	 * patient's OWN ACTIVE ORDERS whatever the question and the answer name. (Enumerated for the same
+	 * patient's OWN ACTIVE ORDERS, scoped to what the response is about — either the drug or the
+	 * recorded finding must be named by the question, the answer or a cited record. (Enumerated for the same
```
dkayiwa, line 46: Eight lines down, the same comment still says the second join exists for a prescribing error "which no wording of a question or an answer should be able to hide". After this change the wording of the question, the answer and the cited records is precisely what decides, so the paragraph now argues against its own opening sentence.

Three more sites the second commit's sweep left standing:

- `QueryScopeRouter.java:50` promises "Public only so the drug-safety layer can reuse `isInteractionScreening` ... Every other member stays package-private", and this PR adds two more public members.
- `ContraindicationToggleContextTest.java:36` says "This arm fires on EVERY question about an allergy-bearing patient on a related drug". Its fixture happens to ask "What are her current medications?", so it still passes, which is itself worth a word: that question string is load-bearing now and reads as incidental.
- `docs/drug-kb-demo.md:422-458` is the operator-facing one, and probably the costliest. It tells the reader that "Every question additionally has this patient's own active orders checked against her own allergies and conditions", that "All three are EXTRA only on a query that names neither drug", and signs off with "no wording of a question should hide it". The paracetamol and naproxen cheat-sheet rows carry no medication or allergy cue and name neither contraindicated order, so those three extra chips are no longer expected on them, and whoever follows the recipe will read the difference as a broken deployment.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
```diff
@@ -3527,18 +3655,81 @@ private List<List<DrugReference>> recordedAllergens(PatientClinicalContext conte
...
+	/**
+	 * @return whether the chart records anything an active order could be contraindicated BY. Extracted
+	 *         from {@link #addActiveOrderContraindications}'s own guard rather than copied to its call
+	 *         site: the call site has to know the same answer to decide whether parsing the cited-record
+	 *         corpus can pay for itself, and two spellings of "this patient records nothing to check
+	 *         against" would drift into a corpus built for an arm that returns, or an arm that runs
+	 *         without one. Both token sets, for the reason the guard's own javadoc gives — the curated
+	 *         arm's condition leg is half of what the original scoping suppressed.
+	 */
+	/**
```
dkayiwa, line 3735: Two javadoc blocks in a row here, so `hasContraindicationRecords`'s block is dropped and the method below it ends up undocumented. `PatientClinicalContext.java:411` has the same slip, and it costs more there: the new visibility note becomes `containsToken`'s only javadoc, and the long original block (the fragment rationale, the #141 diacritic measurement, the "not the allergen's comment or reactions" correction) becomes a dead comment. Confirmed by running `javadoc` over a two-block probe with a positive control on the second block: only the last block reaches the output, the first is dropped entirely.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
```diff
@@ -687,14 +703,111 @@ private static List<String> citedRecordTextsLower(String answer, List<RecordMapp
...
+
+		/**
+		 * Whether the finding a MATCHED rule fired on is what this response is about — asked of the
+		 * rule's own token and never of the patient's whole list, because a response citing one
+		 * condition while a rule fires on another is the reported defect again at token granularity.
+		 * Through {@link PatientClinicalContext#containsToken}, the matcher that decided the rule
+		 * matched at all, so the two cannot drift.
+		 */
+		private boolean names(DrugReference.Contraindication c) {
+			if (coversRecordedAllergies && isAllergyRule(c)) {
```
dkayiwa, line 821: Was leaving conditions out of the widening deliberate? `Intent.CONDITIONS` already exists with its own cues ("conditions", "diagnosis/diagnosed", "problem list"), and the argument in `SubjectMatter`'s javadoc, that a question about her medications or her allergies makes that whole list the topic even where the prose writes no individual name, reads the same for a question about her conditions.

Where it shows is the pre-answer pass, which has no answer text at all: "Does she have any allergies?" promotes a finding into the prompt and "What conditions does she have?" does not, though the ulcer rule is as much the topic in the second as the allergy is in the first. `theFindingSideFiresForADrugTheResponseNeverMentions` picks a conditions-domain question and reaches its chip through the answer, so nothing goes red either way. If the omission is deliberate, the reason belongs in `SubjectMatter`'s javadoc beside the other two.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReferenceInjector.java
```diff
@@ -155,14 +155,20 @@ void setDrugSafetyValidator(DrugSafetyValidator drugSafetyValidator) {
...
 	 * </ul>
-	 * Neither can therefore differ between this pre-answer pass and the post-answer chips pass, which
-	 * is the property that keeps a finding in the prompt from ever being asserted without a chip beside
-	 * the answer. (The #143 arm skips a drug already in play, so a question naming one of the patient's
+	 * The first cannot differ between this pre-answer pass and the post-answer chips pass. The second
+	 * CAN, and only in the safe direction: this pass calls {@code validate} with an EMPTY answer, so
+	 * its subject matter is the question alone, while the chips pass adds the answer and the records it
+	 * cited — a superset of the same texts, with the two question-derived widenings identical either
+	 * way. Every test {@code SubjectMatter} applies is monotone in those texts, so the findings of this
+	 * pass are a SUBSET of the chips beside the answer. That is the direction this property exists for:
```
dkayiwa, line 168: A third path carries this patient's own contraindication findings into the prompt, and it is neither of the two enumerated above: the "Recorded for this patient:" clause that `contraindicationSections` puts on an injected `drug_reference` record. Its own javadoc says a clause is marked "when ANY rule folded into it matched, which is exactly when the ledger raises a chip for that key", and the order leg no longer satisfies that. `matchingEntries` admits an active order that merely shares an ATC level-4 subgroup or a curated group with a question-named drug; such an entry is not in `inPlay`, so its chip now goes through `SubjectMatter` while the clause is still marked off `recordedContraindicationKind`.

I could not reproduce it on anything shipped, which is why this is a question and not a finding: no two entries in the bundled curated file share a level-4 subgroup or the NSAID group, and `ddinter`/`atc` publish no contraindication rules at all, so it takes a deployment-authored file, which `cross-reactivity-groups.json` does invite ("expand per deployment"). Is the record's reading meant to outlive the chip here, and if so, is the "exactly when" sentence over there worth softening?

Line 171 also went unwrapped after the edit, at 154 characters against 112 for the next longest line in the file.

api/src/test/java/org/openmrs/module/chartsearchai/reference/SubjectMatterScopedContraindicationTest.java
dkayiwa: Fixed in 9012c9da. The case now runs on `"What are her current medications?"` with an answer about her tumour, so the pre-answer set is non-empty and the loop has something to compare, and an `assertEquals(1, preAnswer…)` guard sits above it so a later change that empties the pass reddens instead of silently going no-op. Mutation-checked: making `SubjectMatter` non-monotone now fails the loop's own assertion rather than being caught incidentally by `AllergenNameResolutionTest`. The old arrangement survives as `theQuestionAloneCanLeaveThePreAnswerPassWithNothingToPromote`, since its `0` is the scoping assertion and not the subset one.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
dkayiwa: Fixed in 9012c9da. The wording claim is gone and the containment argument is recorded in its place: `validate` hands the same `citedTextsLower` list to the `SubjectMatter` constructor, which only ever adds the question and the answer on top of it, so `isEchoOfCitedRecord(ref)` implies `SubjectMatter.names(ref)`. The javadoc now also says that is what a later change has to preserve, since narrowing what the arm reads to less than the cited records would reopen #143 with nothing going red.

api/src/main/java/org/openmrs/module/chartsearchai/reference/SafetyWarning.java
dkayiwa: Fixed in 9012c9da, all four sites. This paragraph now says the ANSWER's wording alone must not hide the finding, which is the true and narrower claim. `QueryScopeRouter`'s class javadoc states the actual rule for what earns a member `public` instead of a promise this branch breaks. `ContraindicationToggleContextTest` drops "fires on EVERY question" and records that its fixture's question string is load-bearing. `docs/drug-kb-demo.md` had all three blocks rewritten, and the chip counts it now promises were verified against the real validator over the bundled curated dataset with Margaret's chart shape.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
dkayiwa: Fixed in 9012c9da. `hasContraindicationRecords`'s block sits on its own method, and in `PatientClinicalContext` the visibility note is merged into the existing block as a `<p>` rather than displacing it. A repo-wide scan confirms neither site remains. Three others exist but are all pre-existing on `main` and outside this branch, so I left them.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
dkayiwa: Not deliberate. Implemented in 9012c9da rather than documented away, since the argument behind the other two widenings applies unchanged to conditions and `Intent.CONDITIONS` already existed: `QueryScopeRouter.asksAboutConditions` plus `SubjectMatter.coversRecordedConditions`, with `isConditionRule` extracted beside `isAllergyRule`. Written test-first, and a second case pins that the widenings stay per-list so a conditions question cannot reach her allergy records. This is the only runtime behaviour change in the two commits, so it is the one worth a second opinion if you would rather the arm stayed at two widenings.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugReferenceInjector.java
dkayiwa: Both addressed in 9012c9da and 3ef3c4ca. `preAnswerFindings` now names the third channel with its reachability bound, and `contraindicationSections` is corrected from "exactly when the ledger raises a chip" to "when", with the reason it is not gated too: the marking states what the chart records, and scoping it would make the record report an absence it cannot substantiate. The reachability claim is verified, no two entries of the bundled curated file share a level-4 subgroup or the shipped NSAID group, and the `ddinter` and `atc` sources both document that they emit no contraindication rules. Line 171 is rewrapped, the file's longest line is back to 112 from 154.

The harden pass afterwards found the same false claim in production javadoc one file over: `addActiveOrderContraindications` rested on "issue #110 turns every chip into a citable pre-answer record", which is the converse of what its argument needs. It now states the forward direction. Ten further sites still state the converse and I left them alone, since each is about a drug-in-play chip where a drug only the answer names has never had a pre-answer record either, so that looseness predates this change.

Issue #275, opened 2026-08-17
The demo instance is 60 commits stale: deploy has failed since 2026-07-28 on a server-side permission denial, and --no-health-check means nothing said so


dkayiwa:
## What

**The demo instance has not been deployed since 2026-07-28, and the workflow that would have said so cannot fail loudly enough to be noticed.** Two facts, one causally downstream of the other.

**The deploy is broken on the server.** [Run 32069013083](https://github.com/openmrs/openmrs-module-chartsearchai/actions/runs/32069013083/job/95507687401) (`main` @ c39cc524, 2026-08-17 21:02 UTC), step *Deploy via scoped SSH*:

```
# ***:22 SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.16
/usr/local/bin/deploy-compose: line 17: cd: /root/docker/emr-3-ai: Permission denied
##[error]Process completed with exit code 1
```

The keyscan and the login both worked — the banner is in the log, and the error is the server-side wrapper's own `cd`. `deploy` is an unprivileged user; `/root/docker/emr-3-ai` sits under root's home, which is not traversable by anyone else. The script dies before it touches a container, so the failure is inert: nothing was torn down, and whatever was started on 07-28 is presumably still serving.

**The instance is 60 commits behind.** The last run that deployed anything is [30367235516](https://github.com/openmrs/openmrs-module-chartsearchai/actions/runs/30367235516), 2026-07-28T14:12:43Z, on `83cc33e4` ("Add DDInter-backed drug-reference source"). `git log 83cc33e4..origin/main` is **60 commits** — the whole drug-safety chip and injector series from [#86](https://github.com/openmrs/openmrs-module-chartsearchai/issues/86) onward has never run on chartsearchai.openmrs.org.

## This is not a workflow bug

`.github/workflows/deploy.yml` has not changed since 56cd9403 (2026-04-27), which only replaced `/opt/deploy.sh` with `deploy` in the command string. That exact command — `deploy emr-3-ai --no-health-check` — succeeded on every run from 2026-06-14 through 2026-07-28.

The run before this one, [31501343023](https://github.com/openmrs/openmrs-module-chartsearchai/actions/runs/31501343023) (2026-08-11), failed **differently**:

```
sudo: a terminal is required to read the password; either use the -S option ...
sudo: a password is required
```

[...]

PR #282, opened 2026-08-18
fix(deploy): a reset deploy left OpenMRS serving the setup wizard, and querystore sweeping with no embedder

dkayiwa:
## What

`https://chartsearchai.openmrs.org/openmrs` has been serving the **OpenMRS installation wizard** since the `reset` deploy on 2026-08-18. Run [32150623566](https://github.com/openmrs/openmrs-module-chartsearchai/actions/runs/32150623566) at 14:48 UTC is the one that did it — the only one of that day's three that passed `--destroy-volumes`. The two before it ([14:44](https://github.com/openmrs/openmrs-module-chartsearchai/actions/runs/32150176400), [14:47](https://github.com/openmrs/openmrs-module-chartsearchai/actions/runs/32150505463)) logged `Container emr-3-ai-backend-1  Running` — compose only recreates a container when the image digest changes, so they changed nothing.

The host is not inspectable from here (SSH to the deploy box is refused, and Cloudflare answers every external probe with `403 cf-mitigated: challenge`, so origin health cannot be read from outside). So this was diagnosed by pulling the published `nightly-chartsearch` image and booting it on wiped volumes, which reproduces it exactly.

## Two independent failures, both first-boot-after-a-wipe only

**1. The seed and OpenMRS's installer fight over the schema.**

`--destroy-volumes` deletes `openmrs-data`, which is where `openmrs-runtime.properties` lives. `backend-init.sh` then imports the complete 5,284-patient dump into the empty database. OpenMRS boots, finds no runtime properties, and runs the base image's automated installer — `startup.sh` drives it via `-DOPENMRS_INSTALLATION_SCRIPT` plus a `curl` to `/openmrs/` — with `create_tables=true`, taken from `OMRS_CONFIG_CREATE_TABLES`. Against a schema the seed just built:

```
Table 'allergy' already exists [Failed SQL: (1050) CREATE TABLE openmrs.allergy ...]
WARN InitializationFilter$InitializationCompletion — Error while trying to create tables and demo data
```

The install is abandoned and `/openmrs/` redirects to `/openmrs/initialsetup` for the life of the container. The application does come up behind that redirect — modules load, the REST API answers 200 — which is part of why nothing complained. It clears on a restart, because the abandoned install leaves the properties file behind; that is why only the first boot after a wipe is affected and why steady-state checks never see it.

**2. The retrieval wiring only ever existed in the database the wipe deleted.**

`querystore.embedding.modelFilePath` and `vocabFilePath` were a manual operator step — this file's own header called it a pending "follow-up" — while the seed still sets `querystore.bootstrap.autostart=true`. So the next start swept every record of all 5,284 seeded patients against an unconfigured embedder:

```
35,715 × java.lang.IllegalStateException: Required global property is not configured:
         querystore.embedding.modelFilePath          (each with a full stack trace)
  at org.openmrs.module.querystore.embedding.OnnxEmbeddingProvider.resolvePath(OnnxEmbeddingProvider.java:349)
  at org.openmrs.module.querystore.bootstrap.TypeBootstrapper.serializeAndEmbed(TypeBootstrapper.java:266)
```

112MB of stack traces in eight minutes and still climbing — roughly 20GB/day into a host that also ships its logs to Loki. And `chartsearchai.querystore.enabled` came back `false`, so chart search was off regardless.

## The fix

**`backend-init.sh`**
- Decide `create_tables` from the database rather than from the environment. Also covers a wipe of `openmrs-data` alone (db-data kept), where the schema likewise predates the installer.
- Write `openmrs-runtime.properties` when this entrypoint hands OpenMRS a database it prepared, from the credentials the seed just used successfully — so no wizard is involved at all and the first boot after a wipe is indistinguishable from a steady-state one. Narrow: only when the schema is really present and the file is really absent; an existing file is never rewritten, and a genuinely empty database still gets the ordinary first-install path.
- Point the retrieval properties at the files this script provisions, **only where the property is blank**, so a deliberately-set value survives — the same courtesy the seed already extends by snapshotting `chartsearchai%`/`querystore%` across the import.
- Never leave the combination that floods: a bootstrap sweep enabled with no embedder to run it.
- Assert `chartsearchai.querystore.enabled` on the seed path, where the demo's configuration is being established from scratch. The dump carries `false` baked in, so blank-only wiring cannot reach it.

**`Dockerfile.backend` + `docker-compose.yml`** — the healthcheck probed `/openmrs/`, so it was *passed by the very redirect that was the symptom*. It now probes `/openmrs/health/started`, which is 200 only once `Listener.isOpenmrsStarted()` flips, with a `start_period` that covers a genuine first boot (with runtime properties present OpenMRS starts synchronously at context init, so nothing answers until the seed is brought up to the current schema).

**`README.md`** — no longer tells operators to set those properties by hand on the Docker path.

## Verification

Real published image, fresh volumes, measured before and after:

| | before | after |
|---|---|---|
| `/openmrs/` on first boot | `302 → /openmrs/initialsetup` | **200** |
| `"Initial setup is needed"` in the log | present | **0 occurrences** |
| embedder exceptions | 35,715 | **0** |
| log volume in 8 minutes | 112 MB | **260 KB** |
| `chartsearchai.querystore.enabled` | `false` | `true` |
| `querystore.embedding.modelFilePath` | `NULL` | `querystore/model.onnx` |

Also confirmed: idempotent on a second start (seed skipped, properties untouched); an operator-set `modelFilePath=custom/operator-model.onnx` and an explicit `enabled=false` are both left alone; and on the second run the script reads its own generated properties file back through `rtp_get`, which cross-checks the format OpenMRS then parsed successfully.

No tests were modified; no Java changed.

## Not in scope

`deploy.yml` still passes `--no-health-check` on both branches and has no step after the SSH call, so the job still cannot go red for a server that never came up — the second half of #275. A post-deploy gate has to poll from somewhere Cloudflare does not challenge, and an SSH-based probe cannot be tested without host access, so it is deliberately left out of this PR rather than shipped untested.

## Recovering the live instance

This only takes effect in a newly built image. After merge, `build-docker` publishes `nightly-chartsearch`; then run the deploy **with `reset` checked**. That is the path verified end to end here, and it is needed rather than a plain deploy because `chartsearchai.querystore.enabled` is currently an explicit `false` from the dump — the blank-only wiring deliberately will not override it, but the seed path asserts it. A plain deploy would clear the wizard and stop the flood, but leave chart search disabled.


backend-init.sh
```diff
@@ -339,6 +355,16 @@ SQL
...
 
+  # The dump carries chartsearchai.querystore.enabled=false baked in, so the
+  # blank-only wiring below cannot reach it and a re-seeded demo would come back
+  # with retrieval switched off. Asserted here rather than there because this is
+  # where the demo's intended configuration is being established from scratch —
+  # it runs after the operator-GP snapshot is restored, alongside the autostart
+  # line above, and only on the seed path, so an ordinary boot never overrides a
+  # deliberate false.
+  seed_sql "$DB_NAME" -e \
+    "INSERT INTO global_property (property, property_value, uuid) VALUES ('chartsearchai.querystore.enabled','true', UUID()) ON DUPLICATE KEY UPDATE property_value='true';" || true
```
dkayiwa, line 366: This needs to come out before merge: `chartsearchai.querystore.enabled` no longer exists. It was deleted in 46b4eb1e (Phase 5a of #51) along with `ChartSearchAiUtils.isQueryStoreEnabled` and `LlmInferenceService.resolveQueryStoreEnabled`, and that commit's own message gives the reason: "querystore is a required module, so the toggle no longer had a meaning". Nothing at this head reads it, in Java, in `config.xml`, in querystore, or in the reference frontend. The `false` row the reset boot found is real but vestigial: the dump is tagged `5284-2026-05-19` and the GP was removed on 2026-06-17, so the dump predates the removal.

Where that bites is "Recovering the live instance". If merged as-is, an operator follows it and runs the deploy with `reset` checked, i.e. `--destroy-volumes` a second time, on the stated grounds that a plain deploy "would leave chart search disabled". A plain deploy is enough: it recreates the container, so the leftover runtime properties clear the wizard, and `configure_retrieval_gps` fills in `modelFilePath`/`vocabFilePath`, which are the blank properties that actually broke retrieval. The wipe buys nothing and costs another ~8GB of model downloads plus a full re-import of the 5,284-patient dump, which is the operation that caused this outage.

So I would drop this INSERT, the `gp_set_if_blank` for the same property at line 564, its place in the `[retrieval-wiring]` log line and in the file header, and rewrite the README paragraph, which as written contradicts the table three rows above it ("querystore is a required module and is always the retrieval path, there is no toggle to disable it"). Deleting the stale row would be a bonus, since it surfaces under Chartsearchai in **Admin > Settings** looking like a live knob. Not for this PR, but `build-standalone.yml:209` writes it too, so this is a stale belief in three places rather than a typo.

Nothing else in the verification table leans on it: the 35,715 exceptions and the 112MB are the embedder properties.

backend-init.sh
```diff
@@ -347,6 +373,216 @@ SQL
...
+  # '=' inside the value need no escaping; the wizard escapes them only because
+  # Properties.store() always does.
+  ( umask 077; cat > "$_rtp_file" <<EOF
+#Written by backend-init.sh for a database this entrypoint prepared
+connection.driver_class=com.mysql.jdbc.Driver
+connection.url=jdbc:mysql://${DB_HOST}:${_db_port}/${DB_NAME}${_db_args}
+connection.username=${DB_USER}
+connection.password=${DB_PASS}
+auto_update_database=${OMRS_CONFIG_AUTO_UPDATE_DATABASE:-true}
+module.allow_web_admin=${OMRS_CONFIG_MODULE_WEB_ADMIN:-true}
```
dkayiwa, line 512: These two read only the deprecated environment names, and they should read the current ones too before this merges. The base image's `startup-init.sh` treats `OMRS_AUTO_UPDATE_DATABASE` and `OMRS_MODULE_WEB_ADMIN` as the primary names (lines 15 and 41) and `OMRS_CONFIG_*` as "deprecated environment names" that override them (lines 53 and 68). So a deployment that sets `OMRS_MODULE_WEB_ADMIN=false` gets `module.allow_web_admin=true` written here instead. `ModuleUtil.allowAdmin()` reads that key straight out of the runtime properties and defaults it to `false` when absent, so the effect is that web module upload comes on against the operator's setting, silently, and permanently, because `write_runtime_properties_if_missing` returns early once the file exists and never corrects it. `OMRS_AUTO_UPDATE_DATABASE=false` goes the same way into `DatabaseUpdater.allowAutoUpdate()`.

This repo's compose uses the `OMRS_CONFIG_*` spelling so it is unaffected, but the deployed compose is not this file (the Dockerfile comment says so), and the credential block at lines 250-253 already reads both families.

```suggestion
auto_update_database=${OMRS_CONFIG_AUTO_UPDATE_DATABASE:-${OMRS_AUTO_UPDATE_DATABASE:-true}}
module.allow_web_admin=${OMRS_CONFIG_MODULE_WEB_ADMIN:-${OMRS_MODULE_WEB_ADMIN:-true}}
```

That ordering matches the base image, where `OMRS_CONFIG_*` wins when both are set. I ran both lines through `dash -n`, `sh -n` and `shellcheck -s sh -S warning`, so the new lint job stays green.

backend-init.sh
```diff
@@ -347,6 +373,216 @@ SQL
...
+  _db_args="?autoReconnect=true&sessionVariables=default_storage_engine=InnoDB&useUnicode=true&characterEncoding=UTF-8"
+  # Java's Properties parser splits on the FIRST unescaped separator, so ':' and
+  # '=' inside the value need no escaping; the wizard escapes them only because
+  # Properties.store() always does.
+  ( umask 077; cat > "$_rtp_file" <<EOF
+#Written by backend-init.sh for a database this entrypoint prepared
+connection.driver_class=com.mysql.jdbc.Driver
+connection.url=jdbc:mysql://${DB_HOST}:${_db_port}/${DB_NAME}${_db_args}
+connection.username=${DB_USER}
+connection.password=${DB_PASS}
```
dkayiwa, line 510: Worth escaping the value here. OpenMRS reads this file with `Properties.load` (via `OpenmrsUtil.loadProperties`), where a backslash is an escape character rather than a literal. The comment above is right that `:` and `=` need no escaping, but `\` does: I ran `Properties.load` on `connection.password=pa\ss` and it parses to `pass`, so a password containing one produces a file OpenMRS reads as a different password, it cannot connect, and the file is never rewritten afterwards. Low odds with `openmrs` or a typical generated password, but the boot after a wipe is the one that would fail.

```suggestion
connection.password=$(printf %s "$DB_PASS" | sed 's/\\/\\\\/g')
```

`DB_USER` and `DB_NAME` are the same shape if you would rather do it once for all three.

docker-compose.yml
```diff
@@ -48,18 +48,33 @@ services:
...
+    # inside the window do not count toward retries, and a success ends it.
+    #
+    # Note for whoever closes the other half of #275: this widens the legitimate
+    # not-yet-healthy window from ~2 min to as much as 30. Re-enabling the deploy
+    # wrapper's own poll by dropping --no-health-check, without giving that poll a
+    # matching timeout, would make a reset deploy report failure while the
+    # instance is starting perfectly normally.
     healthcheck:
-      test: ["CMD", "curl", "-fsS", "-m", "5", "-o", "/dev/null", "http://localhost:8080/openmrs/"]
+      test: ["CMD", "curl", "-fsS", "-m", "5", "-o", "/dev/null", "http://localhost:8080/openmrs/health/started"]
```
dkayiwa, line 73: Did you check that the empty-database path still installs itself after this? Probing `/openmrs/` was quietly doing a second job. `InitializationFilter.skipFilter` returns false while initialization is required, so each 30s probe reached `doGet`, matched `install_method=auto` and called `autoRunOpenMRS`. `/openmrs/health/started` is handled earlier, in `StartupFilter.doFilter`, and only reads `Listener.isOpenmrsStarted()`, so it triggers nothing.

What is left is the base image's `startup.sh`: `sleep 15; curl -sL http://localhost:8080/openmrs/ > /dev/null || true`. One attempt at a fixed 15s, before Tomcat necessarily has its connectors up (they start after the contexts), and openmrs-core deleted exactly that line in d74673b1f, replacing it in the same commit with a `/openmrs/health/alive` healthcheck at `interval: 5s` plus `StartupFilter.triggerSetup`, which GETs the context root. Upstream deliberately moved this responsibility onto the retrying health probe, which is the probe being repointed here. The image we build keeps the 2.8.x scripts, so the one-shot curl is still present, but it is now the only trigger.

I could not boot the image to see which side of the 15s the connector lands on, so this is a question rather than a claim. It only matters when the seed does not produce a schema, which `maybe_seed_demo_data` tolerates and continues from (failed dump download, failed import). On that path the comment above says backend-init.sh "keeps that state from arising", and it does not: the wizard is still the design there, so the first visitor becomes the trigger and the installation screen is what they would see.

`/health/started` still looks like the right endpoint for the check itself, since `/health/alive` answers 200 while setup is needed, which is the state you are trying to stop rewarding. It is the trigger that needs its own home. Backgrounding a short retry loop against `/openmrs/` from the branch that did not write runtime properties would make it explicit instead of incidental.

PR #287, opened 2026-08-19
fix(streaming): a long answer is no longer cut off by a proxy that has read nothing yet

dkayiwa:
## The defect

`searchStream` commits its response headers and then writes nothing until the LLM's first `thinking` or `token` event. A reverse proxy cannot tell that silence from a hung origin, so it closes the connection on its read timeout — **Cloudflare at ~120s, stock nginx's `proxy_read_timeout` at 60s**. The client is left with `HTTP 200` followed by **zero bytes**: no error event, no audit row, nothing on the server saying a clinician got nothing. Silent at both ends, and the module's own `llm.timeoutSeconds` default of 300 is longer than either proxy window, so the configuration invites it.

## Measured

On the `chartsearchai.openmrs.org` demo, 2026-08-19, on its 4-core box:

| served model | first `thinking` | outcome |
|---|---|---|
| Gemma 4 E4B | never arrived in the window | **cut at ~125s, 0 bytes** — `/search` as a Cloudflare `524 origin_response_timeout` (125124ms), the stream as a dropped connection (125403ms) |
| Gemma 4 E2B | **27.6s / 38.0s** | completed at 149.2s / 154.2s, closed cleanly, 2004 bytes |

Both E2B runs finished *well past* the 125s cut. What decides whether a long answer survives is therefore whether something is written **early**, not whether the answer finishes inside the window — so a keep-alive is the fix, and a smaller model is not. The demo is pinned to E2B only because that was the sole way to get a byte out in time; with this merged, the served model goes back to being chosen on answer quality (README's Evals gate) rather than by a proxy timeout.

## The fix

`streamAnswer` writes an SSE comment before generation begins and every 15s until the answer is finished. A comment is what the SSE spec requires clients to ignore, so **no client change is needed** and no phantom event can reach the UI.

Four details carry the correctness:

- **Writes serialize on the `OutputStream` monitor `writeSseEvent` takes.** The timer runs on its own thread, and an unsynchronized comment can land between an event's `event:` line and its `data:` lines, splitting one event into two malformed ones for every client.
- **`stop()` sets its flag under that same monitor.** A comment *can* still land after `done` — a task parked on the monitor during the final write may take it before `stop()` does, which is harmless — but no write can happen after `streamAnswer` returns. `shutdownNow()` alone cannot give that, because interrupting a thread parked on a monitor does nothing; only the flag closes that window.
- **`scheduleWithFixedDelay`, not `scheduleAtFixedRate`.** A write to a slow client can block for longer than the interval; at a fixed rate the executor then owes several runs and fires them back to back into the same congested socket.
- **`stop()` is called from a `finally`, not from the tail of the try block.** Every error exit leaves through a catch — a disconnect, a chart too large, a misconfiguration, an LLM timeout — and only a `finally` covers those. Unstopped, one daemon thread per disconnected request goes on writing every interval for the life of the JVM, since both catches in `SseKeepAlive.write` keep it alive through whatever the recycled response throws at it; where the write succeeds instead of throwing, that is 14 bytes into whichever request owns that stream object next.

15s is a constant rather than a global property deliberately: the only requirement on the number is that it be smaller than every proxy default, and one large enough to need tuning would already be broken.

## Spec change, called out explicitly

`ChartSearchAiStreamingTest` banned `new Thread(` in this controller outright. That ban is **narrowed, not dropped**. Its reason is the one the test's other three assertions name — OpenMRS binds authentication to the request thread, so work done off it either loses `Context` or shares it unsafely. A timer that only writes bytes carries none of that risk. The assertion now requires that every `new Thread(` **and every `Executors` call** in the controller is the keep-alive's, and that `SseKeepAlive` never touches `Context`. All fail under mutation (a rogue thread elsewhere; a factory-less `Executors.newSingleThreadExecutor` running `Context.getAuthenticatedUser`, which the `new Thread(` scan alone could not see; a `Context.getAuthenticatedUser()` inside the keep-alive), and a further assertion fails loudly if the brace-matched class body is ever short — a guard that weakens in silence is worse than none. The javadoc names those two spellings rather than "every thread", because a thread can still arrive by a route neither reaches (`new Timer()`, `CompletableFuture.runAsync`, a Spring-managed pool).

Every existing thread in this module lives in the `api` layer (`LocalLlmEngine`, `PrewarmRefreshExecutor`); this is the first in the controller, which is precisely why that test existed. Hiding the timer in a helper class to keep the grep green would have satisfied the letter and left the real invariant unguarded.

## Tests

Seven behavioural tests through the real `streamAnswer`: the early byte, the periodic writes, the timer stopping at return, the timer stopping on a client disconnect too, no frame ever split, a keep-alive that cannot be written not failing the answer, and one that throws not cancelling the schedule. Each behaviour's test is killed by a mutation of exactly its own mechanism and nothing else.

Two are worth flagging:

- The interleaving test deliberately does **not** write to a `ByteArrayOutputStream`. Every one of its methods is synchronized, so a whole `write(byte[])` is already atomic there and the hazard is unobservable — written that way the test stayed green with the production lock deleted. Over a byte-at-a-time sink, which is what a servlet output stream is permitted to be, it fails without the lock.
- No timing-dependent test here can fail from a slow runner, and none of them counts on a span elapsing to prove its point. Where the assertion is that keep-alives ARRIVE, the stub waits until they have actually been written, so a timer starved by a loaded runner or a GC pause cannot fail it spuriously. Where the assertion is that they STOP, there is a fixed 200ms sleep, ten intervals, and that direction is safe: a longer wait only gives an unstopped timer more room to be caught, so it can strengthen the assertion and never break a healthy one.

`mvn clean install`: **api 1281/1281, omod 85/85**, no failures.

All four of the details above are now pinned. The fixed delay's guard matches on the CALL rather than the method name, because `SseKeepAlive.write`'s own catch comment names `scheduleWithFixedDelay` in prose and an assertion on the bare name passes with the call swapped. The `finally` was the last to be pinned and the one whose absence was quietest: moving `stop()` into the tail of the try block left the suite green at 84 of 84, `noKeepAliveIsWrittenOnceTheAnswerIsFinished` included, because that test drives the happy path where the tail runs anyway. `aClientDisconnectStopsTheTimerToo` drives a sink that refuses the ANSWER's frames instead, so the generation loop unwinds exactly as a mid-stream disconnect does, and it is the only test that reddens on that mutation.

## Client compatibility — verified, not assumed

All seven in-repo Python SSE consumers (`eval/latency/*`) match on `event:`/`data:` prefixes, so a comment line falls through both branches. The frontend's parser (`openmrs-esm-chartsearchai`, `src/api/chartsearchai.ts`) is line-based and ignores unknown lines, and its `dispatchEvent()` returns early when no data lines are buffered — so the blank line after a comment dispatches nothing. The README's SSE section now documents the comment frames as part of the wire contract.

## What this does not fix

Non-streaming `POST /search`. Its total response time (~150s) exceeds the proxy window regardless of when the first byte goes out, so direct API callers — including `eval/latency/prod_probe.py` — stay broken until inference itself gets faster. Note also that this change takes away the proxy cut without anything replacing it. `llm.timeoutSeconds` is not that replacement: it becomes a JDK `HttpRequest.timeout()`, which stops applying once the inference server's response headers arrive, after which the token stream is read lazily and bounded by nothing. Measured 2026-08-20 against a local llama-server on CPU (`-ngl 0`, Llama-3.2-3B-Q4_K_M): a cold prefill does raise `HttpTimeoutException`, at 2.0 s and at 8.0 s against timeouts of exactly those lengths, while a cache-warm request under a 20 s timeout returned its headers at 117 ms and then streamed for 23.8 s without raising one. So a slow query holds its request thread and llama-server's single slot for the prefill, which `timeoutSeconds` caps, plus the whole of generation, which nothing caps. README, its global-property table, `docs/adr.md` and `LlmEngine.inferStreaming`'s `@param` all say that now; an actual bound is an api-layer change and is not in this PR.


omod/src/test/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiStreamingTest.java
```diff
@@ -40,12 +45,44 @@ public void searchStreamMethod_shouldExist() throws NoSuchMethodException {
...
+			String creation = source.substring(at, Math.min(source.length(), at + 120));
+			assertTrue(creation.contains("\"chartsearchai-sse-keepalive\""),
+					"the only thread this controller may create is the SSE keep-alive's, because "
+							+ "OpenMRS authentication is bound to the request thread; found: " + creation);
+		}
+		String keepAlive = nestedClassBody(source, "SseKeepAlive");
+		assertTrue(keepAlive.contains("stopped = true"),
+				"the extracted class body must reach SseKeepAlive.stop, or the region is short and the "
+						+ "Context assertion below is passing on text it never read — a guard that weakens "
+						+ "in silence is worse than no guard");
```
dkayiwa, line 150: This pins the flag's write half but not the read, and the read is the half that closes the window. I deleted `if (stopped) { return; }` from `SseKeepAlive.write()` and the omod suite stayed green at 81/81, `noKeepAliveIsWrittenOnceTheAnswerIsFinished` included, even though that test's own message names the hazard ("writes into a response the container has already recycled for someone else"). Deleting `stopped = true` does redden the assertion here, so the mechanism reads as covered when only one of its two halves is.

What the regression looks like: with the read gone, a task that parked on the monitor while the terminal `done` frame was being written takes the monitor once `stop()` releases it, so its write can land after `streamAnswer` has returned and the container has recycled the response. `shutdownNow()` cannot stop it, for exactly the reason your comment gives, since a thread blocked entering a `synchronized` block ignores the interrupt. That puts 14 bytes of `: keep-alive` into whichever request owns that stream object next, and in the case that actually hurts, where the write succeeds instead of throwing, nothing says so.

A deterministic assertion in the genre this file already uses. I checked it: passes on your HEAD, reddens on the deletion above.

```suggestion
		assertTrue(keepAlive.contains("stopped = true"),
				"the extracted class body must reach SseKeepAlive.stop, or the region is short and the "
						+ "Context assertion below is passing on text it never read — a guard that weakens "
						+ "in silence is worse than no guard");
		assertTrue(keepAlive.contains("if (stopped)"),
				"stop() setting the flag is worthless unless write() reads it: without the read, a task "
						+ "parked on the monitor writes after streamAnswer returns, into a response the "
						+ "container may already have recycled, and shutdownNow cannot stop it because a "
						+ "thread blocked entering a synchronized block ignores interrupt");
```

A behavioural version is possible (tiny interval, a terminal-event write slow enough to guarantee a parked task, then assert the byte count never grows after return), but the last step is a monitor race you cannot force either way, so it would only redden probabilistically. I would take the text assertion. Nothing here is wrong as merged; the gap is that the next edit can remove the read in silence.

README.md
```diff
@@ -450,6 +450,8 @@ SSE events:
...
...parser must skip any line beginning with `:` (as the SSE spec requires) rather than treat it as a frame. This matters because on a CPU-only server the entire pre-answer wait is silent, and a proxy cannot tell that silence from a hung origin: measured on this project's own demo 2026-08-19, Gemma 4 E4B queries were closed at ~125 s having delivered **zero** bytes and no `error` event, while the same question served by E2B (first `thinking` at 27–38 s) completed at 149–154 s. Cloudflare closes a silent origin connection at ~120 s and nginx's `proxy_read_timeout` defaults to 60 s — both shorter than `chartsearchai.llm.timeoutSeconds` (300), which until this keep-alive made the network the effective ceiling. It no longer is: the stream is never read-idle, so a slow query now runs to `timeoutSeconds`, holding its request thread and llama-server's single slot (`--parallel 1`) for that long instead of being cut at ~120 s. Set `timeoutSeconds` on that basis. The blocking [`/search`](#search) endpoint has no equivalent protection, because it writes nothing until the answer is complete: on hardware where one answer outlasts the proxy's timeout, that endpoint fails where this one now succeeds.
```
dkayiwa: `docs/adr.md` carries this same wire contract in its "Streaming endpoint (SSE)" section (line 822) and says nothing about the comment frames. It defers payload detail to this table, but it states client obligations in its own voice at line 840 ("A client must therefore keep reading past `done`, and must not assume `done` is terminal"), so the new obligation to skip lines beginning with `:` now lives in only one of the two places someone writing a client will read. A sentence beside that one would close it. Your call whether it is worth the edit.

dkayiwa: The sentence went in at `adr.md:840` in c58ee1f2, beside the keep-reading-past-`done` obligation, and it defers the proxy timeout numbers and the demo measurements to this paragraph rather than restating them.

omod/src/main/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiRestController.java
```diff
@@ -1119,15 +1179,119 @@ private void sendReferencesEvent(OutputStream out, List<RecordReference> referen
...
+		/**
+		 * Writes the first comment and schedules the rest.
+		 *
+		 * <p>That first write is on the CALLING thread deliberately. The property that matters is
+		 * that a byte has left before generation begins; scheduling it would make that a race
+		 * against the model's own first token, which is the race this whole mechanism is about.</p>
+		 */
+		static SseKeepAlive start(OutputStream out, long intervalMillis) {
+			ScheduledExecutorService timer = Executors.newSingleThreadScheduledExecutor(runnable -> {
+				Thread thread = new Thread(runnable, "chartsearchai-sse-keepalive");
```
dkayiwa, line 1257: This line creates the background thread that `searchStream`'s own javadoc at line 366 says the design avoids: *"This avoids the need for `SseEmitter`, background threads, async servlet support, and proxy privileges."*

You renegotiated that invariant deliberately on the test side, renaming `streamingEndpoint_shouldNotUseBackgroundThreads` and rewriting the class javadoc to draw the "on the request thread" rather than "without background threads" distinction. The production javadoc on the endpoint still states the version the old assertion enforced, so the code and the doc now disagree about the one invariant this PR is about, and the javadoc is the half a reader of `searchStream` sees first. Nothing breaks at runtime; the cost is a maintainer being told the controller has no background thread by the method that starts one.

The same narrowing you applied to the test would do it, roughly: "Writes SSE events directly to the response output stream in the request thread, which is where OpenMRS binds authentication. The only other thread is the keep-alive timer below, which writes comment frames and does no OpenMRS work. This avoids the need for `SseEmitter`, async servlet support, and proxy privileges."

While you are in that javadoc: its event list is the third place someone writing a client reads this contract, and it already carries a client obligation in its own voice ("Clients must keep consuming the stream after `done` to receive it"), so the sentence you put at `adr.md:840` for that reason belongs here too. Same edit either way.

omod/src/main/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiRestController.java
```diff
@@ -1119,15 +1179,119 @@ private void sendReferencesEvent(OutputStream out, List<RecordReference> referen
...
+				return thread;
+			});
+			SseKeepAlive keepAlive = new SseKeepAlive(out, timer);
+			keepAlive.write();
+			// Fixed DELAY, not fixed rate: the question a keep-alive answers is "has anything been
+			// written lately", so the clock should start when a write finishes. A write to a slow
+			// client can block for longer than the interval, and at a fixed rate the executor then
+			// owes several runs and fires them back to back into the same congested socket, growing
+			// its queue for as long as the congestion lasts.
+			timer.scheduleWithFixedDelay(keepAlive::write, intervalMillis, intervalMillis,
```
dkayiwa, line 1269: Of the three details your description says carry the correctness, this is the one no test holds. I switched it to `scheduleAtFixedRate` and the omod suite stayed green at 83/83; no api test reads this file, so that is the whole guard set.

What makes it worth an assertion rather than a general "more mutation testing" grumble is that everything around it does die. Dropping `synchronized (out)` from `writeSseEvent` takes `aKeepAliveNeverSplitsAnEventFrame` from 200 token events to 199. Dropping `if (stopped)` reddens the flag assertion in `ChartSearchAiStreamingTest`. Periodicity is held (a one-shot `timer.schedule` kills two tests), and so are both catches: rethrowing the `IOException` kills `aKeepAliveThatCannotBeWrittenDoesNotFailTheAnswer`, and deleting the `RuntimeException` catch kills `aKeepAliveThatThrowsDoesNotCancelTheRestOfTheSchedule`. So the unguarded thing is exactly the delay-versus-rate choice, and the two methods read as interchangeable to anyone who has not been bitten by the catch-up burst, which is what makes it the plausible next edit.

I would take a source-text assertion in the `SseKeepAlive` region, the same remedy the stop-flag read got last round, and I would not hold the merge for it. A behavioural version needs a write that blocks longer than the interval and then a judgment about what counts as back to back, so it would not fail cleanly.

```java
assertTrue(keepAlive.contains("scheduleWithFixedDelay"),
		"the keep-alive must be scheduled at a fixed DELAY, not a fixed rate: a write to a slow "
				+ "client can block for longer than the interval, and at a fixed rate the executor "
				+ "then owes several runs and fires them back to back into the same congested "
				+ "socket. Measured: swapping the two leaves the whole omod suite green");
```

README.md
dkayiwa: "instead of being cut at ~120 s" is not the baseline this had. On `main` nothing writes to `out` between `flushBuffer()` and the model's first output, so the ~120 s cut never reached the request thread at all. It learned of the disconnect at its first write attempt, which in the E4B case this paragraph is about is the ~194 s prefill your own test message cites. The thread and llama-server's slot were held that long, not 120 s.

So the change in resource holding is "first write attempt" to "full answer, or `timeoutSeconds`", not "~120 s" to "`timeoutSeconds`". Telling an operator to set `timeoutSeconds` on that basis is the right advice; it is the figure beside it that is off, by ~74 s in the very case the paragraph measures. Something like "...for that long, where before a cut connection went unnoticed until the first write attempt and unwound the request there" keeps the point and drops the wrong number.

`adr.md` defers the numbers to here, so this is the only place that needs it.

omod/src/main/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiRestController.java
dkayiwa: Narrowed in f97ab1d0. It now says what holds — writes go out on the request thread because that is where OpenMRS binds authentication, `SseKeepAlive` runs a timer, and no thread but the request thread does OpenMRS work — and names `streamingEndpoint_shouldNotRunOpenmrsWorkOnBackgroundThreads` as where that is enforced. `SseEmitter`, async servlet support and proxy privileges stay in the avoided list, since those parts were never false.

The comment frames went into the same javadoc as a paragraph after the event list rather than a `<li>`, because a comment is not an event. It states the skip-lines-beginning-with-`:` obligation and points at README for the read timeouts. It pointed at `adr.md` too when I wrote this; a later pass measured that adr.md carries none of those numbers and defers them to README, so that half was wrong and is gone (96d78849).

omod/src/main/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiRestController.java
dkayiwa: Guarded in f97ab1d0, but not with the assertion sketched above: that one is vacuous. `SseKeepAlive.write`'s catch comment names `scheduleWithFixedDelay` in prose, so `contains("scheduleWithFixedDelay")` is satisfied by the comment with the call swapped, and I confirmed it — 84 of 84 green under the mutation. The needle shipped is `timer.scheduleWithFixedDelay(`, whitespace-stripped and scoped to the nested class body, which does redden and is the only thing that does.

It sits in its own test rather than joining the assertions in `streamingEndpoint_shouldNotRunOpenmrsWorkOnBackgroundThreads`, whose stated scope is OpenMRS work on threads; congestion behaviour is a different claim. No region canary beside it: this is a positive containment, so a short region fails it rather than passing it, which is the opposite of the case the canary above exists for.

README.md
dkayiwa: Corrected in f97ab1d0. The clause now reads that the increase is real but the old ceiling was not the ~120 s cut: nothing was written before the model's first output, so a cut connection went unnoticed until the request's first write attempt, which on the E4B runs the paragraph already measures landed after the cut and is where the request unwound. The instruction to set `timeoutSeconds` on that basis stayed, with "rather than on the proxy's window" added.

I left the ~194 s out of the README. It is the prefill figure `keepAlivesKeepArrivingWhileTheModelStaysSilent` cites, and I did not measure it myself; the general statement is what an operator needs, and it rests only on the E4B measurement already in that paragraph. `adr.md` defers the numbers here, so it needed nothing.

omod/src/main/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiRestController.java
```diff
@@ -589,6 +650,18 @@ void streamAnswer(final OutputStream out, Patient patient, String sanitizedQuest
...
+			// Every exit stops the timer, including a client disconnect, which unwinds through the
+			// catch above rather than returning. Once stop() returns, no keep-alive can be in flight
+			// or begin (see SseKeepAlive.stop) — which is what lets the flush below take no lock.
+			// A comment can still land after the terminal event, because a task parked on the monitor
+			// during the final write may take it before stop() does. That is harmless: a comment
+			// carries no data, and with async grounding a client already has to keep reading past
+			// done. What must not happen is a write after this method returns, which is the window
+			// SseKeepAlive's stopped flag closes — shutdownNow alone cannot, since interrupting a
+			// thread parked on a monitor does nothing.
+			keepAlive.stop();
```
dkayiwa, line 667: "Every exit stops the timer, including a client disconnect" is the one claim in this block that no test holds. I moved `keepAlive.stop()` out of the `finally` and into the tail of the try block, and the omod suite stayed green at 84 of 84, `noKeepAliveIsWrittenOnceTheAnswerIsFinished` included: that test drives the happy path, where the try's tail runs anyway, so it cannot tell the two placements apart. Every other exit leaves through a catch, and that is most of them (a disconnect, chart too large, a misconfiguration, an LLM timeout). Nothing is wrong as merged, so I would not hold the merge for this, but I would land the test with it.

What the regression would cost, whenever someone makes that edit: a disconnected request never reaches `stop()`, so `stopped` stays false and `shutdownNow()` is never called either, and one daemon thread per such request goes on writing `: keep-alive` every 15s for the life of the JVM. Both catches in `write()` keep it alive through whatever the recycled response throws at it, and in the case that actually hurts the write succeeds instead of throwing, putting 14 bytes into whichever request owns that stream object next. Nothing is logged above DEBUG. The other three details your description calls load-bearing each picked up a guard this round; this is the fourth, and its failure is the least visible of the four.

A deterministic case, in this file's idiom. It needs a sink that refuses event frames while letting comments through, which is the inverse of `RefusingSink`. I ran it: 85 of 85 on your HEAD, and under the mutation above it is the only test that reddens (28 bytes at return, 140 after the sleep).

```java
	/**
	 * The exit that is NOT a return: a client that has gone away unwinds through
	 * {@code streamAnswer}'s catch, and the timer has to be stopped there too.
	 *
	 * <p>{@link #noKeepAliveIsWrittenOnceTheAnswerIsFinished} covers only the happy path, and every
	 * error exit in {@code streamAnswer} — disconnect, chart-too-large, misconfiguration, an LLM
	 * timeout — leaves through a catch instead. Measured: moving {@code keepAlive.stop()} out of the
	 * {@code finally} into the tail of the try block leaves the whole omod suite green at 84/84, while
	 * a disconnected request then keeps a daemon thread writing every interval for as long as the JVM
	 * lives, into a response object the container recycles for whoever asks next.</p>
	 */
	@Test
	public void aClientDisconnectStopsTheTimerToo() throws Exception {
		DisconnectedClientSink gone = new DisconnectedClientSink();
		// Waits for the keep-alives it asserts on rather than a fixed span, for the reason
		// SilentThenAnswerStub.awaitingComments gives.
		controller.setChartSearchService(SilentThenAnswerStub.awaitingComments(gone.sink(), 2));

		controller.streamAnswer(gone, patient(), "any allergies?", user(), false, 20L);
		int settled = gone.sink().size();

		assertTrue(countKeepAlives(gone.text()) >= 2,
				"the timer must have been running when the disconnect happened, or this test proves "
						+ "nothing: with only the synchronous comment there is no schedule left to leak. "
						+ "Got " + countKeepAlives(gone.text()));
		Thread.sleep(200L); // ten intervals; a timer still running would have written several times
		assertEquals(settled, gone.sink().size(),
				"a client disconnect leaves through a catch, not a return, so only a finally stops the "
						+ "timer: unstopped it writes into a response the container has already recycled "
						+ "for someone else, and shutdownNow cannot reach a task parked on the monitor");
	}

	/**
	 * A client that has gone away: the answer's own frames are refused so the generation loop unwinds
	 * through {@code writeSseEventOrThrow} exactly as a mid-stream disconnect does, while comment
	 * frames still land, so the test can see whether the timer kept writing after the unwind.
	 */
	private static final class DisconnectedClientSink extends OutputStream {

		private final ByteArrayOutputStream sink = new ByteArrayOutputStream();

		@Override
		public void write(int b) {
			sink.write(b);
		}

		@Override
		public void write(byte[] frame, int off, int len) throws IOException {
			if (len > 0 && frame[off] != ':') {
				throw new IOException("client gone");
			}
			sink.write(frame, off, len);
		}

		ByteArrayOutputStream sink() {
			return sink;
		}

		String text() {
			return new String(sink.toByteArray(), StandardCharsets.UTF_8);
		}
	}
```

omod/src/test/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiStreamingTest.java
```diff
@@ -40,12 +53,73 @@ public void searchStreamMethod_shouldExist() throws NoSuchMethodException {
...
+
+		assertTrue(source.contains("new Thread("),
+				"the keep-alive's thread must be created HERE, in the controller, or the loop below "
+						+ "iterates nothing and passes: an empty scan is indistinguishable from a "
+						+ "compliant one. Extracting the thread factory to a shared helper is the "
+						+ "refactor that does this — the module hand-rolls the same daemon factory in "
+						+ "LocalLlmEngine and PrewarmRefreshExecutor, so it looks like obvious reuse — "
+						+ "and it would leave the next background thread added to this controller "
+						+ "unguarded while the suite stayed green");
+		for (int at = source.indexOf("new Thread("); at >= 0; at = source.indexOf("new Thread(", at + 1)) {
```
dkayiwa, line 138: This loop can only see a thread created by a literal `new Thread(` in the controller, which is narrower than the message under it and narrower than the javadoc's "A thread that does OpenMRS work still fails here". I put this in the controller, outside `SseKeepAlive`:

```java
private static final ExecutorService ROGUE_POOL = Executors.newSingleThreadExecutor();

void rogue(final Patient patient) {
	ROGUE_POOL.submit(() -> Context.getAuthenticatedUser());
}
```

84 of 84 green. The default thread factory puts no `new Thread(` in this file, so the loop iterates once over the keep-alive's own site and passes, and the `Context.` assertion below is scoped to the nested class body so it never looks here.

The gap predates the PR, but this is the change that puts `Executors` in the controller's imports, so a factory-less pool is now the shortest way to add the next background thread here. Scoping `Executors.` the way you already scope `timer.scheduleWithFixedDelay(` closes it. I checked this one too: green on your HEAD, and it reddens on the mutation above with `expected: <1> but was: <2>`.

```java
assertEquals(occurrences(keepAlive, "Executors."), occurrences(source, "Executors."),
		"every executor this controller creates must be the keep-alive's: the default thread "
				+ "factory writes no `new Thread(` for the loop above to find, so a pool created "
				+ "outside SseKeepAlive is a background thread with no guard on it at all");
```

with (and `assertEquals` added to the static imports):

```java
/** How many times {@code needle} occurs in {@code haystack}. */
private static int occurrences(String haystack, String needle) {
	int count = 0;
	for (int at = haystack.indexOf(needle); at >= 0; at = haystack.indexOf(needle, at + 1)) {
		count++;
	}
	return count;
}
```

A thread can still arrive by a route neither check reaches (`new Timer()`, `CompletableFuture.runAsync`, a Spring-managed pool), so beyond that the message is the thing that has to stay honest about its own reach. Take it or leave it; nothing is wrong today.

omod/src/main/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiRestController.java
```diff
@@ -1119,15 +1192,119 @@ private void sendReferencesEvent(OutputStream out, List<RecordReference> referen
...
+				// The client is gone. The generation loop discovers that on its own next write and
+				// unwinds through writeSseEventOrThrow — a keep-alive is never the reason a request
+				// fails, and throwing from here would only cancel the schedule in silence.
+				log.debug("Could not write SSE keep-alive, client likely disconnected");
+			}
+			catch (RuntimeException e) {
+				// scheduleWithFixedDelay silently unschedules a task that throws — the documented
+				// behaviour of both periodic schedule methods — so without this the rest of a long
+				// answer would run with no keep-alive and nothing to say why.
+				log.warn("SSE keep-alive failed; the stream continues without one", e);
```
dkayiwa: The catch two lines up is what keeps the schedule alive, and `aKeepAliveThatThrowsDoesNotCancelTheRestOfTheSchedule` asserts that the next comment does arrive, so "the stream continues without one" tells an operator the opposite of what happens. On a persistently failing write it also fires every 15s, each occurrence disproving the previous one's claim.

```suggestion
				log.warn("SSE keep-alive failed; the schedule continues", e);
```

omod/src/main/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiRestController.java
dkayiwa: Fixed in a9927431. `aClientDisconnectStopsTheTimerToo` went in beside `noKeepAliveIsWrittenOnceTheAnswerIsFinished`, and it is still the only test that reddens when `stop()` moves to the tail of the try block — re-measured after the comment edit below, 28 bytes at return against 126 after.

Three departures from the sketch above. `DisconnectedClientSink` sits beside `RefusingSink` rather than next to the test, because this file puts tests first and sinks after, and its javadoc says it is that class's inverse. It recognises an event frame by NOT opening with the `:` of a comment rather than by matching `event:`, so a frame shape the controller grows later is refused too instead of quietly turning the test green. And the 84-of-84 figure is dated in the javadoc the way `theKeepAliveIsScheduledAtAFixedDelayAndNotAFixedRate` dates its own, since the suite is 85 now and the number would otherwise read as current.

The `finally` comment also names the test now. That was the half missing on the production side: the claim sat here with no way for a reader to tell whether anything held it, which is the same reason `searchStream`'s javadoc names its own enforcing test.

omod/src/test/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiStreamingTest.java
dkayiwa: Guarded in a9927431, with the assertion as sketched: every `Executors.` call in the controller must sit inside the `SseKeepAlive` body. It reddens on the factory-less pool with `expected: <1> but was: <2>` and is green otherwise.

The javadoc changed with it, and that is the part worth flagging. It said "every thread this controller creates must be the keep-alive's" and "a thread that does OpenMRS work still fails here", which is the claim the mutation falsified, so it now names the two spellings the scan actually reaches and states outright that `new Timer()`, `CompletableFuture.runAsync` and a Spring-managed pool are not among them. Adding the `Executors` half while leaving that sentence alone would have closed the measured hole and kept the unmeasured over-claim.

No region canary beside it, and the message says why: this is an equality against the same needle counted in the whole file, so a short region drops the keep-alive's own call and the two counts stop matching. It fails rather than passes, which is the opposite of the case the `stopped = true` canary exists for.

omod/src/main/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiRestController.java
dkayiwa: Applied verbatim in a9927431. Nothing else read the old string — grepped the java, md, py and xml — so the message was the whole of it.

ONBOARDING.md
```diff
@@ -38,14 +38,16 @@ Base path: `/ws/rest/v1/chartsearchai`. Every endpoint gates on a privilege up f
...
 | **GET** | **`/prewarmstatus`** | **Manage AI Prewarm** | **Bulk-prewarm progress/status** |
 | GET | `/auditlog` | View AI Audit Logs | Query the AI audit log |
 | POST | `/feedback` | AI Query Patient Data | Submit thumbs-up/down on an answer |
 | GET | `/drugreferencestatus` | Get Global Properties | Which drug-reference dataset is actually loaded |
 
+**Writing a `/search/stream` client.** Events arrive in this order: `preliminary`, `thinking`, `token`, `references`, `done`, `grounded`, `error`. Two framing rules, and a client that breaks either does so in a way the server cannot detect. `grounded` is a *trailing* event after `done` (async grounding only), so keep consuming the stream past `done` rather than treating it as terminal. And between events the stream carries SSE *comments* — lines opening with `:`, written so a reverse proxy never sees a read-idle connection — which must be skipped rather than read as a frame; `EventSource` does that for you. README's [Streaming search (SSE)](README.md#streaming-search-sse) section has each event's payload and the proxy read timeouts behind the comments.
```
dkayiwa: `EventSource` cannot reach this endpoint. It is mapped `RequestMethod.POST` with an `@RequestBody`, and the browser API only ever issues a GET with no body, which is why the frontend hand-rolls a `fetch` + `getReader()` parser (`openmrs-esm-chartsearchai`, `src/api/chartsearchai.ts`) instead of using it. So in the one paragraph whose job is to tell a client author what to do, the only thing said about who skips the comment lines points at an API they cannot use, and a reader who takes it at face value writes the parser the clause just before it warns against.

README and `docs/adr.md` carry the same clause, but each of those follows it with the hand-rolled obligation, so this copy is the one where it actually misleads. The `searchStream` javadoc has it too. Naming the property rather than the API covers all four:

```suggestion
**Writing a `/search/stream` client.** Events arrive in this order: `preliminary`, `thinking`, `token`, `references`, `done`, `grounded`, `error`. Two framing rules, and a client that breaks either does so in a way the server cannot detect. `grounded` is a *trailing* event after `done` (async grounding only), so keep consuming the stream past `done` rather than treating it as terminal. And between events the stream carries SSE *comments* — lines opening with `:`, written so a reverse proxy never sees a read-idle connection — which must be skipped rather than read as a frame, as the SSE spec requires. `EventSource` would do that for you but cannot reach this POST endpoint, so every client here is a hand-rolled reader that has to skip them itself. README's [Streaming search (SSE)](README.md#streaming-search-sse) section has each event's payload and the proxy read timeouts behind the comments.
```

omod/src/test/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiStreamingTest.java
dkayiwa: The loop is keyed on the thread's NAME rather than on where it sits, so one route past it is still open: a raw `new Thread(...)` outside `SseKeepAlive` that reuses the name literal. Measured on the head, this in the controller beside `formatDate`

```java
void rogue() {
	new Thread(() -> log.info("{}", Context.getAuthenticatedUser()),
			"chartsearchai-sse-keepalive").start();
}
```

leaves omod green at 85 of 85. The same block with the name changed to `"chartsearchai-rogue"` reddens this test, so the scan itself works and it is the name equality that admits the thread. Nothing else reaches it either: the `Executors.` counts stay at 1 because the thread is started directly, and the `Context.` assertion is scoped to the `SseKeepAlive` body. That leaves the message under the loop, "the only thread this controller may create is the SSE keep-alive's", claiming more than the check delivers, which is the shape this class's own javadoc calls a guard weakening in silence.

The cheapest close is the assertion already sitting below it, applied to threads as well as pools:

```java
assertEquals(occurrences(keepAlive, "new Thread("), occurrences(source, "new Thread("),
		"every thread this controller creates must sit INSIDE SseKeepAlive, not merely wear "
				+ "its name: a raw new Thread(runnable, \"chartsearchai-sse-keepalive\") started "
				+ "outside it passes the name check above, and neither the Executors. count nor "
				+ "the Context. scan reaches it");
```

I ran that one: green on the head as it stands, and it fails the rogue block above with `expected: <1> but was: <2>`. It needs no region canary of its own for the reason the `Executors.` assertion gives, and the name loop is worth keeping beside it, since the name is what makes the thread identifiable in a dump. Production is unaffected either way, so this is only about how far the guard reaches.

README.md
```diff
@@ -450,6 +450,8 @@ SSE events:
...
...-08-19, Gemma 4 E4B queries were closed at ~125 s having delivered **zero** bytes and no `error` event, while the same question served by E2B (first `thinking` at 27–38 s) completed at 149–154 s. Cloudflare closes a silent origin connection at ~120 s and nginx's `proxy_read_timeout` defaults to 60 s — both shorter than `chartsearchai.llm.timeoutSeconds` (300), which until this keep-alive made the network the effective ceiling. It no longer is: the stream is never read-idle, so a slow query now runs to `timeoutSeconds`, holding its request thread and llama-server's single slot (`--parallel 1`) for that long. Longer than before, but the old ceiling was not the ~120 s cut — nothing was written before the model's first output, so a cut connection went unnoticed until the request's first write attempt, which on the E4B runs above landed *after* the cut and is where the request unwound. Set `timeoutSeconds` on that basis rather than on the proxy's window. The blocking [`/search`](#search) endpoint has no equivalent protection, because it writes nothing until the answer is complete: on hardware where one answer outlasts the proxy's timeout, that endpoint fails where this one now succeeds.
```
dkayiwa: This needs correcting before merge: `timeoutSeconds` is not a ceiling on a streaming query, so the paragraph names a replacement bound that does not exist.

`HttpRequest.timeout()` covers the arrival of the response, not the reading of its body, and both engines read the body lazily (`LocalLlmEngine:242` and `RemoteLlmEngine:118`, both `BodyHandlers.ofInputStream()`; the only other bound the client carries is `connectTimeout`). I measured it against a server that sends headers and then stalls, with the timeout set to 2s: `send()` returned in 30ms and the body read was still blocked when the server closed at 12.0s, with no `HttpTimeoutException`. Same on Java 11, 17 and 21, which is all three CI builds against, and the same with `ofString()`, which threw a plain `IOException` at 12.0s rather than timing out. Nothing higher up adds a bound either, since `LlmProvider.searchStreaming` calls the engine synchronously on the request thread with no future or executor around it.

So if merged as-is, an operator who reads this and lowers `chartsearchai.llm.timeoutSeconds` to cap how long one slow query can hold a request thread and llama-server's single slot gets no cap at all. The keep-alive has taken away the proxy cut, and what the paragraph names in its place bounds only the wait for llama-server's response, not the generation the request then sits through. On the demo's one-slot box that is every other clinician's query queued behind a request nothing will end, and it fails quietly: the setting is accepted, nothing logs, and the operator believes they are covered.

I would fix the sentence and not grow the PR. The honest version is that the request now runs for as long as generation runs, and there is currently no ceiling on a streaming query at all, because `timeoutSeconds` bounds the wait for the inference server's response and not the token stream read from it. An actual bound is an api-layer change and belongs on its own. Worth knowing that the same belief is written at `docs/adr.md:925` ("Each request times out after `chartsearchai.llm.timeoutSeconds` (default 300s)"), which predates this PR but is the second place a reader would go to confirm the sentence.

omod/src/test/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiStreamingTest.java
```diff
@@ -40,12 +54,116 @@ public void searchStreamMethod_shouldExist() throws NoSuchMethodException {
...
+						+ "guarding it. Measured before this assertion existed: an "
+						+ "Executors.newSingleThreadExecutor in this class running "
+						+ "Context.getAuthenticatedUser left the whole module green at 84 of 84. "
+						+ "This assertion needs no region canary of its own — a short region drops the "
+						+ "keep-alive's own call, the two counts stop matching, and it fails rather "
+						+ "than passes. It matches call sites, so a mention of Executors in prose "
+						+ "outside the class reddens it too; that is the safe direction");
+		assertTrue(!keepAlive.contains("Context."),
+				"the keep-alive thread must write bytes and nothing else — reading Context off it is "
+						+ "exactly the unsafe sharing this test exists to prevent");
```
dkayiwa, line 187: A static import walks straight past this one, and it seems worth closing while the guard is being written. The needle is the text `Context.`, so `import static org.openmrs.api.context.Context.getAuthenticatedUser` lets the keep-alive read authentication state as a bare `getAuthenticatedUser()` and the scan never sees it.

Measured on the head, both directions. Adding that static import and putting `log.debug("keep-alive written for {}", getAuthenticatedUser())` after the write in `SseKeepAlive.write` leaves omod green at 85 of 85. The identical read spelled `Context.getAuthenticatedUser()` fails this assertion and nothing else, 85 run and 1 failure at line 163, so the scan itself works and it is the spelling that admits it. Placement matters for the measurement: the same read put *before* the write reddens six behavioural cases instead, because then no comment is ever written, which is the case `RestControllerContext`'s javadoc already records.

The controller carries no static imports at all today, so a whole-file negative costs nothing. `source` rather than `keepAlive` because an import sits above the class body, and it needs no region canary of its own: the positive `source.contains("new Thread(")` at the top of this method already fails if the source ever reads empty. I ran it, green on the head and the only failure under the mutation above.

```suggestion
		assertTrue(!keepAlive.contains("Context."),
				"the keep-alive thread must write bytes and nothing else — reading Context off it is "
						+ "exactly the unsafe sharing this test exists to prevent");
		assertTrue(!source.contains("import static org.openmrs.api.context.Context"),
				"Context must stay QUALIFIED in this controller: the assertion above matches the "
						+ "text \"Context.\", so a static import lets the keep-alive read authentication "
						+ "state as a bare getAuthenticatedUser() and the scan never sees it. Whole file "
						+ "rather than the class body, because an import is file-scoped");
```

README.md
dkayiwa: llama-server withholds its response headers until the first token, so `timeoutSeconds` does bound the prefill, and "no cap at all" above is wrong. What it does not bound is generation, which is the half that matters here. Measured 2026-08-20 against a local llama-server on CPU (`-ngl 0`, Llama-3.2-3B-Q4_K_M): a cold prefill raised `HttpTimeoutException` at 2.0 s and at 8.0 s against timeouts of exactly those lengths, while a cache-warm request under a 20 s timeout returned its headers at 117 ms and then streamed for 23.8 s without raising one. curl put that same prompt's first byte at 14.3 s and its whole response at 62.5 s, which is why the 2 s and 8 s runs fired at all.

Corrected in 5c9bfd55, in four places rather than one, because the belief had four homes: this paragraph, the global-property table at README:241 ("Maximum seconds to wait for LLM inference"), `docs/adr.md`'s concurrency section, and `LlmEngine.inferStreaming`'s own `@param`. The adr one was wrong in a second way the comment did not reach. It was step 3 of the queueing story, and the timer does not start until a request already holds the engine lock, so it never bounded the queue wait either; with this keep-alive running through that wait, nothing now cuts a queued request off at all, and the section says so.

Two boundaries kept deliberately. `infer` and `warmup` keep "maximum wall-clock seconds for the request", because whether a server that does not stream withholds its headers until the answer is complete was not measured, and if it does then that wording holds there. Replacing one unmeasured claim with another is not the fix. And no actual bound was added, per your last paragraph: the text now says a slow query holds its thread and the single slot for the prefill, which `timeoutSeconds` caps, plus the whole of generation, which nothing caps.

The figures come from a standalone client issuing the same two calls both engines make, `HttpRequest.timeout(...)` and `send(request, BodyHandlers.ofInputStream())`, rather than through the module, which needs a running server; the javadoc records that provenance rather than leaving the numbers to look like they came off the production path.

omod/src/test/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiStreamingTest.java
dkayiwa: Applied in 5c9bfd55, with the assertion as sketched. The javadoc changed with it, and that is the part worth flagging. The scope paragraph listed what the guard requires and named the two creation spellings as what the scan reaches; it now carries the qualified-`Context` requirement as well, and a new paragraph records the measurement in both directions, so the next reader can tell which half is the hole and which is only the harness. Placement is the harness half: the same read put before the write reddens six behavioural cases instead, because then no comment is ever written, and that is `RestControllerContext`'s case rather than this one.

One route past both is left, and I stated it rather than closed it, on the footing `new Timer()` already has in that javadoc: a helper method elsewhere in this controller that reads `Context` and is called from inside `SseKeepAlive`. The qualified read then sits outside the scanned body and no import is involved, so neither check reaches it and a text scan cannot.

Re-measured against the shipped wording rather than the sketch, since the message was reflowed: 85 run, 1 failure, and it is this assertion.

docs/adr.md
```diff
@@ -922,9 +922,9 @@ When multiple users submit queries simultaneously to the local engine, requests
 
 1. The first request acquires the engine lock and begins inference.
 2. Subsequent requests queue on the `synchronized` block and wait.
-3. Each request times out after `chartsearchai.llm.timeoutSeconds` (default 300s).
+3. `chartsearchai.llm.timeoutSeconds` (default 300s) does not bound that wait. It is a JDK `HttpRequest.timeout()` on the call to the inference server, so it does not start until a request already holds the lock, and it stops applying the moment that server's response headers arrive. On the streaming endpoint those arrive with the first token, so it bounds the prefill and leaves the queueing and the answer's own generation uncapped. Measured 2026-08-20 — README's [Streaming search (SSE)](../README.md#streaming-search-sse) keep-alive paragraph carries the figures.
```
dkayiwa, line 925: Ten lines down, in this same section, `### Existing mitigations` still says the opposite of what this line now says:

> - **Configurable timeout**: Prevents requests from waiting indefinitely.

That is adr.md:935, and it came in with the step 3 you just replaced (323960bb), so it is that same belief in one more home. It is also the line an operator scanning for mitigations actually reads.

If merged as-is, someone reading Decision 12 to decide whether the queue needs a bound gets "does not bound that wait" here and "prevents requests from waiting indefinitely" ten lines later, takes the bullet, and adds nothing. It reads worse after this merge than before it: the keep-alive removes the proxy cut that used to end such a request at the client, so now nothing ends it, and nothing logs or errors while it doesn't.

I checked both halves rather than reading them off the text above. `LocalLlmEngine.inferStreaming` is `synchronized` and builds its `HttpRequest.timeout(...)` inside the method, so the timer cannot start until the lock is already held. And against a raw socket that sends headers and then stalls, a 2s request timeout let the body read block for 12s with no `HttpTimeoutException`, on Java 11, 17 and 21 alike, which is what your figures on `LlmEngine.inferStreaming` say too. So the bullet is wrong for `/search` as well, not only for the stream: what it claims to bound is the queue wait, and the timer never covers that on either endpoint.

Worth correcting rather than deleting, since it does cap the prefill of whoever holds the lock:

```
- **Configurable timeout** (`chartsearchai.llm.timeoutSeconds`): caps the prefill of whichever request holds the lock, so a stalled prefill cannot block the queue indefinitely. It does not bound the queue wait itself, or the generation after the prefill (see point 3).
```

omod/src/test/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiStreamKeepAliveTest.java
```diff
@@ -0,0 +1,662 @@
...
+		SilentThenAnswerStub stub = SilentThenAnswerStub.awaitingComments(out, 3);
+		controller.setChartSearchService(stub);
+
+		controller.streamAnswer(out, patient(), "any allergies?", user(), false, 20L);
+
+		int written = countKeepAlives(stub.writtenAtEntry);
+		assertTrue(written >= 3,
+				"the silence must carry several keep-alives, not just the opening one: a proxy's read "
+						+ "timeout restarts on every byte, so one early byte does not save a prefill that "
+						+ "outlasts the timeout — E4B's was ~194s against a ~120s window. Got " + written
```
dkayiwa: `~194s` is the wrong quantity for this sentence. The only place that figure exists in the repo is README:140, where it is the **total** cold-query latency for E4B, in a table whose other row is E2B at `~63 s` cold and `~8.5 s` warm, and whose column header says "Cold query". Cited here as a prefill, it reads as a time to first output that nobody measured.

The claim it supports is fine, and it is measured somewhere better: this class's own javadoc already has "every Gemma 4 E4B query died at ~125s having delivered *zero* bytes", which says the prefill outlasted the window directly rather than by way of a total. Leaning on that keeps the point and drops the number.

```suggestion
						+ "outlasts the timeout, as E4B's did on the demo. Got " + written
```

I applied that and ran the class: 8 of 8, and the line stays at 90 columns. Not worth holding the merge for.

docs/adr.md
dkayiwa: Fixed in ecd58a87. One departure from the wording sketched above: it says "the inference server's first output" rather than "the prefill", which is the phrase README's property table and config.xml already use. "Prefill" is only right where the headers arrive with the first token, and whether a non-streaming server withholds them until the answer is complete is the boundary 5c9bfd55 deliberately left unmeasured on `infer`/`warmup` — so the streaming-only word would have generalised the same fact this section was just scoped away from. "First output" is also what lets the second sentence disclaim "the generation that follows that first output" without asserting anything about the blocking path.

"(see point 3 above)" rather than "(see point 3)", since the list sits under a different heading.

omod/src/test/java/org/openmrs/module/chartsearchai/web/rest/ChartSearchAiStreamKeepAliveTest.java
dkayiwa: Applied verbatim in ecd58a87. Only the message moved, so no expectation did.

One thing worth adding to the record, since a message is only as good as its rendering: I checked it against the failure it exists for rather than only that the class stays green. With `scheduleWithFixedDelay` swapped for a one-shot `timer.schedule`, this is the test that reddens, and it now reads `... outlasts the timeout, as E4B's did on the demo. Got 2 in ": keep-alive\n\n: keep-alive\n\n"` — the two being the synchronous comment plus the single scheduled one, which is exactly the shape a reader needs to see to know the schedule died rather than the lock.

PR #288, opened 2026-08-20
feat(drug-reference): the load status says which safety arms the loaded dataset can serve

### LINKED ISSUE #285: No default install can reach a dosing ceiling or a hand-authored contraindication, and warnOnDoseExcess still reads true
## What

Since #277 the shipped default is `sourceFormat=ddinter` with the whole DDInter knowledge base bundled — 2283
substances of interaction data, no configuration needed. DDInter carries no dosing ceilings and no
hand-authored allergy/condition rules. The curated `drug-reference.json` carries both, and seeds four drugs.
`sourceFormat` selects **one** parser for **one** file, so no shipped configuration reaches both: the default
install has 2283 substances and a dead dose arm, and the only alternative has live rules for Ibuprofen,
Paracetamol, Amoxicillin and Gentamicin and no interaction coverage at all.

This is not a regression and not a surprise — `config.xml` says the dose arm is dormant under the default
dataset, and ADR Decision 24's data-availability matrix says dosing maxima have no free authoritative source.
What is filed here is the **either/or**, and the fact that two arms ship *enabled* while being unable to fire,
with nothing observable saying so.

## Measured, not reasoned

Read directly from the two datasets `main` @ `b0cfe545` ships, for one substance that exists in both.

`api/src/main/resources/chartsearchai/drug-reference.json`, entry `Gentamicin`:

```
drugClass:        Aminoglycoside antibiotic
contraindication: {"type":"allergy","token":"aminoglycoside","note":"aminoglycoside hypersensitivity"}
contraindication: {"type":"condition","token":"renal impairment","note":"significant renal impairment (dose adjustment required)"}
ageBands:         [{0–11y, mgPerKg 5–7.5, maxDailyDoseMg 360}, {12–120y, mgPerKg 5–7.5, maxDailyDoseMg 480}]
```

`api/src/main/resources/chartsearchai/ddi-knowledge-base.json` — every drug row carries exactly these keys:

```
id, name, rxcui, rxnorm_name, drugbank_id, atc, ciel
```

plus top-level `metadata`, `mechanisms`, `interactions`. No ageBands, no maxDailyDoseMg, no contraindications,
for gentamicin or for any of the other 2282 substances. `config.xml`'s own `sourceFormat` description states
the same thing ("carries interaction rules but no dosing/contraindications (V1 DDI-only scope)"), so this is
the documented shape, confirmed against the shipped bytes rather than re-derived.

Consequence for the arms, on a default install:

- `chartsearchai.drugSafety.warnOnDoseExcess` defaults to **true** and has no published ceiling to compare a
  parsed dose against, for any drug.
- The hand-authored allergy/condition rule arm has no rule to evaluate, for any drug. The same-drug,
  shared-subgroup and curated-cross-reactivity arms still work — they need only ATC codes or nothing.

Observable on a real chart, this box's `chartsearchai_audit_log`, patient `a7090f70-…` (Betty Williams),
question "Is gentamicin appropriate for this patient?": the 2026-07-31 runs (curated era) answered around the
injected gentamicin record — renal-function-dependent risk, creatinine 146.5 µmol/L, "verify renal dosing" —
while every run since 2026-08-11 answers with the DDInter interaction verdict alone. The July wording is the
model paraphrasing an injected record, so the claim here is narrow: the **record content** that supported that
answer is no longer loadable alongside interaction data, not that a specific chip disappeared.

## What a fix has to decide

1. Whether the loader **composes** sources — curated rules layered over the DDInter base — and if so, which
   row owns substance identity when both name a drug (`substanceKey` is resolved per substance-name family, and
   two datasets disagreeing about a `drugbank_id` is a new shape for it), and what happens when a curated rule
   and a DDInter row contradict each other.
2. Or, if composition is out of scope for now, whether an arm that **cannot fire on the loaded dataset** is
   reported — a `DrugReferenceValidity` finding and/or a field on `GET /chartsearchai/drugreferencestatus`
   saying the dose arm has no ceilings to check. A GP that reads `true` while the arm is structurally inert is
   the "looks healthy, checks nothing" state #149 exists to make loud, and today `drugreferencestatus` only
   flags `inert` when zero entries load.
3. Neither answer should require an operator to discover the trade-off by reading a GP description: the
   default install is the one that has to be honest about what it is checking.

dkayiwa:
Fixes #285.

Takes item 2 of the issue. Item 1 (composing the curated rules over the DDInter base) is deferred to its own issue — it has to answer which row owns substance identity when two datasets name one drug and disagree about `drugbank_id`, which is a new shape for `substanceKey` and touches what #164/#185/#187 rest on.

## What it does

`isInert()` answers whether a dataset yielded anything. It cannot answer whether what it yielded is enough for a given arm, and the shipped default is exactly that case: 2283 entries load, so nothing is inert, while DDInter publishes no age band and no hand-authored allergy/condition rule at all.

`DrugReferenceLoad` now reports a `Coverage` verdict per arm — `published`, `absent`, `unloaded` — with the count beside it, carried by `toMap()` to `GET /chartsearchai/drugreferencestatus`:

```json
"arms": {"doseCeilings":      {"coverage": "absent",    "entriesPublishing": 0},
         "handAuthoredRules": {"coverage": "absent",    "entriesPublishing": 0},
         "atcCodes":          {"coverage": "published", "entriesPublishing": 1839},
         "interactions":      {"coverage": "published", "entriesPublishing": 2283}}
```

A verdict and not a bare count, because `notLoaded()` zeroes every field: `0` is what both "nothing was read" and "a dataset was read and publishes none" look like — the "count of 0 printed as cheerfully as 2283" failure ADR Decision 32 was written against, one level down.

Measured through the production accessors over the shipped 19 MB knowledge base, which is where those figures come from.

## Why not a `DrugReferenceValidity` finding

The issue offers "a finding and/or a field". The finding half is closed already, by the suite rather than by argument: **a dataset that publishes no dosing is VALID to this loader**, and three existing cases say so — `healthyLoadIsNotReportedAtWarnOrError` puts a dosing-less operator file in appdata and asserts no WARN (its javadoc refusing weakening in advance), and two fixtures with no age bands assert `getFindings().isEmpty()`. Taking the finding half means weakening those, which is changing what the loader considers valid. Absent dosing is inside the boundary by design — Decision 24's matrix records that no free authoritative dataset publishes dosing maxima. So this reports a **capability**, not a defect. There is no new WARN and no new finding, and the shipped default stays silent.

## Each arm asks what its own check asks

Not whether the field is populated. That distinction is the whole feature, and getting it wrong reintroduces the state the feature exists to remove:

- **dosing** asks `DrugSafetyValidator.publishesACeiling`, named in this PR as the patient-independent half of `actionableBand` and now shared with it. A band carrying only an age range parses fine — the ceiling fields are primitives defaulting to `0` — and can never fire for anyone.
- **hand-authored rules** ask `evaluatesAgainstTheChart`: a rule typed neither `allergy` nor `condition`, or carrying no matchable token, is unaskable.
- **class codes** count `atcSubgroups()`, the reduction the class comparison consumes, not the raw codes.

Each of those is mutation-verified: swapping in the obvious weaker predicate reddens exactly one named test and nothing else. So is the serialization — folding `absent` into `unloaded` inside `toMap`, or dropping either null-element guard, each reddens exactly one case.

## Notes for review

- The constructor now takes the loaded entry list instead of a count, so the count and the verdicts derive from one list rather than two arguments a caller must keep in step. There was exactly one construction site.
- `actionableBand` gained an early return on `publishesACeiling`. It is **behaviour-neutral by construction** — the predicate is strictly weaker than `dailyArm || weightArm`, so it fires only where the method already returned null — and its javadoc says so rather than leaving it to look tested. It exists so the arm and the report read their shared premise from one statement. Inverting it is caught by every dose-excess case.
- **Test files are edited**, which this project's rules are emphatic about, so: three edits extend existing assertions and none weakens one. `DOCUMENTED_FIELDS` and the api `toMap` key set each gain one element with every prior element keeping its position; the omod arm-key list is new and exact. A fourth strengthens a precondition that did not establish what its message claimed (`isInert()` is `loaded && entryCount == 0`, so it is also false for a load that never happened).
- Three new fixtures, each existing to make one assertion discriminating rather than trivially true: a rule the module cannot ask, a field populated with nothing actionable, and a null element inside `contraindications`/`ageBands` — the last of which would otherwise NPE inside `ensureLoaded`, which has no catch.
- Docs updated together: README's endpoint block, `ONBOARDING.md`'s contract statement, the controller javadoc, and Decision 32's retained-field list. The README one is the only wire contract statement nothing in the build would have caught as stale.

## Out of scope, worth its own issue

`DrugReferenceInjector:1468` gates dosing rendering on `band != null` alone, so for the data shape this PR newly calls `absent` it renders "0-0 mg/kg per dose (no pediatric daily maximum published…)" into a citable record while the endpoint reports `doseCeilings: absent` for the same dataset. Pre-existing — the injector never used `actionableBand` — but it is the one place the shared method's notion of "publishes a ceiling" is contradicted on the wire.

Also not addressed: `getCrossReactivityGroups()` is a second, independently loaded dataset that can be empty, and `arms` says nothing about it, since it is not entry-derived.

---

## What three review rounds changed

The commits after the first are rounds of clean-context review — a fresh agent that had not seen the code being written reviewed each pushed head, a second agent applied its findings, and the loop ran until a round reported **zero blocking findings**. Terminating count: round 3, 0 blocking, 3 non-blocking (applied).

**Round 1 — a live NPE the original change did not fix.** The reviewer drove the real `DrugSafetyValidator.validate` over this PR's own new fixture and got `NullPointerException: Contraindication.getType()`. The guards the first commit added fixed the *count* and left the crash: every `/search` putting that drug in play lost its whole safety pass behind the answer path's catches, while the status endpoint reported the arm cleanly `absent`. Guarding a bad-data shape at its call site is what `CLAUDE.md` forbids in terms, and it names the remedy — drop the offending value at the loader. Hence the `null-list-element` rule.

**Round 2 — the log leg was sold as removing the need to poll, and prints nothing by default.** Asserted unqualified in five places, false in all of them: `log4j2.xml` in openmrs-api sets `org.openmrs=WARN`, and the test that "proved" it passes only because `LogCapture` forces the level the claim assumes. Qualified rather than made loud, since ADR Decision 36 refuses to make an unactionable every-install state a WARN. The endpoint field — the channel #285's own comment prescribed — is what carries the answer.

**Round 3 — structure and coverage.** `Arm` now declares `abstract boolean publishedBy(DrugReference)`, so a fifth arm cannot be added without a predicate; before, it would have compiled and silently reported `absent (0)`. And the null drop covers six of an entry's lists where only two were tested — the untested `interactions` leg is dereferenced without a guard in two places, so a regression there reproduced round 1's crash class with nothing reddening.

## Verified on a running server, not only in tests

Merging head `5a78ab67`, deployed to a 3.7.0 standalone at platform 2.9.0-SNAPSHOT:

- the `arms` object returned character-for-character as documented above, unchanged by rounds 2–3;
- an operator dataset carrying a `null` inside `contraindications`, `ageBands` and `interactions` produced `null-list-element` / `dropped` / `occurrences: 3` naming the entry, the load still succeeding (`loaded: true`, `entryCount: 1`) — the finding reaching both the endpoint and a WARN, correctly loud because the file is the operator's own;
- a `/search` naming that drug returned HTTP 200 with the drug-reference load happening lazily *inside* the request, so the safety pass ran over the null-dropped dataset with zero NPEs.

One thing the runtime check establishes that the tests cannot: the arm-summary INFO line was observed only because that standalone's logging config had been raised to INFO. On a stock install it does not appear — which is what round 2's correction says.


dkayiwa:
**Disclosure: this PR was hardened by an automated clean-context review loop**, not by a single pass. Recording what it did so a human reviewer knows what has already been through, and — more usefully — what it could not judge.

Three rounds. Each round a fresh agent that had not seen the code being written reviewed the pushed head, a second agent applied the findings, and the loop ran until a round reported zero blocking findings. Terminating count: round 3, **0 blocking**, 3 non-blocking (all applied). Nothing was declined in any round.

| round | head | blocking | what it caught |
|---|---|---|---|
| 1 | `f999ef7c` | 1 | a live NPE the change did not fix |
| 2 | `7229a10b` | 1 | the log leg's purpose asserted false in five places |
| 3 | `07f656e3` | 0 | structure and coverage only |

Round 1 is the one worth knowing about. The reviewer drove the real `DrugSafetyValidator.validate` over this PR's own new fixture and got `NullPointerException: Contraindication.getType()`. The guards the first commit added fixed the *count* and left the crash — every `/search` putting that drug in play would lose its whole safety pass behind the answer path's catches, while the status endpoint reported the arm cleanly `absent`. Fixed the way `CLAUDE.md` prescribes for a bad-data shape: dropped at the loader, not guarded where it happened to throw.

Round 2 caught something the tests structurally could not: the arm-summary INFO line was sold in five places as removing the need to poll, and prints nothing on a stock install (`org.openmrs=WARN`). The test that "proved" it passes only because `LogCapture` forces the level the claim assumes.

**Verified on a running server**, merging head `5a78ab67`: the `arms` object returned character-for-character as documented, and an operator dataset carrying a `null` inside `contraindications`, `ageBands` and `interactions` produced `null-list-element` / `dropped` / `occurrences: 3` with the load still succeeding, the finding reaching both the endpoint and a WARN. A `/search` naming that drug returned HTTP 200 with the load happening lazily inside the request, so the safety pass ran over the null-dropped dataset with zero NPEs.

## What the loop could not establish — where review is still wanted

- **Whether the two deviations from this issue's own prescribed accessors are acceptable.** The comment on #285 named `normalizedAtcCodes()`; the code counts `atcSubgroups()`. It named the age-band accessor; the code asks `publishesACeiling`. Both are strict tightenings, both documented, both unobservable on either shipped dataset — but they are departures from what the issue asked for, and that is a maintainer's call, not a reviewer agent's.
- **Whether a fourth arm belongs in the map at all.** `INTERACTIONS` was added because `sourceFormat=atc` withholds interactions entirely. It is the one arm counted by field presence rather than by its own check, for reasons argued on the constant. Reasonable people could want it out of scope for #285.
- **Whether the log leg is worth keeping** now that its stated purpose is gone. It is invisible by default and the endpoint is what answers the question. It was added on a review suggestion and then qualified; deleting it is defensible.
- **Anything a clean-context reviewer cannot see because it is not in the diff** — whether this is the shape the module wants for capability reporting, and whether `getCrossReactivityGroups()` having no representation is a gap worth closing now rather than later.

Deliberately out of scope, each noted in the PR body: `DrugReferenceInjector:1468` renders a dosing line for a band with no ceiling, contradicting `doseCeilings: absent` for the same dataset (pre-existing; wants its own issue), and item 1 of #285, which the issue itself defers.

Issue #315, opened 2026-08-25
A stopped drug order can be reported without its stop date, and on one question shape the answer denies the status its own cited record states


dkayiwa:
An answer can name a drug from a **stopped** drug-order record while silently dropping the stop date the same cited record carries. On one of the question shapes it also asserts that no current medication status is confirmed, which is the opposite of what the record it cites states.

This is answer fidelity, not the safety layer. The deterministic arms behave correctly throughout — a stopped order leaves `getActiveOrders`, so no chip is raised (verified separately). The defect is that the prose reports a discontinued prescription as if its end were unrecorded.

## Measured

Local standalone (RefApp 3.7.0), shipped defaults (`chartMode=queryScoped`, `topK=12`), bundled DDInter KB (2283 entries). One synthetic drug order for one patient: Nevirapine, activated 30 days ago, `date_stopped` = yesterday, reindexed through `POST /querystore/reindex` so the record is genuinely in the index.

The indexed document text carries the stop date explicitly:

```
Drug order: Nevirapine. Action: NEW. Urgency: ROUTINE. Stopped: 2026-08-24
```

Every run below cited exactly that record (`references: [drug_order]`) and raised zero safety chips. Answers were identical across repeats of a shape.

| Question | n | Names Nevirapine | States it was stopped | Answer |
|---|---|---|---|---|
| `is he currently taking any medications?` | 3 | 3/3 | **0/3** | "No current medication status is explicitly confirmed in the records. The following drug order is recorded: Nevirapine [1]." |
| `what medications has he been prescribed?` | 2 | 2/2 | **0/2** | "Nevirapine was prescribed on 2026-07-26 [1]." |
| `what medications is he taking?` | 2 | 2/2 | 2/2 | "Nevirapine was ordered on 2026-07-26 [1] and was stopped on 2026-08-24 [1]." |
| `when was his nevirapine stopped?` | 1 | 1/1 | 1/1 | "Nevirapine was stopped on 2026-08-24 [1]." |

The last two shapes prove the stop date is legible and citable from the record. The first two show it dropped from the same record, in the same chart, on the same patient.

## Why the first row is the worst of the four

[...]

PR #316, opened 2026-08-25
refs(#315): no two-class ENDED/CURRENT prompt clause classifies an unmarked drug order — the measurement, and the tooling for the next attempt

dkayiwa:
Refs #315 — **this PR does not close it.** The issue stays open, and the reason is the deliverable:
the fix it proposed was built, measured, and does not work. What ships is that measurement, the
structural remedy it points at, and the tooling that makes the next attempt cheaper.

`LlmProvider.java` and `README.md` are **byte-identical to `origin/main`**; the compiled
`DEFAULT_SYSTEM_PROMPT` `cmp`s equal to base at 7637 chars. There is no runtime behaviour change in
this PR at all — the only production edit is javadoc.

## What was tried

#315's own "Likely root cause" is that the prompt says nothing about an order's ended state. So one
record-class clause was added to `DEFAULT_SYSTEM_PROMPT`, keyed on the two end markers
`DrugReferenceInjector.describesEndedOrder` already uses for the #118 reconciliation. It fixed the
ticket's reported defect: on the ticket's own arrangement — one Nevirapine order stopped two days
earlier, nothing active — `is he currently taking any medications?` went from
*"Yes — the patient was ordered Nevirapine on 2026-07-26 [1]"* to
*"No, the patient is not currently taking any medications. The order for Nevirapine was stopped on
2026-08-24 [1]"*, 4/4.

## Why it cannot ship

Five wordings, measured on the live 3.7.1 standalone, one binary, GP-swapped, **interleaved** — a
decoy question between every sample, so each re-prefills instead of reusing the KV cache. Two cells
decide it: **A** = the ticket's own case; **B** = a Simvastatin order lapsed by `auto_expire_date`
with `date_stopped` NULL (so its record carries *no* marker) beside live Bupivacaine and Lidocaine.

| wording | A — ticket's case | B — auto-expiry |
|---|---|---|
| base, no clause | ❌ "…**was ordered** Nevirapine on 2026-07-26" | ✅ "…Bupivacaine and Lidocaine. The patient **also has an older drug order** for Simvastatin" |
| `…neither marker is CURRENT, and a question … is answered from those` | ✅ | ❌❌ "Yes — currently taking Simvastatin" — **both live drugs dropped** |
| `…neither marker records no end` | ❌❌ "**is currently taking** Nevirapine" — worse than base | ➖ live drugs kept, lapsed one called current |
| `…records no end and is CURRENT — the patient is taking every one of those drugs` | ✅ 4/4 | ➖ same fold |
| the currency sentence **deleted entirely** | ❌ 4/4 | ➖ same fold |

Nine further wordings were measured; none recovered B. What the ledger establishes:

1. The `CURRENT`/`ENDED` **contrast** is what makes the ended branch fire — remove it or delete it and
   A reverts to its defect, 4/4 both times.
2. **The fold is not caused by that sentence.** The wording with it deleted folds identically, while
   base does not fold at all. It is caused by the clause *as a whole*: teaching an ENDED-vs-not
   classification over text in which a lapsed order is indistinguishable from a live one.
3. So no wording of a **two-class ENDED/CURRENT** clause satisfies both, whatever it says: the
   auto-expired record carries no discriminator for it to key on, which is a property of the
   information in the record text rather than a search that has not yet succeeded. Scope that claim
   as stated — the prompt *does* tell the two apart where `date_stopped` is set (row 4 above, 4/4).
   What is refuted is the binary framing, not "a prompt cannot classify order state". A third class
   (*a record carrying neither marker does not state whether the drug is current*) would supply the
   contrast point 1 needs without the assertion point 2 clears of blame; it is **unmeasured**, named
   in ADR Decision 45 as where the next prompt-side attempt starts, and it is not what this PR
   claims to have refuted.

Both failures are the same clinical error — *a drug reported as current when it isn't*. The clause
moves it from stopped orders onto auto-expired ones, and `auto_expire_date` is set by every
duration-based prescription, so that is not the rarer case in production. `CLAUDE.md`: *never commit
code with known regressions … diagnose the shared problem before patching stages.*

## The shared problem, and the remedy

The chart text shown to the model cannot express order status, while the module knows it
authoritatively — `Order.isActive()` excludes an expired order, so `getActiveOrders` already has the
answer. querystore never writes `auto_expire_date` into the record *text*, pinned by
`QuerystoreOrderTextMarkerTest.anAutoExpireDateAloneIsNotVisibleInTheRenderedText`.

So the fix is to mark the drug-order records whose orders are not in the active set. **That is now
#317**, filed with the measurement, the two obstacles below and the fail-closed precedent to
follow. Two things make it a separate ticket rather than another round here:

- **It cannot live in `DrugReferenceInjector`.** `chartsearchai.drugReference.enabled` defaults to
  `false`, so the injector is off on a default install while a prompt clause is always on. The mark
  belongs in the always-on chart-assembly path (`QueryStoreChartBuilder` / `PatientChartSerializer`),
  which today reads only querystore and takes no `OrderService` dependency.
- **The naive version fails CLOSED.** `PatientClinicalContextBuilder` swallows a failed active-order
  read into an empty set at DEBUG with no flag — unlike the two contraindication reads beside it — so
  "not in the active set ⇒ ended" would mark every drug the patient has as stopped when the read
  fails. `PatientClinicalContext.contraindicationRecordsRead()` is the precedent to follow.

## What ships

- **A yes/no medications gate cell** (`dc8560c9|probe-current-meds`). No committed probe cell was a
  medications list, which is why an earlier round's verdict-lead regression was invisible to every
  gate in the repo. Verified by mutation: under the reverted wording it scores 0/1.
- **Three always-on eval-harness defects.** `capture_probe_yesno.sh` printed `CAPTURE_DONE` to stdout
  but never wrote the marker *file* its comparers read, so every A/B over its captures reported a
  killed run and exited 3. And Tier-B completeness counted against cells from a cohort no single host
  has. The claim that `score_directness.py` reads that marker was false in **four** places, not the
  three an earlier round claimed to have fixed — the fourth was `capture_probe_safety.sh`'s own
  marker comment, in a file this PR does not otherwise touch. All four are corrected; the real
  readers are `compare_arms.py` and `score_probe_safety.py`. The lesson is in the ADR: grep for a
  claim's homes, do not count them from memory.
  The marker guard took four attempts across four rounds, and each round's fix was right about the
  defect it saw and blind to the one it made. It now records the cells the run INTENDED to fire
  beside how many landed, is *refused* when none landed, and **clears any marker already in the
  directory — below the refusals, above the first cell fired**. Every clause is load-bearing:
  without the clear, the documented resume path left the previous arm's answers *and* its
  completeness marker standing after a wholesale failure, and `compare_arms.py` — which reads
  marker presence and never the body — reported "class flips: 0" and exited 0 over the OLD arm's
  answers; with the clear above the refusals instead, a mistyped `CAPTURE_TIER_B=yes` silently
  destroyed a complete arm. Both reproduced and both re-run after the fix. **The same guard is now
  in `capture_probe_safety.sh`**, which had the identical hole in the family that is actually
  gated — a stub answering GET 200 / POST 500, which is what a wedged LLM leaves, wrote a marker
  asserting 20 cells and scored exit 0 with every column zero. (An arm that failed
  wholesale otherwise wrote `cells=0` and read as a clean, empty A/B — reproduced against a closed
  port), and the cohort scope can be STATED with `--cohort` rather than only inferred from the
  capture, because an inferred scope cannot tell a cohort that is absent from the host from one whose
  every cell failed. Both are covered by `score_directness.py --selftest`.
- **`EndedOrderMarkerContractTest`** pins two facts about querystore's rendered drug-order format
  against the real `DrugOrderRecordSerializer` that nothing else in the repo pins and this module's
  matcher does not read: the `"Drug order:"` record prefix — the only cue **guaranteed** to be
  present that separates a prescription from the other order classes carrying the same end markers,
  since the four others (`. Dose:`, `. Duration:`, `. Quantity:`, `. PRN`) are each emitted behind a
  null check on an optional field — and that a DISCONTINUE record carries no end date at all. The
  prefix case is a deliberate FORWARD contract for the structural remedy, with no present-day
  reader, and says so rather than claiming one. Three raw-casing/self-consistency cases were written and dropped
  in review: the matcher lowercases, so the change class they uniquely covered cannot reach a decision
  here, and their three package-private constants went with them rather than sit in production with no
  production reader.
- **A stale javadoc corrected**: `QUERYSTORE_STOPPED_MARKER` is emitted for `getDateStopped()` only,
  not also for `getAutoExpireDate()` — measured by running the serializer, and true before this PR.
- **ADR Decision 45**, status **REJECTED**, carrying the ledger, the residues in both directions, the
  structural remedy and its two obstacles.

## A methodology finding that outlives this ticket

`LocalLlmEngine`'s own javadoc records that `cache_prompt` "introduces a low-level non-determinism on
borderline argmax decisions". So **consecutive repeats of one question measure cache stability, not
answer stability** — `n=3 byte-identical` can be a warm-cache artifact on a borderline cell. That is
why one wording passed its gates and was then measured, interleaved, to have reinstated the very
defect it was written to fix. Every figure above is interleaved.

**Scoped, because this entry's own data narrows it, and an earlier revision of both this body and the
ADR overstated it.** Only ONE of the two refuting cells needed interleaving. The auto-expiry cell was
found *without* it — what had hidden that one was the arrangement set, since every chart measured
before the review loop carried an ENDED order. The ticket's own cell is the one interleaving exposed.
And what is *not* established: there is no paired A/B of the two protocols on that cell, so the cache
is the mechanism `LocalLlmEngine`'s javadoc supplies rather than a measured contrast. The discipline
is still worth adopting — a warm-cache repeat cannot tell you an answer is stable — but that is the
honest strength of it.

## Also found, not fixed here

Rendered stop dates run one day early where the local timestamp is after 21:00 (DB
`2026-08-16 00:29:12` renders as `2026-08-15`), consistent with a UTC conversion in querystore's
serializer. Pre-existing and outside this module — worth its own issue.


dkayiwa:
Closing as superseded. Everything here has either shipped, been re-cut against current `main`, or is deliberately deferred — nothing is dropped silently.

**Shipped in #318** (`1d76eee`, closing #317): the structural remedy this PR's ADR entry recommends. The mark lives in `QueryStoreChartBuilder.toSerializedRecords` — the always-on chart-assembly funnel, which is the first of the two obstacles named here — and `SerializedRecord.getOrderActive()` is three-valued with the mark written as a positive fact rather than inferred from absence, which is the second. #318 also landed the `QUERYSTORE_STOPPED_MARKER` javadoc correction from this branch, in its own wording.

That leaves this PR with no production change at all: its only remaining edits are javadoc on `DrugReferenceInjector`, which `main` now states better, and those are exactly where it conflicts.

**Re-cut as #319** — the eval-harness work, none of which ever depended on the prompt clause:
- `capture_probe_yesno.sh` writing the `CAPTURE_DONE` file `compare_arms.py:82` actually reads, plus the refusal and the pre-fire clear that make the marker mean something. This one is worth landing soon: every A/B over that script's captures on `main` today exits 3.
- the same guard in `capture_probe_safety.sh`
- the `probe-current-meds` medications gate cell
- cohort-scoped Tier-B completeness with `--cohort`

**Re-cut as #320** — ADR Decision 45, at the number #318 left free for it, with its forward-looking sections updated: what became of each obstacle, where each piece landed, and #315's trade-off pointing at Decision 46 rather than asserting a present-tense answer the module no longer gives. The ledger and the methodology finding are unchanged.

**Deferred, not dropped:** `EndedOrderMarkerContractTest`. Its `"Drug order:"` prefix case was kept as a forward contract for the structural remedy, and that remedy shipped keyed on `OrderService` and `RESOURCE_TYPE_DRUG_ORDER` rather than on the record text, so the reader it was written for never arrived. The DISCONTINUE-carries-no-end-date case still has a live subject — `describesEndedOrder` survives as the AND-ed prose fallback under Decision 46 — and wants re-reading on its own terms. #320 records it as unlanded and why.

**#315 stays open.** #318 gives the model the discriminator this PR established a prompt cannot supply, but the ticket's second half — a stopped order named without the stop date its own cited record carries — is not closed by it; #318's own A/B notes that on the one-stopped-order patient the shipped wording gives the verdict bare.

Issue #337, opened 2026-08-30
The answer paraphrases a deterministic safety finding and deletes the hazard: a Major methemoglobinemia warning reaches the clinician as "can also increase the risk"


dkayiwa:
## What happens

The safety validator produces a deterministic finding whose text is taken verbatim from the knowledge base. That text is injected as a citable `safety_finding` record, and the model then **rewrites it** in the answer. In the two Major findings measured below, the rewrite deletes precisely the clause that names the hazard, and what reaches the clinician is a sentence with a citation marker on it that no longer says anything clinically.

The chips beside the answer are correct. Nothing reconciles the two.

## Measured

OpenMRS RefApp standalone 3.7.1, chartsearchai `main` @ `09717dc7`, bundled DDInter KB (2283 entries, classpath origin), stock GPs, local Gemma E4B.

Patient `763e6e5f-c489-4bab-8a55-c379f085dd1c` (Susan Young) — active orders Tiotropium, Botulinum toxin type A, Lidocaine, Neomycin, Metoclopramide.

```
POST /ws/rest/v1/chartsearchai/search
{"patient":"763e6e5f-c489-4bab-8a55-c379f085dd1c",
 "question":"Please screen her current medications for drug interactions."}
```

**Finding 1 — the hazard is deleted outright.**

Chip (`safetyWarnings[].detail`, correct, straight from the KB):

> Lidocaine interacts with active order metoclopramide — Major. Coadministration of local anesthetics with other oxidizing agents that **can also induce methemoglobinemia** such as antimalarials (e.g., chloroquine, primaquine, quinine, …

Answer:

> Lidocaine interacts with active order metoclopramide [7] — Major: Coadministration of local anesthetics with other oxidizing agents that **can also increase the risk** [47].

"induce methemoglobinemia" became "increase the risk". The risk of *what* is now absent from the sentence. A clinician reading the answer learns that two drugs interact and is told nothing about methaemoglobinaemia — which is the entire content of this Major rating, and an oxygen-saturation problem they would need to watch for.

**Finding 2 — the term that makes the partner relevant is deleted.**

Chip:

> Botulinum toxin type A interacts with active order neomycin — Major. The effects of botulinum toxin may be potentiated by **neuromuscular blockers, aminoglycoside antibiotics,** and other drugs that can interfere with neuromuscular transmission…

Answer:

[...]

Issue #336, opened 2026-08-30
A capped interaction screen is indistinguishable from a complete one: 8 of 18 pairs were withheld with no signal anywhere in the response


dkayiwa:
## What happens

An interaction screen that hits `chartsearchai.drugSafety.maxPairChips` truncates its result and tells the clinician nothing about it. The withheld pairs appear **only** in a server-side `WARN` line. Nothing in the `/search` response — not the answer, not the chips, not the `references` array — distinguishes a complete screen from a truncated one.

The validator's own source already says this is the problem it does not solve:

```java
// Named, not counted: a clinician reading the reported chips has no way to tell a capped
// screen from a complete one, so the withheld pairs must at least be recoverable from the
// log rather than vanishing.
```
(`DrugSafetyValidator.addActiveOrderPairInteractions`, ~line 4430)

A clinician who asks "screen her medications for interactions" and receives ten chips has been given a list that reads as exhaustive. It is not, and they have no way to know.

## Measured

OpenMRS RefApp standalone 3.7.1, chartsearchai `main` @ `09717dc7`, bundled DDInter KB (2283 entries, `origin: classpath:/chartsearchai/ddi-knowledge-base.json`), stock GPs (`maxPairChips=10`, `minInteractionSeverity=minor`), local Gemma E4B.

Patient `dc8560c9-6d2b-45bf-861c-8fcf562ec9b1` (Sarah Taylor), 8 active oral drug orders — Ibuprofen, Celecoxib, Dexamethasone, Diclofenac, Hydrocortisone, Prednisone, Budesonide, Methylprednisolone.

```
POST /ws/rest/v1/chartsearchai/search
{"patient":"dc8560c9-6d2b-45bf-861c-8fcf562ec9b1",
 "question":"Please screen her current medications for drug interactions."}
```

Server log:

```
WARN - DrugSafetyValidator.addActiveOrderPairInteractions(4436)
Interaction screening across 17 active-order reference entries found 18 pair(s) above the
severity floor; reporting the 10 most severe and WITHHOLDING 8: Celecoxib x Dexamethasone
(Moderate); Celecoxib x Diclofenac (Moderate); Celecoxib x Hydrocortisone (Moderate);
Dexamethasone x Diclofenac (Moderate); Diclofenac x Methylprednisolone (Moderate);
Diclofenac x Prednisone (Moderate); Diclofenac x Budesonide (Moderate);
Diclofenac x Hydrocortisone (Moderate)
```

[...]

PR #350, opened 2026-08-31
fix(drug-safety): the drug-in-play interaction arm states its most severe finding first

### LINKED ISSUE #346: The drug-in-play interaction arm never sorts by severity, so a Major warfarin/ibuprofen bleeding interaction sits at position 8 and never reaches the answer
## What happens

The drug-in-play interaction arm (`DrugSafetyValidator.addInteractionWarnings` — the question's drug checked against the patient's active orders) appends one chip per partner **in the knowledge base's own row order, with no severity sort**. Both *other* interaction arms sort most-severe-first before emitting:

```
DrugSafetyValidator.java:4306   Collections.sort(found, PAIR_SEVERITY_DESCENDING);          // question-pair arm
DrugSafetyValidator.java:4979   Collections.sort(pairs, SCREENED_PAIR_SEVERITY_DESCENDING); // screening arm
```

`addInteractionWarnings` (from line 2360) contains **no `Collections.sort` at all** — it iterates `rules` and calls `warnings.add`.

The consequence is not only that the chip list is out of order. The answer enumerates the findings in the order it is given them, and it stops before the end — so **the most severe interaction is dropped from the answer entirely** while five Moderate ones are spelled out at length.

## Measured

OpenMRS RefApp standalone 3.7.1, chartsearchai `main` @ `77c0f9a2`, bundled DDInter KB (2283 entries, classpath origin), stock GPs, local Gemma E4B.

Patient `dc8560c9-6d2b-45bf-861c-8fcf562ec9b1` (Sarah Taylor), 8 active oral orders.

```
POST /ws/rest/v1/chartsearchai/search
{"patient":"dc8560c9-6d2b-45bf-861c-8fcf562ec9b1","question":"Can I give her warfarin?"}
```

The eight interaction chips, in the order the module emits them (severities now readable thanks to #340):

```
 1. sev=Moderate  Methylprednisolone
 2. sev=Moderate  Prednisone
 3. sev=Moderate  Budesonide
 4. sev=Moderate  Celecoxib
 5. sev=Moderate  Dexamethasone
 6. sev=Major     Diclofenac
 7. sev=Moderate  Hydrocortisone
 8. sev=Major     Ibuprofen      <-- never reaches the answer
```

That is exactly the KB's row order for the Warfarin entry, confirmed by reading the dataset directly.

The answer (2841 characters, **byte-identical on three consecutive runs**) walks chips 1 → 7 and stops mid-enumeration:

> No — Warfarin should not be given: Warfarin interacts with active order Methylprednisolone [349], a Moderate problem because … Warfarin interacts with active order Diclofenac [354], a Major problem because … Warfarin interacts with active order Hydrocortisone [355], a Moderate problem because … induced by corticosteroids [355].

```
mentions Diclofenac: True    mentions Ibuprofen: False
occurrences of "Major": 1    occurrences of "Moderate": 6
```

So **Warfarin × Ibuprofen — Major — is absent from the answer**, on a patient who is actively prescribed ibuprofen. That is NSAID potentiation of an oral anticoagulant's hypoprothrombinemic effect and bleeding risk: one of the most consequential interactions in the dataset, and arguably the single thing a clinician asking "can I give her warfarin?" most needs told. The chip exists and carries `severity: Major`; the prose the clinician reads does not contain it.

Had the list been sorted most-severe-first, both Majors would have been stated first and the truncation would have cost only Moderate corticosteroid findings.

**Second, smaller instance of the same cause** — same patient, `"Can I give her ziprasidone?"`:

```
 1. sev=Minor     Dexamethasone
 2. sev=Major     Hydrocortisone   (QT prolongation / torsade de pointes)
```

A Minor is presented ahead of a Major. Here the list is short enough that the answer reached the Major and led with it correctly, so nothing was lost — which is precisely why the warfarin case is the one that shows the cost.

## Scope — what is and is not broken

- The **verdict** logic is fine. Both answers correctly led with "No", so #283's "strongest governs" clause is working.
- The **question-pair arm** is fine. Verified in isolation on a patient prescribed neither drug (`"Can I give her warfarin and ibuprofen together?"` → `interactionPairs {found:1, reported:1}`, one `Major` chip, correct lead).
- The **screening arm** is fine — it sorts, and its cap now reports its extent (#336).
- What is broken is only the third arm, and it is the one that answers the commonest question shape: "can I give this patient drug X?"

Note this arm is also **uncapped**, so unlike the two sorted arms it has no `maxPairChips` cut that would have forced an ordering decision — which is likely why the sort was never added.

## Suggested fix

Sort the drug-in-play findings most-severe-first before appending, reusing `PAIR_SEVERITY_DESCENDING` (or the same `severityPriority` ordering) so all three arms agree. The comparator already exists and already handles unrated rules ranking above `major`.

The principle the two pairwise arms follow explicitly — *"what is dropped is always the least severe"* — should hold for this arm too. It currently does not hold even by accident, because nothing orders the list at all.

Worth noting that #340 (severity on the wire) makes this newly visible to clients, and a client that sorts chips by severity itself would mask the chip-ordering half of this while leaving the dropped-from-the-answer half untouched. The fix belongs in the validator, not the client.

dkayiwa:
Fixes #346

## What was broken

`DrugSafetyValidator.addInteractionWarnings` (the drug-in-play arm — the question's drug checked
against the patient's active orders) appended one interaction chip per partner in the knowledge base's
own row order, with no severity sort. The other two interaction arms already sort most-severe-first
before emitting (`Collections.sort(found, PAIR_SEVERITY_DESCENDING)` in
`addQuestionPairInteractions`, `Collections.sort(pairs, SCREENED_PAIR_SEVERITY_DESCENDING)` in
`addActiveOrderPairInteractions`).

Live on `main` @ `77c0f9a2`, a patient with 8 active orders asked "Can I give her warfarin?" got 8
interaction chips in KB row order; the LLM's answer enumerates the chips it is given and stops
partway through, so the Major warfarin×ibuprofen bleeding interaction (chip #8) never reached the
answer while five Moderate corticosteroid interactions did.

## The fix

Sorts the arm's candidate rules with a new `RULE_SEVERITY_DESCENDING` comparator on `SubjectRule`,
built from the same shared `severityPriority` ordering the other two arms' comparators already use, so
all three arms agree on "most severe first, unrated ranks above major." `Collections.sort` is stable,
so partners tied on severity keep the knowledge base's own row order, matching how the other two arms
already describe their own sort.

One javadoc paragraph on `ruleAbout` (which scans the arm's candidate list to decide which rule's chip
absorbs a class-arm sentence, for the one documented case where two distinct rows can name one active
order) was corrected: it used to say "the first in dataset order takes the fold," which is no longer
literally true now that the list arrives sorted by severity first.

A pre-existing test, `CoMedicationResolutionPerPassTest`, pins an exact golden list of chip
order for a two-drug question; its expected order was reordered to match the newly-correct
severity-first order (same content — same subject, partner, name and rating for every chip — verified
against the real `validate()` output, not hand-derived). A second pre-existing test's comment,
`DdiDrugReferenceSourceTest.replacingAGroupsWinnerLeavesTheChipsInDatasetOrderOfFirstAppearance`,
was also corrected: it used to explain a Major-before-Minor chip order by dataset position, which this
fix's severity sort now also explains on its own, so the comment was stale about *why* — the assertion
itself still holds and needed no change.

## Scope

Only the drug-in-play arm's chip order, per the ticket. Not #336 (bounded-screen extent) or #337
(mechanism-text garbling), both mentioned in the ticket's reproduction but explicitly out of scope.

One tangential, deliberately-deferred item: the corrected `DdiDrugReferenceSourceTest` case used to
double as a regression guard for `bestRulePerPartner`'s `LinkedHashMap` preserving a group's
dataset-first-appearance position across a winner replacement (verified by reverting that map to a
plain `HashMap` and confirming the test still passes either way, since its two chips differ in
severity and the new sort now decides their order on its own). Restoring that guard would need a new
fixture with two partners tied on severity; left as a follow-up rather than folded into this PR, since
it's a pre-existing implementation-detail risk unrelated to chip ordering by severity.

## Testing

- New test `DrugInPlayInteractionSeverityOrderTest`, with a dedicated fixture whose KB row order
  (Moderate, Minor, Major) does not match severity order — fails on `main` (chips return
  `[Moderate, Minor, Major]`), passes after the fix (`[Major, Moderate, Minor]`).
- `mvn -o clean install` from the repository root: 1694 api tests + 105 omod tests, 0 failures.

### DIFF
--- a/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
+++ b/api/src/main/java/org/openmrs/module/chartsearchai/reference/DrugSafetyValidator.java
@@ -2367,6 +2367,13 @@ public class DrugSafetyValidator {
 		DrugReference ref = subjects.subjectOf(rows.get(0));
 		List<SubjectRule> rules = new ArrayList<SubjectRule>(
 				bestRulePerPartner(rows, context, severityFloor, orderEntries));
+		// Most severe first, matching addQuestionPairInteractions and addActiveOrderPairInteractions
+		// (issue #346) — this arm used to leave the chips in the knowledge base's own row order, so an
+		// answer that enumerates them and stops partway through could drop the single most severe
+		// finding while stating several milder ones. Collections.sort is stable, so two rules tied on
+		// severity keep the dataset order the loop above produced them in, exactly as the other two
+		// arms already do.
+		Collections.sort(rules, RULE_SEVERITY_DESCENDING);
 		// Which rule row carries which class sentence, decided before anything is emitted: the class
 		// arm is walked per active-order CO-MEDICATION (issue #171 — it used to walk per CODE, so a
 		// substance filed under several codes reached this loop once per code) while the chips are one
@@ -2907,6 +2914,18 @@ public class DrugSafetyValidator {
 		}
 	}
 
+	/** Orders the drug-in-play arm's candidate rules most-severe first (issue #346); see
+	 *  {@link #severityPriority}. The same ordering {@link #PAIR_SEVERITY_DESCENDING} and
+	 *  {@link #SCREENED_PAIR_SEVERITY_DESCENDING} already apply to the other two interaction arms, on
+	 *  {@link SubjectRule} rather than on either of those arms' own candidate types. */
+	private static final Comparator<SubjectRule> RULE_SEVERITY_DESCENDING = new Comparator<SubjectRule>() {
+
+		@Override
+		public int compare(SubjectRule a, SubjectRule b) {
+			return Integer.compare(severityPriority(b.rule.getSeverity()), severityPriority(a.rule.getSeverity()));
+		}
+	};
+
 	/**
 	 * The (substance, partner) pairs an interaction chip has already been raised for in this pass, so the
 	 * screening arm ({@link #addActiveOrderPairInteractions}) can stand down from a pair the drug-in-play
@@ -3095,9 +3114,12 @@ public class DrugSafetyValidator {
 	 *         both name one order: across the full KB exactly one such pair exists, {@code enalapril}
 	 *         and {@code enalaprilat}, two genuinely different entries which that grouping deliberately
 	 *         keeps as two chips — the population that figure counts, and its base, are named at
-	 *         {@link #bestRulePerPartner} (issue #263). The first in dataset order takes the fold; the
-	 *         other keeps its rule chip unfolded, which is the conservative direction, since the
-	 *         alternative is stating one duplicate-therapy relationship twice.
+	 *         {@link #bestRulePerPartner} (issue #263). Since issue #346 {@code rules} arrives sorted
+	 *         most-severe-first ({@link #RULE_SEVERITY_DESCENDING}), so the first of the two to match
+	 *         takes the fold — the more severe of the two against this order where they differ, else
+	 *         whichever the dataset lists first, since the sort is stable. The other keeps its rule
+	 *         chip unfolded, which is the conservative direction, since the alternative is stating one
+	 *         duplicate-therapy relationship twice.
 	 */
 	private SubjectRule ruleAbout(Set<String> orderCodes, List<SubjectRule> rules,
 			CoMedications coMedications) {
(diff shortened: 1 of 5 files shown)


dkayiwa:
Closing in favour of #351, which was merged for #346.

Both PRs fix the same defect. #351 orders on `licensesWithholding(SafetyWarning)` first and `severityPriority` within each side, rather than on the rule's rating alone — which matters for a folded chip, where `getSeverity()` deliberately keeps reporting the rule's rating while the finding itself withholds. It also verified the ticket's own reproduction on a running standalone, and restored the `bestRulePerPartner` guard this change disarms rather than deferring it.

The analysis here was sound and the deferred-guard note in this PR is what identified that follow-up.

Issue #391, opened 2026-09-08
Proposal: consume the knowledge base's drug-disease rows and its derived tier (schema 1.2)


pbiondich:
The `ddinter` source ships the whole knowledge base byte-identical to a release of
[openmrs-ddi-knowledge-base](https://github.com/pbiondich/openmrs-ddi-knowledge-base) (ADR Decision 36).
That knowledge base has grown two tables since the copy the module bundles (schema 1.0, July). This
proposes that the module read them. It is two changes of different size, and I would be happy for them
to be split into separate issues if that is easier to review.

## The case that motivated it

A patient on stavudine is asked about metformin. DDInter's drug-drug row for that pair is `Unknown`
with no text, so the rule arm has nothing to raise and the default severity floor (`minor`) filters
it anyway. The clinician sees no chip. Yet DDInter carries both halves of the concern in its
drug-disease table:

- stavudine, `Liver Diseases`, Major: "Hepatotoxicity including **lactic acidosis**, severe hepatomegaly
  with steatosis ... has been associated with the use of some nucleoside reverse transcriptase inhibitors"
- metformin, `Acidosis, Lactic`, Major: "The use of metformin is contraindicated in patients with ... any
  condition associated with hypoxemia" (seven references, from Luft 1978 to Wiholm 1993)

DDInter never joins its two tables, so the pair reads as unrated. The knowledge base now carries the
drug-disease rows and, on top of them, a derived tier that records exactly that chain.

## What changed in the knowledge base

Two tables were added, both built reproducibly by `build_kb.py` from committed inputs.

| Table | Rows | What a row is |
|---|---|---|
| `disease_interactions` | 8,346 across 1,480 drugs and 470 conditions | `[drug_id, condition, severity, note_id]`: DDInter's drug-disease rating, with the description and references interned once in `disease_notes` (3,942) |
| `derived_interactions` | 111,138 linking 97,493 pairs through 185 conditions | `[cause_drug, cause_condition, cause_severity, cause_note, rated_drug, condition, rated_severity, rated_note]`: a causal chain, one drug's text says it causes a condition the other drug is rated for |

The derived tier is an inference, and the knowledge base says so in its metadata, its schema, and its
README. A row exists only when a sentence in the cause drug's drug-disease text names the condition on a
word boundary, carries a causal cue ("may cause", "associated with", "induce", "worsen", "have been
reported with"), and carries no precaution phrasing ("in patients with", "history of"). Two drugs
[...]

PR #392, opened 2026-09-08
ddinter source: read brand_names as aliases (knowledge-base schema 1.3), refresh the bundled knowledge base

pbiondich:
Closes the module side of pbiondich/openmrs-ddi-knowledge-base#5 (@dkayiwa's live case: a patient on simvastatin, asked *"Is it safe to give her panadol?"*, had nothing injected and nothing checked, because `panadol` was no alias of the Acetaminophen entry). Related to #391, which proposes reading the other two tables the refreshed file carries; this PR reads only `brand_names`.

## What changes

**`DdiDrugReferenceSource.DrugRow.of` reads `drugs[].brand_names` into the alias list** (knowledge-base schema 1.3). One loop beside the `rxnorm_name` and CIEL-name reads; on a schema 1.0 file `path("brand_names")` is a missing node and the loop runs zero times, so an older operator file loads exactly as before. A brand of a combination product is filed under every ingredient it contains (Vytorin under simvastatin and ezetimibe), the same many-to-one shape the bridged CIEL names already give the list.

**The bundled knowledge base is refreshed to the current release** (schema 1.3, byte-identical to `pbiondich/openmrs-ddi-knowledge-base` at `66489ff`, `out/ddi_knowledge_base.json` — 34,882,241 bytes, sha256 `b55f8b147a11a6b394f669f3d007253d595f9016005e42f92a689f249e4f1b2a`; the earlier `9938a2d` / `e1ebea3b…` this paragraph named was superseded by review round 1's exclusion commit, and the release and hash now sit in `DdiDrugReferenceSource`'s javadoc, which survives the squash merge). That is the documented refresh path (ADR Decision 36), kept as its own commit for readability; it is not separable from the parser change in this direction, since the new test's precondition and the read itself need a file that declares brand names. The parser reads by key, so the two tables this PR does not read (`disease_interactions`, `derived_interactions`) are ignored. Cost: the file grows 18.9 → 34.9 MB; retained memory is unchanged until those tables are read.

## What the knowledge base did about the hazard this repo already guards

`ShippedAliasVocabularyTest` pins that every alias of five characters or fewer is a substance name, the property ADR Decision 68 declines to give up: an alias that is also an ordinary word makes the word resolve a drug, and `matchesText` has no length floor. The raw RxNorm brand list broke it (`Align`, `Today`, `Sleep Aid`, `Ella`), so the knowledge base now excludes, at source and with the dropped list committed beside the kept one (`src/brand_names_excluded.json`, 203 names, each with the rule it fell to):

- a brand of three characters or fewer (`Alo`, `DOK`, `Ery`: acronyms);
- one whose every word is a lower-case dictionary word (`Align`, `Today`, `Sleep Aid`, `Correct (New Formula)`); capitalised dictionary entries are kept, because dictionaries list Demerol and Dilantin too;
- one that is a first name (`Ella`, `Julie`, `Kirsty`, `Dotti`).

What survives is 3,861 distinct brands over 1,642 of 2,283 drugs (71.9%) — the brand count is the head's, produced by `ShippedAliasVocabularyTest.theShippedVocabularyCarriesExactlyTheBrandNamesPinnedBesideIt` over the real load, including the short ones clinicians actually ask by: `advil`, `aleve`, `cipro`, `coreg`, `lasix`, `xanax`, `zocor`. RxNorm is US-centric (Panadol yes, Calpol no); CIEL synonyms would cover some of that and are a separate fetch.

## Tests

TDD, through the real parser, resolution and injector, over `ddi-brand-name-aliases.json`, a verbatim slice of the shipped file (acetaminophen, simvastatin, ezetimibe and the rows among them):

- `DdinterBrandNameAliasTest` (4 cases): `panadol` resolves the Acetaminophen entry; `Vytorin` resolves both ezetimibe and simvastatin; the panadol question injects acetaminophen's reference record for a patient on simvastatin; the schema 1.0 excerpt still loads and honestly resolves nothing for a brand. Three of the four fail on `main` without the parser change (measured by stashing the source file).

Three existing cases moved with the refresh, each read before it was touched:

- `ShippedAliasVocabularyTest`: the assertion is restated as the residue the property still holds. Every short alias that is not a brand the shipped file itself declares must be one of the seven substance names (it is), and no brand alias is three characters or fewer (none is). Brand names are read off the raw JSON, not the parser, because the question is which aliases are brands.
- `BridgedConceptLegBoundsTest`: `intersectionNarrows` 243 → 244. Diffed through the same walk over both files: the one concept that joined is CIEL 74185 *Cytotect* (a CMV immunoglobulin), whose name now also reaches Misoprostol because misoprostol's brands carry *Cytotec* and the order-name rule's two-letter inflection allowance reads one as an inflection of the other. A near-identical brand is a collision shape generic names rarely produce, and the intersection bound keeps Misoprostol out of the answer, which is what it is for. Recorded in the test beside the constant.
- `SlicedReferenceRowProvenanceTest`: `ddi-issue338-allergy-cross-reactivity.json` regenerated field-for-field from the refreshed file (its two rows gained `brand_names`); interactions and mechanisms unchanged.

Full api suite on JDK 11: 1,996 run, 0 failures, 0 errors, 57 skipped (the same 57 as `main`).

## Review round 1 (2026-09-09)

- `stye`/`propel`: excluded by name upstream (knowledge base `66489ff`), the residue enumerated (16 candidates, read one by one, these two are the ordinary words), and the 193 short brand aliases now asserted as a sorted list so the next lacuna lands in a diff.
- ADR Decision 68 restated for the refreshed file.
- `ddi-brand-name-aliases.json` added to `SlicedReferenceRowProvenanceTest.SLICES`.
- Packed-jar figure moved with the file: 4,031,048 bytes for the entry, read off the built api jar.

## Not in this PR

- Reading `disease_interactions` or `derived_interactions` (#391).
- CIEL synonyms as aliases (a fresh OCL pull on the knowledge-base side).
- Any change to how aliases are matched.

## Review round 2 (2026-09-14)

- **The duplicate-therapy defect on `DdiDrugReferenceSource.java:600` is fixed.** A brand is an alias of every row that carries it, so one prescription resolved to two substances at one rank and each became a co-medication; the restating-existing-therapy skip asks whether a partner IS the substance in play, which the sibling reading is not. The skip gains a third leg, `OrderPartner.readingsOfOneOrder` — the substances an order's own names imply WITHOUT NAMING are readings of that one prescription. Naming and not counting, the distinction `substancesNamedByBridge` already draws for the bridge, so an order spelling its constituents keeps their chips. → `AmbiguousBrandNamedOrderTest` (6 cases, the real `validate` and `injectRecords` over the shipped file).
- **Provenance**: the release and sha256 above are now in the javadoc, verified by blob id against the knowledge base's own tree entry.
- **ADR Decision 36's two copies of the packed-jar figure** moved with the file: 4,031,048 bytes, read off the built api jar with `unzip -v`.
- **The brand vocabulary is pinned whole** rather than at its short tail, which could not see `propel` (six characters). `api/src/test/resources/chartsearchai-test/shipped-brand-aliases.txt`, 3,861 lines, asserted as the brands that joined and left.

Full api suite on JDK 11 after round 2: 2,003 run, 0 failures, 0 errors, 57 skipped.


api/src/test/java/org/openmrs/module/chartsearchai/reference/ShippedAliasVocabularyTest.java
```diff
@@ -62,32 +69,75 @@ private static Set<String> distinctAliases() {
...
+				}
+			}
+		}
+		return brands;
+	}
+
+	/**
+	 * The vocabulary's SHORT tail, which is what the argument turns on: every one of these is either a
+	 * substance in its own right or a brand name the knowledge base declares, so nothing in the shipped
+	 * alias set is a word that ordinary clinical prose carries in another sense. That is the property
```
dkayiwa, line 94: This sentence is what the safety argument rests on, and the refreshed file falsifies it. `stye` is a brand of Mineral oil (a real one, a lubricant eye ointment, which is why the upstream rule's intent covers it), so an ordinary clinical sentence resolves a drug through `matchesText`, which has no length floor. Driving the real service over the shipped entries:

```
findImpliedByQuery("The patient has a stye on the right upper eyelid.")
  main:        []
  this branch: [Mineral oil]
```

and through the real `injectRecords` with the safety validator behind it, patient's only active order Levofloxacin, question "How should I manage the stye on her eyelid?":

```
Safety finding - Mineral oil: Mineral oil interacts with active order Levofloxacin - Moderate.
Bowel cleansing as well as overuse of certain laxatives may cause electrolyte loss and increase
the risk of torsade de pointes ventricular arrhythmia ... This finding is a reason to withhold it.
```

The same arrangement on main injects no record and raises no finding. So merged as is, a question carrying the word "stye" puts a laxative into the prompt as a citable `safety_finding` with a withholding clause, about a drug nobody asked about. That is the `alert` failure Decision 68 declines to accept, on the module's own default dataset, and nothing throws, so CI cannot see it and the cost lands on whoever reads the answer.

It looks like a dictionary lacuna rather than a design mistake: `fetch_brand_names.py`'s ordinary-word rule reads `/usr/share/dict/words`, which carries `sty` but not `stye`, and `propeller`/`propellant` but not `propel`. `propel` is kept too, and resolves the three Mometasone rows on "She is unable to propel her wheelchair unaided." `src/curation.json` `brand_exclusions` plus `--refilter` drops both with no fetch. I did not try to enumerate the rest; with the dictionary that lossy I could not do it from here.

Worth keeping this case able to see the next one as well. Neither new assertion can: a short brand is exempted from the substance-name list, and only a brand of three characters or fewer fails, so `stye` (4) and `propel` (6) both pass. Asserting the five-character-or-fewer brand aliases as a sorted list too, which is what the javadoc below asks for ("a knowledge-base refresh reports WHICH word joined rather than that a number moved"), puts the 194 the shipped file carries into the diff where they can be read. A list like that is how `Align`, `Today` and `Ella` were caught.

api/src/test/java/org/openmrs/module/chartsearchai/reference/ShippedAliasVocabularyTest.java
```diff
@@ -10,15 +10,22 @@
...
+import java.util.Locale;
 import java.util.Set;
 
 import org.junit.jupiter.api.Test;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 /**
  * What the shipped knowledge base's alias vocabulary is SHAPED like, measured through the real parser
```
dkayiwa, line 30: `docs/adr.md` still carries the pre-refresh version of this figure as current. Decision 68 reads "every alias of five characters or fewer is itself a substance name — `clove`, `hemin`, `iron`, `kava`, `opium`, `urea`, `yeast`, and nothing else", and names this case as what asserts it. On the refreshed file that set is 201 aliases, and the case now asserts something narrower. That paragraph is the argument for refusing the CIEL-synonym harvest the ticket asked for, so it wants restating with the refresh rather than leaving the difference to be found in the test.

api/src/test/resources/chartsearchai-test/ddi-brand-name-aliases.json
```diff
@@ -0,0 +1,916 @@
+{
+ "metadata": {
+  "schema_version": "1.3",
+  "title": "OpenMRS DDI knowledge base — verbatim slice for brand-name alias tests",
```
dkayiwa, line 4: This wants adding to `SlicedReferenceRowProvenanceTest.SLICES`. That list's javadoc says a new slice is added to it rather than given a guard of its own, and it is the only thing holding a fixture that calls itself verbatim to the shipped rows. I added it and ran the case, green, so the slice is faithful today: rows field for field, in the shipped file's own order, with the interaction rows falling wholly inside it and their mechanisms.

Without it the drift is silent. `DdinterBrandNameAliasTest` builds its expectations out of this file, so if a later refresh dropped `Panadol` from acetaminophen's brands upstream, the case that proves the reported bug fixed would stay green while the shipped default stopped resolving it.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
```diff
@@ -40,7 +40,8 @@
  * duplicated:
  * <ul>
  *   <li>{@code mechanisms}: {@code {groupId: {text, categories}}}, each description stored once;</li>
- *   <li>{@code drugs}: {@code {id, name, rxcui, rxnorm_name, drugbank_id, atc[], ciel[]}};</li>
+ *   <li>{@code drugs}: {@code {id, name, rxcui, rxnorm_name, drugbank_id, atc[], brand_names[], ciel[]}}
+ *       ({@code brand_names} since knowledge-base schema 1.3; read as aliases, absent on older files);</li>
```
dkayiwa, line 44: The cost sentence further down this javadoc still says "2.1 MB of packed jar", which is main's file. The resource packs to 3.8 MB now, read off the built api jar's entry for it, against 2.10 MB for main's. All but about 40 KB of that increase is `disease_notes`, `disease_interactions` and `derived_interactions`, which nothing reads yet (attributed by re-serializing the file with those three tables removed). No objection to the refresh, but that figure is what Decision 36 argues the whole-KB default from, so it is worth moving with the file.

On the description's offer to drop this commit and re-stage it alongside #391: dropping it reddens `ShippedAliasVocabularyTest`'s new precondition (`brands.size() > 1000`) and leaves the parser read with nothing to read on a default install, so the two are not separable in that direction.

api/src/test/java/org/openmrs/module/chartsearchai/reference/ShippedAliasVocabularyTest.java
pbiondich: Thank you, that is a real one, and the dictionary is the culprit exactly as you read it. Both are excluded by name upstream now (`src/curation.json` `brand_exclusions`, knowledge base `66489ff`), and the bundled file here is that release (7e6de4c0).

Rather than stop at the two, I enumerated the residue. Of the 3,028 single-word brands the knowledge base keeps, 16 are words a 370k-entry English list knows and `/usr/share/dict/words` does not know in any case; read one by one, the ordinary words among them are exactly `Stye` and `Propel` (`Isochron` and `Fungistat` are technical words I left in; the other twelve are trademarks that larger list lowercases, Actos, Valium, Prolia, Benadryl among them, which is why it cannot replace the rule). So the next lacuna is found the way this one was, by reading, and the case now gives that reading a diff to happen in: the 193 short brand aliases are asserted as a sorted list, produced through `DrugReference.getAliases()` over the real load and intersected with the raw `brand_names`, beside the seven substance names and the three-character floor. The javadoc says what a newcomer on that list is to be read for, and that the knowledge base's exclusion list is updated before this list is.

Measured after the fix, through the real `findImpliedByQuery` over the refreshed entries: "The patient has a stye on the right upper eyelid" resolves nothing.

api/src/test/java/org/openmrs/module/chartsearchai/reference/ShippedAliasVocabularyTest.java
pbiondich: Restated in `docs/adr.md` under Decision 68 rather than left in the test: the original figure is now dated to the schema 1.0 file, and a new paragraph carries the schema 1.3 state (200 short aliases, the seven substance names and 193 brands), the two halves the property is now stated as, the exclusion rule the second half rests on, the `stye` measurement and the enumeration of the residue. The argument for refusing the CIEL-synonym harvest stands on the first half, which is unchanged: no short alias that is not a declared brand is anything but one of the seven.

api/src/test/resources/chartsearchai-test/ddi-brand-name-aliases.json
pbiondich: Added. Both slices were regenerated field-for-field from the refreshed file and the case is green on the pair.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
pbiondich: Moved with the file. I measured it myself rather than copy yours: 4,031,048 bytes for the entry, read off the built api jar with `unzip -v`, and the javadoc says so and attributes the increase to the three unread tables as you did. On the description's offer to drop the refresh commit: you are right that it is not separable in that direction, and I have taken that sentence out of the description.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
```diff
@@ -580,12 +585,23 @@ static DrugRow of(JsonNode d) {
...
 			if (d.path("rxnorm_name").isTextual()) {
 				addAlias(aliases, d.get("rxnorm_name").asText());
 			}
+			// Brand names (knowledge base schema 1.3, openmrs-ddi-knowledge-base issue #5): a clinician's
+			// question is where brand names live, and without them "is it safe to give her panadol?" resolved
+			// to nothing, so the asked-about drug was never injected or checked. A brand of a combination
+			// product is filed under every ingredient it contains (Vytorin under simvastatin AND ezetimibe),
+			// which is the same many-to-one shape the bridged CIEL names below already give the alias list.
+			// Absent on a schema 1.0 file: path() yields a missing node, and the loop runs zero times.
+			for (JsonNode b : d.path("brand_names")) {
```
dkayiwa, line 600: This needs fixing before merge. A brand the knowledge base files under two rows this module counts as two substances makes one prescription resolve twice, and neither leg of the restating-existing-therapy skip can see the second one, so a patient on a single order gets a duplicate-therapy finding against her own prescription.

Driving the real `validate` over the shipped entries, one active order whose concept the dictionary did not map (the shape `addPartnersForUnmappedOrders` exists for), 0 chips on `main` in every row:

```
"Nexium 40mg"     / "Can I give her esomeprazole?"
  Esomeprazole is in the same ATC class (A02BC) as active order Omeprazole — possible duplicate therapy
"Neurontin 300mg" / "Can I give her gabapentin?"
  Gabapentin is in the same ATC class (N02BF) as active order Gabapentin enacarbil — possible duplicate therapy
"Tricor 145mg"    / "Can I give her fenofibrate?"
  Fenofibrate is in the same ATC class (C10AB) as active order Fenofibric acid (fenofibrate) — possible duplicate therapy
"Nasonex"         / "Can I give her mometasone?"
  Mometasone is in the same ATC class (D07AC) as active order Mometasone furoate — possible duplicate therapy
```

There is no second drug in any of those four charts. Through the real `injectRecords` it reaches the prompt as citable evidence carrying the withholding clause:

```
Safety finding — Esomeprazole: Esomeprazole is in the same ATC class (A02BC) as active
order Omeprazole — possible duplicate therapy. This finding is a reason to withhold it.
```

Same root, second arm: with `nexium` also recorded as an allergen, that pass raises `Esomeprazole is contraindicated by a recorded allergy to "nexium".` and `Omeprazole is contraindicated by a recorded allergy to "nexium".`, two chips off one record.

Three controls put the cause here rather than in the arm. The same order with its concept mapped to `A02BC05` raises nothing, because the code walk governs it and the name leg never runs. `Zocor 20mg` asked about simvastatin raises nothing, one substance. And `esomeprazole 40mg` asked about esomeprazole raises nothing, which is why `main` cannot reach this at all: the generic spelling resolved to one row because `nameMatchStrength`'s display-name rung broke the tie, `findImpliedByDrugName("esomeprazole")` being `[Esomeprazole]` on both branches while `findImpliedByDrugName("Nexium")` is `[Esomeprazole, Omeprazole]` here. A brand matches both rows only as an alias, at equal strength, so the tie-break that used to hide DDInter's shared-RxCUI rows is gone. 44 rxcui values in the shipped file are carried by rows with differing `drugbank_id` that declare brand names; I drove four of them.

The skip then has nothing to fire on. `addPartnersForUnmappedOrders` gives each partner `partner.substances.add(named.getKey())`, its own key and no other, and sets `codesFromDataset = true`, which scopes out the exact-code leg, so `classRelationships`' `partner.substances.contains(refSubstance) || (!partner.codesFromDataset && ...)` is false in both legs for a sibling substance resolved from the same order. The comment there, "This partner IS this substance, so restating it is not duplicating it", held while one order's name implied one substance.

What I would do is give a name-leg partner every substance THAT ORDER's name implies, `substanceRowsNamedBy(order, ...).keySet()` rather than only its own. That is the scoping `CLAUDE.md` already states for the unnameable-code rung ("over EVERY order `ordersCarrying` returns rather than the first"), and it reads right for a real combination too, where the constituents are one prescription rather than two therapies. Fixing the `Omeprazole` row upstream would take out this instance (its `rxnorm_name` is `esomeprazole` and its `rxcui` 283742, which is how esomeprazole's brands landed on it) but not the shape, since the other three families are distinctions the KB is right to keep apart.

Nothing throws on the way through: the chip is `severity: null`, so nothing rates it, and the rendered clause above is what the model reads. Scope is installs with `chartsearchai.drugReference.enabled` on.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
```diff
@@ -65,7 +66,11 @@
  * openmrs-ddi-knowledge-base release it came from, so a refresh is a file copy and its provenance is
  * checkable by hash. It used to be a 16-drug excerpt with the full file left to the operator to download,
```
dkayiwa, line 67: "its provenance is checkable by hash" has nothing to check against for this file. The head bundles sha256 `b55f8b14…` with 3,861 declared brands, while the description and the refresh commit both still record `9938a2d`, sha256 `e1ebea3b…` and 3,863 brands, which is the file as it stood at 850b2d07 (I computed both hashes). The commit that replaced it names knowledge base `66489ff` and no hash, and since this repo squash-merges the commit messages, the only hash in the permanent record will be the superseded one. Nothing in-tree carries the release either: the file's own `metadata` block has `schema_version` but no source ref. Worth putting the head's release and its hash in one of the two places a later reader looks, the description or the javadoc beside this sentence.

api/src/test/java/org/openmrs/module/chartsearchai/reference/ShippedAliasVocabularyTest.java
dkayiwa: The list does what it was asked for on `stye`, and it cannot see `propel`. Putting `Propel` back as a brand of the three Mometasone rows in the shipped file, nothing else changed, leaves the whole api suite green: `Tests run: 1996, Failures: 0, Errors: 0, Skipped: 57`, the same totals this PR reports. `Stye` back the same way reddens `everyShortAliasTheShippedVocabularyCarriesIsItselfASubstance` on the list, as intended. So of the two words this round removed, one can come back unseen.

There is nothing else to read it in. The shipped file has no newlines, so a refresh is `1 insertion(+), 1 deletion(-)` over one 34.9 MB line, and that 193-item list is the only surface a newcomer shows up on. The band is set where the hazard is not: `matchesText` has no length floor, which is what `SHORT`'s own comment says, and `propel` is 6.

I would pin the whole declared brand set instead of the short tail, as a sorted test resource of 3,861 lines asserted against `getAliases()` intersected with the raw `brand_names`, so a refresh diffs to exactly the brands that joined and left and no band has to be picked. That also turns "the knowledge base's exclusion list is updated before this list is" into something the diff shows rather than something a reader has to remember.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
dkayiwa: Decision 36's own two copies of that figure did not move with it. The measurement paragraph still reads "**2.1 MB** of packed jar" and the consequences list "**−** +2.1 MB of packed jar, ~30 MB of heap, and 0.6 s", so the decision that argues the whole-KB default understates the cost it authorizes by 1.8× (2,204,469 bytes for the entry on `main` against 4,031,048 here, both read off the built api jar). The omod sentence beside it, "Narrowing that would recover ~2.1 MB", doubles the same way.

The paragraph's other two figures do hold on the refreshed file, so they need no re-measuring: 30-31 MB retained and 0.41-0.48 s warm through `DdiDrugReferenceSource.parse`, against 0.34-0.41 s for `main`'s file in the same JVM.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
dkayiwa: Fixed in 83fa510, on this branch. The cause is where you put it: the brand is an alias of both rows, so `findImpliedByDrugName("Nexium 40mg")` returns both at one claim rank, each becomes a co-medication, and the skip asks whether a partner IS the substance in play — which the sibling reading is not.

The remedy is a third leg on the skip rather than the widening you proposed, because I measured the widening and it is not free: filling `partner.substances` with everything the order's name implies reddens `AmbiguousBrandNamedOrderTest.anUnmappedCombinationSpellingBothItsSubstancesStillReportsTheOther` — an unmapped `Metronidazole and secnidazole`, asked about metronidazole, loses the chip naming Secnidazole — while `DuplicateTherapySelfChipTest`, `PartialOrderCoveragePartnerTest` and `OneOrderNameAcrossOneResponseTest` stay green under it. That is the loss `OrderPartner.substances`' javadoc argued and nothing pinned; it is pinned now.

So `OrderPartner.readingsOfOneOrder` carries the substances an order's own recorded names imply WITHOUT NAMING, and the skip reads it beside identity. Naming and not counting — the distinction `DrugReferenceService.substancesNamedByBridge` already draws for the dictionary bridge, on this very pair of substances: a name that SPELLS its constituents names each of them and they keep their own chips; a brand names neither, so the module knows the prescription is one of them and cannot say which.

It is read by that skip alone, deliberately not by `alreadyACoMedication`. Collapsing the two readings into one co-medication would close more, but it loses every class only the dropped reading's codes reach — and for a combination BRAND, which RxNorm files under each ingredient it really contains, that is a duplicate-therapy chip about a drug the patient genuinely is on.

Measured after the fix, through the real `validate` over the shipped file: all four of your rows raise nothing; the same `Nexium 40mg` chart asked about pantoprazole still raises chips, so the arm is live on it; `Vytorin 10/20` asked about ezetimibe keeps its Simvastatin interaction chip; the unmapped `Metronidazole and secnidazole` keeps its chip in both directions; and through the real `injectRecords` the `safety_finding` you quoted is no longer injected. `AmbiguousBrandNamedOrderTest` pins all of that, 6 cases. Full api suite: 2,003 run, 0 failures, 0 errors, 57 skipped.

Two things I did not close, both recorded at `OrderPartner.readingsOfOneOrder` rather than left to be rediscovered:

- **The residue.** Both readings remain co-medications, so `Nexium 40mg` asked about pantoprazole raises one chip naming `active order Omeprazole` and one naming `active order Esomeprazole`. One prescription, two names — the names are a guess, the class claim is not, and nothing in the knowledge base separates that shape from a combination brand where both chips are true.
- **The second arm does not reproduce as a defect.** With `nexium` also recorded as an allergen the pass now raises exactly two chips: `Esomeprazole is contraindicated by a recorded allergy to "nexium".` and the same sentence for Omeprazole. That is #238's relationship wording rather than the asserting form ("The patient has a recorded allergy to X."), and two chips off one ambiguous recorded allergen is #193/#195's deliberate additive resolution — on `main` that chart raises nothing at all, because `nexium` resolved nothing. The duplicate-therapy chip that used to sit between them is gone.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
dkayiwa: Recorded in fc610cd, in the javadoc beside the sentence rather than only in the description, for the squash-merge reason you give.

The head's file is openmrs-ddi-knowledge-base `66489ff3207bcb809968c3de81cfba94d05692ae`, `out/ddi_knowledge_base.json`: 34,882,241 bytes, sha256 `b55f8b147a11a6b394f669f3d007253d595f9016005e42f92a689f249e4f1b2a` (`shasum -a 256` on the bundled resource — the same `b55f8b14…` you computed).

I verified the release rather than taking the claim on trust, and without fetching 34 MB: the bundled resource's git blob id is `021ee454697b20cedfddb0d918737f91f23bd8d2` (`git hash-object`), which is exactly the id and size that release's own `out` tree entry carries (`gh api repos/pbiondich/openmrs-ddi-knowledge-base/git/trees/…`), so the two files are byte-identical. That commit is "Exclude Stye and Propel: ordinary clinical words the dictionary lacks", which is the round-1 exclusion.

The description is corrected too: it named the superseded `9938a2d` / `e1ebea3b…` and 3,863 brands, and now names the head's release, hash and 3,861 — that count produced by `ShippedAliasVocabularyTest.theShippedVocabularyCarriesExactlyTheBrandNamesPinnedBesideIt` over the real load. The javadoc paragraph also says a refresh restates all of it.

api/src/main/java/org/openmrs/module/chartsearchai/reference/DdiDrugReferenceSource.java
dkayiwa: Both copies moved, in 11c37c8.

I measured the figure myself rather than taking it from here or from the javadoc: `mvn -o package -pl api`, then `unzip -v` on the built jar's own entry — `chartsearchai/ddi-knowledge-base.json`, 34,882,241 bytes uncompressed, **4,031,048 bytes** packed. That is what the measurement paragraph and the consequences list now carry, each saying what produced it, and the omod sentence beside them ("Narrowing that would recover …") is restated with the same figure.

The paragraph also says what the figure is a share of, so the gap is not left implicit: the decision was quoting 2.1 MB for a resource that packs to 3.8 MB, understating the packed cost it authorizes by about 1.8x of that 2.1 MB. It attributes all but about 40 KB of the increase to `disease_notes`, `disease_interactions` and `derived_interactions`, and says the retained-heap figure does not move with it. Your other two figures are left as they stand, with a note that they were re-measured on the refreshed file in this review and hold — I did not re-take them.

api/src/test/java/org/openmrs/module/chartsearchai/reference/ShippedAliasVocabularyTest.java
dkayiwa: Taken, and taken the way you proposed: the whole declared brand set is pinned, in 2760f19.

I reproduced your measurement first. With `Propel` restored as a brand of the three Mometasone rows (nothing else changed), the short-tail list does not see it — `propel` is six characters and that list's band is five. So the band was the wrong unit, and it is not a band a reading can pick: the hazard is a brand that is also an ordinary word, and `matchesText` has no length floor, which is `stye`'s own argument.

`api/src/test/resources/chartsearchai-test/shipped-brand-aliases.txt` now carries all 3,861, sorted, produced through the real load (`DrugReference.getAliases()` over `shippedEntries()`, intersected with the raw `brand_names`), and `ShippedAliasVocabularyTest.theShippedVocabularyCarriesExactlyTheBrandNamesPinnedBesideIt` asserts it. It asserts the brands that JOINED and the ones that LEFT rather than list equality, so a refresh fails naming exactly those instead of printing 3,861 lines: under the `Propel` mutation it fails with `[propel]`, while `everyShortAliasTheShippedVocabularyCarriesIsItselfASubstance` stays green beside it — which is the measurement that the short list was a projection rather than a second guard.

So I removed the short brand list rather than keeping both, as you suggested. It is subsumed (the short set is a function of this one) and keeping it is a second thing to update on a refresh. The case it lived in keeps the two statements a whole-set pin does not make: the non-brand short aliases are the seven substance names, and no brand alias is three characters or fewer. One assertion was added to keep the intersection honest — every brand the file declares must reach the alias vocabulary, so nothing can leave the pin by being dropped between the file and `getAliases()` (measured: 3,861 declared, 3,861 carried, none dropped).

The javadoc says what a reader does with the diff and that the exclusion is made UPSTREAM before the list is regenerated — a name dropped from this file alone still resolves. Decision 68 is restated with it.

